PR #2, opened 2020-04-24
pull list into client library

lkingland:
Adding the new list command to the client library, and creating a new knative concrete implementation package!


lkingland:
By the way, in addition this PR includes a removal of a DeleteExplicit test, as the API no longer supports it, and a downgrade from yaml.v3 to yaml.v2 as it is unstable, was not decoding as expected, and this was fortunately caught by the test.

matejvasek:
LGTM

PR #5, opened 2020-04-27
Describe function

matejvasek:


matejvasek:
@lkingland I think this is ready for review.

client/client.go
```diff
@@ -105,6 +106,22 @@ type Lister interface {
 	List() ([]string, error)
 }
 
+type Subscription struct {
+	Source string `json:"source" yaml:"source"`
+	Type string `json:"type" yaml:"type"`
+	Broker string `json:"broker" yaml:"broker"`
```
matejvasek, line 112: Maybe we don't want to display broker now, probably we could assume `default` for now, WDYT?

client/client.go
```diff
@@ -105,6 +106,22 @@ type Lister interface {
...
+}
+
+type FunctionDescription struct {
+	Name          string `json:"name" yaml:"name"`
+	Routes        []string `json:"routes" yaml:"routes"`
```
matejvasek, line 117: Do we need to assume multiple routes?

client/client.go
lkingland: That sounds like a good idea to me.  I expect we'll want to expose the concept later, but for now, simpler the better.  Once we expose something, we're stuck with it.  Better to start small and expose as-needed.

client/client.go
lkingland: I think that, at this point, a single route per function is sufficient.  Can you think of a reason for multiples at this time?

cmd/describe.go
```diff
@@ -0,0 +1,76 @@
...
+		return
+	}
+
+	description, err := client.Describe(name)
+	if err != nil {
+		return
+	}
+
+
+	formatFunctions := map[string] func(interface{})([]byte, error) {
```
lkingland, line 59: I love being able to choose output format.  should do this for every command...

knative/describer.go
```diff
@@ -0,0 +1,100 @@
...
+	eventingv1client "knative.dev/eventing/pkg/client/clientset/versioned/typed/eventing/v1alpha1"
+	servingv1client "knative.dev/serving/pkg/client/clientset/versioned/typed/serving/v1alpha1"
+)
+
+type Describer struct {
+	Verbose        bool
+	namespace      string
+	servingClient  *servingv1client.ServingV1alpha1Client
+	eventingClient *eventingv1client.EventingV1alpha1Client
+	config 		   *rest.Config
```
lkingland, line 19: This is strange; how did formatting get off on this line?  You have `gofmt` run on save I presume?

knative/describer.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	describer.eventingClient, err = eventingv1client.NewForConfig(config)
+	if err != nil {
+		return
+	}
+	describer.config = config
+	return
+}
+
+func (describer *Describer) Describe(name string) (description client.FunctionDescription, err error) {
```
lkingland, line 48: The name which is actually used in-cluster is a "DNS subdomain" (DNS 1035 label) encoded version of the fully-qualified service name as presented to the faas developer.  I.e. `www.example.com` becomes `www-example-com`.  There are utility functions for going back-and-forth between these representations in `k8s/names.go` such that you can `k8s.ToSubdomain(displayName)` when describing, and then `k8s.FromSubdomain(serviceName)` when exporting.

knative/describer.go
```diff
@@ -0,0 +1,100 @@
...
+	servingClient := describer.servingClient
+	eventingClient := describer.eventingClient
+
+	service, err := servingClient.Services(namespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		return
+	}
+
+	serviceLabel := fmt.Sprintf("serving.knative.dev/service=%s", name)
+	routes, err := servingClient.Routes(namespace).List(metav1.ListOptions{LabelSelector: serviceLabel})
```
lkingland, line 60: I agree with your earlier comment that we probably want to assume a 1-to-1 between function and route at this point; though we may eat our words later.

knative/describer.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return
+	}
+
+	triggerMatches := func(t *v1alpha1.Trigger) bool {
+		return (t.Spec.Subscriber.Ref != nil && t.Spec.Subscriber.Ref.Name == service.Name) ||
+			(t.Spec.Subscriber.URI != nil && service.Status.Address != nil && service.Status.Address.URL != nil &&
+				t.Spec.Subscriber.URI.Path == service.Status.Address.URL.Path)
+
+	}
```
lkingland, line 80: 😳

lkingland:
@matejvasek I am going to go ahead and merge this in irrespective of my above comments.

Issue #12, opened 2020-05-26
replace appsody build with buildpacks


lkingland:
The `build` step presently uses the `appsody` binary for its implementation, and makes use of custom "stacks" referenced through the repository `boson-project/stacks`.   Appsody offers many features which are not applicable for our "faas" context, and can be omitted for speed improvements and clarity.  Additionally, breaking the dependency on a locally installed binary is of course preferable.  The tentative choice of technology that has emerged is buildpacks.  

We would ideally replace the `appsody.Builder` implementation in `./appsody/builder.go` (which implements the `faas.Builder` interface) with a `buildpacks.Builder` implementation in `./buildpacks/builder.go`.  This could utilize the `libcnd` library https://github.com/buildpacks/libcnb or similar.

Other references that might help are https://buildpacks.io/docs/install-pack/ and https://github.com/lance/buildpacks

matejvasek:
some progress: https://github.com/matejvasek/buildpacks/commit/6ebf74e719f67a80d8c84787762c33b74ce1e5e5
https://github.com/matejvasek/faas/commit/825e1608690bcf8f1144e2277d49b9e92bb8413c

PR #19, opened 2020-06-05
feat: buildpacks

matejvasek:


matejvasek:
relates to:  #12

buildpacks/builder.go
```diff
@@ -0,0 +1,67 @@
...
+	registry  string
+	namespace string
+}
+
+func NewBuilder(registry, namespace string) *Builder {
+	return &Builder{registry: registry, namespace: namespace}
+}
+
+var lang2pack = map[string]string{
+	"java": "quay.io/mvasek/faas-quarkus-builder",
```
lkingland, line 25: Do you want to upload these to the boson project at Quay?

appsody/builder.go
```diff
@@ -26,7 +26,7 @@ func NewBuilder(registry, namespace string) *Builder {
 }
 
 // Build an image from the function source at path.
-func (n *Builder) Build(name, path string) (image string, err error) {
+func (n *Builder) Build(name, language, path string) (image string, err error) {
```
lkingland, line 29: Deriving the language from the .faas.config didn't work out?

client.go
```diff
@@ -57,7 +57,7 @@ type Deployer interface {
 // Updater of a deployed service function with new image.
 type Updater interface {
 	// Deploy a service function of given name, using given backing image.
-	Update(name, image string) error
+	Update(name, language, image string) error
```
lkingland, line 60: I'm not seeing where the language is being used for update?  Is this an early addition that should be part of replacing the KN binary dependency later?

matejvasek: Update calls `Build()`.

appsody/builder.go
matejvasek: I am not sure if this method should be aware of the config file. The config is read by caller of this function.

client.go
lkingland: Ah, it's down in build.  Resolving.

buildpacks/builder.go
matejvasek: Probably that's better palce.

appsody/builder.go
lkingland: That's fine.  We can just remember that the source code has a place for metadata like this if we want to lean towards introspection in the future.

matejvasek: Maybe we could pass Function{} struct there?

client.go
matejvasek: You are right, it's unused.

appsody/builder.go
lkingland: We could, but there are purposely zero dependencies from the parent into the implementation packages such that they are free to use exported utilities from the parent package.  So any implementation of a "Builder", for instance, can create a "Function" about a given path to get a typesafe interface to a meta-object representing that function, for instance.

lkingland: By structuring in this way, with the "Domain Types" in the root, but the root having no dependencies on the implementing packages, we can share certain abstractions like this between the subcomponents.  For instance, I could see us adding more and more metadata to the .faas.yaml as time goes on.  Any implementer of Builder, Updater, etc., by just creating a new `Function(path)` can then have access to this metadata.
I don't think passing language in as a parameter is necessarily wrong.  But it seems redundant since, in theory, the path points to a function whose language should be derivable.

matejvasek: Isn't `name` redundant too then?

lkingland: That's a good point.  It's "Build the function at this path", not "Build the function at this path and use this name".
That was already set during the initialization phase when it was asked to "Initialize a function at this location with this name...".

buildpacks/builder.go
lance: @matejvasek looking forward to a PR of that buildpack :smiley:

matejvasek: @lance  https://github.com/boson-project/buildpacks/pull/1

lance: Hahah - I shold have looked first :)

PR #21, opened 2020-06-16
Function Templates

### LINKED ISSUE #18: Replace appsody init
The `init` step currently utilizes the `appsody` binary and custom stacks to initialize new Service Functions.  This should be replaced by either built-in templates or a git-repo-based solution.  Architecture TBD.

lkingland:
Added is support for Function Templates

A base set of "core" templates are built into the binary, which refers to the function signature and contextual environment of a Function.  The default template is presently the CloudEvents listener context, which can be initialized via:
```
faas create $RUNTIME
```
Alternate templates are available by specifying the `--template` flag
```
faas create $RUNTIME --template $CONTEXT
```
The available templates are sourced from the directory `/templates` of this repository.

Support for extensible templates is provided by searching a config directory for additional templates.  This location is by default `$XDG_CONFIG_HOME/faas/templates` (usually `~/.config/faas/templates`) and can be explicitly configured with either the environment variable `$FAAS_TEMPLATES` or the `--templates` flag.

These extensible templates are namespaced unlike the core built-ins.  For instance, to include a hypothetical Boson Experimental templates repository which includes a JSON HTTP listener for Node, one would first install the templates with:
```
git clone git@github.com:boson-project/experimental ~/.config/faas/templates/experimental
```
This would result in something similar to the following structure:
```
~/.config/faas/templates
└── experimental
    └── node
        └── json
            ├── index.js
            ├── local.js
            ├── package.json
            ├── README.md
            └── test
                ├── integration.js
                └── unit.js
```
Which would then allow the use of the new template via:
```
faas create node --template experimental/json
```

Closes #18.


matejvasek:
I’m not sold on the `context` term, I like `template` more or maybe `archetype` (like in maven).

matejvasek:
I am afraid that the resource bundling is not honouring Unix access rights; e.g. mvn wrapper script is not executable.

lkingland:
> I’m not sold on the `context` term, I like `template` more or maybe `archetype` (like in maven).

Changed them back to `templates` for now, and we can float the `archetype` (or other) name to the team...

lkingland:
> I am afraid that the resource bundling is not honouring Unix access rights; e.g. mvn wrapper script is not executable.

The missing file modes are now retained; thanks for pointing this out.

cmd/create.go
```diff
@@ -24,25 +28,29 @@ func init() {
...
 
 // The create command invokes the Service Funciton Client to create a new,
 // functional, deployed service function with a noop implementation.  It
 // can be optionally created only locally (no deploy) using --local.
 var createCmd = &cobra.Command{
-	Use:        "create <language>",
-	Short:      "Create a Service Function",
-	SuggestFor: []string{"init", "new"},
+	Use:               "create <language>",
+	Short:             "Create a Service Function",
```
lance, line 41: Nit: could this just be "Create a Function"?

cmd/create.go
```diff
@@ -175,6 +199,7 @@ func gatherFromUser(config createConfig) (c createConfig, err error) {
 	config.Registry = prompt.ForString("Image registry", config.Registry)
 	config.Namespace = prompt.ForString("Namespace at registry", config.Namespace)
 	config.Language = prompt.ForString("Language of source", config.Language)
```
lance, line 201: I'm wondering if we need a term like "language runtime" since Quarkus is not a language in and of itself.

matejvasek:
@lkingland  When trying to compile I got:
```
# github.com/boson-project/faas/cmd
../create.go:31:44: undefined: embedded.DefaultContext
```
It's needed to change that to DefaultTemplate.

lkingland:
> @lkingland When trying to compile I got:
> 
> ```
> # github.com/boson-project/faas/cmd
> ../create.go:31:44: undefined: embedded.DefaultContext
> ```
> 
> It's needed to change that to DefaultTemplate.

Thanks @matejvasek I had a few commits unpushed.

cmd/create.go
lkingland: Thanks @lance ; If that seems more obvious to you, I can go with just "Functions" for now.  I will summarize the method to my madness in having called them thus and pop the question on Slack to see if there are others with an opinion.
Since your gut reaction was immediately that the term is redundant or unnecessary, I will go ahead and remove the prefix unless the team comes down squarely in favor of the full "Service Function".

cmd/create.go
lkingland: I like the idea of "Runtime" better than language, because the developer is specifying not just the language of the resultant function, but also the entire context within which it operates.  I will update all references to "runtime".  This works well with the switch back to "Template" from "Context" as well, which was half solving the problem. @lance

cmd/create.go
lkingland: After getting some feedback from the team as well, Functions works for now, and we can kick the question about whether the 'faas' term itself is applicable, down the proverbial road.

PR #24, opened 2020-07-01
fix: remove dependency on `kubectl` binary

### LINKED ISSUE #14: use k8s client for deploy
Replace the POC implementation of deploy which wraps kubectl with the use of a k8s client.

matejvasek:
Resolves #14


matejvasek:
@lkingland I decide to use `kn` here instead of `kubectl`.

PR #44, opened 2020-07-21
build: utilize docker latest tag for releases only

### LINKED ISSUE #43: Container latest tag
We currently us `tip` when tagging the container to ensure that directly pushing artifacts made during development do not become the latest default image pulled if running `docker run quay.io/boson/faas`.  We should, however, go ahead and set a `latest` tag such that a user can run this command and be assured they are pulling the latest released version.

lkingland:
This PR configures the system to update the container `latest` tag on release.

All image builds create a `latest` tag locally, but the specific make target `release` must be run to push it to the repository, such that we can continue to use the other targets for development builds (which use the `tip` tag).

This new target is invoked only by the release workflow, when there is a commit to `main` with a semver tag (release).

Closes #43

### DIFF
--- a/.github/workflows/releases.yaml
+++ b/.github/workflows/releases.yaml
@@ -23,7 +23,9 @@ jobs:
         env:
           USER: ${{ secrets.QUAY_USERNAME }}
           PASS: ${{ secrets.QUAY_PASSWORD }}
-        run: docker login -u "$USER" -p "$PASS" quay.io && make push
+        run: |
+          docker login -u "$USER" -p "$PASS" quay.io 
+          make push && make latest
       - name: Compress Release Binary
         run: gzip ./faas
       - name: Create Release
--- a/Makefile
+++ b/Makefile
@@ -17,20 +17,26 @@ test:
 	go test -cover -coverprofile=coverage.out ./...
 
 image: Dockerfile
-	docker build -t $(REPO):$(VERS) \
+	docker build -t $(REPO):latest  \
+	             -t $(REPO):$(VERS) \
 	             -t $(REPO):$(HASH) \
 	             -t $(REPO):$(DATE)-$(VERS)-$(HASH) .
-
-push: image
-	docker push $(REPO):$(VERS)
-	docker push $(REPO):$(HASH)
-	docker push $(REPO):$(DATE)-$(VERS)-$(HASH)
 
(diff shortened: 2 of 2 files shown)


lance:
I wonder if the Makefile target could be `latest` so that `make release VTAG=vx.x.x` in https://github.com/boson-project/faas/pull/45 doesn't conflict, since semantically I think the latter is more of an actual release.

Makefile
```diff
@@ -26,6 +27,11 @@ push: image
 	docker push $(REPO):$(HASH)
 	docker push $(REPO):$(DATE)-$(VERS)-$(HASH)
 
+release:
+	# push the local 'latest' tag as the new public latest version
+	# (run by CI only for releases)
+	docker push $(REPO):latest
```
matejvasek, line 33: Don't we also want to push `$(VERS)` tag?

.github/workflows/releases.yaml
```diff
@@ -23,7 +23,9 @@ jobs:
         env:
           USER: ${{ secrets.QUAY_USERNAME }}
           PASS: ${{ secrets.QUAY_PASSWORD }}
-        run: docker login -u "$USER" -p "$PASS" quay.io && make push
+        run: |
+          docker login -u "$USER" -p "$PASS" quay.io 
+          make push && make release
```
lkingland, line 28: Notice here @matejvasek that, when the push is for a versioned release, it pushes all tags and then additionally tags the :latest.

Makefile
lkingland: Yes, and that's a good question; I should clarify: there is a make target `make push` for pushing VERS, HASH, and teh verbose DATE-VERS-HASH, which can be invoked at any time, even from an unreleased, unversioned commit such as from a feature branch.  In this case VERS will just be `tip`.  However, when the GitHub action (`.github/workflows/releases.yaml`) is run, it will `make push` but, since it only matches versioned commits, it will _also_ run this release target, which essentially tags the already-created `:$(VERS)` image as `:latest`.

lkingland:
> I wonder if the Makefile target could be `latest` so that `make release VTAG=vx.x.x` in #45 doesn't conflict, since semantically I think the latter is more of an actual release.

This sounds like a good idea, and the change has been made.

PR #48, opened 2020-07-27
Adding multiple platforms to Makefile

markito:
Adding multiple platforms to the makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
 REPO := quay.io/boson/faas
 BIN  := faas
 
-WINDOWS=$(BIN)_windows_amd64.exe
-LINUX=$(BIN)_linux_amd64
 DARWIN=$(BIN)_darwin_amd64
+LINUX=$(BIN)_linux_amd64
+WINDOWS=$(BIN)_windows_amd64.exe
 
 CODE := $(shell find . -name '*.go')
 DATE := $(shell date -u +"%Y%m%dT%H%M%SZ")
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,17 +1,37 @@
...
-	go build -ldflags "-X main.date=$(DATE) -X main.vers=$(VERS) -X main.hash=$(HASH)" ./cmd/$(BIN)
+
+windows: $(WINDOWS) ## Build for Windows
+
+linux: $(LINUX) ## Build for Linux
+
+darwin: $(DARWIN) ## Build for Darwin (macOS)
+
+$(WINDOWS):
+	env GOOS=windows GOARCH=amd64 go build -i -v -o $(WINDOWS) -ldflags "-X main.date=$(DATE) -X main.vers=$(VERS) -X main.hash=$(HASH)" ./cmd/$(BIN)
```
matejvasek, line 25: This `install` flag is causing problems for me. It tries to install something to /usr where normal user cannot write. @lkingland can you please verify it's not working for you?

matejvasek: Otherwise lgtm.

lkingland, line 25: Removing the `-i` also fixed the build for me.

Makefile
```diff
@@ -1,17 +1,37 @@
...
+
+$(LINUX):
+	env GOOS=linux GOARCH=amd64 go build -i -v -o $(LINUX) -ldflags "-X main.date=$(DATE) -X main.vers=$(VERS) -X main.hash=$(HASH)" ./cmd/$(BIN)
+
+$(DARWIN):
+	env GOOS=darwin GOARCH=amd64 go build -i -v -o $(DARWIN) -ldflags "-X main.date=$(DATE) -X main.vers=$(VERS) -X main.hash=$(HASH)" ./cmd/$(BIN)
+
+# $(BIN): $(CODE)
+# 	go build -ldflags "-X main.date=$(DATE) -X main.vers=$(VERS) -X main.hash=$(HASH)" ./cmd/$(BIN)
```
lkingland, line 34: I would go ahead and leave this one in place, such that defaults apply for developers running local `make` on different combinations.

Makefile
```diff
@@ -1,17 +1,37 @@
 REPO := quay.io/boson/faas
 BIN  := faas
 
+WINDOWS=$(BIN)_windows_amd64.exe
+LINUX=$(BIN)_linux_amd64
+DARWIN=$(BIN)_darwin_amd64
```
lkingland, line 6: While this may sound pedantic, I would consider putting these in alphabetical order to avoid any misunderstanding that we have a preference of one platform over another.

Makefile
```diff
@@ -1,17 +1,37 @@
...
+
 CODE := $(shell find . -name '*.go')
 DATE := $(shell date -u +"%Y%m%dT%H%M%SZ")
 HASH := $(shell git rev-parse --short HEAD 2>/dev/null)
 VTAG := $(shell git tag --points-at HEAD)
 VERS := $(shell [ -z $(VTAG) ] && echo 'tip' || echo $(VTAG) )
 
-all: $(BIN)
+all: $(WINDOWS) $(LINUX) $(DARWIN)
```
lkingland, line 14: This new flow always explicitly sets the OS and Architecture, which means someone checking out the repository from a different combination (such as on ARM)) can no longer run `make` as is instructed in the documentation.

I would perhaps consider leaving the current default make target in place, and instead creating a _new_ `make dist` target which has as its contents these three build commands.  Ideally, it would also create the directory `./dist`, and build the binaries into that directory, and remove the directory entirely on `make clean`.

This minor change would also allow for us to avoid having to update the `Dockerfile` and `.dockerignore`.

markito: That's a good idea, but since the current flow already left the faas binary on the main folder, can I consider this a future enhancement ?

markito: I'll leave the default as Linux (as before) but create a new target that is cross-platform.

lkingland:
Thanks for the PR William!  LGTM

lkingland:
@markito Lance enabled fast-forward only mode on the main branch, so I rebased your branch and merged your commits directly into the develop branch.

PR #49, opened 2020-07-28
DO NOT MERGE YET: testing CI

matejvasek:


matejvasek:
@lkingland trying to run CI on multiple OSes.

matejvasek:
@lkingland  there are few tests failing because of file mods (access perms.); I am not sure how we could fix this.

function_unit_test.go
```diff
@@ -23,50 +23,50 @@ func TestPathToDomain(t *testing.T) {
 		{"", noLimit, ""},
 
 		// trailing slashes ignored
-		{"/home/user/example.com/admin/", noLimit, "admin.example.com"},
+		{"/home" + string(os.PathSeparator) + "user" + string(os.PathSeparator) + "example.com" + string(os.PathSeparator) + "admin/", noLimit, "admin.example.com"},
```
lkingland, line 26: This might be cleaner as `filepath.Join(...)`

.github/workflows/pull_requests.yaml
```diff
@@ -5,7 +5,10 @@ on: [pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: ['ubuntu-latest', 'windows-latest', 'macos-latest']
+    runs-on: ${{ matrix.os }}
```
lkingland, line 11: I love that we are dipping our toes in here, but I would caution that we don't spend too much time on cross-compatibility until we're past v1, or we risk spending a lot of time in maintenance of platforms not in active use.

lkingland:
> @lkingland there are few tests failing because of file mods (access perms.); I am not sure how we could fix this.

Sounds like a bug in pkger to me.  I created a new issue #57

.github/workflows/pull_requests.yaml
lance: Although I would say that OSX and Linux are must-have...

lance:
> @lkingland there are few tests failing because of file mods (access perms.); I am not sure how we could fix this.

I think [this test](https://github.com/boson-project/faas/blob/develop/function_unit_test.go#L27-L70) could be changed to be OS agnostic. And [this test](https://github.com/boson-project/faas/blob/develop/embedded/initializer_test.go#L122-L124) could check `runtime.GOOS` for `"windows"` and have a special test just for that platform.

lance:
@matejvasek this PR is now pointing to the incorrect main branch (`develop`) and CI has changed so much that I think this should be closed and reconsidered for a future PR.

PR #54, opened 2020-07-29
CI lint

matejvasek:


Makefile
```diff
@@ -38,6 +58,9 @@ latest:
 	# (run by CI only for releases)
 	docker push $(REPO):latest
 
+check:
+	golangci-lint run --enable=unconvert,prealloc,bodyclose
```
lkingland, line 62: What do you think about taking the installation command from the workflows .yaml files, and instead add a make target for `make bin/golangci-lint` which runs the curl and install task.  This way one can run `make check` themselves without needing to have the linter pre-installed.

.github/workflows/develop.yaml
```diff
@@ -23,3 +23,7 @@ jobs:
         run:  make test
       - name: Build
         run:  make build
+      - name: Lint
+        run: |
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.27.0
+          PATH=$PATH:$(go env GOPATH)/bin make check
```
lkingland, line 29: These could be just `make check` if the addition is made to the Makefile detailed below.

Makefile
```diff
@@ -38,6 +58,9 @@ latest:
 	# (run by CI only for releases)
 	docker push $(REPO):latest
 
+check:
```
lkingland, line 61: This could be changed to include the code necessary to install.  Something like
```
bin/golangci-lint:
     curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ./bin v1.27.0

check: bin/golangci-lint
    ./bin/golangci-lint run --enable=unconvert,prealloc,bodyclose
```
This would enable `make check` to be run by a developer (on a linux host at least).

Makefile
lkingland: The other option would be to go full-on workflows integration, letting the developer be in the dark about lint errors until they push (or set up and run the golangci-lint action locally themselves)
https://github.com/golangci/golangci-lint-action

This is a nice integration, but as a developer I would prefer the ability to run `make check` myself locally before submitting the PR.

PR #56, opened 2020-07-29
chore: add `-race` flag for tests

matejvasek:


matejvasek:
The cli code is pretty much sequential, but still...

PR #65, opened 2020-08-07
feat: add init/build/deploy commands and customizable namespace

lance:
This commit comprises some fairly large changes in the codebase.
The 'create' command has been extracted into 'init', 'bulid' and
'deploy' commands. The 'create' command remains, but now delegates
most of its work to these other three. This also has resulted in
some rework of the various flags. One example is the removal of
the '--local' flag from create, since this can be achieved now by
just running 'faas init'.

There are still a number of changes needed for this to land.

* I've removed some of the 'internal' stub code
* The environment variables have been removed (for the most part)
  from --help text and should be returned
* ~~The progress meter is a little wonky now~~
* The service address is incorrect when printed (EDIT: just confirmed that this was the case before the PR as well)
* ~~There are no tests, and I'm pretty sure I've broken what tests already existed~~

@lkingland @matejvasek I welcome any comments, suggestions, modifications, etc.

It's a start - a work in progress.


lance:
Review requested - but note that it is still a WIP

lance:
One thing that seems a little inconsistent is how configuration parameters are dealt with. Specifically, I mean for example, something like `Deployer`. The struct is created with some values such as `Namespace` and `Verbose`. But the `Deploy()` method accepts `name` and `image`. I think either they should all be parameters to `Deploy()` or all a part of the struct.

cmd/build.go
```diff
@@ -0,0 +1,127 @@
...
+func init() {
+	cwd, err := os.Getwd()
+	if err != nil {
+		panic(err)
+	}
+
+	root.AddCommand(buildCmd)
+	buildCmd.Flags().StringP("path", "p", cwd, "Path to the function project directory")
+	buildCmd.Flags().StringP("tag", "t", "", "Specify an image tag, for example quay.io/myrepo/project.name:latest")
+	buildCmd.Flags().BoolP("push", "u", false, "Push the image to a registry")
```
lkingland, line 42: Should pushing be part of the Deploy step?

cmd/build.go
```diff
@@ -0,0 +1,127 @@
...
+		client, err = faas.New(
+			faas.WithVerbose(config.Verbose),
+			faas.WithBuilder(builder),
+			faas.WithPusher(docker.NewPusher(f.Tag)),
+		)
+	} else {
+		client, err = faas.New(
+			faas.WithVerbose(config.Verbose),
+			faas.WithBuilder(builder),
+		)
```
lkingland, line 119: Something which might simplify this a bit:  the logic of "to push or not to push" could be handled by combining the logic with the deployer. i.e. Deploying implicitly deploys something that has been locally built to a container registry and then creates (or updates) a KN Service with the new version.  Thus pushing is integral to the deployment process.  If deployment includes push, this if statement could be removed along with the associated `--push` flag.

lance:
@lkingland all tests and lint checks are passing. PTAL

lkingland:
Looking great @lance, glad that snapped into place. Doing a bit more in depth looksee now...

client_test.go
```diff
@@ -170,20 +170,16 @@ func TestCreateDelegates(t *testing.T) {
...
-	// function code.  For this test, it is a name derived from the test path.
+	// The builder should be invoked with a path to a function project's source
 	// An example image name is returned.
-	builder.BuildFn = func(name2, path2 string) (string, error) {
-		if name != name2 {
-			t.Fatalf("builder expected name %v, got '%v'", name, name2)
-		}
+	builder.BuildFn = func(name2 string) (string, error) {
 		expectedPath, err := filepath.Abs(root)
+		if expectedPath != name2 {
```
lkingland, line 177: stating `expectedX` is much clearer.

client_test.go
```diff
@@ -416,10 +412,11 @@ func TestUpdate(t *testing.T) {
...
 	// An example image name is returned.
-	builder.BuildFn = func(name2, path2 string) (string, error) {
-		if name != name2 {
-			t.Fatalf("builder expected name %v, got '%v'", name, name2)
-		}
+	// TODO: FIX ME UNDERSTAND ME
+	builder.BuildFn = func(tag string) (string, error) {
+		// if name != name2 {
+		// 	t.Fatalf("builder expected name %v, got '%v'", name, name2)
+		// }
```
lkingland, line 419: should this perhaps be `if tag != expectedTag`?  It's a simple logical passthrough at this point, so the test doesn't add a whole lot of value, merely level-setting for ensuring that future complexity doesn't break the base case, but it may be worthwhile to keep plumbed.

cmd/build.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/boson-project/faas/buildpacks"
+	"github.com/boson-project/faas/docker"
+	"github.com/ory/viper"
+	"github.com/spf13/cobra"
+)
+
+var buildCmd = &cobra.Command{
+	Use:        "build",
+	Short:      "Build an existing function project as an OCI image",
+	SuggestFor: []string{"biuld", "buidl", "built", "image"},
```
lkingland, line 17: AH, adding common misspelling is handy.  It enables sloppy scripting, but probably worth it to the interactive users!

cmd/build.go
```diff
@@ -0,0 +1,125 @@
...
+	// TODO: Add completions for build
+	// ValidArgsFunction: CompleteRuntimeList,
+	RunE: buildImage,
+	PreRunE: func(cmd *cobra.Command, args []string) (err error) {
+		flags := []string{"path", "tag", "push"}
+		for _, f := range flags {
+			err = viper.BindPFlag(f, cmd.Flags().Lookup(f))
+			if err != nil {
+				return err
+			}
```
lkingland, line 27: Ah; nice helper loop for that somewhat ugly BindPFlag call.
While this is a completely stylistic choice, these lines can also be written:
```
	if err = viper.BindPFlag(f, cmd.Flags().Lookup(f)); err != nil {
		return err
	}
```
... which saves a line at the expense of some slightly less readable ternary-ish syntactic sugar.

cmd/build.go
```diff
@@ -0,0 +1,125 @@
...
+		Tag:     viper.GetString("tag"),
+	}
+	return Build(config)
+}
+
+// FunctionConfigForBuild accepts a buildConfig and
+// uses this to load a faas.Function. If the config contains
+// a value for the image tag, this will be used and
+// subsequently written to the config file.
+func FunctionConfigForBuild(config buildConfig) (f *faas.Function, err error) {
```
lkingland, line 64: There's some logic in here in `FunctionConfigForBuild` and `Build` that is not really CLI-specific
- If Name and Runtime are not provided, generate an error
- If a Tag is not provided, use the tag from the config file in the function root
- Tags default to quay.io/%s:latest
- Update the Function config file on disk at function root
- Optionally push on Build

These logic bits are simple, yes, but they are applicable to a developer using the client library directly or the CLI.  So perhaps this is a job for a future refactor, but it would be nice to see these moved into the Client Library, and the API used here in the CLI kept to a very simple "map in booleans and strings, then invoke Build".  This would both remove the need to do a conditional invocation of the client library (one with and one without a Pusher), as well as opens us up to writing tests that validate expected functionality such as:
"Test that Name and Runtime are required or error"
"Test that Tag defaults to DefaultTag (quay.io)"
"Test that a Builder respects the Push boolean configuration"
etc.

buildpacks/builder.go
```diff
@@ -28,14 +28,17 @@ var runtime2pack = map[string]string{
 	"go":      "quay.io/boson/faas-go-builder",
 }
 
-func (builder *Builder) Build(name, runtime, path string) (image string, err error) {
+func (builder *Builder) Build(path string) (image string, err error) {
```
matejvasek, line 31: unused named param `image`

matejvasek: also `err` is returned explicitly

client.go
```diff
@@ -255,79 +254,114 @@ func WithDomainSearchLimit(limit int) Option {
...
+
+func (c *Client) Build(path string) (image string, err error) {
+	image, err = c.builder.Build(path)
+	return image, err
+}
+
+func (c *Client) Deploy(name, tag string) (address string, err error) {
+	err = c.pusher.Push(tag) // First push the image to an image registry
+	if err != nil {
+		return "", err
```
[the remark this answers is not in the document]
lance: This explains so much! I was wondering why there was so much code with just `return` but the signature specified explicit values. :+1:

client.go
```diff
@@ -255,79 +254,114 @@ func WithDomainSearchLimit(limit int) Option {
...
+func (c *Client) Route(name, address string) (route string) {
+	// Ensure that the allocated final address is enabled with the
+	// configured DNS provider.
+	// NOTE:
+	// DNS and TLS are provisioned by Knative Serving + cert-manager,
+	// but DNS subdomain CNAME to the Kourier Load Balancer is
+	// still manual, and the initial cluster config to suppot the TLD
+	// is still manual.
+	c.dnsProvider.Provide(name, address)
+	return name
```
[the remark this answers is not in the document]
lance: This changes the `Provide()` interface to return a string. But I see your point.

client_test.go
lance: Ack - I just missed this when going through the tests.

buildpacks/builder.go
lance: I didn't quite understand how those return statements were working/compiling. This clears it up. TY.

cmd/build.go
lance: Excellent points. I have moved it in to function.go and renamed it to `faas.FunctionConfiguration()`. I think it might benefit from adding a `name` parameter as well in order to update the function name. But to be honest, it's a little clumsy right now and I'm not sure if this is the best way to approach it all in the long run.

lance:
@matejvasek @lkingland your comments are incorporated, and I've removed the WIP status. PTAL and if it looks ok, let's go ahead and merge so that Naina can start experimenting with it.

function.go
```diff
@@ -67,29 +68,72 @@ func (f *Function) Initialize(runtime, context, name string, domainSearchLimit i
...
+	if f.Name == "" || f.Runtime == "" {
+		return nil, fmt.Errorf("Unable to find a function project at %s", path)
+	}
+
+	// Allow caller to override pre-configured t name
+	if tag != "" {
+		f.Tag = tag
+	}
+	if f.Tag == "" {
+		f.Tag = fmt.Sprintf("quay.io/%s:latest", f.Name)
```
matejvasek, line 126: Possible future improvement: I think we could pick better default by looking into docker config file like 
 the `CompleteRegistryList` function does.

cmd/build.go
lkingland: We probably have several solid refactors ahead of us where we can make it less clumsy!

PR #66, opened 2020-08-10
chore: build static binary

matejvasek:


matejvasek:
@lance surprisingly enough the size is even smaller!

Issue #37, opened 2020-07-14
CLI JSON output


lkingland:
Some of the commands support specifying JSON output; it would be nice to have all commands support this flag.

lance:
I see that `describe` has YAML/JSON/XML output options, but it seems that may be the only command that does. How do you see this working for other commands? 

*  `create` optionally spits out YAML for the kn `Service`
* `delete` not sure about this one
* `list` YAML of the deployed `Service`(s)
* `run` ??
* `update` spits out YAML for the updated `Service`

Am I missing anything?

lkingland:
As for the other commands, I was hoping we could create a flag which is short-hand for saying "human readable or machine readable".  In the case of `describe`, the YAML and XML options of course make sense.  Perhaps for the others only JSON does (essentially requesting structured logging)

`run` is a little bit of a special case, as it delegates...   We could respect this flag in faas itself, perhaps setting a well-known environment variable which can optionally be respected by the Function author to choose the way they treat STDOUT and STDERR?

matejvasek:
@lkingland `list` currently returns only names. If we wanted return more, what fields are we interested in?

lkingland:
It's my opinion that we should be seeking the simplest possible functional case at this time, and only add more as specific use-cases require them.  The minimum necessary information to subscribe to an event is its unique identifier, thus the minimum necessary information back out of the system would be those same unique identifiers returned.   If we want to add more information; perhaps augmented with fields that are platform-specific, I would see that as a GET request for more information about the subscription of that ID.
so POSTing a name to `localhost:1111/subscriptions` would append the name to the list.  The list would then be returned by GET, with additional information below the named event.  Something like:
```
GET localhost:1111/subscriptions
[ "newsletter-signups" ]
```
```
GET localhost:1111/subscriptions/newsletter-signups
{ "name": "newsletter-signups",
.... // Other fields lazily added here _only as required_ to fulfill specifically requested use-cases
}
```
By keeping ourselves so simple as to be nearly an oversimplification, I am hoping we can prevent from painting ourselves into a corner unnecessarily.
I may indeed be missing something here, so please let me know if there are other fields than the event's name which are required to implement the basic pub/sub flow!

matejvasek:
@lkingland  by `list` here I meant list of deployed functions/ksvcs not subscriptions.

lkingland:
[...]

PR #67, opened 2020-08-10
feat: JSON output for the `list` sub-command

matejvasek:


matejvasek:
relates to #37

cmd/list.go
```diff
@@ -21,8 +42,45 @@ var listCmd = &cobra.Command{
...
+func fmtYAML(writer io.Writer, names []string) error {
+	encoder := yaml.NewEncoder(writer)
+	return encoder.Encode(names)
+}
+
+var formats = map[string]fmtFn{
+	"plain": fmtPlain,
+	"json":  fmtJSON,
+	"yaml":  fmtYAML,
+}
```
lkingland, line 71: Nitpick, but can we move this up top?  It threw me for a loop for a minute wondering where `formats` came from.

PR #68, opened 2020-08-12
src(templates/node): update Node.js templates

lance:
This commit updates the Node.js templates to use the latest faas-js-runtime
framework, and fixes a couple of dependency issues and broken tests. I have
also added a build time test for these templates so that if/when they change
again, we can be sure to catch any problems.

Fixes: https://github.com/boson-project/faas/issues/63


Makefile
```diff
@@ -34,9 +34,15 @@ $(LINUX):
 $(WINDOWS):
 	env CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build -o $(WINDOWS) -ldflags "-X main.date=$(DATE) -X main.vers=$(VERS) -X main.hash=$(HASH)" ./cmd/$(BIN)
 
-test:
+test: test-binary test-node
+
+test-binary:
 	go test -race -cover -coverprofile=coverage.out ./...
 
+test-node:
```
matejvasek, line 42: Nice. We should add something similar for Quarkus and Go too.

lance: For sure. I just added https://github.com/boson-project/faas/issues/70

lkingland: Definitely nice to have the templates included in the testing pass

PR #72, opened 2020-08-17
test: add test targets for go and quarkus templates

lance:
This commit adds Makefile targets for go and quarkus, and includes these in the regular `make test` target. I've also added a couple of entries to .gitignore.

Fixes: https://github.com/boson-project/faas/issues/70


lkingland (approved):
Looks good to me.  Glad to see we will be testing the templates too.

PR #74, opened 2020-08-19
fix: remove references to unused binaries appsody, kn, kubectl

### LINKED ISSUE #69: Remove appsody from source
We are no longer using appsody - so all of this code should probably be removed from the repo

lkingland:
Also exports the buildpack runtimes map as `Runtimes` in order to be used in the shell completion utilities.  And creates a stubbed Docker runner in place of that provided by appsody.  (Resolves #69)


docker/runner.go
```diff
@@ -7,36 +7,36 @@ import (
...
 	}
 
-	// Extra arguments to appsody
+	// Extra arguments to docker
 	args := []string{"run"}
 
 	// If verbosity is enabled, pass along as an environment variable to the function.
 	if n.Verbose {
-		args = append(args, []string{"--docker-options", "-e VERBOSE=true"}...)
+		args = append(args, []string{"-e VERBOSE=true"}...)
```
lance, line 35: Should this also have `-rm -it -p 8080:8080`?

lance:
@lkingland ping - looks like this needs to be rebased.

lkingland:
Thanks Lance, I'll be reading and merging this (and the other PR in the same situation) shortly. @lance

docker/runner.go
lkingland: Yes.  I believe the interactive flag -i won't be allowed, but the others yes.  This has been updated in the latest version.

Issue #73, opened 2020-08-17
`faas list` does not seem to work


lance:
```
 pwd                                                                                                                          Mon 17 Aug 2020 03:04:34 PM EDT
/home/lanceball/src/go/src/github.com/boson-project/faas/foo

faas/foo on  lance/9-add-path-flag-everywhere [$!?] is 📦 v0.1.0 via ⬢ v14.8.0 
❯ ls -a                                                                                                                        Mon 17 Aug 2020 03:04:34 PM EDT
  .    ..    .faas.yaml    index.js    local.js    package.json    README.md    test

faas/foo on  lance/9-add-path-flag-everywhere [$!?] is 📦 v0.1.0 via ⬢ v14.8.0 
❯ cat .faas.yaml                                                                                                               Mon 17 Aug 2020 03:04:38 PM EDT
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: .faas.yaml
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ name: foo.faas.boson-project.github.com
   2   │ runtime: node
   3   │ tag: quay.io/lanceball/foo-func
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

faas/foo on  lance/9-add-path-flag-everywhere [$!?] is 📦 v0.1.0 via ⬢ v14.8.0 
❯ kn service list                                                                                                      152ms  Mon 17 Aug 2020 03:04:42 PM EDT
NAME                                 URL                                                                  LATEST                                       AGE    CONDITIONS   READY   REASON
foo-faas-boson--project-github-com   http://foo-faas-boson--project-github-com.default.127.0.0.1.nip.io   foo-faas-boson--project-github-com-jbghp-1   7d1h   3 OK / 3     True    

faas/foo on  lance/9-add-path-flag-everywhere [$!?] is 📦 v0.1.0 via ⬢ v14.8.0 
❯ ../faas list -v                                                                                                      142ms  Mon 17 Aug 2020 03:04:51 PM EDT

faas/foo on  lance/9-add-path-flag-everywhere [$!?] is 📦 v0.1.0 via ⬢ v14.8.0 
```

lance:
Update: it's listing only functions deployed to the `faas` namespace.

lkingland:
[...]

PR #78, opened 2020-08-24
The `list` sub-command uses namespace.

matejvasek:
related to #73


lkingland (approved):
LGTM, thanks @matejvasek 

The latest pr #79 makes something similar to this change.  I believe they are logically equivalent but worth a double-check.

Issue #87, opened 2020-08-31
The `--image` flag doesn't work for `create`.


matejvasek:
Call to `func overrideImage(root, override string) (err error)` creates `.faas.yaml`. This causes that subsequent call to `func (c *Client) Create(cfg Function) (err error)` fails.

PR #88, opened 2020-08-31
fix: image override

matejvasek:
resolves: #87

### DIFF
--- a/client.go
+++ b/client.go
@@ -319,6 +319,8 @@ func (c *Client) Initialize(cfg Function) (err error) {
 		return
 	}
 
+	f.Image = cfg.Image
+
 	// Set the name to that provided, defaulting to path derivation if empty.
 	f.Name = cfg.Name
 	if cfg.Name == "" {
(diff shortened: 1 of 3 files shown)


lance:
@matejvasek can you help me understand why the call to `overrideImage()` is causing a failure in `Client.Create()`? Glancing through the code, it seems this should work so it's not obvious to me why there is a failure in the first place.

matejvasek:
> @matejvasek can you help me understand why the call to `overrideImage()` is causing a failure in `Client.Create()`? Glancing through the code, it seems this should work so it's not obvious to me why there is a failure in the first place.

The `overrideImage()` functions creates the config file, but the `Create()` expects the directory to be empty. So maybe another possible fix would be switch order in which they are called. See linked issue.

lance:
OK - I see now. When `Create()` calls `c.Initialize()` the `Initialize()` function calls `assertEmptyRoot()`. I don't really think switching the order would work, because where would you end up calling `overrideImage()`? This change makes more sense IMO.

PR #98, opened 2020-09-02
templates: update faas-js-runtime dependency to 0.2.2

lance:
This makes me think we need a better solution here. The version for
that dependency should not be embedded in the faas binary because in so
doing, we have to (during the early churn) release faas every time there
is a new release of faas-js-runtime. The only reason that module is
included in the template is for running locally outside of a container.

It would be nice instead to enable `npm run local` by having the local
target execute a binary. For example:

```
npx faas-js-runtime ./index.js
```

This would simply be adding a `./bin/faas-js-runtime` to the module.
The binary executable would then just load the function at the path
provided.


lance:
@matejvasek @lkingland can I get a LGTM on this?

matejvasek (approved):

lance:
Closing this in favor of https://github.com/boson-project/faas/pull/99/files

PR #99, opened 2020-09-03
templates(node): make node templates use npx @redhat/faas-js-runtime

lance:
Currently the templates have a `local.js` file and an explicit dependency
on the @redhat/faas-js-runtime module. This removes that build-time
dependency and replaces it with a runtime dependency by using the (pending)
CLI from the module.

This is nice for a couple of reasons.

- Reduces the build time during `faas deploy`
- Eliminates the need for `faas` to bump with updates to faas-js-runtime

See: https://github.com/boson-project/faas-js-runtime/pull/53

~~Marking as WIP until that ships.~~


lance:
@boson-project/core removed the WIP marker. This is ready for review.

lance (commented):
Self approval is important in these difficult times.

PR #100, opened 2020-09-03
chore(actions): add binary uploads to main branch CI

lance:
This will enable non-developers who are interested in trying out the latest builds to easily get a copy pre-release.

image

Once the action has completed, you can click the  Artifacts drop down to pull down a zip archive containing binaries for all three platforms. An optimization would be to have each of the binaries uploaded independently, but I didn't see how to do that in the `upload-artifact` action.

This is a PR intended to land on `main`, so we can start pointing potential testers and fiddlers to https://github.com/lance/faas/actions?query=workflow%3AMain where they can download the latest snapshot.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -8,7 +8,7 @@
 # have unique names:  'Main', 'Develop'
 name: Main
 
-on: 
+on:
   push:
     branches:
       - main
@@ -22,6 +22,15 @@ jobs:
       - name: Test
         run:  make test
       - name: Build
-        run:  make build
+        run:  make cross-platform
       - name: Lint
         run: make check
+      - name: Permissions
+        run: chmod a+x faas_*amd*
+      - uses: actions/upload-artifact@v2
+        with:
+          name: Binaries
+          path: |
+            faas_darwin_amd64
+            faas_linux_amd64
+            faas_windows_amd64.exe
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -22,6 +22,15 @@ jobs:
...
         run: make check
+      - name: Permissions
+        run: chmod a+x faas_*amd*
+      - uses: actions/upload-artifact@v2
+        with:
+          name: Binaries
+          path: |
+            faas_darwin_amd64
+            faas_linux_amd64
+            faas_windows_amd64.exe
```
lkingland, line 36: Regarding your question about separate downloads; from what I can see from how actions are used in other places, the way to upload each binary individually would be to repeat this stanza three times.  Each named something like "Windows Pre-release Binary"... or something like that.

lkingland (approved):
What a nice addition!

.github/workflows/main.yaml
lance: Testing this now

lance: You are correct!

image

lkingland:
Per our discussion about how the `main` branch is really only used for releases anyway, these directives may be very useful on the `develop.yaml` as that branch houses the pre-release code.
This is much cleaner than the way the releases currently deploy assets, I wonder if this would be a better way to implement that workflow as well.

lance:
Closing this in favor of landing it on `develop` branch.

PR #109, opened 2020-09-09
fix: update pkger generated files

matejvasek:
When changing templates I forgot to commit generated file.


lkingland (approved):
LGTM!  This is tricky to remember, but adding it to automation feels like it may backfire.

lance:
I forget this a lot. It would be nice for this to be automated as part of the build, and not included in the source repo - just want to be careful about making the build take too long.

lance (approved):

PR #110, opened 2020-09-09
fix: delete command

matejvasek:
There was namespace issue, there was hard-coded `faas` namespace.


matejvasek:
@lholmquist this is why you have seen problems with `faas delete`

lance (approved):

PR #115, opened 2020-09-10
fix: describe works without Kn Eventing and delete command 

lance:
This PR includes the two fix commits from @matejvasek that landed on `develop` but not `main` prior to https://github.com/boson-project/faas/commit/8a60c5e0c44d28d2ff085e56299217e05e408df8

### DIFF
--- a/client.go
+++ b/client.go
@@ -8,7 +8,6 @@ import (
 )
 
 const (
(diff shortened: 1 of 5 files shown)


lkingland (approved):

matejvasek:
Looks good.

PR #118, opened 2020-09-11
feat: default to no confirmation prompts for CLI commands

lance:
The CLI commands all printed confirmation prompts for the various flags
they exposed. This commit modifies that logic, so that there is no longer
a `-y` flag, but instead a `--confirm` or `-c` flag for each command, and
prompts are only displayed if using this flag. In most cases, the derived
values are printed even if not prompted for.

In all cases where the user is prompted, I have removed the "Verbose"
prompt, as that seems less like a configuration option that needs to be
confirmed, and more like just a CLI option for the current run which we
can just accept as-is.

The text for the prompts has also been reduced to one or two words.

Also added are some checks around image naming and repositories, short
circuiting failures that could occur if these are not specified or are
unknown. For example, if a user does `faas init` and then `faas deploy`
we don't yet know what the image name should be - one hasn't been built.

Fixes: https://github.com/boson-project/faas/issues/91
Fixes: https://github.com/boson-project/faas/issues/90
Fixes: https://github.com/boson-project/faas/issues/89

### DIFF
--- a/client.go
+++ b/client.go
@@ -369,7 +369,7 @@ func (c *Client) Build(path string) (err error) {
 		return
 	}
 
-	// Derive Image from the path (preceidence is given to extant config)
+	// Derive Image from the path (precedence is given to extant config)
 	if f.Image, err = DerivedImage(path, c.repository); err != nil {
 		return
 	}
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/ory/viper"
 	"github.com/spf13/cobra"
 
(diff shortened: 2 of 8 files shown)


matejvasek:
@lance is envvar FAAS_CONFIRM really used anywhere?

lance:
> @lance is envvar FAAS_CONFIRM really used anywhere?

Well, not explicitly. But neither was `FAAS_YES` :laughing:. However, you could set it in your environment to have all commands generate prompts.

```console
faas on  lance/reverse-confirm-logic [$?] via 🐹 v1.14.4 took 3s 
❯ set -x FAAS_CONFIRM true                                                                                            3360ms  Fri 11 Sep 2020 10:07:17 AM EDT

faas on  lance/reverse-confirm-logic [$?] via 🐹 v1.14.4 
❯ mkdir test-fn                                                                                                                Fri 11 Sep 2020 01:21:59 PM EDT

faas on  lance/reverse-confirm-logic [$?] via 🐹 v1.14.4 
❯ ./faas init -p ./test-fn -l node                                                                                             Fri 11 Sep 2020 01:22:09 PM EDT
Project path (./test-fn): 
Project name (test-fn.faas.boson-project.github.com): 
Runtime (node): 
Trigger (http): 

faas on  lance/reverse-confirm-logic [$?] via 🐹 v1.14.4 took 3s 
❯                                                                                                                     3545ms  Fri 11 Sep 2020 01:22:28 PM EDT

```

PR #125, opened 2020-09-16
src(commands): remove --path flag from most commands

lance:
This commit removes the `--path -p` flag for the build, delete, deploy,
describe, run and update commands. The init and create commands have been
left alone for now, pending the outcome of https://github.com/boson-project/faas/issues/123.
The list command doesn't need to know a path, because it's about seeing
all deployed Functions.

Adds a `faas.LoadFunction(path string)` function to load an existing Function
project from disk and return an error if one does not already exist.

Note: I realize this might be controversial. Speak up if you disagree with this proposal!

Edit: The idea here is to enforce the notion that all commands (other than `init` and `create`) are run from within a Function project directory where the configuration file `.faas.yaml` defines the context. Some of the comments from @matejvasek identified areas where functionality has been reduced - e.g. you can't describe or delete a function by name. I think we should actually consider removing these, and possibly even consider removing `faas deploy` and `faas update`. 

BREAKING CHANGE

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -12,27 +12,28 @@ import (
 func init() {
 	root.AddCommand(deleteCmd)
 	deleteCmd.Flags().BoolP("confirm", "c", false, "Prompt to confirm all configuration options - $FAAS_CONFIRM")
-	deleteCmd.Flags().StringP("path", "p", cwd(), "Path to the project which should be deleted - $FAAS_PATH")
 	deleteCmd.Flags().StringP("namespace", "n", "", "Override namespace in which to search for Functions.  Default is to use currently active underlying platform setting - $FAAS_NAMESPACE")
 }
 
 var deleteCmd = &cobra.Command{
-	Use:               "delete <name>",
+	Use:               "delete",
 	Short:             "Delete a Function deployment",
-	Long:              `Removes the deployed Function by name, by explicit path, or by default for the current directory.  No local files are deleted.`,
+	Long:              `Removes the deployed Function in the current project directory.  No local files are deleted.`,
 	SuggestFor:        []string{"remove", "rm", "del"},
 	ValidArgsFunction: CompleteFunctionList,
-	PreRunE:           bindEnv("path", "confirm", "namespace"),
+	PreRunE:           bindEnv("confirm", "namespace"),
 	RunE:              runDelete,
 }
 
 func runDelete(cmd *cobra.Command, args []string) (err error) {
 	config := newDeleteConfig(args).Prompt()
+	function, err := faas.LoadFunction(config.Path)
+	if err != nil {
+		return
+	}
 
 	remover := knative.NewRemover(config.Namespace)
 	remover.Verbose = config.Verbose
-
-	function := faas.Function{Root: config.Path, Name: config.Name}
 
 	client := faas.New(
 		faas.WithVerbose(config.Verbose),
@@ -56,7 +57,7 @@ func newDeleteConfig(args []string) deleteConfig {
 		name = args[0]
 	}
 	return deleteConfig{
-		Path:      viper.GetString("path"),
+		Path:      cwd(),
 		Namespace: viper.GetString("namespace"),
 		Name:      deriveName(name, viper.GetString("path")), // args[0] or derived
 		Verbose:   viper.GetBool("verbose"),                  // defined on root
--- a/cmd/describe.go
+++ b/cmd/describe.go
@@ -19,7 +19,6 @@ func init() {
 	root.AddCommand(describeCmd)
 	describeCmd.Flags().StringP("namespace", "n", "", "Override namespace in which to search for the Function.  Default is to use currently active underlying platform setting - $FAAS_NAMESPACE")
 	describeCmd.Flags().StringP("format", "f", "human", "optionally specify output format (human|plain|json|xml|yaml) $FAAS_FORMAT")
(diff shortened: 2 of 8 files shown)


cmd/delete.go
```diff
@@ -12,28 +12,29 @@ import (
 func init() {
 	root.AddCommand(deleteCmd)
 	deleteCmd.Flags().BoolP("confirm", "c", false, "Prompt to confirm all configuration options - $FAAS_CONFIRM")
-	deleteCmd.Flags().StringP("path", "p", cwd(), "Path to the project which should be deleted - $FAAS_PATH")
 	deleteCmd.Flags().StringP("namespace", "n", "", "Override namespace in which to search for Functions.  Default is to use currently active underlying platform setting - $FAAS_NAMESPACE")
 }
 
 var deleteCmd = &cobra.Command{
-	Use:               "delete <name>",
+	Use:               "delete",
```
matejvasek, line 19: Why not allow to delete function by name?

cmd/describe.go
```diff
@@ -28,18 +27,26 @@ func init() {
 }
 
 var describeCmd = &cobra.Command{
-	Use:               "describe <name> [options]",
+	Use:               "describe [options]",
```
matejvasek, line 30: Why shouldn't be I able to ask for description of function by name (which is not in cwd)?

lance: Basically for the same reasons I wrote in my [comment above](https://github.com/boson-project/faas/pull/125#discussion_r490348425) about `delete`. I think `describe` is another candidate for removal because how is `kn service describe <service-name>` any worse than `kn faas describe`? And again, if config moves upstream, then `kn service describe` === `kn faas describe`.

cmd/delete.go
matejvasek: >kn service delete and kn faas delete doing exactly the same thing

Yes they practically are same thing.

cmd/describe.go
matejvasek: >kn service describe === kn faas describe

Not exactly same, `faas` has slightly different formatting and also pulls info about subscriptions/triggers which `kn` doesn't IIRC.

matejvasek: `kn` has separate sub-command for subscriptions/triggers.

cmd/delete.go
lance: > Not sure what this mean, what conf file, what upstream?

The `.faas.yaml` file, moving to something like `.kn.yaml` in knative/client to generically describe a Knative project.

cmd/describe.go
lance: > Not exactly same, faas has slightly different formatting and it also pulls info about subscriptions/triggers which kn doesn't

Fair point. Let's discuss all of these commands in https://github.com/boson-project/faas/issues/126

zroubalik (approved):
LGTM

matejvasek (approved):

lkingland:
If there is a compelling reason to remove this flag, I am all for agreeing, but I beleve there is a compelling reason it was added in the first place, which I will present below:

The reason for the path flag is primarily to support users who would like to script their usage of the `faas` binary without having to change directories to the target directory before running commands.  For instance, if one must change directory before running the command, a script can no longer be a simple single command, but now must have state:  it must couple the previous command (`cd ...`) with the following command (`faas ...`).  The purpose is very much similar to the `--directory` flag of the `make` command, which I have personally found quite helpful when scripting.

The idea of the --path flag is to run the command as if the command had been run from within that directory.  I believe this to be a valid and quite helpful use case.

Furthermore, by providing this flag as a configuration parameter, we ensure that this variable is explicitly provided in the client API.  By removing it we run the risk of the client library thinking it can sample the current working directory willy-nilly, which could cause problems when used as a library in multi-threaded applications which legitimately change directories during operation.

So If this flag is _only_ being removed to enforce that the client user is running the command from within the directory, I actually would have to disagree.  Commands should take their effective directory as an explicit input, which can be _defaulted_ to the current working directory. 

I hope that adequately presents the case for the existence of this flag, but am happy to capitulate with but a modicum of pushback!

function.go
```diff
@@ -68,6 +68,43 @@ func NewFunction(root string) (f Function, err error) {
...
+// OverrideImage overwrites (or sets) the value of the Function's .Image
+// property, which preempts the default functionality of deriving the value as:
+// Deafult:  [config.Repository]/[config.Name]:latest
+func (f Function) OverrideImage(image string) error {
+	if image == "" {
+		return nil
+	}
+	f.Image = image
+	return f.WriteConfig()
+}
```
lkingland, line 94: FYI, in casd you are wondering why this function was in the CLI codebase not the "core" logic: it was because this overriding is specifically an override of that image prior to the invocation of the client.  In away mutating the state of the function prior to actually invoking the client.  This is a way to provide the ability to override the image name via a flag without "muddying up" the client library such that it can instead always trust what is in .faas.yaml.

lance:
Marking as a draft for now and may rework some of this. @zroubalik will stalling this affect https://github.com/boson-project/faas/pull/127?

zroubalik:
@lance I could finish #127 without any changes wrt to `--path` and the related stuff or I can wait for the resolution on the overall CLI UX decision. I'd prefer the latter, as we can implement that properly

lance:
@zroubalik check out https://github.com/boson-project/faas/pull/128. I think we can use that to iterate on what the CLI behavior should be, with concrete descriptions, etc. I've written it to reflect how the CLI behaves today with the exception of `init` and `create`. Those I wrote to reflect the work in your PR here.

Issue #123, opened 2020-09-16
Proposal to decouple function name from function domain


lance:
Currently, the CLI has been written to conflate the function name with the function domain. For example, creating a project in a directory, `foo`, produces the following:

```console
❯ ../faas init -l node                                                                                                         Wed 16 Sep 2020 09:49:18 AM EDT
Project path: /home/lanceball/src/go/src/github.com/boson-project/faas/foo
Project name: foo.faas.boson-project.github.com
Runtime: node
Trigger: http
```
Where the project name has been derived from the local directory. It is possible to provide a name to the `init` and create commands. For example,

```console
❯ faas init foo -l node                                                                                                541ms  Wed 16 Sep 2020 09:51:01 AM EDT
Project path: /home/lanceball/src/go/src/github.com/boson-project/faas/foo
Project name: foo
Runtime: node
Trigger: http
```
However, this produces problems during deployment.

```console
❯ faas build                                                                                                           545ms  Wed 16 Sep 2020 09:52:07 AM EDT
A repository for Function images is required. For example, 'docker.io/tigerteam'.

Repository for Function images: docker.io/lanceball
Building image: docker.io/lanceball/foo:latest

faas/foo on  main [$!?] is 📦 v0.1.0 via ⬢ v12.16.3 took 35s 
❯ faas deploy                                                                                                          35.8s  Wed 16 Sep 2020 09:52:48 AM EDT
Error: : Invalid value: "foo": should be a domain with at least two segments separated by dots
```

This is, of course, going to be confusing to the end user. "I just wanted to name my function something. Why is it failing?". 

In fact, if I manually edit `./.faas.yaml` to attempt to comply with the error message suggestion, I then get the following.

```console
❯ cat .faas.yaml                                                                                                        5.1s  Wed 16 Sep 2020 09:54:25 AM EDT
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: .faas.yaml
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ name: foo.bar
   2   │ namespace: ""
   3   │ runtime: node
   4   │ image: docker.io/lanceball/foo:latest
   5   │ trigger: http
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

faas/foo on  main [$!?] is 📦 v0.1.0 via ⬢ v12.16.3 
[...]

PR #127, opened 2020-09-18
feat: decouple function name from function domain

zroubalik:

Fixes: #123

```
faas create foo -l node
```
Creates a new directory foo and initialize/create the function inside. 

A directory structure is created according to the specified name, eg. `relative/path/to/foo`, `/absolute/path/to/foo`, `foo` or empty value (in this case it uses current directory).

The 'last' directory name is used as function name.


client.go
```diff
@@ -306,6 +306,12 @@ func (c *Client) Create(cfg Function) (err error) {
 // Initialize creates a new Function project locally using the settings
 // provided on a Function object.
 func (c *Client) Initialize(cfg Function) (err error) {
+
+	// Create project root directory, if it doesn't already exist
+	if err = os.MkdirAll(cfg.Root, 0755); err != nil {
+		return
+	}
+
```
lkingland, line 314: I really like the idea of creating the directory in which a function will be deployed because it would ensure the function is in a clean fresh directory.  However, one small question:  what do we do when a user wants to start over, without also removing their git history (`.git`) or other ancillary config files (`.direnv`)?  This is why the code exists to check if a directory is _effectively_ empty:  contains no visible files, and no .faas.yaml, before allowing a re-run of `create` or `init`.

cmd/create.go
```diff
@@ -110,19 +109,20 @@ func newCreateConfig(args []string) createConfig {
 // Skipped if not in an interactive terminal (non-TTY), or if --confirm (agree to
 // all prompts) was not explicitly set.
 func (c createConfig) Prompt() createConfig {
-	name := deriveName(c.Name, c.initConfig.Path)
```
lkingland, line 113: I am pretty sure that moving this name derivation code below the prompt, means the confirmation prompt will not display the value that will actually be used.  

By deriving `name` first, as the calculated derived and storing it in a temporary variable, it allows the prompt to display the dervied value that would be used if the client library were to be invoked directly (it would derive the value).

cmd/create.go
```diff
@@ -110,19 +109,20 @@ func newCreateConfig(args []string) createConfig {
...
 		return c
 	}
+
+	derivedName, derivedPath := deriveNameAndPath(prompt.ForString("Project name", "", prompt.WithRequired(true)))
 	return createConfig{
 		initConfig: initConfig{
-			Path:    prompt.ForString("Project path", c.initConfig.Path),
-			Name:    prompt.ForString("Project name", name, prompt.WithRequired(true)),
+			Name:    derivedName,
+			Path:    derivedPath,
```
lkingland, line 125: It appears to me that perhaps this omits the prompting of name and path entirely, right?

I was actuall fond of the original idea:  to prompt if the parameter were not provided as a flag.  If that's not the plan anymore, no problem!  But if we wanted to prompt the user when the parameters had not been provided, would we not instead leave things roughly as they were, and insteaed add here a conditional that only prompts if the flag does not exist (as provided by the CLI API)?

cmd/init.go
```diff
@@ -49,10 +48,10 @@ func runInit(cmd *cobra.Command, args []string) error {
 }
 
 type initConfig struct {
-	// Name of the service in DNS-compatible format (ex myfunc.example.com)
+	// Name of the function.
```
lkingland, line 51: Minor nitpick about the comment:  that should probably be `Function`.  In order to avoid confusion with the built-in function primitive of most languages, to name our functions as proper nouns; thus capitalized.

cmd/init.go
```diff
@@ -49,10 +48,10 @@ func runInit(cmd *cobra.Command, args []string) error {
...
 	Name string
 
-	// Path to files on disk.  Defaults to current working directory.
+	// Absolute path to project on disk.
```
lkingland, line 54: While this is still yet to be discussed, it may be worth mentioning:  a "project" is likely to be a _grouping_ of Functions.  Such as the entire `example.com` directory, with a new config file at that parent level.  Our code at persent is just for Functions at this point, not really projects... yet!

cmd/create.go
lkingland: Perhaps this is not the case, as it appears this logic is actually now just moved into the initConfig directly.  I presume this was to make both `create` and `init` work the same?  My comment above may therefore not be applicable.

k8s/names.go
```diff
@@ -10,9 +10,6 @@ import (
 // ToSubdomain converts a domain to a subdomain.
 // If the input is not a valid domain an error is thrown.
```
lkingland, line 11: It appears this is no longer a requirement? If not the comment could use an update.

zroubalik:
@boson-project/core PTAL

cmd/create.go
lance: > It appears to me that perhaps this omits the prompting of name and path entirely, right?

@lkingland I may be misreading your comment, so disregard if so (or correct me). The prompt occurs on line 121.

```js
derivedName, derivedPath := deriveNameAndAbsolutePathFromPath(prompt.ForString("Project path", c.initConfig.Path, prompt.WithRequired(true)))
```

cmd/create.go
```diff
@@ -106,23 +105,24 @@ func newCreateConfig(args []string) createConfig {
...
 		// Just print the basics if not confirming
 		fmt.Printf("Project path: %v\n", c.initConfig.Path)
-		fmt.Printf("Project name: %v\n", name)
+		fmt.Printf("Function name: %v\n", c.initConfig.Name)
 		fmt.Printf("Runtime: %v\n", c.Runtime)
 		fmt.Printf("Trigger: %v\n", c.Trigger)
 		return c
 	}
+
+	derivedName, derivedPath := deriveNameAndAbsolutePathFromPath(prompt.ForString("Project path", c.initConfig.Path, prompt.WithRequired(true)))
```
lance, line 121: This `deriveNameAndAbsolutePathFromPath()` function is called in [`newInitConfig()`](https://github.com/boson-project/faas/pull/127/files#diff-46777a3b1ee86d4cfc7d0bef9b3b7ee8R85) which is called by [`newCreateConfig()`](https://github.com/boson-project/faas/pull/127/files#diff-452662b1e58d63509f34fc4bd164905bL103),  so it seems redundant to call it again here. The values for `c.initConfig.Path` and `c.initConfig.Name` should already be set, shouldn't they?

cmd/root.go
```diff
@@ -165,21 +166,42 @@ func functionWithOverrides(root, namespace, image string) (f faas.Function, err
...
 		return f.Name
 	}
-	maxRecursion := faas.DefaultMaxRecursion
-	derivedName, _ := faas.DerivedName(path, maxRecursion)
-	return derivedName
+
+	return ""
+}
+
+// deriveNameAndPath returns resolved Function name and absolute path
```
lance, line 185: Nit:
```suggestion
// deriveNameAndAbsolutePath returns resolved Function name and absolute path
```

cmd/root.go
```diff
@@ -165,21 +166,42 @@ func functionWithOverrides(root, namespace, image string) (f faas.Function, err
 
 // deriveName returns the explicit value (if provided) or attempts to derive
 // from the given path.  Path is defaulted to current working directory, where
-// a function configuration, if it exists and contains a name, is used.  Lastly
-// derivation using the path us used.
+// a Function configuration, if it exists and contains a name, is used. 
 func deriveName(explicitName string, path string) string {
```
lance, line 170: I wonder if this function could be better named as it no longer really derives the name, but instead just pulls it from the configuration file or returns the passed value.

k8s/names.go
```diff
@@ -26,13 +25,22 @@ func ToSubdomain(in string) (string, error) {
 			out = append(out, c)
 		}
 	}
-	return string(out), nil
+
+	result := string(out)
+
+	if errs := validation.IsDNS1035Label(result); len(errs) > 0 {
```
lance, line 31: What kinds of characters in `in` will cause this to fail?

lance (commented):
This generally LGTM aside from a couple of questions and nits.

cmd/root.go
zroubalik: Any suggestion?

cmd/create.go
zroubalik: This one is for prompt. Yes, the value is there, but user could replace `path` in the prompt, then we need to get the path and name again.

k8s/names.go
zroubalik: Added comment with a description

cmd/root.go
lance: > Any suggestion?

I've been sitting here for 5 minutes trying to come up with a good name. I see why you asked. The best I've come up with is `providedOrConfiguredName(string) string`. But meh.

zroubalik: Hehe, yeah 😄 I don't like the name either, so I am happy to change it to anything else, that makes sense for everyone

Issue #126, opened 2020-09-17
Consider removing `describe` and `delete` & possibly `deploy` and `update`


lance:
When we consider that this project's intent is to run as an inlined plugin in the `kn` CLI, the commands we use now as `faas describe` and `faas delete` become `kn faas describe` and `kn faas delete`. These are really no different than `kn service describe` and `kn service delete`. I don't think we should be duplicating this behavior that already exists in the target binary.

I think we should also discuss the implications of `deploy` and `update`. These commands are very similar to `kn service create` and `kn service update`, but actually do less. The `deploy` command does do add some annotations and labels to the deployment, so it may be necessary to keep. But certainly `kn service update` is hardly different than `faas update`.

/cc @boson-project/core

zroubalik:
Agree on that.

Just a brain dump:
What if `kn faas describe` is just a wrapper around `kn service describe` + printing some additional (functions related) info. I am not sure if this is something, that's even possible in `kn`.

lkingland:
> When we consider that this project's intent is to run as an inlined plugin in the kn CLI

If that is the intent, then yes these commands are redundant.  However, if we would like the faas client library to provide a platform-agnostic set of Function manipulation primitives, which can be used across multiple deployment platforms, then perhaps these individual command should be retained in the `faas` binary and not "plumbed" through `kn`.

nainaz:
This constant switch to kn faas to kn service could be very confusing and jarring. 
I believe consistency for Function Author will be paramount given they might not be from dev background. 
to create function I write ```kn faas init``` or ```kn faas create```. but then to delete or describe I have to write ``` kn service describe``` etc. 
Is there a way to wrap these kn commands in faas cli for better experience?

nainaz:
I agree with the point @lkingland  made here . 
```However, if we would like the faas client library to provide a platform-agnostic set of Function manipulation primitives, which can be used across multiple deployment platforms```. 
This should be our goal, platform agnostic in the long run.

lance:
>  This should be our goal, platform agnostic in the long run.

Why? What other platform do we realistically expect this to be running on any time before 2022?

lance:
>  Is there a way to wrap these kn commands in faas cli for better experience?

They already are in the faas CLI. I am proposing to remove them, because they are redundant. Certainly `delete` is, if not `describe`.

lkingland:
[...]

PR #128, opened 2020-09-18
docs: add documentation for CLI commands

lance:
I've added some documentation for the CLI commands here and called out some odd behavior with the options in a couple of places. I wrote this as I was trying to sort out how I felt about https://github.com/boson-project/faas/pull/125 and https://github.com/boson-project/faas/issues/126 and thought that it would be a good document to use when trying to rationalize any usability changes. For the most part, I have changed my mind about #125 and #126. I'm still on the fence about `faas delete`, but as I wrote in the doc here, there are valid reasons to have the commands that we do and I'll just eat crow. :bird: 

Also - note that for all commands except `init` and `create`, I have documented them as they are today. For the `init` and `create` commands, I documented them as intended for https://github.com/boson-project/faas/pull/127.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -1,0 +1,153 @@
+# CLI Commands
+
+## `init`
+
+Creates a new Function project at _`path`_. If _`path`_ is unspecified, assumes the current directory. If _`path`_ does not exist, it will be created. The user can specify the runtime and trigger with flags.
+
+Similar `kn` command: none.
+
+```console
+faas init <path> [-l <runtime> -t <trigger>]
+```
+
+When run as a `kn` plugin.
+
+```console
+kn faas init <path> [-l <runtime> -t <trigger>]
+```
+
+## `build`
+
+Builds the Function project in the current directory. Reads the `.faas.yaml` file to determine image name and repository. If both of these values are unset in the configuration file, the user is prompted to provide a repository, from there an image name can be derived. The image name and repository may also be specified as flags, as can the path to the project.
+
+The value(s) provided for image and repository are persisted to the `.faas.yaml` file so that subsequent invocations do not require the user to specify these again.
+
+Similar `kn` command: none.
+
+```console
+faas build [-i <image> -r <repository> -p <path>]
+```
+
+When run as a `kn` plugin.
+
+```console
+kn faas build [-i <image> -r <repository> -p <path>]
+```
+
+## `run`
+
+Runs the Function project locally in the container. If a container has not yet been created, prompts the user to run `faas build`. Note: there is no option to specify a path to the project.
+
+Similar `kn` command: none.
+
+```console
+faas run
+```
+
+When run as a `kn` plugin.
+
+```console
+kn faas run
+```
+
+## `deploy`
+
+Deploys the Function project in the current directory. The user may specify a path to the project directory as a flag. Reads the `.faas.yaml` configuration file to determine the image name. Derives the service name from the project name. There is no command line option to specify the image name, although this can be changed in `.faas.yaml`. There is no mechanism by which the user can specify the service name. By default, the image is deployed to the namespace corresponding to the value specified in `.faas.yaml`, or if unset to the currently active Kubernetes namespace. This can be changed with a command line flag. If changed, this value will be persisted in `.faas.yaml`.
+
+Similar `kn` command: `kn service create NAME --image IMAGE [flags]`. This command allows a user to deploy a Knative Service by specifying an image, typically one hosted on a public container registry such as docker.io. The deployment options which the `kn` command affords the user are quite broad. The `kn` command in this case is quite effective for a power user. The `faas deploy` command has a similar end result, but is definitely easier for a user just getting started to be successful with.
+
+```console
+faas deploy [-n <namespace> -p <path>]
+```
+
+When run as a `kn` plugin.
+
+```console
+kn faas deploy [-n <namespace> -p <path>]
+```
+
+## `update`
+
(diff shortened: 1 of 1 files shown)


docs/commands.md
```diff
@@ -0,0 +1,153 @@
...
+
+When run as a `kn` plugin.
+
+```console
+kn faas run
+```
+
+## `deploy`
+
+Deploys the Function project in the current directory. The user may specify a path to the project directory as a flag. Reads the `.faas.yaml` configuration file to determine the image name. Derives the service name from the project name. There is no command line option to specify the image name, although this can be changed in `.faas.yaml`. There is no mechanism by which the user can specify the service name. By default, the image is deployed to the namespace corresponding to the value specified in `.faas.yaml`, or if unset to the currently active Kubernetes namespace. This can be changed with a command line flag. If changed, this value will be persisted in `.faas.yaml`.
```
lkingland, line 55: It might be worth noting that a Function can not be deployed unless it is first built (which results in there being an image name in `.faas.yaml`).  To deploy an image built manually, the `.faas.yaml` may be manually edited.

docs/commands.md
```diff
@@ -0,0 +1,153 @@
...
+
+```console
+kn faas deploy [-n <namespace> -p <path>]
+```
+
+## `update`
+
+Updates the deployed Function project in the current directory. The user may specify the path on the command line with a flag. Reads the `.faas.yaml` configuration file to determine the image name. Derives the service name from the project name. The deployed Function is updated with a new container image that is pushed to a user repository, and the Knative `Service` is then updated.
+
+The namespace defaults to the value in `.faas.yaml` or the namespace currently active in the user's Kubernetes configuration. The namespace may be specified on the command line, and if so this will overwrite the value in `.faas.yaml`. The user may specify a repository on the command line.
```
lkingland, line 73: This is a really helpful paragraph.  Perhaps it should be repeated in the `deploy` section above as well.

docs/commands.md
```diff
@@ -0,0 +1,153 @@
...
+
+```console
+kn faas update [-r <repository> -p <path>]
+```
+
+## `describe`
+
+Prints the name, route and any event subscriptions for a deployed Function. The user may also specify the name of the function to describe, and the namespace in which to search for the function. If provided, this will be updated in the project's `.faas.yaml` configuration file. The user may also supply a path to the Function project.
+
+Similar `kn` command: `kn service describe NAME [flags]`. This flag provides a lot of nice information not available in `faas describe`, such as revisions, age, annotations and labels. Consider a way to combine the two outputs, or rename this command to make it distinct from `kn` - e.g. `faas info`.
```
lkingland, line 93: This is a good suggestion.  I would perhaps rename it to`status`, because that word implies info with the added property that it is transient state.  "the current status of the Function".  In particular I am thinking how we will likely want to add the current number of instances, and aggregate resource usage.  

The output of this `status` command (or `info`, which works too!) could perhaps be thought of as a reflection of the features the client provides. The current status of the "knobs" that `faas` allows to be turned.  So the `describe` of `kn` has more information, because it is showing the status/state of all the knobs that `kn` provides.  We provide less knobs, as we are trying to create a clean and simple abstraction.

However, there _are_ some bits of status which should bubble up from the underlying layer.   Information radiators, rather than knobs. As previously mentioned, number of instances and aggregate resource usage would be the top of my wish-list.

docs/commands.md
```diff
@@ -0,0 +1,153 @@
...
+
+```console
+kn faas describe [-f <format> -n <namespace> -p <path>]
+```
+
+## `list`
+
+Lists all deployed functions. The namespace defaults to the value in `.faas.yaml` or the namespace currently active in the user's Kubernetes configuration. The namespace may be specified on the command line, and if so this will overwrite the value in `.faas.yaml`.
+
+Similar `kn` command: `kn service list [name] [flags]`. This command lists all deployed Knative `Services`. As with other `kn` commands that have similar functionality, there is more information and flexibilty in the `kn` command. However, `kn` will return _all_ `Services`, while `faas list` will only display the boson Functions that have been deployed. Consider improving the output of the `faas list` command so that it is at least as informative as `kn service list`.
```
lkingland, line 109: > Consider improving the output of the `faas list` command so that it is at least as informative as `kn service list`.

I would recommend we leave this as a very plain, simple list of names, such that it can be easily integrated with other systems via a shell script.  The details about the services would then be available through the same mechanism using `faas status <name>`
```
$ faas list
www.boson-project.org
admin.boson-project.org
$ faas list | xargs faas status
... details of each
```

In particular:  `faas list | xargs faas status`

lkingland: Oh, I see that the `faas status` no longer accepts the name as input.  We might want to consider adding that back in.  In particular I am thinking this is useful for someone wanting to get details about a function for which they do not currently have the code locally, but do have access to the cluster.

lkingland (approved):
This is such a readable, and yet thorough write-up.  Great to have as a place for discussion as well.

💯

docs/commands.md
lkingland: Upon further thought, I like the idea of adding the extra data, so long as that first token is the name which is the input to `status` such that it's still easily scriptable.

docs/commands.md
zroubalik: I like the idea of `status` or `info`, great idea @lkingland !

zroubalik (approved):
Looking good @lance !

matejvasek (approved):

lance:
Landing this. @boson-project/contributors, let's be sure to keep it updated as the commands change.

PR #129, opened 2020-09-18
chore: remove [options] from usage string

lance:
The cobra package, magically appends "[flags]" to the usage string if a command has flags. By adding "[options]" to the usage string, we end up with help text that looks like this.

```
  faas init <name> [options] [flags]
```

This commit fixes that.


lkingland (approved):
Ugh, definitely redundant.  Good to know!

PR #130, opened 2020-09-18
fix: correct value for config path and robustify

lance:
The hardcoded, initial value for the configuration path was set to `.faas/config`. But `configPath()` immediately sets this to the correct value of ~/.config. Both the create and init commands use `configPath()` to search for additional templates, if they exist, and were each doing `filepath.Join(configPath(), "faas", "templates")`. This commit also changes `configPath()` so that it is `~/.config/faas` and does so in a cross platform friendly way. If the `$HOME` directory cannot be determined, the config is assumed to be at `./.config/faas`.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -12,8 +12,6 @@ import (
 	"github.com/boson-project/faas"
 )
 
-var config = configPath() // Location of the optional system-wide config file.
-
 // The root of the command tree defines the command name, descriotion, globally
 // available flags, etc.  It has no action of its own, such that running the
 // resultant binary with no arguments prints the help/usage text.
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -11,7 +12,7 @@ import (
 	"github.com/boson-project/faas"
 )
 
-var config = "~/.faas/config" // Location of the optional system-wide config file.
+var config = configPath() // Location of the optional system-wide config file.
```
lkingland, line 15: This line can now be removed, as it's been correctly updated to be directly inlined into the `configPath` function using `filepath` to make it OS agnostic.

Indeed it _should_ probably be removed as the `configPath()` static function should be used preferentially to a global variable.

lkingland (approved):
Good catch, looks like the default initial value was not correctly being used in the function originally; probably a casualty of refactor.  Nice to see the implementation updated to use filepath now too.  That global variable serves no purpose now and can be removed to make it even cleaner.

zroubalik (approved):
LGTM

PR #136, opened 2020-09-23
feat: set builder images in templates and .faas.yaml

lance:
This commit adds a .builder.yaml file to each template directory. In the file
there is at the moment a single key/value pair, "`default: <image>`", where the
actual builder image name is `<image>`. Using a mapping allows the future
possibility that a user may specify a builder image by name via a flag on the
command line.

When a project is initialized, the .builder.yaml file is read, and the default
builder is saved in the project's .faas.yaml file. The .faas.yaml file is then
consulted when building an image with `faas build`. If the builder image is
specified, then the builder will use it. Otherwise, it will fallback to the
defaults. This allows developers to create custom builders, and specify them
in the configuration file.

An open question is whether the .builder.yaml file should be deleted after it
is read during project initialization. The `init` command is the only time the
file is consulted, and it may be confusing for the user if they discover a
second configuration-y file in the project directory.

Fixes: https://github.com/boson-project/faas/issues/120


matejvasek:
>An open question is whether the .builder.yaml file should be deleted after it
is read during project initialization. The init command is the only time the
file is consulted, and it may be confusing for the user if they discover a
second configuration-y file in the project directory.

I think it's OK to delete the file for now.

In future I will add a feature for build modes (like `release`, `debug`), that means the `.builder.yaml` should be renamed to`.builders.yaml` since the file will contain multiple builders. Question is whether this map of builder will be copied to `.faas.yaml` if so then it can be deleted, otherwise it would need to be preserved. But that is question for future, I'll take care that.

templates/go/events/.builder.yaml
```diff
@@ -0,0 +1 @@
+default: nodejs quay.io/boson/faas-go-builder
```
matejvasek, line 1: nodejs ?

zroubalik (approved):
LGTM, there's just the typo in the `go events builder`

+1 to delete the file afterwards

buildpacks/builder.go
```diff
@@ -30,10 +30,17 @@ var Runtimes = map[string]string{
...
-	if !ok {
-		return errors.New(fmt.Sprint("unsupported runtime: ", f.Runtime))
+
+	// Use the builder found in the Function configuration file
+	// If one isn't found, use the defaults
+	var packBuilder string
+	if f.Builder != "" {
+		packBuilder = f.Builder
+	} else {
+		packBuilder = Runtimes[f.Runtime]
```
matejvasek, line 40: `Runtimes` var name seems odd in this context, maybe it could be `Buildpacks` or `Runtime2Buildpack`, I know you haven't introduced that name.

zroubalik:
> In future I will add a feature for build modes (like `release`, `debug`), that means the `.builder.yaml` should be renamed to`.builders.yaml` since the file will contain multiple builders. Question is whether this map of builder will be copied to `.faas.yaml` if so then it can be deleted, otherwise it would need to be preserved. But that is question for future, I'll take care that.

so what if we name the file `.builders.yaml` from the start, since we know that we will add more of them

lance:
I've incorporated all of the feedback @zroubalik @matejvasek - ty!

matejvasek (approved):

zroubalik (approved):

PR #145, opened 2020-10-01
Update to Knative 0.17 + pin some dependencies

rhuss:
Also adjusted some dependencies and overwrite version to align
transitive dependencies.

Not sure why buildpack uses such an older docker dependency but that
clases with the docker dependency that is introduced by knative-dev/test-infra
(which is a dependency of the knative direct dependencies)

If there is a way to exclude a transitive dependency like that on test-infra,
this could be the better way to achieve the same result.

The build now works on macOs natively which was not the case before.


zroubalik (approved):
LGTM

rhuss:
Please note that this will also update buildpacks to 0.14.0 (released today)

go.mod
```diff
@@ -3,23 +3,34 @@ module github.com/boson-project/faas
...
+	github.com/buildpacks/pack v0.14.0
 	github.com/markbates/pkger v0.17.0
-	github.com/mattn/go-colorable v0.1.7 // indirect
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/onsi/gomega v1.10.1 // indirect
 	github.com/ory/viper v1.7.4
-	github.com/spf13/cobra v1.0.0
-	github.com/stretchr/testify v1.5.1 // indirect
-	golang.org/x/sys v0.0.0-20200728102440-3e129f6d46b1 // indirect
+	github.com/spf13/cobra v1.0.1-0.20200715031239-b95db644ed1c
```
matejvasek, line 10: I like to see this. This version contains fixes for `zsh` completion :+1:

lance (approved):

PR #147, opened 2020-10-01
Custom builders

matejvasek:


cmd/deploy.go
```diff
@@ -41,7 +41,7 @@ or -n flag, and if so this will overwrite the value in the faas.yaml file.
 
 func runDeploy(cmd *cobra.Command, _ []string) (err error) {
 	config := newDeployConfig()
-	function, err := functionWithOverrides(config.Path, config.Namespace, "")
+	function, err := functionWithOverrides(config.Path, overrideNamespace(config.Namespace))
```
lance, line 44: Assume this is unrelated, but was in fact a bug?

config.go
```diff
@@ -14,12 +14,13 @@ const ConfigFile = "faas.yaml"
 // Config represents the serialized state of a Function's metadata.
 // See the Function struct for attribute documentation.
 type config struct {
-	Name      string `yaml:"name"`
-	Namespace string `yaml:"namespace"`
-	Runtime   string `yaml:"runtime"`
-	Image     string `yaml:"image"`
-	Trigger   string `yaml:"trigger"`
-	Builder   string `yaml:"builder"`
```
lance, line 22: Pretty sure it was my IDE that formatted these like this. What IDE are you using and do you have a Golang extension/plugin?

cmd/deploy.go
matejvasek: No it wasn't, empty string is ignored, so it's OK. I choose to change the function to be variadic, because I introduced yet another override: `Builder`.

config.go
matejvasek: I am using Goland (Intellij based IDE), but I called `go fmt` from it. Previous formatting was also correct. The indentation changed because I added new member
```go
BuilderMap map[string]string `yaml:"builderMap"`
```

cmd/deploy.go
matejvasek: I didn't wanted to end up with:
```go
function, err := functionWithOverrides(config.Path, config.Namespace, "", "")
```
Plus it's not to difficult to mistake parameter order in the previous version.

cmd/root.go
```diff
@@ -120,46 +120,48 @@ func bindEnv(flags ...string) bindFunc {
 	}
 }
 
+type functionOverride func (root string) error
+
+type strPropSetter func (f *faas.Function, val string)
+
+func createGenericOverride(setter strPropSetter) func (override string) functionOverride {
```
matejvasek, line 127: This is kinda funny. Maybe we could just introduce some new `overrideParam`  struct and make it parameter of `functionWithOverrides`.

lance: > This is kinda funny. Maybe we could just introduce some new `overrideParam` struct and make it parameter of `functionWithOverrides`.

That would probably be a lot more readable.

config.go
lance: Oh of course.

matejvasek:
@lance PTAL

cmd/build.go
```diff
@@ -33,15 +39,19 @@ name will be derived from the project name.
 Any value provided for --image or --repository will be persisted in the
 faas.yaml configuration file. On subsequent invocations of the "build" command
 these values will be read from the configuration file.
+
+It's possible to set custom Buildpacks builder by --builder flag.
```
lance, line 43: Nit
```suggestion
It's possible to use a custom Buildpack builder with the --builder flag.
```

cmd/build.go
```diff
@@ -16,7 +16,13 @@ func init() {
 	buildCmd.Flags().BoolP("confirm", "c", false, "Prompt to confirm all configuration options - $FAAS_CONFIRM")
 	buildCmd.Flags().StringP("image", "i", "", "Optional full image name, in form [registry]/[namespace]/[name]:[tag] for example quay.io/myrepo/project.name:latest (overrides --repository) - $FAAS_IMAGE")
 	buildCmd.Flags().StringP("path", "p", cwd(), "Path to the Function project directory - $FAAS_PATH")
+	buildCmd.Flags().StringP("builder", "b", "default", "Buildpacks builder")
```
lance, line 19: All of the commands have their flags in alphabetical order. I'm not tied to this, but I think it's probably good to be consistent.

lance (approved):
Looks good to me. A couple of minor nits.

templates/quarkus/events/.builders.yaml
```diff
@@ -1 +1,3 @@
-default: quay.io/boson/faas-quarkus-builder
+default: quay.io/boson/faas-quarkus-jvm-builder
```
matejvasek, line 1: @lance note this requires release of buildpacks

lance: https://github.com/boson-project/buildpacks/releases/tag/v0.3.0

PR #148, opened 2020-10-05
fix: stop using manually edited completion

matejvasek:
Depends on PR #145


lance (approved):

matejvasek:
@lance I could add `fish`, but I tested that and it works with `3.1` but not with `3.0`.

PR #151, opened 2020-10-05
feat: fish completion

matejvasek:


matejvasek:
related https://github.com/spf13/cobra/pull/1247

zroubalik (approved):
LGTM

matejvasek:
To make this work on `fish` version `3.0`  we need to  update the `cobra` dependency. @rhuss could it cause any problems in relation to `kn` package?

rhuss:
Sorry for the late reply :) I don't expect any issues with the cobra dependency as cobra is movin very slowly anyway :)

PR #152, opened 2020-10-06
feat!: combine deploy and update commands

zroubalik:

`faas deploy` now builds and deploys a Function.

If the Function is already deployed, it is updated.

Fixes: https://github.com/boson-project/faas/issues/135


zroubalik:
I am planning to unify the rest of the commands once this PR is merged

knative/deployer.go
```diff
@@ -24,52 +25,151 @@ type Deployer struct {
...
+			result = &envs[i]
+			break
 		}
-		return
 	}
-	// TODO: use the KN service client noted above, such that we can return the
-	// final path/route of the final deployed Function.  While it can be assumed
-	// due to being deterministic, new users would be aided by having it echoed.
-	return
+	return result
```
matejvasek, line 174: @lance @rhuss this will let cluster to generate revision name so this: https://github.com/knative/serving/issues/9544 won't affect us, right?

cmd/deploy.go
```diff
@@ -7,62 +7,113 @@ import (
...
 	deployCmd.Flags().StringP("path", "p", cwd(), "Path to the function project directory - $FAAS_PATH")
+	deployCmd.Flags().StringP("repository", "r", "", "Repository for built images, ex 'docker.io/myuser' or just 'myuser'.  - $FAAS_REPOSITORY")
 }
 
 var deployCmd = &cobra.Command{
 	Use:   "deploy",
-	Short: "Deploy an existing Function project to a cluster",
-	Long: `Deploy an existing Function project to a cluster
+	Short: "Deploy or Update an existing Function project to a cluster",
+	Long: `Deploy or Update an existing Function project to a cluster
```
lance, line 29:
```suggestion
	Short: "Deploy an existing Function project to a cluster",
	Long: `Deploy an existing Function project to a cluster
```

I'm not sure the distinction needs to be made here for "Update" since updating is also deploying. The text of the long description clarifies.

cmd/deploy.go
```diff
@@ -7,62 +7,113 @@ import (
...
 
-Deploys the Function project in the current directory. A path to the project
-directory may be provided using the --path or -p flag. The image to be deployed
-must have already been created using the "build" command.
+Builds and Deploys the Function project in the current directory. A path to the project
+directory may be provided using the --path or -p flag. Reads the faas.yaml configuration file
+to determine the image name. An image and repository may be specified on the command line using the  --image or -i
+and --repository or -r flag.
```
lance, line 34: Nit: Since this is going to be displayed in a terminal, I try to keep to 80 columns.
```suggestion
to determine the image name. An image and repository may be specified on the command line
using the  --image or -i and --repository or -r flag.
```

cmd/deploy.go
```diff
@@ -7,62 +7,113 @@ import (
...
+
+If the Function is already deployed, it is updated with a new container image that is pushed to a
```
lance, line 36: Nit: I think this might also be longer than 80 columns.

cmd/deploy.go
```diff
@@ -7,62 +7,113 @@ import (
...
 faas.yaml configuration file. If NAMESPACE is not set in the configuration,
 the namespace currently active in the Kubernetes configuration file will be
 used. The namespace may be specified on the command line using the --namespace
 or -n flag, and if so this will overwrite the value in the faas.yaml file.
+
+
 `,
 	SuggestFor: []string{"delpoy", "deplyo"},
-	PreRunE:    bindEnv("namespace", "path", "confirm"),
+	PreRunE:    bindEnv("image", "namespace", "path", "repository", "confirm"),
```
lance, line 48: Just noting that this is affected by #156 as well.

cmd/deploy.go
```diff
@@ -7,62 +7,113 @@ import (
...
-	config := newDeployConfig()
-	function, err := functionWithOverrides(config.Path, functionOverrides{Namespace: config.Namespace})
+
+	config := newDeployConfig().Prompt()
+
+	// Load Function from the config
+	// function, err := faas.NewFunction(config.Path)
+	// if err != nil {
+	// 	return
+	// }
```
lance, line 60:
```suggestion
```

:wink:

cmd/root.go
```diff
@@ -150,12 +150,6 @@ func functionWithOverrides(root string, overrides functionOverrides) (f faas.Fun
 		}
 	}
 
-	err =  f.WriteConfig()
-	if err != nil {
-		return
-	}
-
-	f, err = faas.NewFunction(root)
```
lance, line 158: I thought this was strange when I encountered it initially. I think I like this change, but I think we should call it out explicitly here, that the behavior is changed. When `functionWithOverrides()` is called, the overrides are __not__ persisted. That needs to be handled explicitly in each command if it's the desired behavior.

docs/commands.md
```diff
@@ -52,40 +52,30 @@ kn faas run
 
 ## `deploy`
 
-Deploys the Function project in the current directory. The user may specify a path to the project directory as a flag. Reads the `faas.yaml` configuration file to determine the image name. Derives the service name from the project name. There is no command line option to specify the image name, although this can be changed in `faas.yaml`. There is no mechanism by which the user can specify the service name. The user must have already built an image for this function using `faas deploy` or they will encounter an error.
+Builds and deploys the Function project in the current directory. The user may specify a path to the project directory using the `--path` or `-p` flag. Reads the `faas.yaml` configuration file to determine the image name. An image and repository may be specified on the command line using the  `--image` or `-i`
+and `--repository` or `-r` flag.
```
lance, line 56: Nit: Odd line break
```suggestion
Builds and deploys the Function project in the current directory. The user may specify a path to the project directory using the `--path` or `-p` flag. Reads the `faas.yaml` configuration file to determine the image name. An image and repository may be specified on the command line using the  `--image` or `-i` and `--repository` or `-r` flag.
```

docs/commands.md
```diff
@@ -52,40 +52,30 @@ kn faas run
...
 
-The namespace defaults to the value in `faas.yaml` or the namespace currently active in the user's Kubernetes configuration. The namespace may be specified on the command line, and if so this will overwrite the value in `faas.yaml`.
+Derives the service name from the project name. There is no mechanism by which the user can specify the service name. The user must have already initialize the  function using `faas init` or they will encounter an error.
```
lance, line 58:
```suggestion
Derives the service name from the project name. There is no mechanism by which the user can specify the service name. The user must have already initialized the  function using `faas init` or they will encounter an error.
```

docs/commands.md
```diff
@@ -52,40 +52,30 @@ kn faas run
...
+container image repository, and the Knative Service is updated.
 
-```console
-faas deploy [-n <namespace> -p <path>]
-```
+The namespace into which the project is deployed defaults to the value in the
+`faas.yaml` configuration file. If `NAMESPACE` is not set in the configuration,
+the namespace currently active in the Kubernetes configuration file will be
+used. The namespace may be specified on the command line using the `--namespace`
+or `-n` flag, and if so this will overwrite the value in the `faas.yaml` file.
```
lance, line 67: Sorry for all of these ridiculous formatting comments! I think this should all be one line like the rest of the file. It's minor, I know.
```suggestion
The namespace into which the project is deployed defaults to the value in the `faas.yaml` configuration file. If `NAMESPACE` is not set in the configuration, the namespace currently active in the Kubernetes configuration file will be used. The namespace may be specified on the command line using the `--namespace` or `-n` flag, and if so this will overwrite the value in the `faas.yaml` file.
```

knative/deployer.go
```diff
@@ -3,19 +3,20 @@ package knative
...
+	"sort"
+	"time"
 
-	"knative.dev/client/pkg/kn/root"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/client/pkg/wait"
+	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
+	v1 "knative.dev/serving/pkg/apis/serving/v1"
```
lance, line 14: You and @dsimansk are taking [different approaches](https://github.com/boson-project/faas/pull/155/files) with this file. I know that you are also dealing with a lot more logic, but I wonder which is right - using the client or using a command?

knative/deployer.go
lance: I think so? I'm starting to get confused about the different approaches. Is it correct to use the [commands](https://github.com/boson-project/faas/pull/155/files) or the client as in this case?

lance (commented):
Looks great! A few minor comments. And one big one about client vs command in kn.

cmd/deploy.go
zroubalik: ?? I have no idea why it is here, sorry.

docs/commands.md
zroubalik: No need to sorry @lance thanks a lot for the feedback, the formatting is important. I should have done better.

cmd/root.go
zroubalik: I'll update the function description. The original solution caused that the `faas deploy` command executed outside of any project directory, created the empty `faas.yaml` file in the directory.
This change makes the `functionWithOverrides()` more clean imho.

knative/deployer.go
zroubalik: 🤷‍♂️

cmd/deploy.go
zroubalik: You are right, fixed.

knative/deployer.go
zroubalik: My impression was, that we should use the client, but I might be wrong.

@rhuss wdyt?

knative/deployer.go
dsimansk: We have briefly discussed with @rhuss that it might be beneficial to avoid direct usage of serving/eventing client and rather go with kn abstraction.
For this particular case it'd mean a lot less mangling with service object directly. And in addition we can expect that some level of test coverage is done on kn side already.

zroubalik: @dsimansk btw, in this PR we are not using serving/eventing client, but the one provided by knative/client: https://github.com/boson-project/faas/pull/152/files#diff-0cff654ab7f6b7619dd3ebd4a8855f70R11-R37
But I suppose that doesn't change anything from your suggestion, right?

rhuss: I strongly recommend choosing either/or. However, if you use the `kn` API that is offered in front of eventing/serving you still have to import serving/eventing APIs like this, since the signature of the kn client API references the serving objects directly.

For consistencies sake and from an architectural POV though I would definitely recommend to use the client API if it offers all that you need. If not, we should fix this in the client.

cmd/deploy.go
```diff
@@ -7,62 +7,109 @@ import (
...
 
 var deployCmd = &cobra.Command{
 	Use:   "deploy",
 	Short: "Deploy an existing Function project to a cluster",
 	Long: `Deploy an existing Function project to a cluster
 
-Deploys the Function project in the current directory. A path to the project
-directory may be provided using the --path or -p flag. The image to be deployed
-must have already been created using the "build" command.
+Builds and Deploys the Function project in the current directory.
```
rhuss, line 31: From this description it looks like that `faas build` and `faas deploy` overlap, so that `deploy` is a superset of `build`. This is fine, but then it would be better to include the _build only_ mode as an option to `faas deploy`, like in `faas deploy --build-only`. This would have several benefits:

* Reducing duplication in options that are the same for `deploy` and `build`
* Less commands, smaller UI interface

Drawbacks are:

* Maybe not easy to find out which options of `faas deploy` are relevant for build and which for creating the service.
* Build can be considered as a high-level concern, so a dedicated command might reflect this more prominently

I'm not sure what's the best way to go. From my experience with `serverless.com` integration, I can say that "Faas" platforms don't expose the "build" aspect so prominently as many FaaS platform don't have that stage anyway (think AWS Lambda). So it's just an aspect of the one single, most important operation: How to get my local code running in the cloud ? Aka "deploy". From that angle, I'd recommend to fold "build" as just another aspect of "deploy" and enable this with a dedicated option that influence the flow (its a bit similar to `--dry-run` option that is used by many commands)

On the other hand, if we really want to expose the build as a first level concept, which e.g. is also used *without* deploying (eg for local development with _containers_), then this separate commands make definitely sense.

knative/deployer.go
rhuss: I recommend using the client APIs. However *not* the commands but the client API which comes as an interface as here --> https://github.com/knative/client/blob/master/pkg/serving/v1/client.go (this is _not_ the serving API but a facade to it)

Please don't use the commands. They can change anytime in incompatible way, there are no precautions to prevent this.

Also we are planning to expose the client API (not the commands!), the way how to parse options, how to parse sink prefixes, ... and generally anything that can be useful for implementing an plugin into a separate repository (`knative/client-pkg` as working title), so that you don't even have a direct dependency to `knative/client` anymore (which then also prevents you technically to access the cobra kn commands)

knative/deployer.go
zroubalik: @rhuss I am starting to be confused a little bit, sorry for that, but I am getting lost in the nomenclature, since the word `client` has multiple meanings in this context 😅 

So your suggestion is to use go client provided byt knative/client? eg:  
```go

client, err := p.NewServingClient(namespace)
...
CreateService(service *servingv1.Service)
```

or to use the kn abstraction as @dsimansk mentioned? eg:
```go
c := service.NewServiceCreateCommand(&params)

args := []string{
	encodedName,
	"--image", f.Image,
	"--env", "VERBOSE=true",
	"--label", "bosonFunction=true",
}
c.SetArgs(args)
err = c.Execute()
```

rhuss: The former. No reference to any cobra commands, but just the services. There is also a better way to create an instance of the service that does not goes over the parameter as in the example above.

rhuss: Mostly to avoid cyclic references, and keep a hierarchical dependency tree.

zroubalik: Thanks for the clarification, that's was my original understanding of this.

knative/deployer.go
matejvasek: I tested this and K_SINK is injected, unlike before this change :+1: .

zroubalik:
@lance @matejvasek Is this ready to be merged? So we can move forward with the follow up PRs, that need to be rebased.

lance:
> @lance @matejvasek Is this ready to be merged? So we can move forward with the follow up PRs, that need to be rebased.

Based on our call today, yes I think so.

matejvasek (approved):

PR #159, opened 2020-10-07
fix: update quarkus templates

### LINKED ISSUE #117: Update readme files of Quarkus templates
The template readmes don't contain any notion of `faas` CLI.

matejvasek:
resolves #117


matejvasek:
Note this is still in progress, I'll have to add templates for `HTTP`.
Question: would it be desirable to have also `Gradle` templates beside `Maven`?

matejvasek:
Diff view won't help much with this review. Just look at the new code.

templates/quarkus/events/src/main/java/functions/Function.java
```diff
@@ -0,0 +1,23 @@
...
+                            "id='" + cloudEvent.id() + '\'' +
+                            ", specVersion='" + cloudEvent.specVersion() + '\'' +
+                            ", source='" + cloudEvent.source() + '\'' +
+                            ", subject='" + cloudEvent.subject() + '\'' +
+                            '}');
+        }
+        return input;
+    }
+
+}
```
lance, line 23: Nit: newline

templates/quarkus/events/src/main/java/functions/Function.java
```diff
@@ -0,0 +1,23 @@
...
+        if (cloudEvent != null) {
+            System.out.println(
+                    "CloudEvent{" +
+                            "id='" + cloudEvent.id() + '\'' +
+                            ", specVersion='" + cloudEvent.specVersion() + '\'' +
+                            ", source='" + cloudEvent.source() + '\'' +
+                            ", subject='" + cloudEvent.subject() + '\'' +
+                            '}');
+        }
+        return input;
```
lance, line 20: I'm not sure what Quarkus is doing with `input` here. What does the object end up being, for example in the testcase?

lance (approved):
Really nice simplifications. A question and nit in comments. I'm also wondering about liveness/readiness endpoints. I suppose those should be added in the buildpack?

templates/quarkus/events/src/main/java/functions/Function.java
matejvasek: @lance This is simple echo function. Whatever you pass in it should be returned back.

matejvasek:
> Really nice simplifications. A question and nit in comments. I'm also wondering about liveness/readiness endpoints. I suppose those should be added in the buildpack?

@lance I don't think that should be in buildpack.  I believe that liveness/readiness should be directly in Quarkus Funqy for Knative. I'll need @patriot1burke blessing for that.

patriot1burke:
@matejvasek Can you talk about what you want to add on the quarkus list?  Quarkus may already be able to provide this.

templates/quarkus/events/src/main/java/functions/Function.java
matejvasek: In test-case it's `Integer`. At HTTP level -- whatever JSON value (object,array,string,bool,number,null) you send you should get equivalent JSON value back.

matejvasek:
https://groups.google.com/g/quarkus-dev/c/-c-AJWXybeA/m/-a2VYggNCQAJ

matejvasek:
@lance I added health check endpoints

templates/quarkus/events/src/main/resources/application.properties
```diff
@@ -1,3 +1,8 @@
 # Configuration file
 # key = value
-quarkus.funqy.export=echo
\ No newline at end of file
+
+quarkus.funqy.export=echo
+
+quarkus.smallrye-health.readiness-path=/readiness
+quarkus.smallrye-health.liveness-path=/liveness
```
matejvasek, line 7: @lance I had to add this, otherwise path is `ready/live` instead of `readiness/liveness`.

Same below.

lance: @matejvasek is the full path `/health/readiness` or just `/readiness`? I kind of like having the `/health` prefix as it minimizes the pollution at the root.

matejvasek: The paths are `/health/readiness` and `readiness/liveness`.
There is property `quarkus.smallrye-health.root-path` which is by default "/health".

matejvasek:
@lance `smallrye-health` works only with `HTTP` template now, it doesn't work with `events`. It will require changes to `Funqy` most likely. See https://groups.google.com/g/quarkus-dev/c/-c-AJWXybeA/m/-a2VYggNCQAJ. For now I could merge this PR without the health check endpoints. Is that OK?

matejvasek:
https://github.com/quarkusio/quarkus/pull/12681

PR #166, opened 2020-10-09
Allow set envvars

### LINKED ISSUE #157: Allow user to set environment variables for a Function
For JVM Quarkus we are using https://github.com/fabric8io-images/run-java-sh

The script can be controlled by environment variables.
In particular `JAVA_DEBUG`, which can enable debugging.

This might be useful for debugging in general.

matejvasek:
resolves #157

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -13,6 +13,7 @@ import (
 func init() {
 	// Add the run command as a subcommand of root.
 	root.AddCommand(runCmd)
+	runCmd.Flags().StringArrayP("env", "e", []string{}, "Sets environment variables for the Function.")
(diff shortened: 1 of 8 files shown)


cmd/run.go
```diff
@@ -60,11 +69,29 @@ type runConfig struct {
...
+			for _, s := range envVarsA {
+				kvp := strings.Split(s, "=")
+				if len(kvp) == 2 && kvp[0] != "" {
+					envVarsM[kvp[0]] = kvp[1]
+				} else if len(kvp) == 1 &&  kvp[0] != "" {
+					envVarsM[kvp[0]] = ""
+				}
+			}
+		}
+	}
```
lance, line 91: Just at a quick glance this looks duplicated from above. Could it be extracted into a function?

zroubalik (approved):

PR #167, opened 2020-10-12
src: use Knative Client libraries

zroubalik:

Using Knative Client lib for manipulation with Knative (Serving/Eventing) objects

Fixes: https://github.com/boson-project/faas/issues/160

### DIFF
--- a/knative/client.go
+++ b/knative/client.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"k8s.io/client-go/tools/clientcmd"
+	clienteventingv1beta1 "knative.dev/client/pkg/eventing/v1beta1"
 	"knative.dev/client/pkg/kn/commands"
 	clientservingv1 "knative.dev/client/pkg/serving/v1"
 )
@@ -16,8 +17,45 @@ const (
 	DefaultWaitingTimeout = 60 * time.Second
 )
 
-func NewClient(namespace string, verbose bool) (clientservingv1.KnServingClient, io.Writer, error) {
+func NewServingClient(namespace string, verbose bool) (clientservingv1.KnServingClient, io.Writer, error) {
 
+	knParams := initKnParams(verbose)
+
+	client, err := knParams.NewServingClient(namespace)
+	if err != nil {
+		return nil, knParams.Output, fmt.Errorf("failed to create new serving client: %v", err)
+	}
+
+	return client, knParams.Output, nil
+}
+
+func NewEventingClient(namespace string, verbose bool) (clienteventingv1beta1.KnEventingClient, io.Writer, error) {
+
+	knParams := initKnParams(verbose)
+
+	client, err := knParams.NewEventingClient(namespace)
+	if err != nil {
+		return nil, knParams.Output, fmt.Errorf("failed to create new eventing client: %v", err)
+	}
+
+	return client, knParams.Output, nil
+}
+
+func GetNamespace(defaultNamespace string) (namespace string, err error) {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
+	namespace = defaultNamespace
+
+	if defaultNamespace == "" {
+		namespace, _, err = clientConfig.Namespace()
+		if err != nil {
+			return
+		}
+	}
+	return
+}
+
+func initKnParams(verbose bool) commands.KnParams {
 	p := commands.KnParams{}
 	p.Initialize()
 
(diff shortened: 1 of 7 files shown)


knative/remover.go
```diff
@@ -22,36 +26,17 @@ type Remover struct {
...
-		remover.Namespace, _, _ = clientConfig.Namespace()
-	}
-	client, err := p.NewServingClient(remover.Namespace)
-	if err != nil {
-		return fmt.Errorf("remover failed to create new serving client: %v", err)
+		return
 	}
 
-	err = client.DeleteService(project, time.Second*30)
+	err = client.DeleteService(serviceName, time.Second*60)
```
matejvasek, line 39: :+1:

matejvasek (approved):

lance (approved):

knative/client.go
```diff
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"k8s.io/client-go/tools/clientcmd"
+	clienteventingv1beta1 "knative.dev/client/pkg/eventing/v1beta1"
 	"knative.dev/client/pkg/kn/commands"
```
rhuss, line 12: would it be possible to avoid this dependency ? This won't work in the future when we carve out the client-pkg repo, there must be no dependency on any cobra command.

knative/client.go
```diff
@@ -16,8 +17,45 @@ const (
 	DefaultWaitingTimeout = 60 * time.Second
 )
 
-func NewClient(namespace string, verbose bool) (clientservingv1.KnServingClient, io.Writer, error) {
+func NewServingClient(namespace string, verbose bool) (clientservingv1.KnServingClient, io.Writer, error) {
 
+	knParams := initKnParams(verbose)
+
+	client, err := knParams.NewServingClient(namespace)
```
rhuss, line 24: You should use the constructor from the client API package to avoid a link into the command line parsing.

Not sure how easy this is, but we should make this possible.

knative/client.go
```diff
@@ -16,8 +17,45 @@ const (
...
+	if defaultNamespace == "" {
+		namespace, _, err = clientConfig.Namespace()
+		if err != nil {
+			return
+		}
+	}
+	return
+}
+
+func initKnParams(verbose bool) commands.KnParams {
```
rhuss, line 58: we shoul get rid of this method

knative/describer.go
```diff
@@ -1,69 +1,57 @@
...
-	eventingClient := describer.eventingClient
+func (d *Describer) Describe(name string) (description faas.Description, err error) {
 
 	serviceName, err := k8s.ToK8sAllowedName(name)
 	if err != nil {
 		return
 	}
 
-	service, err := servingClient.Services(namespace).Get(serviceName, metav1.GetOptions{})
+	servingClient, _, err := NewServingClient(d.namespace, d.Verbose)
```
rhuss, line 39: This is the canonical constructor that should be used to avoid the dependency on the commands.

rhuss (commented):
Thanks ! In general looks very good, except one place where there is still a dependency on the kn commands themselves. Not sure why this is needed but we should get rid of this.

knative/describer.go
zroubalik: This is the constructor from `knative/client.go` https://github.com/boson-project/faas/pull/167/files/62c488923ae8ce21b3d6b49d49e80cc27282763d#diff-cc862f411ab104c6c73b4f523a13b5a516a6ac6ceaf8edd7bcc664ec12b33889R20

knative/client.go
```diff
@@ -28,16 +66,5 @@ func NewClient(namespace string, verbose bool) (clientservingv1.KnServingClient,
 		p.Output = &bytes.Buffer{}
```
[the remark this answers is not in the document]
rhuss: Let me check tomorrow, but I don't think that the service API prints anything out on its own. 

What actuallly do you want to capture ?

zroubalik:
@rhuss I have updated the PR according to your comments, PTAL :)

knative/client.go
```diff
@@ -1,43 +1,68 @@
...
+	eventingv1beta1 "knative.dev/eventing/pkg/client/clientset/versioned/typed/eventing/v1beta1"
+	servingv1 "knative.dev/serving/pkg/client/clientset/versioned/typed/serving/v1"
 )
 
 const (
 	DefaultWaitingTimeout = 60 * time.Second
 )
 
-func NewClient(namespace string, verbose bool) (clientservingv1.KnServingClient, io.Writer, error) {
+func NewServingClient(namespace string) (clientservingv1.KnServingClient, error) {
```
rhuss, line 18: Looks good ! As a future step we might consider to take over those constructors methods to the knative client, too, so that the user doesn't have to specify the config on her own.

knative/client.go
```diff
@@ -1,43 +1,68 @@
...
 
 import (
-	"bytes"
 	"fmt"
-	"io"
-	"os"
 	"time"
 
 	"k8s.io/client-go/tools/clientcmd"
-	"knative.dev/client/pkg/kn/commands"
```
rhuss, line 11: 👍

rhuss (approved):
Looks good to me, thanks !

/approve
/lgtm

PR #172, opened 2020-10-13
feat: pass CloudEvent data as first parameter for Node.js event fns

lance:
The 0.4.0 version of faas-js-runtime extracts the CloudEvent data from
an incoming event and provides that as the first parameter when invoking
a function which receives a CloudEvent. This commit bumps to that version
as well as improves the overall readability and code documentation for the
Node.js CloudEvent function.


zroubalik (approved):
LGTM

PR #173, opened 2020-10-14
chore: tweak output & verbose messages

zroubalik:

Outputs of commands were quite incositent, trying to fix that.

### DIFF
--- a/client.go
+++ b/client.go
@@ -93,6 +93,9 @@ type Describer interface {
 
 type Description struct {
 	Name          string         `json:"name" yaml:"name"`
+	Image         string         `json:"image" yaml:"image"`
+	KService      string         `json:"kservice" yaml:"kservice"`
+	Namespace     string         `json:"namespace" yaml:"namespace"`
 	Routes        []string       `json:"routes" yaml:"routes"`
 	Subscriptions []Subscription `json:"subscriptions" yaml:"subscriptions"`
 }
@@ -363,7 +366,7 @@ func (c *Client) Initialize(cfg Function) (err error) {
 	// TODO: Create a status structure and return it for clients to use
 	// for output, such as from the CLI.
 	if c.verbose {
-		fmt.Printf("OK %v %v\n", f.Name, f.Root)
+		fmt.Println("Function project created")
 	}
 	return
 }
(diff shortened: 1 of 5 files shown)


client.go
```diff
@@ -414,11 +420,17 @@ func (c *Client) Deploy(path string) (err error) {
 	}
 
 	// Push the image for the named service to the configured registry
+	if c.verbose {
+		fmt.Println("\nPushing Function image to the registry:")
```
lance, line 424: Why the ":" at the end of this message (as well as below on 432)?

cmd/describe.go
```diff
@@ -109,25 +110,43 @@ func newDescribeConfig(args []string) describeConfig {
...
 func (d description) Human(w io.Writer) error {
-	fmt.Fprintln(w, d.Name)
+	fmt.Fprintln(w, "Function name:")
+	fmt.Fprintf(w, "  %v\n", d.Name)
+	fmt.Fprintln(w, "Function is built in image:")
+	fmt.Fprintf(w, "  %v\n", d.Image)
+	fmt.Fprintln(w, "Function is deployed as Knative Service:")
+	fmt.Fprintf(w, "  %v\n", d.KService)
+	fmt.Fprintln(w, "Function is deployed in namespace:")
+	fmt.Fprintf(w, "  %v\n", d.Namespace)
```
lance, line 120: Much nicer!

cmd/describe.go
```diff
@@ -109,25 +110,43 @@ func newDescribeConfig(args []string) describeConfig {
...
-	return d.Plain(w)
+	return nil
 }
 
 func (d description) Plain(w io.Writer) error {
 	fmt.Fprintf(w, "NAME %v\n", d.Name)
+	fmt.Fprintf(w, "IMAGE %v\n", d.Image)
+	fmt.Fprintf(w, "KNATIVE SERVICE %v\n", d.KService)
+	fmt.Fprintf(w, "NAMESPACE %v\n", d.Namespace)
+
```
lance, line 141: Why all caps here (and below)? I know it's what was there before, but why?

lance (commented):
A couple of minor comments/questions. Otherwise LGTM.

client.go
zroubalik: I can remove that, but the output of `docker push/build` follows these lines

lance: Oh that makes sense then

cmd/describe.go
lance: I think let's be consistent. Just a suggestion.

lance: Oh sorry that you had to wait 2 hours for a simple answer... Yes I meant remove all caps to be consistent with other output strings.

lance (approved):

Issue #141, opened 2020-09-25
faas build & deploy should use SHA


zroubalik:
`faas build` should note at the end result the SHA e.g. in faas.yaml that then could be picked up by `faas deploy`

rhuss:
Quick question: How is this related to #140 ? Does this already solve the issue that is described in #140 ?

lance:
> Quick question: How is this related to #140 ? Does this already solve the issue that is described in #140 ?

Personally, I think it solves the issue well enough for dev preview. What do you think @zroubalik?

lance:
Hmm- taking a quick look at the Buildpack client, it doesn't seem that they expose the sha after the image is built. https://github.com/buildpacks/pack/blob/main/build.go#L175

I suppose after the build completes, we could query the docker daemon to get the image info and extract it from that. Hrm...

rhuss:
Yes, this getting the SHA typically involves an additional query (could be the docker daemon where the image has been build or also the registry where the image is stored). You would ask for information about the image with the specific tag.

rhuss:
Interestingly the build command also fetches the sha/id just for logging purpose: https://github.com/buildpacks/pack/blob/68dec3e856738048ca848f2addb3858a150e4411/build.go#L310 Not sure whether it can be exposed more easily

zroubalik:
I dealt with the exactly same problem yesterday, seems like docker query is the only viable option.

For Buildpacks, there's https://github.com/buildpacks/imgutil library for inspecting the image (but it is using docker internally) so I think we can use the docker directly.

Issue #140, opened 2020-09-25
Use unique name for built images


zroubalik:
`faas build` currently produces image named eg `docker.io/zroubalik/myfunction:latest`.

We should add some hash to the name, to make the image unique.

Usecase: 
Let’s say, I’d like to create 2 different functions, but both named myfunction (deployed into different namespaces). Current soluiton will result in conflict

rhuss:
The question is also whether could detect that conflict or whether it would just overwrite the other image. Using a random prefix doesn't help much either as this does not guarantee that there is no conflict (if you can't detect the conflict).

Just checking if the image exists in the registry might also not be good enough, as this could have been created just by a previous build from the current function.

rhuss:
I wonder how other container-based faas (e.g. openfaas or riff) deal with this situation. Maybe its worth to have a look at them, too.

PR #175, opened 2020-10-14
src: use image digest

zroubalik:

Using image Digest in the Knative Service. Haven't moved to Docker Go SDK, because with this library it is [quite hard ](https://github.com/moby/moby/issues/39377#issuecomment-503589708) to use some authentication 💩 

We can try to migrate to this lib post Dev Preview. Issue created: https://github.com/boson-project/faas/issues/179

Fixes: #141 
Fixes: #140

### DIFF
--- a/client.go
+++ b/client.go
@@ -42,7 +42,8 @@ type Builder interface {
 // Pusher of Function image to a registry.
 type Pusher interface {
 	// Push the image of the Function.
-	Push(Function) error
+	// Returns Image Digest - SHA256 hash of the produced image
+	Push(Function) (string, error)
 }
 
 // Deployer of Function source to running status.
@@ -423,7 +424,14 @@ func (c *Client) Deploy(path string) (err error) {
 	if c.verbose {
 		fmt.Println("\nPushing Function image to the registry:")
 	}
-	if err = c.pusher.Push(f); err != nil {
+	imageDigest, err := c.pusher.Push(f)
+	if err != nil {
+		return
+	}
+
+	// Store the produced image Digest in the config
+	f.ImageDigest = imageDigest
+	if err = writeConfig(f); err != nil {
 		return
 	}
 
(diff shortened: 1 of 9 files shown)


client.go
```diff
@@ -414,7 +415,14 @@ func (c *Client) Deploy(path string) (err error) {
...
 	// Push the image for the named service to the configured registry
-	if err = c.pusher.Push(f); err != nil {
+	imageDigest, err := c.pusher.Push(f)
+	if err != nil {
+		return
+	}
+
+	// Store the produced image Digest in the config
+	f.ImageDigest = imageDigest
+	if err = writeConfig(f); err != nil {
```
zroubalik, line 425: It is not needed to store the digest curretnly (the push is immediately followed by deploy), so I can remove this part. But I thought we might want to use that in the future, if we introduce a ways how to fire `push` `deploy-only` etc (as was discussed on Slack).

docker/runner.go
```diff
@@ -30,8 +30,8 @@ func (n *Runner) Run(f faas.Function) error {
 		return errors.New("please install 'docker'")
 	}
 
-	if f.Image == "" {
-		return errors.New("Function has no associated image.  Has it been built?")
+	if f.Image == "" || f.ImageWithDigest() == "" {
+		return errors.New("Function has no associated image and SHA356 hash. Has it been built?")
```
matejvasek, line 34: SHA356?

zroubalik: When 256 is not enough 😅  Good catch, thanks!

zroubalik: Fixed

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.14
 
 require (
 	github.com/buildpacks/pack v0.14.0
+	github.com/docker/docker v1.13.1
```
dsimansk, line 7: I don't see it used anywhere, just be aware that the version will be overridden by `replace` block below. In addition I was under impression that `v1.13.1` doesn't work with current buildpack version. 

Dependencies, dependencies, ... :)

zroubalik: Good catch, thanks! This is a leftover from previous iteration.  Fixed

lance (approved):

PR #183, opened 2020-10-20
fix: regenerate pkger files

### LINKED ISSUE #182: Quarkus template still exports "greet"
I'm not sure exactly what's going on here. The template itself exports `echo`, but when I run `faas init`, followed by `faas build` the classes directory contains an application.properties file that has the `greet` function as the export. Could this be an old buildpack is still being used?

```
❯ faas init -l quarkus -t events
Project path: /home/lanceball/demo/quarkus
Function name: quarkus
Runtime: quarkus
Trigger: events

~/demo/quarkus is 📦 1.0.0-SNAPSHOT via ☕ v11.0.8 
❯ faas build
Building image: 
A registry for Function images is required. For example, 'docker.io/tigerteam'.

Registry for Function images: lanceball

~/demo/quarkus is 📦 1.0.0-SNAPSHOT via ☕ v11.0.8 took 25s 
❯ rg greet
target/classes/application.properties
1:quarkus.funqy.export=greet
```

matejvasek:
resolves #182


zroubalik:
Should be the pkger generation part of the Makefile?

matejvasek:
@zroubalik  We have been thinking about that before. It's been decided not to do that, so developer don't have to install `pkger` on her/his machine. But given what problem it's causing we should put that to Makefile and stop versioning generated file after all.

zroubalik:
> @zroubalik We have been thinking about that before. It's been decided not to do that, so developer don't have to install `pkger` on her/his machine. But given what problem it's causing we should put that to Makefile and stop versioning generated file after all.

yeah, or we can run the `pkger` check only in the CI. Make the validation check in the PR, to see whether the files needed to be regenerated. (If I understand it correctly and not all code changes are causing the need for a regeneration).

lance (approved):

PR #184, opened 2020-10-20
fix: examples in readme

matejvasek:


matejvasek:
@lance I updated the readme so URI for `curl/HTTPie` is correct.

lance (approved):

PR #198, opened 2020-10-30
feat: change default runtime to Node.js HTTP

lance:
I did not implement a prompt if the user doesn't choose. It seems odd to me that a flag would be required. I'm happy to discuss this, however.

Fixes: https://github.com/boson-project/faas/issues/191

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -176,9 +176,9 @@ func TestExtensibleTemplates(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// Ensure that a file from that only exists in that template set was actually written 'json.go'
-	if _, err := os.Stat(filepath.Join(root, "json.go")); os.IsNotExist(err) {
-		t.Fatalf("Initializing a custom did not result in json.go being written to '%v'", root)
+	// Ensure that a file from that only exists in that template set was actually written 'json.js'
+	if _, err := os.Stat(filepath.Join(root, "json.js")); os.IsNotExist(err) {
+		t.Fatalf("Initializing a custom did not result in json.js being written to '%v'", root)
 	} else if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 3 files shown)


client_test.go
```diff
@@ -176,9 +176,9 @@ func TestExtensibleTemplates(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// Ensure that a file from that only exists in that template set was actually written 'json.go'
-	if _, err := os.Stat(filepath.Join(root, "json.go")); os.IsNotExist(err) {
-		t.Fatalf("Initializing a custom did not result in json.go being written to '%v'", root)
+	// Ensure that a file from that only exists in that template set was actually written 'json.js'
+	if _, err := os.Stat(filepath.Join(root, "json.js")); os.IsNotExist(err) {
```
matejvasek, line 180: I think this also will require `testdata/templates/boson-experimental/`. Or maybe we don't need to change this file at all since it's just a test.

lance: We don't really need to change this file (or the other in this test template) as they are just empty files with a `.go` extension. But the test relies on the default runtime which is changed to Node.js in this PR. It just seemed wrong for me to have files with `.go` extensions in a Node.js project, even if they are empty.

matejvasek:
```shell
testdata
├── example.com
│   └── region1
│       └── README.md
├── README.md
└── templates
    └── boson-experimental
        ├── go
        │   └── json
        │       ├── handle.go
        │       └── json.go
        └── quarkus
            └── http
                └── mvnw

8 directories, 5 files
```
The `json.js` file doesn't exists, maybe it should be created as part of this PR?

lance: It was! :wink:  https://github.com/boson-project/faas/pull/198/files#diff-41033a046e4d472b14c952e0c898d57d9f4c60e14c8165e0ab71177ab29cb46a

It just appears as a rename in git.

matejvasek: oh, my bad, I didn't notice that.

zroubalik (approved):
LGTM

Issue #189, opened 2020-10-29
[feat] More output on the build/deploy step


markusthoemmes:
**Describe the solution you'd like**
My `kn faas build` command was literally

```console
$ kn faas build
Building image: docker.io/markusthoemmes/node:latest
```

A spinner would've been neat as it has been sitting there for a while. A `Success` message would also be nice on... well... success.

zroubalik:
Discussed this with @lkingland, Implementing the spinner properly (so it is not eating the output) may take some time. So I'll proceed with printing some basic info messages, since we are under such a tight deadline.

And we will revisit the Spinner properly for all the commands as a post Dev Preview task.

PR #206, opened 2020-11-05
feat: Better output of build/deploy/delete commands

zroubalik:

Discussed this with @lkingland, Implementing the spinner properly (so it is not eating the output) may take some time. So I'll proceed with printing some basic info messages, since we are under such a tight deadline.

And we will revisit the Spinner properly for all the commands as a post Dev Preview task.

**Build**
```
./faas build -p func                                        
Building function image
Function image has been built, image: docker.io/zroubalik/func:latest
```

**Deploy**
```
 ./faas deploy -p func                                           
Building function image
Function image has been built, image: docker.io/zroubalik/func:latest
Pushing function image to the registry
Deploying function to the cluster
Function deployed at URL: http://func-default.apps.zroubali.devcluster.openshift.com
```

**Update**
```
./faas deploy -p func                                 
Building function image
Function image has been built, image: docker.io/zroubalik/func:latest
Pushing function image to the registry
Deploying function to the cluster
Function updated at URL: http://func-default.apps.zroubali.devcluster.openshift.com
```

**Delete**
```
./faas delete -p func
Removing Knative Service: func
```

Fixes: #189

### DIFF
--- a/client.go
+++ b/client.go
@@ -384,7 +384,7 @@ func (c *Client) Initialize(cfg Function) (err error) {
 // not contain a populated Image.
 func (c *Client) Build(path string) (err error) {
 
-	fmt.Println("Building Function image")
+	fmt.Println("Building function image")
 
 	f, err := NewFunction(path)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


lance (commented):
LGTM generally. Although, I think we are moving away from using using capital "F" for functions. As @rhuss noted in https://github.com/boson-project/faas/issues/187 

> If its to pronounce the concept, then consequently this should be "Function as a Service". But for readability reasons and to avoid a question mark when reading this, I suggest to keep it lower case

zroubalik:
@lance ok, fixed in the last commit

lance (approved):

PR #211, opened 2020-11-05
fix: return JSON in Node.js event template

lance:
The event template was just returning a string, but the default response
content type is application/json so browsers were failing to parse the string
as JSON.


matejvasek (approved):

templates/node/events/index.js
```diff
@@ -19,13 +19,15 @@ const { CloudEvent, HTTP } = require('cloudevents');
...
+ * @param {Context} context the invocation context
  * @param {Object} user the CloudEvent data. If the data content type is application/json
  * this will be converted to an Object via JSON.parse()
- * @param {Context} context the invocation context
  */
 function verifyUser(context, user) {
   if (!context.cloudevent) {
-    return 'No cloud event received';
+    return {
+      message: 'No cloud event received'
```
matejvasek, line 29: Just out of curiosity, what if it returned:
```js
return '"No cloud event received"'
```

lance: That would probably parse correctly by the browser.

zroubalik (approved):

PR #213, opened 2020-11-06
[WIP] Quarkus Templates

matejvasek:


matejvasek:
This allows to make changes to template files: e.g. set artifactId in pom.xml in accordance with function name.

Makefile
```diff
@@ -34,7 +34,7 @@ $(LINUX):
 $(WINDOWS):
 	env CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build -o $(WINDOWS) -ldflags "-X main.date=$(DATE) -X main.vers=$(VERS) -X main.hash=$(HASH)" ./cmd/$(BIN)
 
-test: test-binary test-node test-quarkus test-go
```
lance, line 37: Why remove quarkus tests during build?

lance (commented):
LGTM - do you intend to land this for Dev Preview?

matejvasek:
> LGTM - do you intend to land this for Dev Preview?

There is a lot of changes so I don't feel much confident. At leas I'll land the other Quarkus Templates PR>

Makefile
matejvasek: @lance the test is directly executing template as it was instantiated project, this won't work for Quarkus anymore because it uses golang templating, so some pre-processing is needed. For instance java files begin with `package {{.GroupId}}.{{.ArtifactId}};` which is not valid java.

lance:
That was my worry. That there were so many changes.￼

PR #220, opened 2020-11-06
fix: `delete`, `deploy, sub-commands respects func.yaml conf

matejvasek:


cmd/delete.go
```diff
@@ -52,7 +52,12 @@ func runDelete(cmd *cobra.Command, args []string) (err error) {
 		return fmt.Errorf("the given path '%v' does not contain an initialized Function.", config.Path)
 	}
 
-	remover, err := knative.NewRemover(config.Namespace)
+	ns := config.Namespace
```
matejvasek, line 55: note `config` in not config from `func.yaml` file, but from cmd flags

lance (commented):
LGTM - but as noted in Slack, should the default value (currently active in k8s config) be written to `func.yaml`? Honestly, it's pretty late in the game to make that change so maybe we just stick with this for dev preview.

zroubalik (approved):

PR #233, opened 2020-11-30
fix: Fix plugin version output

dsimansk:
@lance @rhuss 

The following code should retrieve and set the actual plugin's version based on go.mod definition. The required change/enhancement on kn's midstream is to use `require` correctly.

It seems a bit cleaner than setting another set of compile flags. Anyway the current set of flags is set through `main` package, so we'd need to change that to pass flags during kn's build.

E.g. from my hacks:

```go
require (
	github.com/boson-project/faas v0.9.0
)

replace (
	github.com/boson-project/faas => /home/dsimansk/Git/boson-project/faas
)
```
```bash
➜  client git:(release-v0.17.3) ✗ kn func version
v0.9.0

```

### DIFF
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -1,7 +1,6 @@
 package plugin
 
 import (
-	"fmt"
 	"github.com/boson-project/faas/cmd"
 	"knative.dev/client/pkg/kn/plugin"
 	"os"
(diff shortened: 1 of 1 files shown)


rhuss (approved):
looks good to me, but let's wait for @lance review.

lance (approved):
Seems like a good solution. Thanks @dsimansk !

zroubalik (approved):
LGTM

Issue #236, opened 2020-12-07
`run` shouldn't use image digest but should use image name reference


zroubalik:
There's a bug, `run` uses image digest to reference the image, but digest is being produced by `push` as a reference to an image in the external registry. 

We should use name instead.

If you run these commands in sequence you can see the problem:
```bash
kn func deploy      # this produces image with a Digest, let's say IMAGE1
kn func run         # this will run IMAGE1
kn func build       # this produces another image, eg. IMAGE2
kn func run         # this will use IMAGE1 and not IMAGE2 (it will take the one with a Digest)
```

PR #238, opened 2020-12-07
fix: use image name for run command

zroubalik:

Fixes: #236

### DIFF
--- a/docker/runner.go
+++ b/docker/runner.go
@@ -30,15 +30,15 @@ func (n *Runner) Run(f faas.Function) error {
 		return errors.New("please install 'docker'")
 	}
 
-	if f.Image == "" || f.ImageWithDigest() == "" {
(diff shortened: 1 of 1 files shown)


lance:
I don't understand why we are removing the check for a digest. Is it simply that we don't want to be exposing the digest to the end user?

zroubalik:
> I don't understand why we are removing the check for a digest. Is it simply that we don't want to be exposing the digest to the end user?

As I mentioned on the [issue](https://github.com/boson-project/faas/issues/236), the `run` shouldn't consume the digest (as it is being created once the image is being pushed to the external registry), because if we do `deploy` , then `build` and then `run`, the `run` will use image produced by `deploy` not the subsequent `build`.

lance:
> As I mentioned on the issue...

Oops! I didn't see the issue before asking. :)

lance (approved):

PR #241, opened 2020-12-09
fix: change image to use UBI8-minimal instead of alpine

lance:
Not sure I like this much. It increases the size of the image by about 8x.

Fixes: https://github.com/boson-project/faas/issues/219

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM golang:alpine as build
-RUN apk add make git gcc g++ 
+FROM registry.access.redhat.com/ubi8/ubi:8.2 as build
+RUN dnf install -y make git gcc gcc-c++ golang
 COPY . /src
 WORKDIR /src
 RUN make
 
-FROM alpine:latest
-RUN apk add --no-cache ca-certificates
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.2
+RUN microdnf install -y ca-certificates
 COPY --from=build /src/func /bin/
 ENTRYPOINT ["func"]


matejvasek:
Looks OK. But how is this image supposed to be used? as an builder? What I don't understand is `ENTRYPOINT ["func"]`, how is that useful?

matejvasek (approved):

lance:
> But how is this image supposed to be used? as an builder? What I don't understand is ENTRYPOINT ["func"], how is that useful?

No, it's not a builder. It's meant to be a means to run `func` without requiring that it be downloaded. But to be honest, I'm a bit confused myself. For example, `create` will create a project directory on the container.

```
❯ docker run --rm quay.io/boson/func create foo
Project path: /foo
Function name: foo
Runtime: node
Trigger: http
```

And even though I've got a valid kube context locally, `list` (and presumably other commands that work with kube) does not have access to that.

```
❯ docker run --rm quay.io/boson/func list
Error: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```

Maybe @lkingland can fill us in on his thinking here.

lkingland:
If  I am not understanding the question correctly, please say so and I'll try answering again.

This Dockerfile depcits a two-stage build.  The first stage is a container which includes all the extras necessary to build the final binary `func`.  This first-stage container, called the build container, is named `build`.  In the second stanza, the runtime container is created, which includes only the minimum dependencies necessary to run `func`, and `func` itself.  

The `ENTRYPOINT` line indicates that the default binary to run inside of the container is `func`.  Not including this line results in an error attempting to `docker run [myimage]`.   Using ENTRYPOINT rather than the similar CMD allows this default binary to be overridden using `--entrypoint`, which can be handy for debugging, should one want to have `--entrypoint /bin/sh` or similar.

It is very important that whichever base image is chosen for the build container is identical to that for the run container, as any binary built in the former will have libary links expecting certain paths that will cause ldd errors on run.

I also do not like the fact that we are using a container with 8x the overall size.  However perhaps it's a good call, especially if it has any kind of deliberate hardening?

lkingland:
Ah, I see the confusion now.  I thought we were discussing the Dockerfile used by the buildpacks.

Yes, this container is intended to be used as a way to run `func` (now `faas`) directly from a container image rather than building.  But It's probabaly going to require some volume mounts to work correctly now that we're creating directories.

lkingland:
On further consideration, is this even useful now?  The container was there as a stop-gap until we had the platform-specific binaries being built using automation.

lkingland (approved):
LGTM, but I am all for removing this container, as it is easier to use the new pre-built binaries (no need for cryptic volume mount parameters, etc).

rhuss:
@lball you have to volume-mount the relevant directories (project directory and `~/.kube`) into the container so that it is accessible from within the container.

rhuss:
The benefit of using ubi over alpine are:

* You automatically benefit from security fixes which I would expect to arrive more reliably for a Red Hat based platform then for a community-only driven project.
* You have proper DNS handling (something that alpine was weak on, at least some time ago, not sure that has been fixed)
* You use a "real" libc.so and not musl, which is slimer but, well, not as well test with many packages. There are subtle failure patterns then do not occur often but when they occur it's hard to debug.

Also, I wouldn't be too much concerned about the size of the image, as the image layers are cached and if there is already an image based on ubi loaded in the Docker daemon (maybe not so unlikely in a Red Hat env), then you don't have to download it again. However, when running in a CI context you don't have really control of this caching, so this might lead to grossly different run time (those with the base image cached or when they are need to be fetched from the registry if running on a fresh node)

Dockerfile
```diff
@@ -1,10 +1,10 @@
-FROM golang:alpine as build
-RUN apk add make git gcc g++ 
+FROM registry.access.redhat.com/ubi8/ubi:8.2 as build
+RUN dnf install -y make git gcc gcc-c++ golang
```
rhuss, line 2: why is golang needed for the ubi image but not for the alpine image ?

lkingland:
> The benefit of using ubi over alpine are:
> 
>     * You automatically benefit from security fixes which I would expect to arrive more reliably for a Red Hat based platform then for a community-only driven project.
> 
>     * You have proper DNS handling (something that alpine was weak on, at least some time ago, not sure that has been fixed)
> 
>     * You use a "real" libc.so and not musl, which is slimer but, well, not as well test with many packages. There are subtle failure patterns then do not occur often but when they occur it's hard to debug.
> 
> 
> Also, I wouldn't be too much concerned about the size of the image, as the image layers are cached and if there is already an image based on ubi loaded in the Docker daemon (maybe not so unlikely in a Red Hat env), then you don't have to download it again. However, when running in a CI context you don't have really control of this caching, so this might lead to grossly different run time (those with the base image cached or when they are need to be fetched from the registry if running on a fresh node)

Thanks for this Roland; those are good reasons.

Dockerfile
lkingland: Because the alpine image indicated is actually the variant maintained by the golang team, which is Alpine _plus_ golang (note the full name `FROM golang:alpine`

zroubalik (approved):

lance:
Closing, as I think we should remove the `func` image altogether.

Issue #237, opened 2020-12-07
Golang events triggered function template suggests wrong function signatures


zroubalik:
`create my_function -l go -t events` produces handle.go template with these function signatures in the comment:
```go
// Supported Function signatures:
//   func()
//   func() error
//   func(context.Context)
//   func(context.Context) error
//   func(cloudevents.Event)
//   func(cloudevents.Event) error
//   func(context.Context, cloudevents.Event)
//   func(context.Context, cloudevents.Event) error
//   func(cloudevents.Event, *cloudevents.EventResponse)
//   func(cloudevents.Event, *cloudevents.EventResponse) error
//   func(context.Context, cloudevents.Event, *cloudevents.EventResponse)
//   func(context.Context, cloudevents.Event, *cloudevents.EventResponse) error
```

This is wrong, as cloudevents-sdk expects the response event as the return value of the function

zroubalik:
CC @lkingland

PR #246, opened 2020-12-10
fix: correct fn signatures in Go Events template

zroubalik:

Used signature from Cloudevents Go SDK: https://github.com/cloudevents/sdk-go/blob/e35f432ede703309a1bd4c77a06f4910b3b23582/v2/client/receiver.go#L40-L55
(ust replaced `transport.Result` return value to `error` as `transport.Result` is just [wrapper](https://github.com/cloudevents/sdk-go/blob/5eff2b826cd8f26255f9f6963df66b5a6163a932/v2/protocol/result.go#L9) around `error` type and this is imho more obvious for function developers)

Fixes: #237

### DIFF
--- a/templates/go/events/handle.go
+++ b/templates/go/events/handle.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"os"
 
-	cloudevents "github.com/cloudevents/sdk-go/v2"
+	event "github.com/cloudevents/sdk-go/v2"
 )
 
 // Handle a CloudEvent.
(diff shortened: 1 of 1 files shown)


templates/go/events/handle.go
```diff
@@ -10,18 +10,18 @@ import (
...
-//   func(context.Context, cloudevents.Event) error
-//   func(cloudevents.Event, *cloudevents.EventResponse)
-//   func(cloudevents.Event, *cloudevents.EventResponse) error
-//   func(context.Context, cloudevents.Event, *cloudevents.EventResponse)
-//   func(context.Context, cloudevents.Event, *cloudevents.EventResponse) error
+// * func()
+// * func() error
+// * func(context.Context)
+// * func(context.Context) error
+// * func(event.Event)
```
matejvasek, line 17: I see that cloudevents->event, is that correct given we use renaming import?

lance: > I see that cloudevents->event, is that correct given we use renaming import?

We could just change the import to `event`, right?

zroubalik: You are right, that should be unified.  I like changing the import to shorter `event`.

zroubalik: Updated

lance (approved):

PR #248, opened 2020-12-10
fix: change --format flag to --output for list and describe commands

lance:
Fixes: https://github.com/boson-project/func/issues/223

### DIFF
--- a/cmd/describe.go
+++ b/cmd/describe.go
@@ -18,10 +18,10 @@ import (
 func init() {
 	root.AddCommand(describeCmd)
 	describeCmd.Flags().StringP("namespace", "n", "", "Namespace of the function. By default, the namespace in func.yaml is used or the actual active namespace if not set in the configuration. (Env: $FUNC_NAMESPACE)")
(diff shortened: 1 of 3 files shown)


zroubalik (changes requested):
This should be changed as well: https://github.com/boson-project/func/blob/main/docs/commands.md#describe

lance:
> This should be changed as well: https://github.com/boson-project/func/blob/main/docs/commands.md#describe

@zroubalik done - ptal

matejvasek:
Why did go.mod and files in template change?

lance:
> Why did go.mod and files in template change?

Hmm - strange. I will revert those. I think there might be some problem when testing where `package.json` is updated with a point release dependency, which changes `package-lock.json`, and ultimately `pkged.go`.

lance:
@matejvasek fixed

lance:
@matejvasek yeah - `npm install` was updating package-lock.json. See https://github.com/boson-project/func/pull/251

zroubalik (approved):

PR #260, opened 2021-02-09
Improve parsing using RE

matejvasek:


matejvasek:
What version of golang is the CI using?

matejvasek:
> has no field or method SubexpIndex
This should be in golang 1.14 I believe.

docker/pusher.go
```diff
@@ -68,26 +67,15 @@ func (n *Pusher) Push(f bosonFunc.Function) (digest string, err error) {
 	return
 }
 
+var digestRE = regexp.MustCompile(`digest:\s+sha256:(?P<Digest>\w{64})`)
```
lance, line 70: I don't really understand this syntax `?P<Digest>\w{64}`, but otherwise lgtm.

lance (approved):

docker/pusher.go
matejvasek: `?P<Digest>` names the capture so it can be referenced by that instead of non-descriptive index 1.
Unfortunately the API doesn't work on CI for some reason. `docker/pusher.go:71: digestRE.SubexpIndex undefined (type *regexp.Regexp has no field or method SubexpIndex)`

matejvasek: https://github.com/boson-project/func/pull/260/commits/82aa1c491b02edb413a755849dab523cd9b357be#diff-c2e8a93a8d2fc1d7459fd91cd55e1afb9aa945ca7b77dd00a2dfa7a24df87bd5R70-R79

PR #263, opened 2021-02-12
Fix digest regex

matejvasek:


lkingland (approved):
Nice to see a test in there as well.  LGTM!

Issue #188, opened 2020-10-29
[feat] Support for podman


cardil:
**Is your feature request related to a problem? Please describe.**
On my Fedora box with podman installed I'm unable to use FaaS. I've been using podman happily for more than a year without any trouble. I can do all kinds of stuff with podman and I have never been forced to have docker installed. I would like to use faas without hasle and security pitfalls of having Docker daemon running.

FaaS relying on Docker socket is in conflict with a RH narration of ["no big fat daemons"](https://av.tib.eu/media/43121).

**Describe the solution you'd like**
FaaS should call out to OS installed OCI tool like podman. Things will even work if faas would just call `docker ...` process as there is a podman-docker package on Fedora.

Also, there is an option to vendor https://github.com/containers/image (this one is used by buildah, podman, and skopeo, and have most of the needed code) and use it to build images without any external tool being installed.

**Describe alternatives you've considered**
I need to run a VM with Docker installed to use faas, because of this. I can't install Docker and Podman on the same machine because of cgroups v2 incompatibility.

matejvasek:
This depends on https://github.com/boson-project/func/issues/179

matejvasek:
@cardil I've chosen path of using `podman` as a `docker` deamon: `podman system service`.
User can do:
```
$ podman system service --time=0 tcp:0.0.0.0:1234 & # let run in background or another terminal
$ export DOCKER_HOST=tcp://127.0.0.1:1234
$ func build -v
```

cardil:
But, that's not how Podman is supposed to work. Podman was done to get rid of a daemon. `kn func` should support podman as it is, not by calling an unsecured public port. If you publish that service to `tcp:0.0.0.0:1234` anyone could connect to your machine and do anything you can do on it - that's quite a **BIG** vulnerability. `kn func` shouldn't require opening such vulnerability to operate on RHEL 8/Fedora machines.

A minimal thing would be to support unix sockets (default: `unix:/$XDG_RUNTIME_DIR/podman/podman.sock`) or better to just call podman as a library (libpod) or as child executable.

matejvasek:
@cardil

That was bad example from me, you don't have to run `tcp:0.0.0.0:1234` it would work with `tcp:127.0.0.1:1234` too.

Also the Unix socket works, but you need to setup proper policy (or disable labeling for containers) when using SELinux.
We are looking into it; see https://github.com/containers/podman/discussions/9860.

[...]

PR #266, opened 2021-02-26
fix: using custom docker daemon (e.g podman)

matejvasek:
The `func` binary is now aware of DOCKER_HOST environment variable.


matejvasek:
@dsimansk please check if the changed dependencies don't affect integration with `kn`.

matejvasek:
resolves #188

zroubalik (approved):
Looking good, though I am not sure if this fully solves #188, because for pushing the image, docker binary is still being used.

matejvasek:
@zroubalik good catch, we should call docker API client there, not binary.

matejvasek:
@zroubalik now I recall it's not that easy isn't it?

lance:
> now I recall it's not that easy is it?

Looking at the [docs](https://pkg.go.dev/github.com/docker/docker/client#Client.ImagePush) and a [blog post](https://www.loginradius.com/blog/async/build-push-docker-images-golang/), it seems easy enough. We would just need to handle username and password directly in our CLI instead of relying on the client to be logged in.

matejvasek:
>  We would just need to handle username and password directly in our CLI instead of relying on the client to be logged in.

This make sense. Especially if the user uses both `docker` and `podman`, since the user might be logged in only in one of them (or even might be logged under different usernames in `podman` and `docker`).

dsimansk:
> @dsimansk please check if the changed dependencies don't affect integration with `kn`.

@matejvasek sorry I missed that. Well I'm pretty sure something will brake, because next kn 0.20 is not on k8s `19.x` yet.

Issue #234, opened 2020-12-03
Allow envvars in the `func.yaml` config.


matejvasek:
I think it might be useful to be able to use envvars in `func.yaml`

Sample:
```yaml
envVars:
  SOME_SECRET: {{env.API_KEY}}
```

zroubalik:
@matejvasek this could be closed, right?

matejvasek:
@zroubalik actually no, this is not about setting evvars to ksvc, but about using in config.

matejvasek:
e.g. in java property file you can write
```
my.prop=${env:JAVA_HOME}
```

matejvasek:
So you can fill envvar from envvar.

matejvasek:
@zroubalik Does this make sense?

zroubalik:
@matejvasek gotcha :)

zroubalik:
Referencing a Kubernetes Secret would be nice to have to

matejvasek:
> Referencing a Kubernetes Secret would be nice to have too

Absolutely, but it will decrease level of abstraction from k8s.

zroubalik:
> > Referencing a Kubernetes Secret would be nice to have too
> 
> Absolutely, but it will decrease level of abstraction from k8s.

Yeah, but Secrets are fundamental building blocks heavily used even by non k8s experienced users. Many times, Secrets with creds are preconfigured by admins or ops and devs are consuming them.

matejvasek:
Ian just ask me about such a feature like this.

matejvasek:
@boson-project/core what syntax do we prefer `${env.NAME}` or `{{evn.NAME}}` or something else?

lance:
@matejvasek I don't personally have a preference. Whatever is easiest to code a parser for I guess. :smile:

PR #267, opened 2021-02-26
feat: usage of local evnvvar in func cfg file

matejvasek:
User now can use syntax like:

```yaml
envVars:
  SOME_SECRET: {{env.API_KEY}}
```

### DIFF
--- a/knative/deployer.go
+++ b/knative/deployer.go
@@ -2,6 +2,8 @@ package knative
 
 import (
 	"fmt"
+	"os"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


matejvasek:
resolves #234

zroubalik (commented):
LGTM, but we should document this, before merging this PR

lance (commented):
LGTM except I agree with @zroubalik that there should be some documentation added.

matejvasek:
What is good place to put documentation about this?

lance:
Good question @matejvasek. I suppose we could actually start filling in some of the blanks in the docs stubs here. Maybe add something to the "cookbook" here: https://github.com/boson-project/func/blob/main/docs/functions_cookbook.md. It might end up being moved somewhere else, but that's probably as good a place as any for now.

zroubalik:
Yeah agree and other good place to document this would be in the section describing the "spec" of the cfg file.

lance:
> Yeah agree and other good place to document this would be in the section describing the "spec" of the cfg file.

I've just started to add docs to https://github.com/openshift-knative/docs since this is where everything lives for dev preview and I was just thinking yesterday that those docs needed a `func.yaml` configuration reference doc. That's probably a bigger task than just documenting this particular feature though. I don't want this PR to be held up by something that big. How about we land this and track the documentation here: https://github.com/openshift-knative/docs/issues/51

lance:
Uhh nevermind. Put that doc here, I suppose.

lkingland (approved):
/lgtm
I am interested to know how the documentation process ends up working.  I think we need to have the docs added preferentially to the Boson Project docs before being imported (and tweaked as necessary) for the OpenShift deployment?

zroubalik:
> Regarding docs, I am interested to know how the documentation process ends up working. I think we need to have the docs added preferentially to the Boson Project docs before being imported (and tweaked as necessary) for the OpenShift deployment?

+1 that's my undestanding as well

zroubalik:
I think the best place for Docs for this feature is the (unfortunately not yet existing) Guide mentioned in this PR: https://github.com/boson-project/functions/pull/35/files#diff-406771b961705379986973d48ea9957ca8e1408fde7f33a320ac791b4c98c9edR22

matejvasek:
@lance can I merge this now?

lance:
> @lance may I merge this now?

Yes - please!

PR #273, opened 2021-03-04
chore: Update Knative dependencies to 0.20

dsimansk:
Per title bringing the usual version alignment.

@matejvasek was the update to k8s `0.19.x` required by something specific that may break with earlier version?


rhuss (commented):
/lgtm
/approve

rhuss:
oops, I might have merged too early and overlooked the question to @matejvasek . Apologies if this caused troubles.

Issue #270, opened 2021-03-01
Changing the state of a deployed function with kn CLI causes subsequent deployments to fail


lance:
Using the kn service update command to change the state of an environment variable after deploying the function causes 'func deploy' to subsequently fail after later code changes.

Deploying function to the cluster results in the following error:

```
Error: knative deployer failed to update the service: giving up after 3 retries: admission webhook "validation.webhook.serving.knative.dev" denied the request: validation failed: Saw the following changes without a name change (-old +new): spec.template.metadata.name
{*v1.RevisionTemplateSpec}.Spec.PodSpec.Containers[0].Image:
	-: "quay.io/ilawson/battleships_win@sha256:d84e62776b884e3b0050f7ab0454a8a56373a8b9f1fa3c7bc89f29c9dbad5c1b"
	+: "quay.io/ilawson/battleships_win@sha256:6fe85cdbfe413df4172caa33983dbdee2beb8af4cffc610c5fd4a80c597c9432"
{*v1.RevisionTemplateSpec}.Spec.PodSpec.Containers[0].Env[0].Value:
	-: "20210225T152647"
	+: "20210301T134645"
{*v1.RevisionTemplateSpec}.Spec.PodSpec.Containers[0].Env[1].Value:
	-: "http://watchman-battleships-backend.apps.summit-dev-aws.m7o7.p1.openshiftapps.com/watch"
	+: "http://watchman-jsonquarkussandbox.apps.summit-dev-21.k9pw.p1.openshiftapps.com/watch"
```

According to @markusthommes the error is caused by generating a new revision name in the client during kn service update.

Related: https://github.com/knative/client/issues/994
Related: https://github.com/knative/client/pull/1240

matejvasek:
I see @rhuss resolved this for `kn`. Shouldn't just updating our `kn` dependency help?

matejvasek:
I stumbled on BYO revison once too https://knative.slack.com/archives/C9JP909F0/p1600876741003800

matejvasek:
Now I see you have trouble with `kn` not `func`. Why is this here.

matejvasek:
Back then we also used client-side generated revision, but it has been removed d5839ea6c1e84e843ad643cc0611a82e2e6d2399.

lance:
> Now I see you have trouble with kn not func. Why is this here.

The issue as I understand it is that if a function is updated via `kn` then later updated with `kn func deploy` the latter fails.

zroubalik:
[...]

PR #274, opened 2021-03-04
fix: server-side generated Revision name during the update

zroubalik:

Removing the name so the k8s server can fill it in with generated name, this prevents conflicts in Revision name when updating the KService from multiple places.

Fixes: #270


lance (approved):
LGTM - thanks for hopping on this so quickly!

Issue #272, opened 2021-03-04
Support deploying without implicitly building


lkingland:
Currently, the deploy step (both CLI and client library) implicitly build.

This default behavior should be updated such that the client library _either_ builds or deploys, depending on the `Client` method invoked, and the `func` CLI's `deploy` subcommand updated to invoke Build by default, which can be disabled via flag.

Errors will be expected when attempting to deploy an unbuilt Function.  This potentially may require the inclusion of an explicit --image flag to force completion, which should probably also update the Function's func.yaml if provided.

lance:
Fixed in https://github.com/boson-project/func/pull/279

PR #278, opened 2021-03-10
deploy --build option

lkingland:
Per Issue #272 , this PR adds the option to `deploy --build=false` such that a prebuilt image can be used instead.


lkingland:
This PR pulled in all the pending commits from the other branch as well (it had not yet been merged)  Re-creating to keep things clean.

PR #279, opened 2021-03-10
deploy --build option

lkingland:
Per Issue #272 , this PR adds the option to `deploy --build=false` such that a prebuilt image can be used and building bypassed.


lkingland:
> LGTM, this should be reflected in the docs though

Thanks; `docs/commands.md` is now also updated, which I believe is the only place

zroubalik (approved):

PR #281, opened 2021-03-12
fix: `func run` now uses Docker API, not binary

matejvasek:

### DIFF
--- a/cmd/func/main.go
+++ b/cmd/func/main.go
@@ -1,7 +1,11 @@
 package main
 
 import (
+	"context"
 	"github.com/boson-project/func/cmd"
+	"os"
+	"os/signal"
+	"syscall"
 )
 
 // Statically-populated build metadata set
@@ -9,6 +13,17 @@ import (
 var date, vers, hash string
 
 func main() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+
+	go func() {
+		<-sigs
+		cancel()
+	}()
+
 	cmd.SetMeta(date, vers, hash)
-	cmd.Execute()
+	cmd.Execute(ctx)
 }
--- a/docker/runner.go
+++ b/docker/runner.go
@@ -1,11 +1,18 @@
 package docker
 
 import (
-	"errors"
+	"context"
 	"fmt"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/pkg/stdcopy"
+	"github.com/docker/go-connections/nat"
+	"github.com/pkg/errors"
 	"os"
-	"os/exec"
 	"strings"
+	"time"
+
+	"github.com/docker/docker/client"
 
 	bosonFunc "github.com/boson-project/func"
 )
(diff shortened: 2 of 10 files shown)


docker/runner.go
```diff
@@ -22,52 +29,99 @@ func NewRunner() *Runner {
 }
 
 // Run the function at path
-func (n *Runner) Run(f bosonFunc.Function) error {
-	// Check for the docker binary explicitly so that we can return
-	// an extra-friendly error message.
-	_, err := exec.LookPath("docker")
+func (n *Runner) Run(ctx context.Context, f bosonFunc.Function) error {
```
matejvasek, line 32: IMHO we should add `context.Context` as first parameters to all api functions (e.g. Deploy, Build ...) 
@boson-project/core what do you think?

docker/runner.go
```diff
@@ -22,52 +29,99 @@ func NewRunner() *Runner {
...
 
-	for name, value := range f.EnvVars {
-		if !strings.HasSuffix(name, "-") {
-			args = append(args, fmt.Sprintf("-e%s=%s", name, value))
+	conf := &container.Config{
+		Env:          envs,
+		Tty:          false,
+		AttachStderr: true,
+		AttachStdout: true,
+		AttachStdin:  false,
```
matejvasek, line 65: Only output from function is forwared, no input is sent to the function. I think this is OK. User is not expected to receive anything on stdin anyway. Output is useful for logging.

cmd/func/main.go
```diff
@@ -1,14 +1,29 @@
...
+	"os/signal"
+	"syscall"
 )
 
 // Statically-populated build metadata set
 // by `make build`.
 var date, vers, hash string
 
 func main() {
+	ctx, cancel := context.WithCancel(context.Background())
```
matejvasek, line 16: I wonder how this would play with `kn` plugin system.
@rhuss would the plugin catch the signal, or `kn` would "steal" it?

matejvasek: Well this code is probably not executed in plugin mode at all, am I right?

plugin/plugin.go
```diff
@@ -21,6 +24,17 @@ func (f *funcPlugin) Name() string {
 }
 
 func (f *funcPlugin) Execute(args []string) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	sigs := make(chan os.Signal, 1)
```
matejvasek, line 30: @rhuss @dsimansk
I wonder how this would play with kn plugin system.
Would the plugin catch the signal, or kn would "steal" it?

docker/runner.go
lkingland: You are probably correct; and they should be added as soon as possible, while the API is still easy to change

docker/runner.go
```diff
@@ -22,52 +29,103 @@ func NewRunner() *Runner {
...
+	defer cancel()
+
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
 	if err != nil {
-		return errors.New("please install 'docker'")
+		return errors.Wrap(err, "failed to create docker api client")
 	}
 
-	if f.Image == "" {
-		return errors.New("Function has no associated Image. Has it been built?")
```
lkingland, line 34: I'm not seeing this check for an empty `f.Image` in the new version.  Do we no longer want to explicitly error in this case?

docker/runner.go
```diff
@@ -22,52 +29,103 @@ func NewRunner() *Runner {
...
+	envs := make([]string, 0, len(f.EnvVars)+1)
+	for name, value := range f.EnvVars {
+		if !strings.HasSuffix(name, "-") {
+			envs = append(envs, fmt.Sprintf("%s=%s", name, value))
+		}
```
lkingland, line 45: I can't seem to remember why we need to explicitly exclude environment variables with a trailing dash.  Can you remind me (and perhaps add a comment and/or a test in case I forget again)?

lkingland (approved):
/lgtm

It's so nice to see another gnarly binary dependency dropped in favor of a proper API integration. Such a great step!

I am also insterested to know how being a plugin affects signal handling, and am looking forward to Roland's response here if he has the time to chime in.

cmd/func/main.go
rhuss: Right, `kn` only calls the methods from the `Plugin` interface.

rhuss: Therefore it is important that you only call code in `main()` that is important for standalone execution, anything needed for initializing the plugin should go into the shared code

docker/runner.go
matejvasek: good catch, we still should do the check

plugin/plugin.go
matejvasek: this seems to work

docker/runner.go
matejvasek: We need some way to tell an envvar was explicitly unset. Since envvar name must not contain dash, it can be (ab)used to indicate that envvar is unset (I took inspiration from `kn`).

plugin/plugin.go
matejvasek: looks like `kn` is not explicitly handling sigter or sigint

zroubalik (approved):
LGTM

PR #282, opened 2021-03-15
fix: `func deploy` uses Docker API, not binary

matejvasek:


docker/pusher.go
```diff
@@ -23,46 +29,66 @@ func NewPusher() *Pusher {
...
+		return "", errors.Errorf("invalid image name: %q", f.Image)
+	}
+	registry := parts[0]
+
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create docker api client")
+	}
+
+	credentials, err := config.GetAllCredentials(nil)
```
lance, line 49: Not knowing the Docker API, how does this work? Does it just read your current `~/.docker/config.json`? What happens if a config doesn't exist?

docker/pusher.go
```diff
@@ -23,46 +29,66 @@ func NewPusher() *Pusher {
...
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get credentials")
+	}
+
+	var opts types.ImagePushOptions
+
+	c := credentials[registry]
+	opts.RegistryAuth = base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf(`{"username":"%s","password":"%s"}`, c.Username, c.Password)))
```
lance, line 57: Hmm... my docker config looks like this:

```
{
  "HttpHeaders": {
    "User-Agent": "Docker-Client/19.03.13 (linux)"
  },
  "auths": {
    "https://index.docker.io/v1/": {
      "auth": "REDACTED"
    },
    "quay.io": {
      "auth": "REDACTED"
    },
    "registry.svc.ci.openshift.org": {
      "auth": "REDACTED"
    }
  }
}
```
Will this work? Maybe I'm making too many assumptions about what the docker API is doing...

docker/pusher.go
```diff
@@ -23,46 +29,66 @@ func NewPusher() *Pusher {
...
 	}
 
-	// set up the command, specifying a sanitized project name and connecting
-	// standard output and error.
-	cmd := exec.Command("docker", "push", f.Image)
+	parts := strings.Split(f.Image, "/")
+	if len(parts) < 1 {
+		return "", errors.Errorf("invalid image name: %q", f.Image)
+	}
+	registry := parts[0]
```
lance, line 42: Currently, because the docker CLI supports it, we accepted just the namespace for the docker.io registry. For example `lanceball` instead of `docker.io/lanceball`. With this change, does that default to docker.io still work? At first glance, it doesn't seem like it would, because the value of `registry` here has to match what's in `~/.docker/config.yaml`.

docker/pusher.go
matejvasek: Fixed: it checks docker's and podman's config files. If the user is not logged in one of those two, then the user is prompted for credentials. Is that good @lance ?

docker/pusher.go
matejvasek: Fixed: if the `namespace/image:tag` format is used then `docker.io` registry is assumed.

docker/pusher.go
matejvasek: Fixed: if the `namespace/image:tag` format is used then `docker.io` registry is assumed.

docker/pusher.go
matejvasek: It should. Please try it out.

matejvasek:
@lance PTAL

lance (commented):
@matejvasek thanks for making those changes! Overall, it looks good. But I wonder about the fact that the library code is prompting the user, and reading a config file outside of the context of the project. I think this could be bumping up against the separation of concerns between the client and the library. It might introduce a little complexity, but should the caller of `NewPusher()` ([cmd/deploy.go](https://github.com/boson-project/func/blob/8a91cac6cc78b5cf56d5158f3eb03a4076a34ffe/cmd/deploy.go#L99)) be providing the registry name and credentials?

matejvasek:
> @matejvasek thanks for making those changes! Overall, it looks good. But I wonder about the fact that the library code is prompting the user, and reading a config file outside of the context of the project. I think this could be bumping up against the separation of concerns between the client and the library. It might introduce a little complexity, but should the caller of `NewPusher()` ([cmd/deploy.go](https://github.com/boson-project/func/blob/8a91cac6cc78b5cf56d5158f3eb03a4076a34ffe/cmd/deploy.go#L99)) be providing the registry name and credentials?

definitely worth considering

client_int_test.go
```diff
@@ -155,7 +175,10 @@ func newClient(verbose bool) *boson.Client {
 	builder := buildpacks.NewBuilder()
 	builder.Verbose = verbose
 
-	pusher := docker.NewPusher()
+	pusher, err := docker.NewPusher()
+	if err != nil {
+		panic(err)
```
matejvasek, line 180: The other constructors call have comment like `TODO: remove error from X constructor`, but is that true? Shouldn't they return error? @lkingland

matejvasek:
@lance PTAL

cmd/deploy.go
```diff
@@ -127,12 +138,93 @@ func runDeploy(cmd *cobra.Command, _ []string) (err error) {
...
+		return result, err
+	}
+	password := string(bytePassword)
+
+	result.Username, result.Password = username, password
+
+	return result, nil
+}
+
+func getPassword(ctx context.Context) ([]byte, error) {
```
matejvasek, line 178: getPassword, getUserName -- I needed to make this in order to make those operations cancellable. Usually blocking read can be cancelled by closing the reader. For some reason this is not working with `os.Stdin`.

docker/pusher.go
```diff
@@ -2,67 +2,142 @@ package docker
...
-	var stderr bytes.Buffer
-	cmd.Stderr = &stderr
-
-	// Run the command, echoing captured stderr as well ass the cmd internal error.
-	err = cmd.Run()
-	if err != nil {
-		// TODO: sanitize stderr from docker?
-		err = fmt.Errorf("%v. %v", stderr.String(), err.Error())
+	decoder := json.NewDecoder(r)
+	li := logItem{}
```
matejvasek, line 116: Output of this is not so nice as of `docker` cli, but I think that's OK. I would need to deal with cursor moving I don't want to spend time on that.

client_int_test.go
lkingland: This is just a paradigm which eases the object lifecycle;  By providing constructors with only _data_ necessary for instantiation (sometimes references, if absolutely necessary), and then allowing invocation to be the phase which causes the error.  This can be a nice way to clean up an API. 

It's most certainly not a requirement, and more a note to myself to push things around a bit.  I'll try to whip up a PR to show you what I mean, as it's easier to show rather than explain, and then I can see if you agree with me or not!  Here's the idea, however:

Generally, the object lifecycle is Instantiate, Initialize, Invoke.  By keeping only data (or client references) in the _Instantiate_ and initialize stages (which are combined in the case of constructors), errors can be pushed out to _Invocation_.  This has knock-on effects with the simplification of API client users later in the form of only needing to deal with a single error path.  For a familiar example, think of SQL ORM libraries which tend to not error when you _pass_ them an errant SQL statement, but only when you try to `.Run` them.  The SQL statement is represented as an object which is instantiated, initialized with the raw SQL query, but it's only through invocation that it throws errors.  The lazy evaluation means that there's now only one error path in the code to deal with; the invocation path; which cleans up the code and logic overall.
Again to reiterate; this is just a stylistic choice, more of a note to myself, but perhaps worth the digression, I hope.

cmd/deploy.go
lkingland: It's nice to see the logic for interacting with the user where it belongs, here in `cmd` rather than the client library!

Usually this context creation and channel os.Signal listeners are something that is done at the very root of the process.  While I am not extremely familiar with Cobra, I would assume that this is one of its features, and we may want to change to using the functions in its API which include the context object.  Perhaps that will help.

In particular, I am thinking we may need to update to executing the command using [`ExecuteContext`](https://pkg.go.dev/github.com/spf13/cobra?utm_source=godoc#Command.ExecuteContext).  Perhaps by simply making this change, the command will be run using a cancelable context under the hood, and respond to events correctly?

docker/pusher.go
```diff
@@ -2,67 +2,142 @@ package docker
...
 	}
 
-	// set up the command, specifying a sanitized project name and connecting
-	// standard output and error.
-	cmd := exec.Command("docker", "push", f.Image)
+	var registry string
+	parts := strings.Split(f.Image, "/")
+	switch len(parts) {
+	case 2:
+		registry = "docker.io"
```
lkingland, line 75: Really nice to see we're preserving the docker.io default, at least for now, because it's the one which doesn't require manual intervention to expose the auto-created repository for a new Function.

One suggestion, could we move this "magic string" up top to be `const DefaultRegistry = "docker.io"`?  That helps when skimming later

lkingland (approved):
Looking good; a nice step forward

client_int_test.go
matejvasek: >For a familiar example, think of SQL ORM libraries which tend to not error when you pass them an errant SQL statement, but only when you try to .Run them.

I think that in case of SQL query they cannot simply and reliably validate the query. What is valid query will depend on SQL dialect on the server and on the DB schema. So I am not sure it's just design choice.

cmd/deploy.go
matejvasek: >In particular, I am thinking we may need to update to executing the command using ExecuteContext. 

We are already doing that. Now the context should be available via `cmd.Context()`.

client_int_test.go
matejvasek: >This has knock-on effects with the simplification of API client users later in the form of only needing to deal with a single error path.

Yes it simplify error handling a little bit, still I think it's better to fail fast. Imagine you try to initialize some object and the "constructor" won't return error, then you request some information from user via prompt (like username and pwd).  At the end the call fails and user was bothered unnecessarily. Also other libraries like `docker` or `kn` return errors from "constructors". I am using quotations since I am not sure if there is a thing like constructor in `go`.

cmd/deploy.go
matejvasek: see: https://github.com/boson-project/func/blob/main/cmd/run.go#L69

cmd/deploy.go
```diff
@@ -1,7 +1,14 @@
 package cmd
 
 import (
+	"bufio"
+	"context"
 	"fmt"
+	"github.com/containers/image/v5/pkg/docker/config"
+	"github.com/pkg/errors"
+	"golang.org/x/crypto/ssh/terminal"
```
zroubalik, line 9: Very very minor nit, I like when imports are ordered: 
1. go libs (fmt, context,..) 
2. third party libs (github.com/ory/viper, github.com/spf13/cobra) 
3. project libs (github.com/boson-project/func/knative).

I think that it is nice to follow the convetion at least for the first two points (1 and 2), currently it is quite mixed.

docker/pusher.go
```diff
@@ -2,67 +2,142 @@ package docker
 
 import (
 	"bytes"
-	"errors"
+	"context"
```
zroubalik, line 5: Same remark about imports as above.

zroubalik (commented):
Looking good, just minor nits on the imports.

cmd/deploy.go
matejvasek: I'll try run `goimports`.

PR #283, opened 2021-03-22
src: better ctx propagation

matejvasek:


zroubalik (approved):
LGTM

lance (approved):
LGTM - I guess when `kn` API is modified to accept a `Context` we'll need to revisit some of the commands that use that `kn` library code.

https://github.com/knative/client/issues/1272

PR #286, opened 2021-03-25
feat: add python templates

lance:
This commit adds support for Python by including basic function/test templates for events and http. The event function seems to highlight that creating a simple invocation framework (as opposed to the server.py in buildpacks repo) to help users navigate the event universe might be a good idea.


lance:
I should add runtime documentation to this before we land it.

lkingland (approved):
Looks good to me.  So nice to see we now support Python too; this is an important runtime.

zroubalik (approved):
Great job!

matejvasek:
Looks good to me.

matejvasek (approved):

templates/python/events/func.py
```diff
@@ -0,0 +1,17 @@
+from cloudevents.http import CloudEvent, to_binary
+
+# ---------------------------------------
+# Function template
+# context is a dictionary with the Flask request object and 
+# and a cloud_event key for incoming events
+# ---------------------------------------
+def main(context):
```
matejvasek, line 8: Just idea:
What about using type hints?
```python3
from werkzeug import Request
from typing import TypedDict


class Context(TypedDict):
    request: Request


def main(context: Context):
    """
    Doc here
    :param context: 
    :return: 
    """
    return {"message": "Howdy!"}, 200
```

matejvasek: also I believe that docstring for a function should be written inside `""" """` not using `#`.

matejvasek: To be clear: I am not strong proponent of typehints, but they may be useful for APIs.

lance:
> I discovered that this PR does not appear to work with kn func run, but works fine with docker run. I need to investigate this.

Hmm - this appears to be a problem with me locally. I'm seeing the same thing with all runtimes.

lance:
Never mind. Seems I needed this: https://github.com/boson-project/func/commit/7ed2e86d9672f285c1def490a3d325ceb9e8471f

templates/python/events/func.py
lance: Done. Thanks for the suggestions.

PR #287, opened 2021-03-26
Update pack dependency

matejvasek:


lance (approved):
LGTM - what do we gain with the update?

matejvasek:
@lance this is using released version, before we used commit hash.

PR #289, opened 2021-03-29
ci: run release action prior to checkout

lkingland:
In an attempt to fix the missing version on CI-built binaries, the repository is now checked out for the workflow after running the release action. This should ensure that the release action has fully completed, potentially adding any version tags to the codebase, prior to build's run which utilizes this version tag to populate the resultant binary's version number.


lance (approved):
lgtm - let's give it a spin

PR #290, opened 2021-03-31
chore: fix versioning in CI builds

lance:
For unknown reasons, the git tag for a release CI build via release-please
seems to happen after the build completes. It's a bit uncleear, but the
result is that we get v0.0.0 versions for released builds.

This commit modifies the Makefile to allow for an environment override of
the $VERS variable. Since the release-please-action documents `tag_name`
as output from the action, we can use this in CI to set the environment
variable.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,7 +43,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       # Create a release, or update the release PR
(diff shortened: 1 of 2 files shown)


lance:
From: https://github.com/lance/func/releases/tag/v0.14.0

```
❯ ./func_linux_amd64 --version
v0.14.0
```

lkingland (approved):
Looks good to me.

PR #292, opened 2021-04-07
docs: add template reference docs for Python

lance:

### DIFF
--- a/docs/guides/python.md
+++ b/docs/guides/python.md
@@ -11,9 +11,8 @@ Function name: fn
 Runtime: python
 Trigger: http
 
-~/src/fn via 🐍 v3.8.5
 ❯ tree
-.
+fn
 ├── func.py
 ├── func.yaml
 ├── requirements.txt
--- a/docs/guides/python.md
+++ b/docs/guides/python.md
@@ -5,14 +5,14 @@ looks like a typical Python project. Both HTTP and Event functions have the same
 template structure.
 
 ```
-~/src/pfunk
+~/src/fn
 ❯ kn func create -l python
-Project path: /home/lanceball/src/pfunk
-Function name: pfunk
+Project path: /home/developer/src/fn
+Function name: fn
 Runtime: python
 Trigger: http
 
-~/src/pfunk via 🐍 v3.8.5
+~/src/fn via 🐍 v3.8.5
 ❯ tree
 .
 ├── func.py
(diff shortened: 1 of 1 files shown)


docs/guides/python.md
```diff
@@ -0,0 +1,142 @@
...
+  return body, 200, headers
+```
+
+Note that functions may set both headers and response codes as secondary
+and tertiary response values from function invocation. 
+
+## Dependencies
+Developers are not restricted to the dependencies provided in the template
+`requirements.txt` file. Additional dependencies can be added as they would be
+in any other project by simply adding them to the `requirements.txt` file.
```
abrennan89, line 142: Can you add an example of what that txt file would look like with a python dependency added just for completeness? Or a command that would add it, do they add it manually, etc?

zroubalik: `requirements.txt` could be considered as a standard Python way of describing depenedencies, so every Python dev knows how to manage this file. IMHO I don't think it is neccessary to desribe it for function developers.

zroubalik (approved):
LGTM

docs/guides/python.md
lance: I agree with @zroubalik this will be fundamental knowledge for any Python dev.

Let's hold this PR until @matejvasek's [PR](https://github.com/boson-project/parliament/pull/9) lands since that will affect the docs for `context`.

docs/guides/python.md
```diff
@@ -0,0 +1,142 @@
...
+  The context parameter contains the Flask request object and any
+  CloudEvent received with the request.
+  """
+  print(f"Method: {context['request'].method}")
+  print(f"Event data {context['cloud_event'].data})
+  # ... business logic here
+```
+
+### Return Values
+Functions may return any value supported by Flask, as the invocation framework
```
matejvasek, line 118: So user cannot return just CloudEvent, meaning user has to transform CloudEvent to HTTP response by himself?

lance: Good point. That is correct for now.

lance: Working on adding a `@event` decorator to parliament.

lance: https://github.com/boson-project/parliament/pull/17

docs/guides/python.md
```diff
@@ -0,0 +1,142 @@
...
+Runtime: python
+Trigger: http
+
+~/src/pfunk via 🐍 v3.8.5
+❯ tree
+.
+├── func.py
+├── func.yaml
+├── requirements.txt
+└── test_func.py
```
zroubalik, line 21:
```suggestion
❯ kn func create -l python fn
Project path: /home/developer/projects/fn
Function name: fn
Runtime: python
Trigger: http

❯ tree
fn
├── func.py
├── func.yaml
├── requirements.txt
└── test_func.py
```

Making this consistent with other Guides.

docs/guides/python.md
```diff
@@ -0,0 +1,142 @@
...
+and what APIs are available to you as a developer.
+
+### Invocation parameters
+
+When using the `func` CLI to create a function project, you may choose to generate a project
+that responds to a `CloudEvent` or simple HTTP. `CloudEvents` in Knative are transported over
+HTTP as a `POST` request, so in many ways, the two types of functions are very much the same.
+They each will listen and respond to incoming HTTP events.
+
+When an inacoming request is received, your function will be invoked with a `Context` object as the first parameter. This object is simply a Python `dict` with two keys, `request` and
```
zroubalik, line 101:
```suggestion
When an incoming request is received, your function will be invoked with a `Context` object as the first parameter. This object is simply a Python `dict` with two keys, `request` and
```

lance:
@zroubalik @matejvasek @abrennan89 - ok I think this is ready for a final review before landing. Thanks. :)

docs/guides/python.md
```diff
@@ -0,0 +1,163 @@
+# Python Developer's Guide
+
+When creating a Python function using the `func` CLI, the project directory
+looks like a typical Python project. Both HTTP and Event functions have the same
+template structure.
+
+```
+~/src/fn
+❯ kn func create -l python
```
zroubalik, line 9: The rest of the Guide (as well as the others) shows usage of `func` not `kn func`. Would be nice to make it consistent.
```suggestion
❯ func create -l python fn
```

docs/guides/python.md
```diff
@@ -0,0 +1,163 @@
...
+Project path: /home/developer/src/fn
+Function name: fn
+Runtime: python
+Trigger: http
+
+~/src/fn via 🐍 v3.8.5
+❯ tree
+.
```
zroubalik, line 17:
```suggestion
❯ tree
fn
```
This is what we are using in the other guides, but feel free to not use this suggestion :)

zroubalik (approved):
Looking good, minor nitpicking, feel free to ignore those suggestions :)

docs/guides/python.md
```diff
@@ -0,0 +1,161 @@
+# Python Developer's Guide
```
zroubalik, line 1:
```suggestion
# Python Function Developer's Guide
```

lance:
@zroubalik I appreciate the keen eye - thanks for the feedback!

PR #294, opened 2021-04-12
fix: build needs to use legacy jar

matejvasek:
this should be reverted once our buildpack supports fast-jar


lkingland (approved):
Looks good to me Matej

zroubalik (approved):

lance (approved):

PR #298, opened 2021-04-13
docs: minor chores and fixes

zroubalik:

Minor fixes, modifed Node.js guide a little bit, I removed the reference to Lance's local setup (@lance I hope you don't take it personally 😄 )


lance (approved):
lgtm - thanks for the keen eye! :)

I do not take it personally at all! It was only that way because I was too lazy to change it. :rofl:

PR #299, opened 2021-04-13
src: use go 1.14 in the templates

zroubalik:

Go 1.16 has been released recently, 1.13 is pretty old, so I think it is a time for a bump. I checked the ubi images and they have 1.14 installed.


lance (approved):
lgtm

PR #300, opened 2021-04-13
chore: bump CE version in Go templates

zroubalik:

### DIFF
--- a/templates/go/events/go.mod
+++ b/templates/go/events/go.mod
@@ -2,4 +2,4 @@ module function
 
 go 1.13
 
-require github.com/cloudevents/sdk-go/v2 v2.3.1
+require github.com/cloudevents/sdk-go/v2 v2.4.0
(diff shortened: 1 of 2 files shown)


lance (approved):
I wonder if in the [Golang docs](https://github.com/boson-project/func/pull/297/files#diff-2f61cdd743578e4d6ba79f4ec67a5c8bb9ee326874572a170cbc9b4a99c1ce09R247) it would be better to show pulling in a different dependency, since with this change, that command will accomplish nothing. ;)

zroubalik:
> I wonder if in the [Golang docs](https://github.com/boson-project/func/pull/297/files#diff-2f61cdd743578e4d6ba79f4ec67a5c8bb9ee326874572a170cbc9b4a99c1ce09R247) it would be better to show pulling in a different dependency, since with this change, that command will accomplish nothing. ;)

Fair enough 😄 gonna fix that

PR #303, opened 2021-04-14
src: Quarkus events template uses CloudEvent<>

matejvasek:


matejvasek:
@lance do we want this?

templates/quarkus/events/src/main/java/functions/Function.java
```diff
@@ -3,21 +3,15 @@
...
-            System.out.println(
-                    "CloudEvent{" +
-                            "id='" + cloudEvent.id() + '\'' +
-                            ", specVersion='" + cloudEvent.specVersion() + '\'' +
-                            ", source='" + cloudEvent.source() + '\'' +
-                            ", subject='" + cloudEvent.subject() + '\'' +
-                            '}');
-        }
-        return new Output(input.getMessage());
+    public CloudEvent<Output> function(CloudEvent<Input> input) {
```
matejvasek, line 11: Previous version had one advantage: it accepted also non-CE HTTP (in which case the `cloudEvent` parameter was `null`).

lance (approved):

templates/quarkus/events/src/main/java/functions/Function.java
lance: But non-CE HTTP works fine in the HTTP templates, so...

zroubalik (approved):

PR #309, opened 2021-04-21
fix: default for `--builder` flag

### LINKED ISSUE #308: A non-default builder in func.yaml is overwritten by 'kn func build'
If I run `kn func build -b native` on a quarkus function, the build actually uses the native builder image, and updates `func.yaml` to reflect this. But if I subsequently run `kn func build` without the `-b native` flag, the customized value for `builder` is overwritten with the default builder image, and the default builder image is used.

Steps to reproduce this:

```
❯ kn func build -b native
...

❯ grep builder func.yaml
builder: native
builderMap:
  default: quay.io/boson/faas-quarkus-jvm-builder
  jvm: quay.io/boson/faas-quarkus-jvm-builder
  native: quay.io/boson/faas-quarkus-native-builder

❯ kn func build
...

❯ grep builder func.yaml
builder: default
builderMap:
  default: quay.io/boson/faas-quarkus-jvm-builder
  jvm: quay.io/boson/faas-quarkus-jvm-builder
  native: quay.io/boson/faas-quarkus-native-builder
```

matejvasek:
resolves #308

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -13,7 +13,7 @@ import (
 
 func init() {
 	root.AddCommand(buildCmd)
-	buildCmd.Flags().StringP("builder", "b", "", "Buildpack builder, either an as a an image name or a mapping name.  /n Specified value is stored in func.yaml for subsequent builds.")
+	buildCmd.Flags().StringP("builder", "b", "", "Buildpack builder, either an as a an image name or a mapping name.\nSpecified value is stored in func.yaml for subsequent builds.")
 	buildCmd.Flags().BoolP("confirm", "c", false, "Prompt to confirm all configuration options (Env: $FUNC_CONFIRM)")
 	buildCmd.Flags().StringP("image", "i", "", "Full image name in the orm [registry]/[namespace]/[name]:[tag] (optional). This option takes precedence over --registry (Env: $FUNC_IMAGE")
 	buildCmd.Flags().StringP("path", "p", cwd(), "Path to the project directory (Env: $FUNC_PATH)")
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -13,7 +13,7 @@ import (
 
 func init() {
 	root.AddCommand(buildCmd)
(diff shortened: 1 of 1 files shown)


cmd/build.go
```diff
@@ -13,7 +13,7 @@ import (
 
 func init() {
 	root.AddCommand(buildCmd)
-	buildCmd.Flags().StringP("builder", "b", "default", "Buildpack builder, either an as a an image name or a mapping name as defined in func.yaml")
+	buildCmd.Flags().StringP("builder", "b", "", "Buildpack builder, either an as a an image name or a mapping name as defined in func.yaml\nEmpty value implies default builder will be used.")
```
lance, line 16: This is an improvement, but I was thinking that the proper behavior would be for the CLI to just respect what's in func.yaml if no value is given.

matejvasek: Actually it should.

>Empty value implies default builder will be used.

is not true, it will use what's in the config

matejvasek: @lance I changed the help string

lance: Ah I see now, lgtm

lance (approved):

lance (approved):

cmd/build.go
```diff
@@ -13,7 +13,7 @@ import (
 
 func init() {
 	root.AddCommand(buildCmd)
-	buildCmd.Flags().StringP("builder", "b", "default", "Buildpack builder, either an as a an image name or a mapping name as defined in func.yaml")
+	buildCmd.Flags().StringP("builder", "b", "", "Buildpack builder, either an as a an image name or a mapping name as defined in func.yaml\nIf the value is empty then the value from the config will be used.")
```
lkingland, line 16: This is tricky to explain concisely.  In particular, the idea that func.yaml is more of a side-effect, but can be used to override, leads me to believe we should perhaps first highlight the purpose of writing to func.yaml in the first place:  to allow subsequent builds to remember earlier decisions:

```suggestion
	buildCmd.Flags().StringP("builder", "b", "", "Buildpack builder, either an as a an image name or a mapping name.  /n Specified value is stored in func.yaml for subsequent builds.")
```
Or is this even more confusing?

lkingland (approved):

PR #312, opened 2021-04-22
docs: add func.yaml field documentation

lance:
This adds a simple explanation of each of the fields in the `func.yaml`
configuration file.

### DIFF
--- a/docs/guides/func_yaml.md
+++ b/docs/guides/func_yaml.md
@@ -69,7 +69,7 @@ The language runtime for your function. For example `python`.
 ### `trigger`
 
 The invocation event that triggers your function. Possible values are `http`
-for plain HTTP requests, and `event` for CloudEvent triggered functions.
+for plain HTTP requests, and `events` for CloudEvent triggered functions.
 
 
 ## Environment Variables
@@ -85,4 +85,3 @@ the name with `env.`. For example:
 envVars:
   API_KEY: {{ env.API_KEY }}
 ```
-
--- a/docs/guides/func_yaml.md
+++ b/docs/guides/func_yaml.md
@@ -1,0 +1,88 @@
+# Project Configuration with `func.yaml`
+
+The `func.yaml` file contains configuration information for your function
+project. Generally, these values are used when you execute a `func` CLI
+command. For example, when `func build` is run, the CLI uses the value for
+the `builder` field. In many cases, these values may be overridden by
+command line flags or environment variables. For more information about
+overriding these values, consult the [Commands](command.md) document.
+
+Many of the fields are generated for you when you create, build and deploy
+your function. However there are a few that you may use to tweak things
+such as the function name, and the image name.
+
+## Fields
+
+The following fields are used in `func.yaml`.
+
+### `builder`
+
+Specifies the buildpack builder image to use when building the function.
+In most cases, this value should not be changed.
+
+### `builderMap`
+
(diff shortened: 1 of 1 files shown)


docs/guides/func_yaml.md
```diff
@@ -0,0 +1,88 @@
...
+
+Specifies the buildpack builder image to use when building the function.
+In most cases, this value should not be changed.
+
+### `builderMap`
+
+Some function runtimes may be built in multiple ways. For example, a Quarkus
+function may be built for the JVM, or as a native binary. The `builderMap`
+field will contain all of the available builders for a given runtime. Although
+it's typically unwise to modify the `builder` field, using values from
```
lkingland, line 28: Perhaps the gentler euphemism "unnecessary" rather than "unwise" here?

```suggestion
it's typically unnecessary to modify the `builder` field, using values from
```

docs/guides/func_yaml.md
```diff
@@ -0,0 +1,88 @@
...
+The `envVars` field allows you to set environment variables that will be
+available to your function at runtime. For example, to set a `MODE` environment
+variable to `debug` when the function is deployed, your `func.yaml` file
+may look like this.
+
+```yaml
+envVars: {
+  MODE: debug
+}
+```
```
lkingland, line 42: This ship may already have sailed here, but upon reading this helpful documentation, it seems that the "Vars" suffix is unnecessary, and this could just be "env" (as I have seen in other systems).  Perhaps worth a revisit prior to 1.0 while our API can flex a bit. 

```yaml
env:
  MODE: debug
  API_KEY: {{ env.API_KEY }}
```

docs/guides/func_yaml.md
```diff
@@ -0,0 +1,88 @@
...
+
+The `envVars` field allows you to set environment variables that will be
+available to your function at runtime. For example, to set a `MODE` environment
+variable to `debug` when the function is deployed, your `func.yaml` file
+may look like this.
+
+```yaml
+envVars: {
+  MODE: debug
+}
```
lkingland, line 41: While I much prefer the use of delimiters over indentation, it seems to be more common to use the latter in YAML files, despite this being perfectly valid.  So, should this instead be:
```
envVars:
  MODE: debug
```

lkingland (approved):

docs/guides/func_yaml.md
zroubalik: I like this suggestion

docs/guides/func_yaml.md
```diff
@@ -0,0 +1,88 @@
...
+The Kubernetes namespace where your function will be deployed.
+
+### `runtime`
+
+The language runtime for your function. For example `python`.
+
+### `trigger`
+
+The invocation event that triggers your function. Possible values are `http`
+for plain HTTP requests, and `event` for CloudEvent triggered functions.
```
zroubalik, line 72:
```suggestion
for plain HTTP requests, and `events` for CloudEvent triggered functions.
```

zroubalik (approved):
Looking good!

docs/guides/func_yaml.md
lance: I agree, and am happy to change the docs. That also means we'll need to change the code as well. :)

I will change this doc to reflect the above, then change the code in a separate PR.

matejvasek (approved):

PR #314, opened 2021-04-23
feat: add support for annotations in func.yaml

lance:
This commit adds limited support for annotations in the func.yaml
config file. The feature is limited, because it's only additive. A
user can add an annotation `foo: bar` in the config and deploy the
function, successfully setting that annotation on the Service.
However, if they subsequently remove `foo: bar` from the config
file, it will _not_ be removed from the deployment. This is because
it's not possible to know, from the set of annotations that currently
exist on the deployment, which ones were set by us and which were not.
So, removing any annotations that are not in func.yaml is unsafe.

It may be possible to store in a hidden file somewhere all of the
user-supplied annotations, allowing us to diff func.yaml with that file,
but I'm not sure I want to go down that path. It might just be best to
document this limitation.

We may also want to document that annotations added through func.yaml
should be user supplied settings/values, and not annotations that are
managed by knative (e.g. the autoscaling annotations).

Fixes: https://github.com/boson-project/func/issues/307


zroubalik:
> We may also want to document that annotations added through func.yaml
> should be user supplied settings/values, and not annotations that are
> managed by knative (e.g. the autoscaling annotations).

What is the reason behind this?

lance:
@zroubalik because when you try and set the scaling annotations manually, you get an error from Knative (or I am doing something wrong.

lkingland (approved):
LGTM!

zroubalik (approved):

PR #316, opened 2021-04-23
refactor!: change envVars to env in func.yaml

lance:
There are some places where I've changed variable and function names
where it wasn't strictly necessary. If you don't mind the bit of churn
that results, changing these makes `rg -i envvars` a nice way to check
for anything that could be overlooked.

### DIFF
--- a/config.go
+++ b/config.go
@@ -22,7 +22,7 @@ type config struct {
 	Trigger     string            `yaml:"trigger"`
 	Builder     string            `yaml:"builder"`
 	BuilderMap  map[string]string `yaml:"builderMap"`
-	EnvVars     map[string]string `yaml:"envVars"`
+	Env         map[string]string `yaml:"env"`
 	// Add new values to the toConfig/fromConfig functions.
 }
 
@@ -60,7 +60,7 @@ func fromConfig(c config) (f Function) {
 		Trigger:     c.Trigger,
 		Builder:     c.Builder,
 		BuilderMap:  c.BuilderMap,
-		EnvVars:     c.EnvVars,
+		Env:         c.Env,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


config.go
```diff
@@ -22,7 +22,7 @@ type config struct {
 	Trigger     string            `yaml:"trigger"`
 	Builder     string            `yaml:"builder"`
 	BuilderMap  map[string]string `yaml:"builderMap"`
-	EnvVars     map[string]string `yaml:"envVars"`
+	Env         map[string]string `yaml:"env"`
```
matejvasek, line 25: @lance won't this cause issues for users using newer versions with older conf file?

matejvasek: But not many people use `function` so it is not that much of a problem. We just need to let know people that it's been changed.

lance: Yes - you are right that any existing users would need to change their config file. But this is all so new, and it's still unsupported, so I think perhaps adding to release notes for TP-2 would be OK instead of trying for backwards compatibility. But I will follow up on this to be sure it's OK.

lance:
I should have noted that this change was prompted by the discussion in https://github.com/boson-project/func/pull/312#discussion_r618946318.

config.go
lkingland: LGTM!

Yes, this is technically a backwards incompatible change, but let's get this thing squeaky-clean now, before 1.0, while we still can.  (I also doubt there are others using env vars other than we team members at this point?),

lkingland (approved):

zroubalik (approved):
LGTM

zroubalik:
Nit: imho this shouldn't be a `chore` type of PR, since we are introducing the breaking changes.

lance:
> Nit: imho this shouldn't be a `chore` type of PR, since we are introducing the breaking changes.

Good point!

lance:
@zroubalik how about `refactor!` - adding the exclamation point to a prefix indicates to release-please that there is a breaking change.

PR #322, opened 2021-04-27
feat: add/improve spinner for build and deploy

lance:
This commit modifies the progress meter so that, by default there is no
step counter. It also modifies the responsibility for calling `Done()`,
making it the job of the command rather than the client. This is because
the client does not know how many commands will be executed and therefore
cannot know when the progress bar is done.

Fixes: https://github.com/boson-project/func/issues/296


knative/deployer.go
```diff
@@ -69,39 +69,38 @@ func (d *Deployer) Deploy(ctx context.Context, f bosonFunc.Function) (err error)
...
 
 		route, err := client.GetRoute(f.Name)
 		if err != nil {
 			err = fmt.Errorf("knative deployer failed to get the route: %v", err)
-			return err
+			return "", err
 		}
 
-		fmt.Println("Function updated at URL: " + route.Status.URL.String())
+		return fmt.Sprintf("Function updated at URL: %v", route.Status.URL.String()), nil
```
lkingland, line 102: This return value should probably be the route data (url) only, and the extra natural language English prefix added in the CLI command `cmd/deploy.go`, because that is essentially the UX, whereas here is more for data / client integration.
```suggestion
		return route.Status.URL.String(), nil
```

progress/progress.go
```diff
@@ -149,13 +158,13 @@ func (b *Bar) Done() {
...
 		b.ticker = nil
-		b.overwrite("") // write unindented
+		// b.overwrite("") // write unindented
 	}
 }
 
 // Write a simple line status update.
 func (b *Bar) write() {
-	fmt.Fprintf(b.out, "%v/%v %v\n", b.index, b.total, b.text)
+	fmt.Fprint(b.out, b.format())
```
lkingland, line 167: This looks like a good opportunity to just allow an instance of Bar to always be capable of printing it's current state to String, by using the standard method `.String()`.   If `format` is renamed as such, this line I believe cold become:
```suggestion
	fmt.Fprintln(b.out, b)
```

progress/progress.go
```diff
@@ -179,8 +188,15 @@ func (b *Bar) overwrite(prefix string) {
...
+	//  5 Print linebreak such that subsequent messages print correctly.
+	fmt.Fprintf(b.out, "\r%v%v%v%v", up, clear, prefix, b.format())
+}
+
+func (b *Bar) format() string {
+	if b.printWithStepCounter {
+		return fmt.Sprintf("%v/%v %v\n", b.index, b.total, b.text)
+	}
+	return fmt.Sprintf("%v\n", b.text)
 }
```
lkingland, line 200: If taking the suggestion above to give Bar a standard stringifier public method String, that would necessitate the minor tweak here of removing the trailing linebreaks:
```suggestion
func (b *Bar) String() string {
	if b.printWithStepCounter {
		return fmt.Sprintf("%v/%v %v", b.index, b.total, b.text)
	}
	return b.text
}
```

lkingland (approved):
Lookin good!

knative/deployer.go
lance: This is a good point. Going this route, we would probably want to remove this as well: https://github.com/boson-project/func/pull/322/files#diff-3bc17fd6241d546478336fa19a33f7e7f4c232421ce4e8511512c00fa617860dR52-R54. One side effect of this change however is that `Deployer` is the only part that knows if the function service was created or updated. We would need to either expose this property somehow, or forego the language.

lance:
@lkingland thanks for your suggestions. I've incorporated them. PTAL

client_int_test.go
```diff
@@ -116,7 +116,7 @@ func TestDeploy(t *testing.T) {
 	}
 	defer del(t, client, "deploy")
 
-	if err := client.Deploy(context.Background(), "."); err != nil {
+	if _, err := client.Deploy(context.Background(), "."); err != nil {
```
matejvasek, line 119: this `Deploy()` function really returns only one result, not tuple

buildpacks/builder.go
```diff
@@ -3,14 +3,14 @@ package buildpacks
...
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"os"
 	"runtime"
 
 	"github.com/buildpacks/pack"
 	"github.com/buildpacks/pack/logging"
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
lance: That must have snuck in yesterday when we were talking about handling SIGINT. Removed.

client_int_test.go
lance: My mistake

lance:
> Why does this contains `pkged.go`? it seems that it packaged artifacts that we don't wan't to have there e.g. `target` for java templates.

Hmm - it shouldn't have - and no longer does.

matejvasek (approved):

client.go
```diff
@@ -50,10 +50,23 @@ type Pusher interface {
...
+type Status int
+
+const (
+	Failed Status = iota
+	Deployed
+	Updated
+)
+
+type DeploymentResult struct {
+	Status Status
```
lkingland, line 62: I also like this because it could helps us maintain backwards compatibility going forward

knative/deployer.go
```diff
@@ -16,7 +16,7 @@ import (
 	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 	v1 "knative.dev/serving/pkg/apis/serving/v1"
 
-	bosonFunc "github.com/boson-project/func"
+	fn "github.com/boson-project/func"
```
lkingland, line 19: certainly liking this import name better

progress/progress.go
```diff
@@ -179,12 +189,22 @@ func (b *Bar) overwrite(prefix string) {
 	//  2 Move up one line
 	//  3 Clear to the end of the line
 	//  4 Print status text with optional prefix (spinner)
-	//  5 Print linebreak such that subsequent messaes print correctly.
-	fmt.Fprintf(b.out, "\r%v%v%v%v/%v %v\n", up, clear, prefix, b.index, b.total, b.text)
+	//  5 Print linebreak such that subsequent messages print correctly.
+	fmt.Fprintf(b.out, "\r%v%v%v%v\n", up, clear, prefix, b)
```
lkingland, line 193: Ah hah, handy use of `b`'s `.String` already there.

progress/progress.go
```diff
@@ -99,6 +108,13 @@ func (b *Bar) Increment(text string) {
 		return
 	}
 
+	// Start the spinner if not already started
+	if b.ticker == nil {
+		fmt.Println()
+		b.ticker = time.NewTicker(100 * time.Millisecond)
+		go b.spin(b.ticker.C)
+	}
+
```
lkingland, line 117: I see this is moved such that now the spiner is always running, even in verbose mode (but as a noop due to the verbosity check being moved to the spinner's write).  I'm interested to know what bug this fixed, because fiddling with the line status is finnicky indeed, and of course it would be ideal to not start the goroutine if it's never going to be used.

lkingland (approved):
LGTM!  I am sure this was tricky; fiddling with the line buffer is a descent into madness.

lkingland (approved):
I like it.  It's friendly.

Issue #323, opened 2021-04-29
`func delete` does not undeploy the function/knative service when using the function name as argument


jrangelramos:
delete command help states you can delete a function by just passing the name as argument.
```
This command undeploys a function from the cluster. By default the function from 
the project in the current directory is undeployed. Alternatively either the name 
of the function can be given as argument or the project path provided with --path.

Usage:
  func delete [NAME] [flags]
```
but it does not delete when you are outside of the function project directory

```
$ func list
NAME              NAMESPACE      RUNTIME     URL                                                     READY
myfunc            test-function  node        http://myfunc-test-function.apps-crc.testing            True
$ func delete myfunc
Error: the given path '/home/jeramos' does not contain an initialized function
```

lkingland:
Thanks for pointing this out @jrangelramos .  This was working at one time, but it appears to have been lost during a large refactor.

jrangelramos:
Issue addressed by https://github.com/knative-sandbox/kn-plugin-func/pull/324

lance:
@jrangelramos thanks for the contribution and confirmation!

PR #324, opened 2021-04-30
fix: func delete with explicity name as argument (#323)

jrangelramos:
Allows function to be delete when passing explicity name

```
$ func list | grep gofunc
gofunc            test-function  go          http://gofunc-test-function.apps-crc.testing            True
$ cd ~
$ ls -l func.yaml
ls: cannot access 'func.yaml': No such file or directory
$ func delete gofunc
Removing Knative Service: gofunc
$ func list | grep gofunc
$ 
```

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -42,17 +42,31 @@ kn func delete -n apps myfunc
 	RunE:              runDelete,
 }
 
+
 func runDelete(cmd *cobra.Command, args []string) (err error) {
 	config := newDeleteConfig(args).Prompt()
+	
+	var function fn.Function
 
-	function, err := fn.NewFunction(config.Path)
-	if err != nil {
-		return
-	}
+	// Initialize func with explicit name (when provided)
+	if len(args) > 0 && args[0] != "" {
+		pathChanged := cmd.Flags().Changed("path")
+		if pathChanged {
+			return fmt.Errorf("Only one of --path and [NAME] should be provided")
+		}
+		function = fn.Function{
+			Name: args[0],
+		}
+	} else {
+		function, err = fn.NewFunction(config.Path)
+		if err != nil {
+			return
+		}
 
-	// Check if the Function has been initialized
-	if !function.Initialized() {
-		return fmt.Errorf("the given path '%v' does not contain an initialized function", config.Path)
+		// Check if the Function has been initialized
+		if !function.Initialized() {
+			return fmt.Errorf("the given path '%v' does not contain an initialized function", config.Path)
+		}
 	}
 
 	ns := config.Namespace
--- a/cmd/delete_test.go
+++ b/cmd/delete_test.go
@@ -66,7 +66,10 @@ annotations: {}
 	}
 	defer f.Close()
 
-	f.WriteString(funcYaml)
+	_, err = f.WriteString(funcYaml)
+	if err != nil {
+		t.Fatal(err)
+	}
 	f.Close()
 
 
@@ -74,8 +77,13 @@ annotations: {}
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer os.Chdir(oldWD)
-	err = 	os.Chdir(tmpDir)
+	defer func() {
+		err = os.Chdir(oldWD)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
+	err = os.Chdir(tmpDir)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/docs/guides/commands.md
+++ b/docs/guides/commands.md
@@ -111,7 +111,7 @@ kn func list [-n <namespace> -p <path>]
 
 ## `delete`
 
-Removes a deployed function from the cluster. The user may specify a function by name, path or if neither of those are provided, the current directory will be searched for a `func.yaml` configuration file to determine the function to be removed. The namespace defaults to the value in `func.yaml` or the namespace currently active in the user's Kubernetes configuration. The namespace may be specified on the command line, and if so this will overwrite the value in `func.yaml`.
+Removes a deployed function from the cluster. The user may specify a function by name, path. If both of those are provided the command will not be executed and user will receive an error message. If neither of those are provided, the current directory will be searched for a `func.yaml` configuration file to determine the function to be removed. The namespace defaults to the value in `func.yaml` or the namespace currently active in the user's Kubernetes configuration. The namespace may be specified on the command line, and if so this will overwrite the value in `func.yaml`.
 
 Similar `kn` command: `kn service delete NAME [flags]`.
 


cmd/delete.go
```diff
@@ -45,14 +45,23 @@ kn func delete -n apps myfunc
...
-		return
-	}
-
-	// Check if the Function has been initialized
-	if !function.Initialized() {
-		return fmt.Errorf("the given path '%v' does not contain an initialized function", config.Path)
+	var function bosonFunc.Function
+
+	// Initialize func with explicit name (when provided)
+	if len(args) > 0 && args[0] != "" {
```
lance, line 51: @jrangelramos thanks for the contribution! One thing I'm thinking about with this, is that it has the effect of making the `--path` flag silently ignored. The suggested code here is pretty ugly, but I could not figure out how to determine using cobra whether a user explicitly provided the `--path` flag because it's always set with a default. This is the only way I could work it out.

```suggestion
	if len(args) > 0 && args[0] != "" {
		flags := cmd.Flags()
		if flags.NFlag() > 0 {
			namespace, _ := flags.GetString("namespace")
			confirm, _ := flags.GetBool("confirm")
			switch flags.NFlag() {
			case 1:
				if namespace == "" && !confirm {
					return fmt.Errorf("Only one of --path and [NAME] should be provided")
				}
			case 2:
				if !(namespace == "" && !confirm) {
					return fmt.Errorf("Only one of --path and [NAME] should be provided")
				}
			default:
				return fmt.Errorf("Only one of --path and [NAME] should be provided")
			}
		}
```

zroubalik: Do we want to allow this? My understanding was, that `--path` or `-p` is mandatory in case we don't invoke from within the function project directory. The same as it is for `build`, `deploy` and `describe`.

nainaz: users can delete bu providing -p and function name right?

jrangelramos: I think we could test `os.Args` check for `-p or --path in this case, thus we can ensure it was actually informed or not

jrangelramos: I think it would work with `cmd.Flags().Changed("path")`. Thus we can ensure user actually informed it

lance: > Do we want to allow this? My understanding was, that `--path` or `-p` is mandatory in case we don't invoke from within the function project directory. The same as it is for `build`, `deploy` and `describe`.

@zroubalik I guess now that the function name and service name are the same, I guess it seems trivial to allow `kn func delete [NAME]` - which is how the help was already written.

https://github.com/boson-project/func/blob/5feb0e20f366f8dc46f339257d87419bc852753c/cmd/delete.go#L22
https://github.com/boson-project/func/blob/5feb0e20f366f8dc46f339257d87419bc852753c/cmd/delete.go#L35-L38

I agree, however, that treating `delete` differently than the others seems a bit wrong. I guess I'm on the fence. If we do decide to disallow this, we need to improve the help text at a minimum.

lkingland: > @zroubalik now that the function name and service name are the same, it seems trivial to allow `kn func delete [NAME]`

This has actually always been possible, including before the Function and service names were made equivalent, because the client library made that conversion completely transparently to the user.

I agree with allowing this in principle, because Function name is the only required information to complete the request.   The delete should accept either name or path to a function where name can be found.

This is actually the way the original version I put together worked, so I am of course biased :)
```
var deleteCmd = &cobra.Command{
	Use:               "delete <name>",
	Short:             "Delete a Function deployment",
	Long:              `Removes the deployed Function by name, by explicit path, or by default for the current directory.  No local files are deleted.`,
```
https://github.com/boson-project/func/blob/6fb0116a86a53cecbcf17748a84e0f0636bb3972/cmd/delete.go#L19

I often during testing wanted to delete a Function without having the local files available, so I know it's useful!

lkingland: Hello @jrangelramos, thanks for the submission!

I do like the functionality of being able to delete by name.  The original version of `faas` had this, actually, and I found it quite useful myself:
https://github.com/boson-project/func/blob/6fb0116a86a53cecbcf17748a84e0f0636bb3972/cmd/delete.go#L19
```
var deleteCmd = &cobra.Command{
	Use:               "delete <name>",
	Short:             "Delete a Function deployment",
	Long:              `Removes the deployed Function by name, by explicit path, or by default for the current directory.  No local files are deleted.`,
```

This earlier version did not, however, explicitly error if both path and a name were provided, which I do think  is a good addition.

On to the question of whether this _should_ be here:  I agree with this in principle, because Function name is the only required information to complete the request.   Why request more than we need?  I often during testing wanted to delete a Function without having the local files available, so I know it's useful.

lkingland (dismissed):

cmd/delete.go
lkingland: > Do we want to allow this? My understanding was, that `--path` or `-p` is mandatory in case we don't invoke from within the function project directory. The same as it is for `build`, `deploy` and `describe`.

I think this is indeed true, but in the case of deleting by an explicitly provided name, there is no need to traverse that code path at all, so the value of --path, and the current working directory, should be irrelevant.

lkingland: > now that the function name and service name are the same, it seems trivial to allow `kn func delete [NAME]` - which is how the help was already written.

This was actually available from day one, because the translation from the presented name and the escaped service name was entirely transparent to the user.

lkingland (changes requested):
We'll want to update the command docs in `docs/guides/commands.md`.
This looks good to me otherwise.

cmd/delete.go
zroubalik: > @zroubalik now that the function name and service name are the same, it seems trivial to allow `kn func delete [NAME]` - which is how the help was already written.

Yeah, but as mentioned, our UX is not unified then. I am rather against this particular change, first we should discuss the overall UX and commands' syntax, becase we have introduced the naming change (kservice name == function name). And if needed go ahead with this change, but modify the other commands accordingly.

matejvasek:
@jrangelramos  please add test.

jcrossley3:
I would like to butt in here and say this behavior just bit me and I was completely surprised by it. Once I deployed my function, I assumed I could delete the project on my filesystem. I mean, I pushed my image -- why do I care about its source code anymore? To me, the binary and the source should have independent lifecycles. The deployed function, manifested as a collection of knative services, deployments, etc doesn't need my source code in order to fulfill runtime requests, so I should be able to delete it without any regard to my source code.

When `kn func delete myfunc` failed with _"Error: the given path '/home/jim/src' does not contain an initialized function"_, I was immediately put off -- "WTF kind of janky, fisher-price fresh hell is this?!?!" I thought to myself.

I kid, of course. `kn func` is awesome, but that `func delete` requires a local file definitely violated The Principle of Least Surprise for me. Fortunately, I was able to eventually delete it by running `kn create myfunc` again and retrying, but this felt very hacky to me.

I hope this feedback has been constructive. :stuck_out_tongue:

cmd/delete.go
jcrossley3: I would like to butt in here and say this behavior just bit me and I was completely surprised by it. Once I deployed my function, I assumed I could delete the project on my filesystem. I mean, I pushed my image -- why do I care about its source code anymore? To me, the binary and the source should have independent lifecycles. The deployed function, manifested as a collection of knative services, deployments, etc doesn't need my source code in order to fulfill runtime requests, so I should be able to delete it without any regard to my source code.

When `kn func delete myfunc` failed with _"Error: the given path '/home/jim/src' does not contain an initialized function"_, I was immediately put off -- "WTF kind of janky, fisher-price fresh hell is this?!?!" I thought to myself.

I kid, of course. `kn func` is awesome, but that `func delete` requires a local file definitely violated The Principle of Least Surprise for me. Fortunately, I was able to eventually delete it by running `kn create myfunc` again and retrying, but this felt very hacky to me.

I hope this feedback has been constructive. :stuck_out_tongue:

lkingland: Agree with Zbynek and Jim.  I vote we accept this PR which adds back the ability to delete by name, as the lack of it bugs me as much as it does Jim, and then fast-track the re-addition of the functionality to other pertinent commands.

jrangelramos:
@lkingland doc updated as requested.

zroubalik (approved):
LGTM

zroubalik:
@jrangelramos FYI there are some lint related issues that need to be fixed before merging.

matejvasek:
@zroubalik looks like that was introduced in my commit with tests, @jrangelramos please fix it

matejvasek (approved):

PR #325, opened 2021-05-03
src: update knative deps to 0.22

matejvasek:


go.mod
```diff
@@ -5,26 +5,28 @@ go 1.14
...
+	k8s.io/client-go v0.19.7
+	knative.dev/client v0.22.0
+	knative.dev/eventing v0.22.0
+	knative.dev/pkg v0.0.0-20210331065221-952fdd90dbb0
+	knative.dev/serving v0.22.0
 )
 
 replace (
+	// knative.dev/serving@v0.22.0 and knative.dev/pkg@v0.0.0-20210331065221-952fdd90dbb0 require different versions of go-openapi/spec
+	github.com/go-openapi/spec => github.com/go-openapi/spec v0.19.6
```
matejvasek, line 28: @dsimansk would this work with `kn`?
For some reason `knative.dev/serving@v0.22.0` has indirect dependency on `github.com/go-openapi/spec v0.20.2`, but that breaks something for `sigs.k8s.io/kustomize`.

matejvasek:
this improves cancellability

zroubalik (commented):
I am generally ok with this change. But OpenShift Serverless 1.14 uses Knative 0.20.0, the next version will probably use 0.21.0? But we are bringing dependency on 0.22.0 in here.

Maybe it is ok, but I am not sure 🤷‍♂️

matejvasek:
@rhuss wdyt?

client.go
```diff
@@ -101,7 +101,7 @@ type ProgressListener interface {
 // Describer of Functions' remote deployed aspect.
 type Describer interface {
 	// Describe the running state of the service as reported by the underlyng platform.
-	Describe(name string) (description Description, err error)
+	Describe(ctx context.Context, name string) (description Description, err error)
```
lance, line 104: I'm not sure what to say about bumping knative dependency - probably @rhuss or @markusthoemmes have more insight on that. But I think the changes in this file are probably unrelated, aren't they? I do like adding context in general, but does it need to be in this PR?

matejvasek: I am updating the dep specifically to be able to use ctx.

matejvasek: but I could do separate PR

matejvasek:
We really should use `0.21.0` for upcoming version.

PR #332, opened 2021-05-06
Add 'kn func emit' command

lance:
Adds the `kn func emit` command. Usage is:

```console
Emit event

Emits a CloudEvent, sending it to the deployed function.

Usage:
  func emit [flags]

Examples:

# Send a CloudEvent to the deployed function with no data and default values
# for source, type and ID
kn func emit

# Send a CloudEvent to the deployed function with the data found in ./test.json
kn func emit --file ./test.json

# Send a CloudEvent to the function running locally with a CloudEvent containing
# "Hello World!" as the data field, with a content type of "text/plain"
kn func emit --data "Hello World!" --content-type "text/plain" -s local

# Send a CloudEvent to the function running locally with an event type of "my.event"
kn func emit --type my.event --sink local

# Send a CloudEvent to the deployed function found at /path/to/fn with an id of "fn.test"
kn func emit --path /path/to/fn -i fn.test

# Send a CloudEvent to an arbitrary endpoint
kn func emit --sink "http://my.event.broker.com"


Flags:
  -c, --content-type string   The MIME Content-Type for the CloudEvent data  (Env: $FUNC_CONTENT_TYPE) (default "application/json")
  -d, --data string           Any arbitrary string to be sent as the CloudEvent data. Ignored if --file is provided  (Env: $FUNC_DATA)
  -f, --file string           Path to a local file containing CloudEvent data to be sent  (Env: $FUNC_FILE)
  -h, --help                  help for emit
  -i, --id string             CloudEvent ID (Env: $FUNC_ID) (default "205b5fc6-33b9-42ee-999a-04ca96c59b24")
  -p, --path string           Path to the project directory. Ignored when --sink is provided (Env: $FUNC_PATH) (default "/home/lanceball/src/github.com/boson-project/func")
  -k, --sink string           Send the CloudEvent to the function running at [sink]. The special value "local" can be used to send the event to a function running on the local host. When provided, the --path flag is ignored  (Env: $FUNC_SINK)
  -s, --source string         CloudEvent source (Env: $FUNC_SOURCE) (default "/boson/fn")
  -t, --type string           CloudEvent type  (Env: $FUNC_TYPE) (default "boson.fn")

Global Flags:
  -v, --verbose   print verbose logs
```

Still needed:

- [x] CLI command implementation
- [x] Client tests
- [x] Configurable content type
- [x] Fix lint errors
- [x] Add documentation

### DIFF
--- a/cloudevents/emitter.go
+++ b/cloudevents/emitter.go
@@ -1,0 +1,66 @@
+package cloudevents
+
+import (
+	"context"
+	"fmt"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/cloudevents/sdk-go/v2/client"
+	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/cloudevents/sdk-go/v2/protocol/http"
+	"github.com/cloudevents/sdk-go/v2/types"
+	"github.com/google/uuid"
+)
+
+const (
+	DefaultSource = "/boson/fn"
+	DefaultType   = "boson.fn"
+)
+
+type Emitter struct {
+	Endpoint    string
+	Source      string
+	Type        string
+	Id          string
+	Data        string
+	ContentType string
+}
+
+func NewEmitter() *Emitter {
+	return &Emitter{
+		Source:      DefaultSource,
+		Type:        DefaultType,
+		Id:          uuid.NewString(),
+		Data:        "",
+		ContentType: event.TextPlain,
+	}
+}
+
+func (e *Emitter) Emit(ctx context.Context, endpoint string) (err error) {
+	c, err := newClient(endpoint)
+	if err != nil {
+		return
+	}
+	evt := event.Event{
+		Context: event.EventContextV1{
+			Type:   e.Type,
+			Source: *types.ParseURIRef(e.Source),
+			ID:     e.Id,
+		}.AsV1(),
+	}
+	if err = evt.SetData(e.ContentType, e.Data); err != nil {
+		return
+	}
+	if result := c.Send(ctx, evt); cloudevents.IsUndelivered(result) {
+		return fmt.Errorf(result.Error())
+	}
+	return nil
+}
+
+func newClient(target string) (c client.Client, err error) {
+	p, err := http.New(http.WithTarget(target))
+	if err != nil {
+		return
+	}
+	return client.New(p)
+}
--- a/cloudevents/emitter_test.go
+++ b/cloudevents/emitter_test.go
@@ -1,0 +1,140 @@
+package cloudevents
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/cloudevents/sdk-go/v2/client"
+	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/cloudevents/sdk-go/v2/protocol/http"
+	"github.com/google/go-cmp/cmp"
+)
+
+func makeClient(t *testing.T) (c client.Client, p *http.Protocol) {
+	p, err := http.New()
+	if err != nil {
+		t.Fatal(err)
+	}
+	c, err = client.New(p)
+	if err != nil {
+		t.Errorf("failed to make client %s", err.Error())
+	}
+	return
+}
+
+func receiveEvents(t *testing.T, ctx context.Context, events chan<- event.Event) (p *http.Protocol) {
+	c, p := makeClient(t)
+	go func() {
+		err := c.StartReceiver(ctx, func(ctx context.Context, event event.Event) error {
+			go func() {
+				events <- event
+			}()
+			return nil
+		})
+		if err != nil {
+			t.Errorf("failed to start receiver %s", err.Error())
+		}
+	}()
+	time.Sleep(1 * time.Second) // let the server start
+	return
+}
+
+func TestEmitterDefaults(t *testing.T) {
+	events := make(chan event.Event)
+	ctx, cancel := context.WithCancel(context.Background())
+
+	// start a cloudevent client that receives events
+	// and sends them to a channel
+	p := receiveEvents(t, ctx, events)
+
+	emitter := NewEmitter()
+	if err := emitter.Emit(ctx, fmt.Sprintf("http://localhost:%v", p.GetListeningPort())); err != nil {
+		t.Fatalf("Error emitting event: %v\n", err)
+	}
+
+	// received event
+	got := <-events
+
+	cancel()                    // stop the client
+	time.Sleep(1 * time.Second) // let the server stop
+
+	if got.Source() != "/boson/fn" {
+		t.Fatal("Expected /boson/fn as default source")
+	}
+	if got.Type() != "boson.fn" {
+		t.Fatal("Expected boson.fn as default type")
+	}
+}
+
+func TestEmitter(t *testing.T) {
+	testCases := map[string]struct {
+		cesource string
+		cetype   string
+		ceid     string
+		cedata   string
+	}{
+		"with-source": {
+			cesource: "/my/source",
+		},
+		"with-type": {
+			cetype: "my.type",
+		},
+		"with-id": {
+			ceid: "11223344",
+		},
+		"with-data": {
+			cedata: "Some event data",
+		},
+	}
+	for n, tc := range testCases {
+		t.Run(n, func(t *testing.T) {
+			events := make(chan event.Event)
+			ctx, cancel := context.WithCancel(context.Background())
+
+			// start a cloudevent client that receives events
+			// and sends them to a channel
+			p := receiveEvents(t, ctx, events)
+
+			emitter := NewEmitter()
+
+			if tc.cesource != "" {
+				emitter.Source = tc.cesource
+			}
+			if tc.cetype != "" {
+				emitter.Type = tc.cetype
+			}
+			if tc.ceid != "" {
+				emitter.Id = tc.ceid
+			}
+			if tc.cedata != "" {
+				emitter.Data = tc.cedata
+			}
+			if err := emitter.Emit(ctx, fmt.Sprintf("http://localhost:%v", p.GetListeningPort())); err != nil {
+				t.Fatalf("Error emitting event: %v\n", err)
+			}
+
+			// received event
+			got := <-events
+
+			cancel()                           // stop the client
+			time.Sleep(100 * time.Millisecond) // let the server stop
+
+			if tc.cesource != "" && got.Source() != tc.cesource {
+				t.Fatalf("%s: Expected %s as source, got %s", n, tc.cesource, got.Source())
+			}
+			if tc.cetype != "" && got.Type() != tc.cetype {
+				t.Fatalf("%s: Expected %s as type, got %s", n, tc.cetype, got.Type())
+			}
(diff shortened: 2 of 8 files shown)


cloudevents/emitter.go
```diff
@@ -0,0 +1,60 @@
...
+	e.SetData(emitter.ContentType, emitter.Data)
+	result := c.Send(context.TODO(), e)
+	if result.Error() != "" {
+		return fmt.Errorf(result.Error())
+	}
+	return nil
+}
+
+func newClient(target string) (c client.Client, err error) {
+	p, err := http.New(http.WithTarget(target))
```
cardil, line 54: `kn-event` already supports public sending via http, and in-cluster sending via sender job (thus buggy https://github.com/knative-sandbox/kn-plugin-event/issues/12).

lance: Yes, we have discussed this and want to provide a simple user experience that is tied directly to a user's function project. As `func` replicates many `kn` capabilities but in a simpler way, so too do we want to provide a similar, simple experience with event emission.

cloudevents/emitter.go
```diff
@@ -0,0 +1,60 @@
...
+	return &Emitter{
+		Source:      "/boson/fn",
+		Type:        "boson.fn",
+		Id:          uuid.NewString(),
+		Data:        "",
+		ContentType: event.TextPlain,
+	}
+}
+
+func (emitter *Emitter) Emit(endpoint string) (err error) {
```
matejvasek, line 33: Make `ctx` the first param of the function.

cloudevents/emitter_test.go
```diff
@@ -0,0 +1,132 @@
...
+		if err != nil {
+			t.Errorf("failed to start receiver %s", err.Error())
+		}
+	}()
+	time.Sleep(1 * time.Second) // let the server start
+}
+
+func TestEmitterDefaults(t *testing.T) {
+	events := make(chan event.Event)
+	ctx, cancel := context.WithCancel(context.TODO())
```
matejvasek, line 44: It's ok to use `context.Background()` for tests.

cloudevents/emitter_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	emitter := NewEmitter()
+	emitter.Emit("http://localhost:8080")
+
+	// received event
+	got := <-events
+
+	cancel() // stop the client
+
+	if got.Source() != "/boson/fn" {
```
matejvasek, line 58: What if we used `github.com/stretchr/testify/assert` for prettier assertions?

cmd/emit.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/google/uuid"
+	"github.com/ory/viper"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	e := cloudevents.NewEmitter()
+	root.AddCommand(emitCmd)
+	emitCmd.Flags().BoolP("local", "l", false, "Send the CloudEvent to the function running on the local host")
+	emitCmd.Flags().StringP("source", "s", e.Source, "CloudEvent source")
```
zroubalik, line 14: Would be nice to allow sending to other Sinks (not just a function) as well. I can see a very common ask to send to the Broker/Channel as well.

cloudevents/emitter_test.go
lkingland: I like what assertion frameworks do in some languages, and have used several throughout the years, but (and I stress this is just a humble opinion), I have come to not like them.  especially in Go. Having watched them steadily descend into madness repeatedly throughout the years, they are something I now avoid.

The standard built in language logic operators are are not only accessible to everyone without the need to learn a new API, but they also age better, and tend to teach the reader of the tests about the language, and the package being tested, better as one reads the tests.  The code is being presented in a context similar to how it will be used in their code, rather than a somewhat contrived context of an _almost_, but not quite, natural language assertion syntax that is often just qirky enough to be confusing.

I am happy to go along with the majority opinion on this one, if others want to use an assertions framework I am game, but I feel compelled to make my jaded curmudgeon of a voice heard.  I am also open to being educated on why they are worth the aforementioned problems!

cloudevents/emitter.go
```diff
@@ -0,0 +1,60 @@
...
+
+func NewEmitter() *Emitter {
+	return &Emitter{
+		Source:      "/boson/fn",
+		Type:        "boson.fn",
+		Id:          uuid.NewString(),
+		Data:        "",
+		ContentType: event.TextPlain,
+	}
+}
```
lkingland, line 31: One thing I have seen done, and found to be pleasant for these default values is to turn them into exported constants.  That has a few benefits, among them being the value is included in the auto-generated godoc package documentaiton.  It might be worth adopting if you like that as well:

```suggestion
const (
  // DefaultSource for emitted events.
  DefaultSource = "/boson/fn"
  
  // DefaultType of emitted events.
  DefaultType = "boson.fn"
)

type Emitter struct {
	Endpoint    string
	Source      string
	Type        string
	Id          string
	Data        string
	ContentType string
}

func NewEmitter() *Emitter {
	return &Emitter{
		Source:      DefaultSource,
		Type:        DefaultType,
		Id:          uuid.NewString(),
		Data:        "",
		ContentType: event.TextPlain,
	}
}
```

cloudevents/emitter.go
lkingland: Go idiom is to have the funciton receiver be a single letter.  I used to find this draconian but have come to appreciate the consistency.

```suggestion
func (e *Emitter) Emit(endpoint string) (err error) {
```

cloudevents/emitter.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	e := event.Event{
+		Context: event.EventContextV1{
+			Type:   emitter.Type,
+			Source: *types.ParseURIRef(emitter.Source),
+			ID:     emitter.Id,
+		}.AsV1(),
+	}
+	e.SetData(emitter.ContentType, emitter.Data)
+	result := c.Send(context.TODO(), e)
```
lkingland, line 46: This will end up being the `ctx` from the method signature suggested by Matej

cloudevents/emitter.go
```diff
@@ -0,0 +1,60 @@
...
+	return nil
+}
+
+func newClient(target string) (c client.Client, err error) {
+	p, err := http.New(http.WithTarget(target))
+	if err != nil {
+		return nil, err
+	}
+	c, err = client.New(p)
+	return
```
lkingland, line 59: I might be wrong (it's late) but I think this could be collapsed a little:

```suggestion
	if err != nil {
		return
	}
	return client.New(p)
```

cloudevents/emitter_test.go
```diff
@@ -0,0 +1,132 @@
...
+func TestEmitterDefaults(t *testing.T) {
+	events := make(chan event.Event)
+	ctx, cancel := context.WithCancel(context.TODO())
+
+	// start a cloudevent client that receives events
+	// and sends them to a channel
+	receiveEvents(t, ctx, events)
+
+	emitter := NewEmitter()
+	emitter.Emit("http://localhost:8080")
```
lkingland, line 51: I'm not sure if this is exposed by the `cloudevent` library, but creating a net listener without a port specified allows the OS to choose the port, the final address of the listening service is then available using the `.Addr()` accessor.  This gets around the all-too-common problem of there already being a service attached to `:8080` on the box running the tests.

cmd/emit.go
```diff
@@ -0,0 +1,64 @@
...
+	Use:   "emit",
+	Short: "Emit a CloudEvent to a function endpoint",
+	Long: `Emit event
+
+Emits a CloudEvent, sending it to the deployed function.
+`,
+	Example: `
+# Send a CloudEvent to the deployed function with no data and default values
+# for source, type and ID
+kn func emit
```
lkingland, line 30: This right here is the big benefit of bringing this into the func client library.  We can assume defaults that take into account the current Function context.  If we can keep it this simple, this is definitely going to be a win for Function authors.

cmd/emit.go
```diff
@@ -0,0 +1,64 @@
...
+
+# Send a CloudEvent to the deployed function with the data found in ./test.data
+kn func emit --data ./test.data
+
+# Sed a CloudEvent to the function running locally with an event type of "my.event"
```
lkingland, line 35:
```suggestion
# Send a CloudEvent to the function running locally with an event type of "my.event"
```
spelling

lkingland (commented):
Lookin' pretty good for a draft IMHO

cmd/emit.go
lance: Currently I'm proposing a `--local` boolean flag. To support arbitrary endpoints, this should probably be a little different semantically.

```
func emit [--sink=endpoint]
```
When `--sink` is omitted, the event is by default sent to the deployed function. This means that to send an event to a function running locally via `func run`, the command would be

```
funk emit --sink=http://localhost:8080
```
A possibility is to have a special value for local, e.g. 

```
func emit --sink=local
```

WDYT?

zroubalik: The reason why I want to have the possibility to send to arbitrary endpoints is that when I am developing a function, I might want to test the events flow in my system, eg. that all triggers are correctly set etc. So in that case I don't want to send the CE directly to the function, but mimic what would happen in the real deployment, for example send it to the broker.

And why not make the UX consistent? If this is not implemented, the user would have to switch to different tool (kn-event-plugin) which could have a slightly different syntax and whatnot.

lance: @zroubalik so concretely, are you suggesting something like a `--endpoint` flag that, when provided, overrides any `--path` or `$FUNC_PATH` values and ignores `cwd()`?

zroubalik: Exactly and I like what you proposed above `--sink=local` or `--endpoint=local`, this way we can omit the `-l` flag.

lance:
@boson-project/core I think this is ready for a review. PTAL.

cmd/emit.go
```diff
@@ -0,0 +1,138 @@
...
+		endpoint = desc.Routes[0]
+	}
+
+	emitter := cloudevents.NewEmitter()
+	emitter.Source = config.Source
+	emitter.Type = config.Type
+	emitter.Id = config.Id
+	emitter.ContentType = config.ContentType
+	emitter.Data = config.Data
+	// TODO: Warn user if both --data and --file are provided
```
zroubalik, line 98: Can we print a warning in verbose mode?

zroubalik (approved):
Looking good! Minor nit on warning when both data and file is set.

lkingland (approved):

cmd/emit.go
```diff
@@ -0,0 +1,142 @@
...
+	emitter.Source = config.Source
+	emitter.Type = config.Type
+	emitter.Id = config.Id
+	emitter.ContentType = config.ContentType
+	emitter.Data = config.Data
+	if config.File != "" {
+		var buf []byte
+		if emitter.Data != "" && config.Verbose {
+			// TODO: This made me wonder whether we should switch to a real logging library
+			fmt.Printf("WARN: Found both --data and --file. Using file: %v\n", config.File)
```
lkingland, line 103: @lance I would take the purist approach and say that there is no way this is a state the developer intended, but is due to either a misconfiguration or misunderstanding, and should thus be an error.

PR #340, opened 2021-05-12
src!: revert bump to go 1.16 and template changes

lance:
A couple of problems arose when attempting to build this project for 1.15 as an inlined plugin for `kn`. First, `kn` is still on go 1.15. And second, the templates.tgz file is not included in source, so when `kn` is built and dependencies are pulled via `go mod vendor`, that file does not exist. 

This is the first of a couple of different PRs to address this situation. Assuming that the `kn` build will not bump to go 1.16 this build reverts to 1.15 and removes the changes to how we bundle templates.


matejvasek (approved):

matejvasek:
lgtm

zroubalik (approved):

PR #341, opened 2021-05-12
src: add templates.tgz to git

lance:
If the serverless 1.15 build can be bumped to golang 1.16, then this commit can be used to solve build problems for `kn` noted in https://github.com/boson-project/func/pull/340


lance:
Closing in deference to #340

PR #345, opened 2021-05-13
src: check that func.yaml is valid

zroubalik:

Currently contents of `func.yaml` are not validated, so users are able to specify anything in there (and that content ignored).
- adding validation check
- polishing the error message a little bit:
```
$ func deploy                                                                                                                                          
Error: yaml: unmarshal errors:
  line 10: field foo not found in type function.config
  line 11: field baz not found in type function.config
```
->
```
$ func deploy 
Error: 'func.yaml' config file is not valid:
  line 10: field foo is not valid
  line 11: field baz is not valid
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,9 +1,12 @@
 package function
 
 import (
+	"errors"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


lance (approved):
lgtm

I assume versioning will be a separate PR?

Issue #349, opened 2021-05-17
replace pkger with builtin go:embed


lkingland:
The pkger dependency, while a good stop-gap, needs to be replaced with a tarfile generated on `go generate` and embedded as a `fs.FS` using the `go:embed` introduced in go 1.16.

UPDATE:  This work is complete, along with several other updates necessary for the transition (see associated PR).  Unfortunately our productization pipeline currently supports Go1.15 max.  This issue's resolution is therefore blocked until such time as 1.16 is supported throughout.

PR #350, opened 2021-05-17
src(refactor): replace pkger with embed

lkingland:
See #349

NOTE: Merging on hold until confirmation that productization pipeline supports Go 1.16


lance:
Happy to see this back in the queue. Sorry for my brutal surgery last week!

lkingland:
> Happy to see this back in the queue. Sorry for my brutal surgery last week!

Oh it's no problem at all @lance! we can keep this one on life-support until 1.16 is available.  Thanks for helping roll it back.  Thank goodness for git.

lkingland:
Leaving this PR in draft mode, as it's contents are likely to diverge substatively between now and when it is able to be actually accepted.

lkingland:
Please not:  `go embed` will not solve our use case, because `go embed` works downstream of the source code:  it takes source and builds artifacts into binaries.

`pkger` works upstream from the source code: it encodes non-go files as Go source code, and exposes this as a filesystem at runtime.

It is important that we use a tool upstream from the source code, creating .go artifacts which are compiled, because our project is as much a client library as a CLI (likely more important in the long run).  If we upgrade to `go embed`, users of the client library will not receive any templates.  Using a tool like `pkger` is more appropriate in this situation.

To me the solution to this problem is to use a Go library which performs the same action as the `pkger` binary, as part of the build step, removing the need for the third-party binary, and sidestepping some of the difficulties with `pkger` such as not being able to omit certain directories from its initial static code analysis step.  Since this is such a focused use case, it might make sense to simply perform the serialization ourselves:  this would also give us a chance to upgrade the filesystem exposed at runtime to the Go1.16-introduced FS type, which would simplify the implementation of the templates Writer.

PR #355, opened 2021-05-19
feat: add typescript templates

lance:
This commit also bumps the faas-js-runtime to a github commit hash, since
this is needed for typescript support.

Signed-off-by: Lance Ball <lball@redhat.com>

PR #356, opened 2021-05-19
src(templates)!: modify the nodejs event template to accept a cloudevent

jcrossley3:
Pretty-printing the contents of the event and its envelope for each
request. This is handy when invoked as a Knative event sink as some
sources, e.g. Kafka, ignore the response body.

### DIFF
--- a/templates/node/events/index.js
+++ b/templates/node/events/index.js
@@ -15,21 +15,16 @@ const { CloudEvent, HTTP } = require('cloudevents');
  *  -H'Ce-specversion: 1.0' \
  *  http://localhost:8080
  *
- * For convenience, the runtime will extract the event data from the
- * CloudEvent envelope and pass it in the second parameter to the
- * function. The entire envelope, along with the HTTP request
- * metadata, is available from the first parameter, 'context'.
- *
  * @param {Context} context the invocation context
- * @param {Object} event the CloudEvent data
+ * @param {Object} event the CloudEvent 
  */
 function handle(context, event) {
 
+  context.log.info("context");
+  console.log(JSON.stringify(context, null, 2));
+
   context.log.info("event");
   console.log(JSON.stringify(event, null, 2));
-
-  context.log.info("cloudevent envelope");
-  console.log(JSON.stringify(context.cloudevent, null, 2));
 
   return HTTP.binary(new CloudEvent({
     source: 'event.handler',
--- a/templates/node/events/index.js
+++ b/templates/node/events/index.js
@@ -2,8 +2,10 @@
 const { CloudEvent, HTTP } = require('cloudevents');
 
 /**
- * An example function that responds to incoming CloudEvents over HTTP. For example,
- * from the Knative event Broker. Try invoking with a request such as this.
+ * A function that responds to incoming CloudEvents over HTTP from,
+ * for example, a Knative event Source, Channel or Broker.
+ *
+ * If running via 'npm run local', it can be invoked like so:
  *
  * curl -X POST -d '{"name": "Tiger", "customerId": "0123456789"}' \
  *  -H'Content-type: application/json' \
@@ -13,37 +15,27 @@ const { CloudEvent, HTTP } = require('cloudevents');
  *  -H'Ce-specversion: 1.0' \
  *  http://localhost:8080
  *
- * The event data is extracted from the incoming event and provided as the first
- * parameter to the function. The CloudEvent object itself may be accessed via the
- * context parameter, For example:
- *
- * const incomingEvent = context.cloudevent;
+ * For convenience, the runtime will extract the event data from the
+ * CloudEvent envelope and pass it in the second parameter to the
+ * function. The entire envelope, along with the HTTP request
+ * metadata, is available from the first parameter, 'context'.
  *
  * @param {Context} context the invocation context
- * @param {Object} user the CloudEvent data. If the data content type is application/json
- * this will be converted to an Object via JSON.parse()
+ * @param {Object} event the CloudEvent data
  */
-function verifyUser(context, user) {
-  if (!context.cloudevent) {
-    return {
-      message: 'No cloud event received'
-    };
-  }
+function handle(context, event) {
 
-  context.log.info('Processing user', user);
-  context.log.info(`CloudEvent received: ${context.cloudevent.toString()}`);
+  context.log.info("event");
+  console.log(JSON.stringify(event, null, 2));
 
-  user = verify(user);
+  context.log.info("cloudevent envelope");
+  console.log(JSON.stringify(context.cloudevent, null, 2));
+
   return HTTP.binary(new CloudEvent({
-    source: 'function.verifyUser',
-    type: 'user:verified',
-    data: user
+    source: 'event.handler',
+    type: 'echo',
+    data: event
   }));
 };
 
-function verify(user) {
-  // do something with the user
-  return user;
-}
-
-module.exports = verifyUser;
+module.exports = handle;
--- a/docs/guides/nodejs.md
+++ b/docs/guides/nodejs.md
@@ -86,20 +86,27 @@ If you prefer another, that's no problem. Just remove the `tape` dependency from
 
 ## Function reference
 
-Boson Node.js functions have very few restrictions. You can add any required dependencies
-in `package.json`, and you may include additional local JavaScript files. The only real
-requirements are that your project contain an `index.js` file which exports a single function.
-In this section, we will look in a little more detail at how Boson functions are invoked,
-and what APIs are available to you as a developer.
+Boson Node.js functions have very few restrictions. You can add any
+required dependencies in `package.json`, and you may include
+additional local JavaScript files. The only real requirements are that
+your project contain an `index.js` file which exports a single
+function.  In this section, we will look in a little more detail at
+how Boson functions are invoked, and what APIs are available to you as
+a developer.
 
 ### Invocation parameters
 
-When using the `func` CLI to create a function project, you may choose to generate a project
-that responds to a `CloudEvent` or simple HTTP. `CloudEvents` in Knative are transported over
-HTTP as a `POST` request, so in many ways, the two types of functions are very much the same.
-They each will listen and respond to incoming HTTP events.
+When using the `func` CLI to create a function project, you may choose
+to generate a project that responds to a `CloudEvent` or simple
+HTTP. `CloudEvents` in Knative are transported over HTTP as a `POST`
+request, so in many ways, the two types of functions are very much the
+same.  They each will listen and respond to incoming HTTP events.
 
-When an incoming request is received, your function will be invoked with a `Context` object as the first parameter. If the incoming request is a `CloudEvent`, any data associated with the `CloudEvent` is extracted from the event and provided as a second parameter. For example, a `CloudEvent` is received which contains a JSON string such as this in its data property, 
+When an incoming request is received, your function will be invoked
+with a `Context` object as the first parameter. If the incoming
+request is a `CloudEvent`, it will be provided as the second
+parameter. For example, a `CloudEvent` is received which contains a
+JSON string such as this in its data property,
 
 ```json
 { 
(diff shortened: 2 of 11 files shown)


templates/node/events/index.js
```diff
@@ -13,37 +15,27 @@ const { CloudEvent, HTTP } = require('cloudevents');
...
- * this will be converted to an Object via JSON.parse()
+ * @param {Object} event the CloudEvent data
  */
-function verifyUser(context, user) {
-  if (!context.cloudevent) {
-    return {
-      message: 'No cloud event received'
-    };
-  }
+function handle(context, event) {
```
zroubalik, line 26: I tend to agree with Lance on the naming of this param. I'd prefer `data` or `eventData`.

jcrossley3: Why?

templates/node/events/index.js
```diff
@@ -13,37 +15,27 @@ const { CloudEvent, HTTP } = require('cloudevents');
  *  -H'Ce-specversion: 1.0' \
  *  http://localhost:8080
  *
- * The event data is extracted from the incoming event and provided as the first
- * parameter to the function. The CloudEvent object itself may be accessed via the
- * context parameter, For example:
- *
- * const incomingEvent = context.cloudevent;
+ * For convenience, the runtime will extract the event from the
```
lance, line 18:
```suggestion
 * For convenience, the runtime will extract the event data from the
```

templates/node/events/index.js
```diff
@@ -13,37 +15,27 @@ const { CloudEvent, HTTP } = require('cloudevents');
...
+ * CloudEvent envelope and pass it in the second argument to the
```
lance, line 19: Nit
```suggestion
 * CloudEvent envelope and pass it in the second parameter to the
```

templates/node/events/index.js
```diff
@@ -13,37 +15,27 @@ const { CloudEvent, HTTP } = require('cloudevents');
...
+ * function. The entire envelope, along with the HTTP request
+ * metadata, is available from the first argument, 'context'.
```
lance, line 21:
```suggestion
 * metadata, is available from the first parameter, 'context'.
```

templates/node/events/index.js
lance, line 26: I prefer `data` or `eventData` as @zroubalik mentioned.

lance (commented):
Overall lgtm. The overall cleanup is nice, as is the intent. I have a couple of minor wording suggestions. And of course, you know how I feel about the parameter naming. Holding this open for more input on that.

templates/node/events/index.js
jcrossley3: What does the sentence mean to you without the word "data"? What potential misinterpretation are you trying to prevent?

templates/node/events/index.js
jcrossley3: Why?

I know it's just a variable name, but it's obviously important to both of us, so I'm going to reiterate the reasons I gave in slack here:

- The name `data` is not descriptive enough. It's analogous to other poor variable names such as `x` or `object` -- it conveys no domain meaning whatsoever.
- The name `eventData` is too descriptive. It risks confusion with the CloudEvent envelope's `data` field, which is an encoded string. The CloudEvent envelope is an implementation detail that I'd prefer we not expose in the function signature.
- The name `event` is just right. It represents the domain event reconstituted from the CloudEvent envelope according to its `data`, `datacontenttype`, and `dataschema` fields.
- The name `event` is as specific as we can get without knowing which domain is triggering the function. This PR is replacing a function that assumed the domain was user verification. And its second parameter was logically named `user`. It wasn't called `userData` because of course it has data. Otherwise, it wouldn't be a parameter at all!

I genuinely cannot fathom why you guys feel so strong about this. My only guess is that you feel your users will be expecting to work directly with CloudEvent envelopes. And that when they see the parameter name, `event`, they will expect that it's the envelope. If that's true, then we should be passing the envelope there, not the event reconstituted from its data. Personally, I'd rather have the latter, which is why I stand by the name.

But hey, I like to think I'm reasonable -- sell me!

matejvasek: Don't know what are anybody else expectations, but when I see `event` I expect it's the whole CloudEvent (including id,source,type etc.).

matejvasek: I am not sure if envelope is good word. Bill once compared CloudEvent to  SOAP envelope, but I don't recall the term from CloudEvent spec.

matejvasek: btw `data` is optional field for CloudEvent it might be nil

matejvasek: but I totally agree that just `data` doesn't tell much

matejvasek: > I genuinely cannot fathom why you guys feel so strongly about this. My only guess is that you feel your users will be expecting to work directly with CloudEvent envelopes. And that when they see the parameter name, event, they will expect that it's the envelope. If that's true, then we should be passing the envelope there, not the event reconstituted from its data.

Well IIRC in Go we are working directly with whole CloudEvent, user has to extract details (data) by themself.

matejvasek: I am not helping, am I.

lance: I think we are mostly coming at this with an awareness of the `CloudEvent` specification and that is influencing how we approach naming. I went back to the spec to see if it offered any illumination for us, and this is what I found under [Terminology](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#terminology).

> An "event" is a data record expressing an occurrence and its context.
> An "occurrence" is the capture of a statement of fact during the operation of a software system. 

The parameter we are discussing is not, in the CloudEvent specification's definition, the event. What is being provided in that parameter is the "statement of fact", expressed as a JavaScript object, extrapolated from the event's representation of the "fact" in some encoded form, typically JSON in the `data` field. While I agree that `data` is not in any way descriptive, I would argue that `event` is barely more so, and in this case I think it would be misleading.

I can understand the concern that adhering to this concept of an event leaks implementation details into the semantics of our API. Frankly, I think that's OK, as the CloudEvent is integral to what we are offering, is a proposed standard, and is the lingua franca of Knative Serverless components. In naming this parameter `event` I think we would be giving the impression that it is, in fact, the CloudEvent. It would lead me to believe that when I call `event.source` to determine where this "statement of fact" originated, I should get a value conforming to the CloudEvent spec's definition of `CloudEvent#source`. But in fact, I will get `undefined`.

jcrossley3: Helping as much as I am! :)

I don't think we should necessarily strive for cross-language consistency. We shouldn't impose conventions in one language due to limitations in another. I doubt any nodejs developers will care about the quarkus templates.

Regarding the term "envelope", I heard @nainaz and @slinkydeveloper use it, and I like it as a way to distinguish between a conceptual _event_ and its digital _representation_. For me, *envelope* == *CloudEvent*. For you guys building the stuff, it seems like *event* == *CloudEvent*. Not sure about your users (other than myself).

lance: In the world of CloudEvents, `event == CloudEvent` based on my understanding of this sentence in the spec.

> An "event" is a data record expressing an occurrence and its context.

The "context" in this sense is all of the event metadata, e.g. `source`, `type`, `id` and the "occurrence" is the data record itself, i.e. `event.data`. 

Providing `event.data` as the second parameter was done as a convenience, assuming that the function developer is primarily interested in the data that was generated by the "occurrence". But maybe there is equal interest in the other event properties as well. For example, `source` and `type` may be used to filter events or perform conditional actions in the function. If that's the case, I would be in favor of an `event` parameter that corresponds to the actual CloudEvent.

jcrossley3: I'm +1 on the 2nd parameter being an actual CloudEvent if that's the lingua franca of most users. I'm not a "CloudEvent guy" so I apologize for all the noise around this.

I'm assuming for a 2-argument handler, the js-runtime will pass the CE in the second parameter and it will *NOT* be in the context.

And for a 1-argument handler, only the context will be passed, and it will include the CE as it currently does.

Correct? Want me to create an issue in the js-runtime for that?

zroubalik: Agree with Lance and Matej, if I read `event` I expect a whole `CloudEvent`, but I might be biased. And because we are using Knative Eventing and users (fn developers) will need to know at least some basic stuff about KEventing and it's concepts, in order to write the functions properly, so therefore I think that a term `CloudEvent` would be at least somehow familiar for them, so eventually they might get the same impression as I do. ie. expect a `CloudEvent` in `event` parameter.

Not sure if changing the 2nd parameter to be a full CE is somehow useful, I suppose that parameter was introduced so users can easily get the actual data 🤷‍♂️

lkingland: When implementing a Handler with Naina (having no familiarity with the function signatures of the language beforehand), we were both thrown off by the second parameter not being a CloudEvent.  It seemed an obvious assumption that the _event_ would be the argument.  Or, at least this would be a supported default.  I would certainly advocate for that change, while at the same time completely agreeing with Jim that we should not force consistency between languages; instead working to their strengths.

templates/node/events/index.js
```diff
@@ -13,37 +15,22 @@ const { CloudEvent, HTTP } = require('cloudevents');
...
 
-  user = verify(user);
   return HTTP.binary(new CloudEvent({
-    source: 'function.verifyUser',
-    type: 'user:verified',
-    data: user
+    source: 'event.handler',
+    type: 'echo',
+    data: event
   }));
```
lkingland, line 33: So with an eye on our related conversation about accepting an actual CloudEvent named _event_ as the second parameter, it may gain some support from its effect on the default implementation.  If we assume that the default handler echoes a cloud event, this cleans up nice:
```
function handle(context, event) {
  return event
}
```

lkingland (commented):
Looks nice to me! Pending resolution on the CloudEvent parameter and naming of course.

templates/node/events/index.js
jcrossley3: That absolutely appeals to my aesthetic!! :)

However, I kinda dig showing a bit of the `HTTP` library, and I think it's important to set the source & type fields on the returned event.

It might be more useful to the new `func` user to see `data: event.data` though, just to reinforce that the event is an actual `CloudEvent`.  But I'd rather not do that until after the runtime is changed to do that. As it stands, this template will work with either the current or the future runtime.

jcrossley3: Also, if the event source is a broker, won't returning the event result in an infinite loop?

lance: > Also, if the event source is a broker, won't returning the event result in an infinite loop?

That's a good question. I suppose only if the function is subscribed via Trigger, and only filtering on event type - and even then, that assume's the broker is not smart enough to deal with this.

lance: > However, I kinda dig showing a bit of the `HTTP` library, and I think it's important to set the source & type fields on the returned event.

```js
function handle(context, event) {
  return event.cloneWith({ source: 'echo.fn', type: 'echo' });
}
```

lance: >  But I'd rather not do that until after the runtime is changed to do that.

If this lands, https://github.com/boson-project/faas-js-runtime/pull/96 - will you coordinate with this PR? ([TypeScript](https://github.com/boson-project/func/pull/355) needs to change as well).

jcrossley3: > If this lands, [boson-project/faas-js-runtime#96](https://github.com/boson-project/faas-js-runtime/pull/96) - will you coordinate with this PR? ([TypeScript](https://github.com/boson-project/func/pull/355) needs to change as well).

Sure, but... define "coordinate" in this context. I think they can each be merged independently. Or I can wait until after #355  lands and adjust those templates, too?

lance: @jcrossley3 I mean, primarily, that this PR is updated to reflect those changes. Due to the fact that these are user facing API changes, we'll also want to update docs/guides/nodejs.md.

jcrossley3: Ah, gotcha. I just pushed a commit that updates the nodejs guide.

lance (approved):
This LGTM. Thank you @jcrossley3 for being so patient with this process. In the end I think this is a nice improvement!

lkingland (approved):

lance:
@jcrossley3 I have just realized that the `faas-js-runtime` dependencies in both event/package.json and http/package.json should be updated to 0.7.1 before this lands.

jcrossley3:
> @jcrossley3 I have just realized that the `faas-js-runtime` dependencies in both event/package.json and http/package.json should be updated to 0.7.1 before this lands.

Done

PR #359, opened 2021-05-25
chore: bump to Go 1.15

zroubalik:

Is there a particular reason why we are still at 1.14? Knative Client is using 1.15 - we should sync that.


lance (approved):
Huh - I guess we prematurely went from 1.14 to 1.16 earlier (for templates which ultimately had to be rolled back).

PR #365, opened 2021-05-27
chore(builders): add explicit version numbers for builder images

lance:
This commit adds specific version numbers to each of the builder images
referenced in function templates, func.yaml file. Because the API for
at least some of the runtimes has changed over time (looking at you,
faas-js-runtime), we should consider publishing our func.yaml files with
known-to-be-working-with-this-release versioned builder images.


zroubalik (approved):

lkingland (approved):
As much as it is a bit of a PITA; pegging versions is the probably the right way to go here.  LGTM!

PR #373, opened 2021-06-03
feat: add support for labels in func.yaml

lance:
This change adds support for setting labels on deployed functions via `func.yaml`. The CLI will assume that `func.yaml` is the definitive source of information about what labels should exist on the deployed Service, and each deploy ensures that the labels in the deployed Service and what's in `func.yaml` are in sync.

Fixes: https://github.com/boson-project/func/issues/261


zroubalik (commented):
Looking good, minor nit: do we want to validate the specified labels in the `func.yaml` or do we want to keep the validation on the actual deployment process via Knative?

Documentation is missing for this field.

zroubalik:
`kn`  allows users to set and unset labels this way, I wonder whether we might want to use similar setting:
```
  -l, --label stringArray                 Labels to set for both Service and Revision. name=value; you may provide this flag any number of times to set multiple labels.
                                          To unset, specify the label name followed by a "-" (e.g., name-).
```

lkingland (approved):

lance:
> Looking good, minor nit: do we want to validate the specified labels in the `func.yaml` or do we want to keep the validation on the actual deployment process via Knative?

What kind of validation are you thinking of?

lance:
> `kn` allows users to set and unset labels this way,

For us, I suppose the `-l` flag would be only on `deploy`. Is this what you had in mind?

zroubalik:
> > Looking good, minor nit: do we want to validate the specified labels in the `func.yaml` or do we want to keep the validation on the actual deployment process via Knative?
> 
> What kind of validation are you thinking of?

Just a check that the input value is a valid k8s label, as per: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

But that's just a question, if we don't do the check it will fail during the deployment anyway.

zroubalik:
> > `kn` allows users to set and unset labels this way,
> 
> For us, I suppose the `-l` flag would be only on `deploy`. Is this what you had in mind?

I just want to give users way how to remove a label, because the current solution allows only adding/updating labels, not removing (they would have to undeploy and deploy the function again in order to get rid of labels that were specified before). 

I am not sure that the flag is the best solution, but what else do we have? 🤷‍♂️

lance:
I've updated this so that it's integrated with the interactive prompt that Zbynek recently introduced, and I have added tests. I think it should be ready for review. PTAL

/cc @zroubalik @lkingland @matejvasek

cmd/config_labels.go
```diff
@@ -0,0 +1,286 @@
...
+			{
+				Name:   "name",
+				Prompt: &survey.Input{Message: "Please specify the label name:"},
+				Validate: func(val interface{}) error {
+					return utils.ValidateLabelName(val.(string))
+				},
+			},
+			{
+				Name:   "value",
+				Prompt: &survey.Input{Message: "Please specify the label value:"},
```
zroubalik, line 170: Shouldn't be the value validated as well?
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

zroubalik (commented):
Looking good, minor nits only.

config.go
```diff
@@ -42,13 +42,13 @@ func (v Volume) String() string {
 	return ""
 }
 
-type Envs []Env
-type Env struct {
+type Tuples []Tuple
+type Tuple struct {
```
matejvasek, line 46: Maybe `Pair`, `NamedValue` or `KeyValuePair` would be better name for the type.

zroubalik: or `NameValuePair` bc the fields are Name` and `Value`

zroubalik: And another point, `Labels` are `key/value` pairs not `name/value` pairs. So I am not 100% sure we should name it this way in the config.

matejvasek: But isn't tuple ordered sequence of zero or more elements, not just two?

docs/guides/func_yaml.md
```diff
@@ -51,17 +51,40 @@ envs:
 - value: '{{ configMap:myconfigmap2 }}'     # (4) all key-value pairs in ConfigMap as env variables
 ```
 
-### `volumes`
-Kubernetes Secrets or ConfigMaps can be mounted to the function as a Kubernetes Volume accessible under specified path. Below you can see an example how to mount the Secret `mysecret` to the path `/workspace/secret` and the ConfigMap `myconfigmap` to the path `/workspace/configmap`. This Secret/ConfigMap needs to be created before it is referenced in a function.
+### `image`
```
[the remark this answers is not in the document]
lance: > I do understand that alphabetical order may make sense in here, but shouldn't we instead sort the section of this docs as the fields are really sorted in `func.yaml`?

Personally I would prefer alphabetical, since that's more intuitive if you don't have `func.yaml` open.

config.go
lance: I'm fine changing it. I didn't look up the exact definition when I used the `Tuple` type name. I just used it based on past experiences. But when reading the formal definitions, I see your points. (TY for the support Luke). I know `NameValuePair` is probably the most descriptive. But in the interest of fancy type names, what about `Dyad`?

zroubalik: And for `Labels` do you still wan to keep the exposed config as:
```yaml
labels:
- name: role       
  value: backend
```

But if we follow kubernetes naming it should be:
```yaml
labels:
- key: role
  value: backend
```

WDYT?

lance: I had it that way and changed it back for consistency. I'm open to do what the hive mind thinks is best on this.

zroubalik: My vote would be for key/value, so the user facing config is consistent with k8s. The internal code consistency is lower priority imho.

matejvasek: As for `key vs. name` I vote for keeping it consistent within our config. Having it different is IMHO confusing for our user.

lance: @matejvasek just to be clear, you think that the config should look like this?
```
labels:
- name: role       
  value: backend
envs:
- name: API_KEY
  value: somesecret
```

Or this?
```
labels:
- key: role       
  value: backend
envs:
- key: API_KEY
  value: somesecret
```
Or something else?

matejvasek: I like the first one more. I just prefer have it consistent within the file as is could be confusing having it different. but maybe users wouldn't mind.

lance:
@zroubalik @matejvasek I settled on `Pair`. And I tend to agree with Matej about the file consistency. A function developer isn't going to care if it's `key` or `name`, and most probably won't know that Kube treats these differently, so having differences in func.yaml for no apparent (to the developer) reason.

Also added label value validation. PTAL

utils/names.go
```diff
@@ -40,3 +43,27 @@ func ValidateEnvVarName(name string) error {
...
+// a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed
+// by a slash (/).
+// Valid label values must be 63 characters or less (can be empty),
+// unless empty, must begin and end with an alphanumeric character ([a-z0-9A-Z]),
+// could contain dashes (-), underscores (_), dots (.), and alphanumerics between.
+// Label values may also come from the environment and therefore, could be enclosed with {{}}
+// Treat this as a special case.
+func ValidateLabel(name, value string) error {
+	errs := validation.IsQualifiedName(name)
+	if !strings.HasPrefix(value, "{{") {
```
lance, line 61: The only issue here is that if the eventual value of the environment variable is an invalid label value, we won't catch it. Maybe I should evaluate the environment variable, and try to validate that.

zroubalik: +1

matejvasek:
why is `pkged.go` modified?

cmd/config_labels.go
```diff
@@ -0,0 +1,290 @@
...
+				Prompt: &survey.Input{Message: "Please specify the label name:"},
+				Validate: func(val interface{}) error {
+					return utils.ValidateLabel(val.(string), "")
+				},
+			},
+			{
+				Name:   "value",
+				Prompt: &survey.Input{Message: "Please specify the label value:"},
+				Validate: func(val interface{}) error {
+					return utils.ValidateLabel("", val.(string))
```
matejvasek, line 172: This will always fails since name must be non-empty.

cmd/config_labels.go
```diff
@@ -0,0 +1,290 @@
...
+				Prompt: &survey.Input{Message: "Please specify the label name:"},
+				Validate: func(val interface{}) error {
+					return utils.ValidateLabel(val.(string), "")
+				},
+			},
+			{
+				Name:   "value",
+				Prompt: &survey.Input{Message: "Please specify the local environment variable:"},
+				Validate: func(val interface{}) error {
+					return utils.ValidateLabel("", val.(string))
```
matejvasek, line 205: This will always fail since name must be non-empty.

utils/names.go
matejvasek: >Maybe I should evaluate the environment variable, and try to validate that.

but that can be only done in `deploy` right?

lance: > > Maybe I should evaluate the environment variable, and try to validate that.
> 
> but that can be only done in `deploy` right?

It can be evaluated with `func config` as well.

```
> ./func config -p viewer
Error: 'func.yaml' config file is not valid:
  label entry #0 with name 'foo' has invalid value when the environment is evaluated: '-badvalue': a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

matejvasek: yeah but then it might be re-set to something else latter before `deploy`

lance: > yeah but then it might be re-set to something else latter before `deploy`

That would be OK, right? If my func.yaml looks like this

```
labels:
- name: foo
  value: {{ env:FOO }}
```
And in my local environment I have
```
echo $FOO
-bar # this is an invalid label value
```
When I'm using `func` CLI, I think it's preferable to know before I try to deploy that whatever the value is in my environment is invalid. If I then change the value to something valid, subsequent CLI commands will work. What other did you have in mind?

matejvasek: >When I'm using func CLI, I think it's preferable to know before I try to deploy that whatever the value is in my environment is invalid.

OK makes sense, but what do we do in opposite case -- somebody unset the var latter, it makes deployment fail automatically (because server won't accept that)?

lance: > OK makes sense, but what do we do in opposite case -- somebody unset the var latter, it makes deployment fail automatically (because server won't accept that)?

The func.yaml file is read and validated with each command. So, if later the env var is set to something invalid, when `func deploy` is run, the first thing that happens is the func.yaml file is read and validated.

zroubalik: Yeah, agree with Lance on this. I don't think there anything else that we can do.

config.go
```diff
@@ -42,13 +42,13 @@ func (v Volume) String() string {
...
-type Envs []Env
-type Env struct {
+type Pairs []Pair
+type Pair struct {
 	Name  *string `yaml:"name,omitempty"`
 	Value *string `yaml:"value"`
 }
 
-func (e Env) String() string {
+func (e Pair) String() string {
```
matejvasek, line 51: The name of the type has been change to something more generic than `Env` but stringer still contains `Env \"%s\"` in fmt string

lance: Good catch!

lance:
OK - one more time @lkingland @zroubalik and @matejvasek - sorry this has been so long in the making. I think finally it's good for a final review.

lkingland:
@lance a recent PR fixes the Makefile such that `build` again correctly omits updating `pkged.go` unless there has been changes in `./templates`.

config.go
```diff
@@ -61,18 +61,18 @@ func (e Env) String() string {
 	} else if e.Name != nil && e.Value != nil {
 		match := regKeyFromSecret.FindStringSubmatch(*e.Value)
 		if len(match) == 3 {
-			return fmt.Sprintf("Env \"%s\" with value set from key \"%s\" from Secret \"%s\"", *e.Name, match[2], match[1])
+			return fmt.Sprintf("Pair \"%s\" with value set from key \"%s\" from Secret \"%s\"", *e.Name, match[2], match[1])
```
zroubalik, line 64: Nit, is `Pair` the right message from users point of view? It might be a little bit misleading imho, what about `Entry` or something similar?

zroubalik (commented):
Lookin good, just a minor nit

config.go
lance: That's a reasonable point. Fixed.

Issue #264, opened 2021-02-18
RFE: Add the ability to set min/max concurrency when deploying


lance:

zroubalik:
Request for the general support for kn flags is tracked here: https://github.com/boson-project/func/issues/164

Issue #164, opened 2020-10-09
`faas deploy` should support options available for `kn service create`


zroubalik:
Users might want to tweak the deployed Function (Knative Service).

There are bunch of [options](https://github.com/knative/client/blob/master/docs/cmd/kn_service_create.md#options) available for `kn service create` command, such as cpu/memore requests, scaling configuration etc.

Let's start listing here the options we'd like to support, and link to any PRs or Issues

- [x]  `deploy`: min/max concurrency, scaling related settings:  [Issue](https://github.com/boson-project/func/issues/264), [PR](https://github.com/boson-project/func/pull/374) 
- [x] `deploy`: resource requests/limits & container concurrency limit: [issue](https://github.com/boson-project/func/issues/383), PR

PR #374, opened 2021-06-04
feat: allow setting autoscaling options to deployed KService

zroubalik:

Users can specify (autoscaling) options in `func.yaml`. The options are added/update/removed based on the content in `func.yaml` file.

If there's a demand, we can add these options as flags to `func deploy` command (same way they are exposed in `kn service create/update` -> with removal option), but I think that specification in config file is enough.

```yaml
name: test
namespace: ""
runtime: go
...
options:
  scale:
    min: 0
    max: 10
    metric: rps
    target: 75
    utilization: 75
```

`concurrency.limit` will be added in the following PR, which will handle resource limits/requests as proposed here: https://github.com/boson-project/func/pull/374#discussion_r649070413

Relates: #164
Fixes: #264

### DIFF
--- a/config.go
+++ b/config.go
@@ -34,15 +34,15 @@ type Options struct {
 }
 
 type ScaleOptions struct {
-	Min    *int64  `yaml:"min,omitempty"`
-	Max    *int64  `yaml:"max,omitempty"`
-	Metric *string `yaml:"metric,omitempty"`
+	Min         *int64   `yaml:"min,omitempty"`
+	Max         *int64   `yaml:"max,omitempty"`
+	Metric      *string  `yaml:"metric,omitempty"`
+	Target      *float64 `yaml:"target,omitempty"`
+	Utilization *float64 `yaml:"utilization,omitempty"`
 }
 
 type ConcurrencyOptions struct {
-	Limit       *int64   `yaml:"limit,omitempty"`
-	Target      *float64 `yaml:"target,omitempty"`
-	Utilization *float64 `yaml:"utilization,omitempty"`
+	Limit *int64 `yaml:"limit,omitempty"`
 }
 
 // Config represents the serialized state of a Function's metadata.
@@ -308,6 +308,21 @@ func validateOptions(options Options) (errors []string) {
 					*options.Scale.Metric))
 			}
 		}
+
+		if options.Scale.Target != nil {
+			if *options.Scale.Target < 0.01 {
+				errors = append(errors, fmt.Sprintf("options field \"scale.target\" has value set to \"%f\", but it must not be less than 0.01",
+					*options.Scale.Target))
+			}
+		}
+
+		if options.Scale.Utilization != nil {
+			if *options.Scale.Utilization < 1 || *options.Scale.Utilization > 100 {
+				errors = append(errors,
+					fmt.Sprintf("options field \"scale.utilization\" has value set to \"%f\", but it must not be less than 1 or greater than 100",
+						*options.Scale.Utilization))
+			}
+		}
 	}
 
 	// options.concurrency
@@ -316,21 +331,6 @@ func validateOptions(options Options) (errors []string) {
 			if *options.Concurrency.Limit < 0 {
 				errors = append(errors, fmt.Sprintf("options field \"concurrency.limit\" has value set to \"%d\", but it must not be less than 0",
 					*options.Concurrency.Limit))
-			}
-		}
-
-		if options.Concurrency.Target != nil {
-			if *options.Concurrency.Target < 0.01 {
-				errors = append(errors, fmt.Sprintf("options field \"concurrency.target\" has value set to \"%f\", but it must not be less than 0.01",
-					*options.Concurrency.Target))
-			}
-		}
-
-		if options.Concurrency.Utilization != nil {
-			if *options.Concurrency.Utilization < 1 || *options.Concurrency.Utilization > 100 {
-				errors = append(errors,
-					fmt.Sprintf("options field \"concurrency.utilization\" has value set to \"%f\", but it must not be less than 1 or greater than 100",
-						*options.Concurrency.Utilization))
 			}
 		}
 	}
--- a/config.go
+++ b/config.go
@@ -8,7 +8,6 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
-	"time"
 
 	"gopkg.in/yaml.v2"
 )
@@ -30,13 +29,20 @@ type Env struct {
 }
 
 type Options struct {
-	ScaleMin              *int    `yaml:"scale-min,omitempty"`
-	ScaleMax              *int    `yaml:"scale-max,omitempty"`
-	ScaleInit             *int    `yaml:"scale-init,omitempty"`
-	AutoscaleWindow       *string `yaml:"autoscale-window,omitempty"`
-	ConcurrencyLimit      *int64  `yaml:"concurrency-limit,omitempty"`
-	ConcurrencyTarget     *int    `yaml:"concurrency-target,omitempty"`
-	ConcurrencyUtilizatin *int    `yaml:"concurrency-utilization,omitempty"`
+	Scale       *ScaleOptions       `yaml:"scale,omitempty"`
+	Concurrency *ConcurrencyOptions `yaml:"concurrency,omitempty"`
+}
+
+type ScaleOptions struct {
+	Min    *int64  `yaml:"min,omitempty"`
+	Max    *int64  `yaml:"max,omitempty"`
+	Metric *string `yaml:"metric,omitempty"`
+}
+
+type ConcurrencyOptions struct {
+	Limit       *int64   `yaml:"limit,omitempty"`
+	Target      *float64 `yaml:"target,omitempty"`
+	Utilization *float64 `yaml:"utilization,omitempty"`
 }
 
 // Config represents the serialized state of a Function's metadata.
@@ -189,7 +195,7 @@ func writeConfig(f Function) (err error) {
 }
 
 // validateVolumes checks that input Volumes are correct and contain all necessary fields.
-// Returns array of error messages, empty if none errors are found
+// Returns array of error messages, empty if no errors are found
 //
 // Allowed settings:
 // - secret: example-secret              		# mount Secret as Volume
@@ -219,7 +225,7 @@ func validateVolumes(volumes Volumes) (errors []string) {
 }
 
 // ValidateEnvs checks that input Envs are correct and contain all necessary fields.
-// Returns array of error messages, empty if none errors are found
+// Returns array of error messages, empty if no errors are found
 //
 // Allowed settings:
 // - name: EXAMPLE1                					# ENV directly from a value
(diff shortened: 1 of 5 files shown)


config.go
```diff
@@ -28,6 +29,16 @@ type Env struct {
...
 }
 
+type Options struct {
+	ScaleMin              *int    `yaml:"scale-min,omitempty"`
+	ScaleMax              *int    `yaml:"scale-max,omitempty"`
+	ScaleInit             *int    `yaml:"scale-init,omitempty"`
+	AutoscaleWindow       *string `yaml:"autoscale-window,omitempty"`
+	ConcurrencyLimit      *int64  `yaml:"concurrency-limit,omitempty"`
+	ConcurrencyTarget     *int    `yaml:"concurrency-target,omitempty"`
+	ConcurrencyUtilizatin *int    `yaml:"concurrency-utilization,omitempty"`
```
zroubalik, line 39: I am not sure whether we should use the same notation as it is used in `kn` ie. dashes in the name or whether we should use CamelCase notation that's being used in the rest of the file...

lkingland: tl;dr  I see no compelling reason to maintain exact equivalence, and lean towards CamelCase at a minimum.

My first thought on reading this was much the same:  Glad we have it; a bit less enthusiastic about the naming.  I would even go a little further... Following is perhaps overly opinionated, but why not share!

I am sure that many would encourage us to maintain homogeneity between the two.  However, I think this actually leads to a couple challenges:  First, it encourages a false equivalence between the systems, causing confusion by users and developers alike; especially when we do deviate.  Second, we risk grandfathering in naming which, with the benefit of hindsight, could be easily improved or made to fit our use case a bit better.  I see the argument for exact equivalence as valid, but weaker than either:  "it's easier to remember if you're familiar with both systems".

We have a golden and ephemeral opportunity to lay down consistent, coherent naming now, before we are restrained by the iron yoke backwards compatibility.  I say we take it whenever the change is a clear improvement and not just bike-shedding.

For instance, would anyone argue that this is clearer semantics?
```
scaling:
  initial: 2
  min: 1
  max:  10
  window: 10s
concurrency:
  target: 75
  utilization: 75
  limit: 100
```
(This is of course just a conversational example more than an actual suggestion; illustrating a point.  I like the practical original suggestion of adopting CamelCase just fine)

lance: I actually like your "conversational example" a bit more than using an `options` keyword. My first thought upon seeing the PR was, "why an `options` key" instead of just having `scale-min`, for example, at the top level? I suppose there is benefit in the fact that, as new flags/options are added they can all just be stuffed into one `options` bag instead of being scattered across the file. I do agree, however, that your example, @lkingland is more readable/understandable. What about something like this (also a conversational example, but maybe also a suggestion :shrug:)

```yaml
options:
  scaling:
    initial: 2
    min: 1
    max:  10
    window: 10s
  concurrency:
    target: 75
    utilization: 75
    limit: 100
```

config.go
```diff
@@ -166,7 +189,7 @@ func writeConfig(f Function) (err error) {
 }
 
 // validateVolumes checks that input Volumes are correct and contain all necessary fields.
-// Returns array of error messages, empty if none
+// Returns array of error messages, empty if none errors are found
```
lance, line 192:
```suggestion
// Returns array of error messages, empty if no errors are found
```
Such a tiny nit here and below...

config.go
lance, line 39:
```suggestion
	ConcurrencyUtilization *int    `yaml:"concurrency-utilization,omitempty"`
```
Spelling nit

config.go
```diff
@@ -241,7 +264,27 @@ func ValidateEnvs(envs Envs) (errors []string) {
...
 			}
+		}
+	}
+
+	return
+}
+
+// validateOptions checks that input Options are correctly set.
+// Returns array of error messages, empty if none errors are found
+func validateOptions(options Options) (errors []string) {
```
lance, line 275: Why no validation for the other options? Seems like you could check them, for example, similar to this.

- `minScale >= 0`
- `maxScale >= minScale`
- `initialScale >= minScale`
- `0s <= window <= 1h`
- `utilization <= 100`

docs/guides/func_yaml.md
```diff
@@ -62,6 +62,27 @@ volumes:
   path: /workspace/configmap
 ```
 
+### `options`
+Options allows you to set specific configuration for the deployed funciton, ie. it allows you to tweak Knative Service options related to autoscaling and other properties, if these options are not set in the default ones will be used. This options are based on the ones provided by `kn service` [command](https://github.com/knative/client/blob/main/docs/cmd/kn_service_create.md#options).
```
lance, line 66:
```suggestion
Options allows you to set specific configuration for the deployed funciton, allowing you to tweak Knative Service options related to autoscaling and other properties. If these options are not set, the Knative defaults will be used. 
```

Minor wordsmithing, and I think it might be best to avoid referencing the `kn` options, just to avoid confusion among users - but I'm not firm on this suggestion.

docs/guides/func_yaml.md
```diff
@@ -62,6 +62,27 @@ volumes:
...
+- `scale-min`: Minimum number of replicas.
+- `scale-max`: Maximum number of replicas.
+- `scale-init`: Initial number of replicas with which a Service starts. Can be 0 or a positive integer.
+- `autoscale-window`: Duration to look back for making auto-scaling decisions. The Service is scaled to zero if no request was received in during that time. (eg: 10s)
+- `concurrency-limit`: Hard Limit of concurrent requests to be processed by a single replica.
+- `concurrency-target`: Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to `concurrency-limit` when given.
+- `concurrency-utilization`: Percentage of concurrent requests utilization before scaling up.
```
lance, line 73: It might be beneficial to have the default values listed as well.

lance (commented):
In general looks great! A few minor comments/suggestions.

docs/guides/func_yaml.md
lance: Oh also a typo... `funciton`.

config.go
zroubalik: Yeah, I created the `options` exactly for this, to not have all the settings scattered across the func.yaml file.

I like this suggestion, will refactor the config this way, thanks a lot for the input.

config.go
zroubalik: Thx!

docs/guides/func_yaml.md
zroubalik: I was thinking about documenting default values, but those are based on a entries in a config in knative-serving namespace, so they could vary across individual installations.

zroubalik:
@boson-project/core  PTAL, I have refactored this PR per feedback, thanks for that!

@markusthoemmes suggested that we should dropp `autoscale window` and `init scale` as those are not yet stable and might break, I have introduced `scale metric` instead.

config.go
```diff
@@ -28,6 +28,23 @@ type Env struct {
...
+	Concurrency *ConcurrencyOptions `yaml:"concurrency,omitempty"`
+}
+
+type ScaleOptions struct {
+	Min    *int64  `yaml:"min,omitempty"`
+	Max    *int64  `yaml:"max,omitempty"`
+	Metric *string `yaml:"metric,omitempty"`
+}
+
+type ConcurrencyOptions struct {
```
markusthoemmes, line 42: With the introduction of `metric` above, is this really still limited to `concurrency`? I'd guess only `Limit` is?

zroubalik: How would you map those values into the config then so it is not bloated and easy to read? I am open to any suggestions :)

markusthoemmes: maybe `target` and `utilization` should move into the scale options (that's what they are after all), and the concurrency limit should get a special place somehow (as it also is in Knative: It's inferred as an autoscaling setting (inferred into `target` actually) but also significant as a routing and traffic behavior setting)

rhuss: Actually, I think one design goal would be to allow to map the config names here directly to `kn` options. Unfortunately, we have chosen `--concurrency-target` and `--concurrency-utilization`, so its not so easy to move those around without confusing this UX. Of course we could do this hand-in-hand with a corresponding change on `kn`, but this would be a more tricky because of backwards compatibility issues ...

lance: > Actually, I think one design goal would be to allow to map the config names here directly to `kn` options. Unfortunately, we have chosen `--concurrency-target` and `--concurrency-utilization`, so its not so easy to move those around without confusing this UX.

```yaml
options:
  scaling:
    initial: 2
    min: 1
    max:  10
    window: 10s
  concurrency:
    target: 75
    utilization: 75
    limit: 100
```

Have we not already made enough of a leap with the YAML in this format to assume the user experience does not have a direct correlation to `kn` without at least a little interpretation? The function development experience is meant to be clear, concise, opinionated, and targeted towards developers new to Knative. Does a strict adherence to `kn` command semantics necessarily help this goal? I'm not sure. :shrug: 

I know that, as a Linux user I can never remember if I should use `adduser` or `useradd`, so I understand the UX concern! :smile: 

> Of course we could do this hand-in-hand with a corresponding change on kn, but this would be a more tricky because of backwards compatibility issues ...

What if we think of `func` (and any other `kn` plugin, I suppose) as testing ground for user experiences that __may__ make their way into `kn`, with proper deprecation timelines and `--experimental` flags? So that instead of needing to make a hand-in-hand corresponding change, there is a process for `kn` in adopting features/usability/etc from plugins.

markusthoemmes: Ack, I don't think we need to inherit all design choices from `kn` (or even Knative!) if we can do better.

zroubalik: @markusthoemmes so you'd like to have the config like this?
```
name: test
namespace: ""
runtime: go
...
options:
  scale:
    min: 0
    max: 10
    metric: rps
    target: 75
    utilization: 75
  concurrency:
    limit: 100
```

lkingland: > Ack, I don't think we need to inherit all design choices from `kn` (or even Knative!) if we can do better.

We can do better indeed!  We have the benefits of both hindsight and a higher level of abstraction, thanks in no small part to kn and Knative in general.  I most certainly like the clarity and simplicity of these options as they are shaping up here.

lkingland (approved):
Well, I like it.

config.go
rhuss: Ok, fine with it and let's fix in on `kn` and also support other types of scaling then only concurrency (rps, cpu)

Just to be sure about this particular change: `concurrency.limit` is really only about the concurrency while `scale.target` is about any of concurrency, rps or cpu depending of which scaling type has been selected with the annotation `autoscaling.knative.dev/metric`. How would one specify "hard limits" then for RPS or CPU based scaling ? (if we stay with the analogy of soft and hard limits for target/limit). Is it also via `...spec.containerConcurrency` (then that field would be misnamed) or is this just not be possible ? If it is the former case, then I would put the `concurrency.limit` also to the scale group, as it feels unnatural to have target and limit in different sections.

rhuss: > Ok, fine with it and let's fix in on kn and also support other types of scaling then only concurrency (rps, cpu)

Introducing the scaling metric (`--scale-metric`) is then also a good hook for changing the `kn` option name to `--scale-target`.

I still believe having user-faced wording aligned across abstractions is a good thing.

markusthoemmes: There are no hard limits for RPS. CPU limits would be set via the container's resource block.

I would not conflate those though. The concurrency limit is not just impacting scale, but it's also impacting traffic behavior. Likewise, CPU-limits are not only impacting scale but also the behavior of the container. Target on the other hand, is strictly a scaling concern.

zroubalik: I have moved `target` and `utilization` to `scale` section in https://github.com/boson-project/func/pull/374/commits/7a8b7ef6bcf4a13c1f6858190b7db8fd8dedafa6

rhuss: Ok, I understand. Then I wonder why not then grouping all limits together ? Like something as:

```yaml
resources:
  limits:
    cpu: ....
    memory: ....
    concurrency: ....
  requests:
    cpu: ...
    memory: ...
```

@zroubalik How do you specify the resource limits currently for the container ?

zroubalik: Resource requests/limits are not being exposed at the moment. But my idea was to add them once we agree on the scaling options and my intention was to add them to the options exactly the way as you proposed. The concurrency added to limits seems nice! Great idea.

markusthoemmes: Nice idea indeed! I like!

zroubalik: I have removed `concurrency.limits` from this PR and will handle that in another one together with resource requests/limits.

lance (approved):

PR #375, opened 2021-06-07
feat: typed errors for templates use cases

lkingland:
Typed errors when dealing with both embedded and external template repositories.


client_test.go
```diff
@@ -180,37 +180,107 @@ func TestExtensibleTemplates(t *testing.T) {
 
 	// Create a new client with a path to the extensible templates
 	client := bosonFunc.New(
-		bosonFunc.WithTemplates("testdata/templates"),
+		bosonFunc.WithTemplates("testdata/repositories"),
```
lkingland, line 183: This will be renamed to "packages" now that we have decided on a rename of --tempates per #381

lance (commented):
I like the typed errors. That's a nice addition. I want to make sure we're in agreement on a repository/package namespacing.

client_test.go
lkingland: Opeining this as a separate issue, as it has a fair bit of discussion around it.

lkingland:
> I like the typed errors. That's a nice addition. I want to make sure we're in agreement on a repository/package namespacing.

Sounds good.  Our agreed-upon solution is in #384

lance (approved):

zroubalik (approved):

PR #376, opened 2021-06-07
feat: Rust templates

jcrossley3:
Depends on https://github.com/boson-project/buildpacks/pull/95

Open questions:

1. I added a `test-rust` make target, but didn't add it to the existing `test` target yet. Does CI run that? Is `cargo` available there?
2. I did not reference a specific tagged version of the builder image in `.builders.yaml` and I'm not sure who makes `latest` available. Is that also CI?
3. Should I squash these to make a more cohesive commit message?


lance:
> Open questions:
> 
>     1. I added a `test-rust` make target, but didn't add it to the existing `test` target yet. Does CI run that? Is `cargo` available there?

We use GH Actions for CI, and yes you can add cargo https://github.com/actions-rs/cargo.

> 
>     2. I did not reference a specific tagged version of the builder image in `.builders.yaml` and I'm not sure who makes `latest` available. Is that also CI?

Yes when we publish a new version CI pushes to quay.io

> 
>     3. Should I squash these to make a more cohesive commit message?

Your choice. We can squash when landing too.

templates/rust/http/src/main.rs
```diff
@@ -0,0 +1,52 @@
+use actix_web::{http::Method, web, App, HttpRequest, HttpResponse, HttpServer};
+
+use env_logger as elog;
+use log::info;
+
+async fn index(req: HttpRequest) -> HttpResponse {
```
lance, line 6: I'm curious why this function is async. I apologize for my Rust ignorance. How is this async and how does that work in an HTTP request?

lance (commented):
Thanks for this contribution @jcrossley3! One of the things we try to do is hide all of the plumbing from the function developer so that they are presented with only a function in their source file. Would it be possible to modify this template/buildpack combo so that it looks a little more like the Go template/buildpack?

templates/rust/http/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+            .route(
+                "/health/{_:(readiness|liveness)}",
+                web::get().to(HttpResponse::Ok),
+            )
+    })
+    .bind(("127.0.0.1", port))?
+    .workers(1)
+    .run()
+    .await
+}
```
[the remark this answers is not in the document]
jcrossley3: That was actually my initial approach. Ultimately, I decided against it, mostly for three reasons:

1. There are so many different ways that `App` can be configured depending on the handler function signature that I couldn't justify hiding that from the user. The template is just scratching the surface of what actix provides.
2. The handler would be referring to actix types, so there will always be a potential version mismatch: both the runtime and the user's handler will have to depend on compatible versions of actix. I couldn't justify introducing that obstacle.
3. Library dependencies in rust are expressed as "crates" so I would either have to publish a crate containing these few lines of configuration code that a user will likely want to change anyway... or work up some hackery in the buildpack to make cargo think it's building a runtime crate that depends on the user's app for its handler.

In light of the latter, and the prior art of the java-based buildpacks expecting a fully-formed app from the user, I thought initially making the buildpack capable of compiling *any* rust app might serve the rust developer better. Heck, there are lots of alternatives to actix out there. You know as well as I do how many web frameworks we can swing a dead cat and hit. 

I'm just guessing. I like magic, but only if I'm sure I'll never want to change it, and at this point, I'm not sure what I may want to change. So I figure we take a conservative approach initially and see if/how folks might use it. We can always introduce a new template later that hides the server config (`App`) once doing that seems obvious in hindsight.

One thing I could do is split the code into two files: `main.rs` and `handler.rs`. And we'll move the handler and its tests into the latter. Would that make things more clear?

templates/rust/http/src/main.rs
jcrossley3: I'm no expert, but rust async is inspired by erlang async. In rust, you have to bring your own runtime to schedule the lightweight processes defined by that async keyword. Actix uses the tokio runtime, which ultimately gets created by that "macro attribute" on line 15: `#[actix_web::main]`.

I'm not familiar with the actix-web internals but I expect at some point, something is invoking `await` on the `Future` returned by the `index` function. I know it says it's returning an `HttpResponse`, but because that `async` keyword is there, it's really returning a `Future`.

Were you looking for more detail than that?

jcrossley3:
> We use GH Actions for CI, and yes you can add cargo https://github.com/actions-rs/cargo.

Where do I add that, exactly?

zroubalik:
> > We use GH Actions for CI, and yes you can add cargo https://github.com/actions-rs/cargo.
> 
> Where do I add that, exactly?

[Here](https://github.com/boson-project/func/tree/main/.github/workflows) you can find the GH Actions definitions, that are being used by CI and for PR validation.

jcrossley3:
> > Where do I add that, exactly?
> 
> [Here](https://github.com/boson-project/func/tree/main/.github/workflows) you can find the GH Actions definitions, that are being used by CI and for PR validation.

Thanks. I saw that, but couldn't figure out where the other langs were being configured and then thought, "maybe rust is available, by default?" So I just updated the `test` target and everything worked. Yay!

templates/rust/http/src/main.rs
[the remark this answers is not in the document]
lance: > I'm just guessing. I like magic, but only if I'm sure I'll never want to change it, and at this point, I'm not sure what I may want to change. So I figure we take a conservative approach initially and see if/how folks might use it. We can always introduce a new template later that hides the server config (`App`) once doing that seems obvious in hindsight.

That seems reasonable, I suppose.


> One thing I could do is split the code into two files: `main.rs` and `handler.rs`. And we'll move the handler and its tests into the latter. Would that make things more clear?

Yes, I like this change. Thanks. :)

lance (approved):

templates/rust/http/src/main.rs
jcrossley3: I think so.

jcrossley3:
Ok, I squashed my commits and I think I'm generally happy with it as an initial strawman.

zroubalik (commented):
Looking good, my Rust experience is limited though... :)

Do you think that you could add Rust Function developer guide? Something similar that has been done for other runtimes: https://github.com/boson-project/func/blob/main/docs/guides/developers_guide.md#language-guides

Thanks!

jcrossley3:
> Do you think that you could add Rust Function developer guide?

@zroubalik Is there something in particular I missed in the `README.md` included in each template? I'm assuming the intended audience already knows rust well enough to have their environment set up and should already be pretty familiar with the `cargo` subcommands.

Is the guide still useful if it's mostly restating what's in the README's?

lance:
@jcrossley3 it's a lot of what's in the readme, but it's also other stuff depending on the runtime/framework. Currently we try to have a language guide for each set of templates. You are probably right that this ultimately probably lands in the template itself - especially as these things start to move upstream.

lkingland (approved):

jcrossley3:
> Do you think that you could add Rust Function developer guide?

Done.

PR #378, opened 2021-06-08
refactor!: trigger is template, function signature implied

lkingland:
The system includes two templates, 'http' and 'event' for each of the runtimes.  These templates illustrate the two currently supported function signatures.

These signatures happen to differ only insofar as their triggering event, but explicitly enumerating the triggering event is not necessary, as this is implied by the function signature written out.

This is the original intention of the innards of the client library:  supported function signatures are documented clearly, templates write an example Function implementation which showcases one of the available options.  The library includes one example of each supported signature for each language.

This PR hopefully illustrates this more clearly than I can in words, as it shows this realization that trigger is ultimately the same concept as template, results in generally deleted lines and other simplifications.

### DIFF
--- a/client.go
+++ b/client.go
@@ -15,7 +15,12 @@ import (
 const (
 	DefaultRegistry = "docker.io"
 	DefaultRuntime  = "node"
-	DefaultTrigger  = "http"
+
+	// DefautlTemplate is the default Function signature / environmental context
+	// of the resultant function.  All runtimes are expected to have at least
+	// one implementation of each supported funciton sinagure.  Currently that
+	// includes an HTTP Handler ("http") and Cloud Events handler ("events")
+	DefaultTemplate = "http"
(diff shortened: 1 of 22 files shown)


lance:
Poo! I was working on this at the same time as you! My duplicate has a few different changes. Let's discuss the merits of those before closing one of these two: https://github.com/boson-project/func/pull/379

lkingland:
Fortunately, the changes from the two PRs look like they are not in opposition; just not entirely equivalent.  Per our quick base-touching on Slack, I am applying both PRs, and will bubble up any questions about individual modifications in Slack.

lkingland:
This PR's changes, along with #379's are now included in PR #381, so the two can probably be closed in favor of working off #381.  My mistake for not mentioning this PR's existence in team chat.

PR #379, opened 2021-06-08
refactor!: change --trigger and --templates flags

lance:
This commit is a breaking change.

Change the `--trigger` flag to be `--template` and the `--templates` flag
to be `--packages`. This is being done in anticipation of future work focused
on making `func` extensibility friendlier, and in an attempt to finalized some
of the naming conventions we have used to date.

In fact, the `--trigger` flag used to be `--template` but we decided to
change that a few months ago. This commit reverses that decision. The reason
behind this is twofold.

1. Using 'trigger' has proved to be confusing. Even if I create a function
with an HTTP trigger, it will still be invoked when a CloudEvent is sent
to the function process. Or alternatively, it is possible to send a raw
HTTP request to a function with an event trigger. Using 'template' instead
implies that the incoming request does not determine how the function is
invoked - rather it is the structure of the function signature that informs
the invocation.

2. The `trigger` terminology is not inclusive enough for our use cases. For
example, a third party provider of function templates may provide a template
for multiplexing incoming HTTP requests in Go using `gorilla-mux`. It doesn't
really make sense to say that `gorilla-mux` is the trigger. It's just a
defining feature of how the template is structured. I think this:

```sh
func create --runtime go --template gorilla-mux
```

Makes more sense than this:

```sh
func create --runtime go --trigger gorilla-mux
```

In changing this flag to be `--template`, we then need to come up with
another name for our existing `--templates` flag. I chose `--packages`
because what is being specified here is more than just the template. The
user sees only the function template when they run `func create...` but
the filesystem from which this template is pulled also contains metadata
about the template - most importantly right now, `.builders.yaml`. It is
conceivable that we may ultimately want to stuff these directories with
event more metadata in the future.

Something like `--packages` makes sense to me, but I am open to suggestion.

Thinking of these as a package also allows for better extensibility features
down the road. For example, users could reference packages at a URI like so.

```
func create --packages https://mycompany.com/function/templates.tgz
```

This would result in `func` downloading the tarball, extracting it to the
config directory, and using it for additional templates.

### DIFF
--- a/client.go
+++ b/client.go
@@ -15,7 +15,7 @@ import (
 const (
 	DefaultRegistry = "docker.io"
 	DefaultRuntime  = "node"
-	DefaultTrigger  = "http"
+	DefaultTemplate = "http"
 )
 
 // Client for managing Function instances.
@@ -29,7 +29,7 @@ type Client struct {
 	lister           Lister   // Lists remote services
 	describer        Describer
 	dnsProvider      DNSProvider      // Provider of DNS services
-	templates        string           // path to extensible templates
+	packages         string           // path to extensible templates
 	registry         string           // default registry for OCI image tags
 	progressListener ProgressListener // progress listener
 	emitter          Emitter          // Emits CloudEvents to functions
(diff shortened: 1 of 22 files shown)


lance:
@lkingland apparently we were both inspired to work on this at the same time! :laughing: Mine is a duplicate of https://github.com/boson-project/func/pull/378/files but also includes changing `--templates` to `--packages`. I would normally close a duplicate PR that hit second, but I will leave this open for now so we can discuss the merits of that additional change.

lkingland:
Looks good to me as well.  Yes, we did step on each other's toes a bit.  I'll try to straighten things out.

lkingland (approved):
LGTM!

zroubalik (approved):
LGTM!

lkingland:
This PR's changes, along with #378's are now included in PR #381, so the two can probably be closed in favor of working off #381.  My mistake for not mentioning the former's existence in team chat.

lkingland:
This PR's changes, along with #378's are now included in PR #381, so I am closing the two in favor of working off #381. My mistake for not mentioning the former's existence in team chat.

PR #381, opened 2021-06-09
refactor!: trigger to template, templates to packages

lkingland:
This PR contains the updates from PR #378 and PR #379.  There were substantive differences between the two, but contained a very high level of overlap, so the process was:  Apply the changes from 378, then apply 379, rebasing and resolving conflicts.

I found nothing during the merge that I think would be contentious, but since they were not identical in their approach, this PR gives us a chance to thumb through the summation of the changes before... committing (pun intended).

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -16,8 +16,8 @@ func init() {
 	root.AddCommand(createCmd)
 	createCmd.Flags().BoolP("confirm", "c", false, "Prompt to confirm all configuration options (Env: $FUNC_CONFIRM)")
 	createCmd.Flags().StringP("runtime", "l", bosonFunc.DefaultRuntime, "Function runtime language/framework. Available runtimes: "+utils.RuntimeList()+" (Env: $FUNC_RUNTIME)")
-	createCmd.Flags().StringP("templates", "", filepath.Join(configPath(), "templates"), "Path to additional templates (Env: $FUNC_TEMPLATES)")
-	createCmd.Flags().StringP("template", "t", bosonFunc.DefaultTemplate, "Function template. For eample 'http' or 'events' (Env: $FUNC_TEMPLATE)")
+	createCmd.Flags().StringP("packages", "p", filepath.Join(configPath(), "packages"), "Path to additional template packages (Env: $FUNC_PACKAGES)")
+	createCmd.Flags().StringP("template", "t", bosonFunc.DefaultTemplate, "Function template. Available templates: 'http' and 'events' (Env: $FUNC_TEMPLATE)")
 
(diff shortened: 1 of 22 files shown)


docs/guides/func_yaml.md
```diff
@@ -87,12 +87,6 @@ The Kubernetes namespace where your function will be deployed.
 
 The language runtime for your function. For example `python`.
 
-### `trigger`
```
lance, line 90: This should be replaced with an entry for `template`.
```
### `template`
The source code template tailored for the invocation event that triggers
your function. Possible values are `http` for plain HTTP requests, and `event`
CloudEvent triggered functions.
```

lance (approved):
lgtm with one minor doc concern. thanks for taking care of ths

cmd/create.go
```diff
@@ -16,8 +16,8 @@ func init() {
 	root.AddCommand(createCmd)
 	createCmd.Flags().BoolP("confirm", "c", false, "Prompt to confirm all configuration options (Env: $FUNC_CONFIRM)")
 	createCmd.Flags().StringP("runtime", "l", bosonFunc.DefaultRuntime, "Function runtime language/framework. Available runtimes: "+utils.RuntimeList()+" (Env: $FUNC_RUNTIME)")
-	createCmd.Flags().StringP("templates", "", filepath.Join(configPath(), "templates"), "Path to additional templates (Env: $FUNC_TEMPLATES)")
-	createCmd.Flags().StringP("trigger", "t", bosonFunc.DefaultTrigger, "Function trigger. Available triggers: 'http' and 'events' (Env: $FUNC_TRIGGER)")
+	createCmd.Flags().StringP("packages", "p", filepath.Join(configPath(), "packages"), "Path to additional template packages (Env: $FUNC_PACKAGES)")
```
zroubalik, line 19: Minor concern, `-p` is being used for path with the other commands, this might get a little bit confusing from UX. But not sure if it is really an issue or just my nit picking 🤷‍♂️

zroubalik (approved):
Looking good, one minor q on `-p` parameter.

cmd/create.go
lkingland: We could bike-shed a bit and rename the whole parameter 'repositories'... Or I can just change it.

I think for now I'll just change it to 'a' for "Additional Packages".  That seems reasonable?

docs/guides/func_yaml.md
lkingland: Thanks very much for pointing out the obvious omission; a casualty of merge it appears.

cmd/create.go
lance: @lkingland I think it's maybe worth bike shedding a bit. See https://github.com/boson-project/func/pull/375#discussion_r648669324

PR #385, opened 2021-06-11
fix: password read on windows

matejvasek:

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -233,7 +233,7 @@ func getUserName(ctx context.Context) (string, error) {
 		ch <- struct {
 			u string
 			e error
-		}{u: strings.TrimRight(username, "\n"), e: nil}
+		}{u: strings.TrimRight(username, "\r\n"), e: nil}
 	}()
 
 	select {
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -201,7 +201,7 @@ func getPassword(ctx context.Context) ([]byte, error) {
 	})
 
 	go func() {
-		pass, err := term.ReadPassword(int(syscall.Stdin))
+		pass, err := term.ReadPassword(int(syscall.Stdin)) // nolint: unconvert
 		ch <- struct {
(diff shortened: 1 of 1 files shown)


matejvasek:
Beware of the above code; I didn't tested it.

cmd/deploy.go
```diff
@@ -200,7 +201,7 @@ func getPassword(ctx context.Context) ([]byte, error) {
 	})
 
 	go func() {
-		pass, err := term.ReadPassword(0)
+		pass, err := term.ReadPassword(int(syscall.Stdin)) // nolint: unconvert
```
lkingland, line 204: Good to know!

lkingland (approved):

matejvasek:
@lkingland I don't even know how to write integration test for this.

First idea might be `echo  "mvasek\npasswordhere\n" | f deploy --build=false -v`. But that won't work.
It does causes `read` syscall to get expected input. However for password you also need to invoke `ioctl` syscall to disable character echoing. result is `inappropriate ioctl for device`. :disappointed:

matejvasek:
@ramo

> @lkingland I don't even know how to write integration test for this.
> 
> First idea might be `echo "mvasek\npasswordhere\n" | f deploy --build=false -v`. But that won't work.
> It does causes `read` syscall to get expected input. However for password you also need to invoke `ioctl` syscall to disable character echoing. result is `inappropriate ioctl for device`. 
> 
> EDIT:
> it can be done by `expect -c 'spawn -noecho func deploy --build=false -v; send "mvasek\npwdhere\n"; interact'`

@jrangelramos it might be good to test this in our e2e tests if possible.

lance (approved):

matejvasek:
@boson-project/core I added just tiny amendment: trimming of `\r` character.

jrangelramos:
Yea. @matejvasek That's a good point to test and ensure it won't break on future.  I think using `expect` as you mentioned could work as well

PR #386, opened 2021-06-15
feat: allow setting resource requests/limits

zroubalik:

Follow up on https://github.com/boson-project/func/pull/374

Users can specify resource requests/limits (cpu&memory) and concurrency limit, which are applied on deployed KService.

```yaml
name: test
namespace: ""
runtime: go
...
options:
  scale:
    min: 2
    ...
  resources:
    requests:
      cpu: 100m
      memory: 128Mi
    limits:
      cpu: 1000m
      memory: 256Mi
      concurrency: 100
```

Fixes: #383

### DIFF
--- a/config.go
+++ b/config.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 // ConfigFile is the name of the config's serialized form.
@@ -29,7 +30,8 @@ type Env struct {
 }
 
 type Options struct {
-	Scale *ScaleOptions `yaml:"scale,omitempty"`
+	Scale     *ScaleOptions     `yaml:"scale,omitempty"`
+	Resources *ResourcesOptions `yaml:"resources,omitempty"`
 }
 
 type ScaleOptions struct {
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -317,5 +335,56 @@ func validateOptions(options Options) (errors []string) {
 		}
 	}
 
+	// options.resource
+	if options.Resources != nil {
```
lkingland, line 339: As you are probably well aware, this giant if statement is a bit unwieldy.
As a stylistic thing, keeping the happy path as far left as possible is favorable.  This can be accomplished, sometimes, by failing fast when preconditions on a logic branch are not met.  For example, this _might_ help in this situation:
```
if options.Resources == nil {
  return
}
...
```
And extracting logic into separate functions to make that return statement possible.

And as an additional note, the use of pointers may not even be necessary on the options struct.  That would just avoid the problem of having to check for nil constantly entirely.  Go relies heavily on zero values precisely to avoid nil pointers and the requisite defensive programming they engender.

Ah, the price we pay to get a nicely-nested yaml :/

config.go
```diff
@@ -29,7 +30,8 @@ type Env struct {
 }
 
 type Options struct {
-	Scale *ScaleOptions `yaml:"scale,omitempty"`
+	Scale     *ScaleOptions     `yaml:"scale,omitempty"`
+	Resources *ResourcesOptions `yaml:"resources,omitempty"`
```
lkingland, line 34: I don't think these should be pointers.  The zero value of the struct should probably be sufficient?

lkingland (approved):
Pretty happy with all this.  I'd remove the pointers if possible, because dealing with that many potentially nil pointers is going to be a bit of a chore.  Hooray for zero values.

config.go
zroubalik: Yeah, but I need to make these pointers, so we can distinguish when the value was actually set in yaml. If we rely on default values (for example for `int` fields), then I am not able to find out whether the user wanted to set the value of the field to `0` or if it is just a default value.

And the similar situation with structs, if I am not mistaken, if those are not pointers, then those fields are set in the yaml file as part of serialization, even if they don't have any values set. I prefer to omit them if user doesn't want to set the specific field, just to make the yaml file more pretty :)

lance: I think this is a fair concern about the pointers. These fields should not be used with default values, unless we are specific about what those values are (and that they match the existing knative defaults - which seems error prone itself).

zroubalik: @lance so is this PR okay from your POV?

lkingland: Ah, yes I understand.  I mistakenly thought this was the final canonical set of settings.  We have to be able to distinguish between "unspecified" (use the default) and "specifically the zero value".

This is why I have never understood the position held by some that "nil is never correct".

PR #389, opened 2021-06-15
feat: add a URL output type for `func describe`

### LINKED ISSUE #387: Need a programmatic way to get deployed function's URL
Knowing that a deployed function is actually a knative service, I can get its deployed URL using either `kn` or `kubectl`, but I'd prefer to use the same tool I deployed it with: `func`.

I'd like to be able to do something along these lines:

```shell
func deploy
curl $(func describe -o url)
```
I got the `url` output format idea from `kn service describe ...`

jcrossley3:
Fixes #387

### DIFF
--- a/cmd/describe.go
+++ b/cmd/describe.go
@@ -18,7 +18,7 @@ import (
 func init() {
 	root.AddCommand(describeCmd)
 	describeCmd.Flags().StringP("namespace", "n", "", "Namespace of the function. By default, the namespace in func.yaml is used or the actual active namespace if not set in the configuration. (Env: $FUNC_NAMESPACE)")
-	describeCmd.Flags().StringP("output", "o", "human", "Output format (human|plain|json|xml|yaml) (Env: $FUNC_OUTPUT)")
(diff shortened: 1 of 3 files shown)


cmd/describe.go
```diff
@@ -161,3 +161,10 @@ func (d description) XML(w io.Writer) error {
 func (d description) YAML(w io.Writer) error {
 	return yaml.NewEncoder(w).Encode(d)
 }
+
+func (d description) URL(w io.Writer) error {
+	if len(d.Routes) > 0 {
+		fmt.Fprint(w, d.Routes[0])
```
zroubalik, line 167:
```suggestion
		fmt.Fprintf(w, "%s\n", d.Routes[0])
```

zroubalik (commented):
Looking good! Just a minor nit in output.

cmd/describe.go
jcrossley3: I was thinking the linefeed might mess up `curl $(func describe -o url)` but it doesn't seem to. Thanks!

zroubalik (approved):
LGTM

lance (approved):
TY!

Issue #362, opened 2021-05-26
Allow function developers to access Kubernetes Secrets


zroubalik:
Function developers should be able to specify secrets that needs to be mounted or added as ENV variable to the deployed function.
They can do that via editing `func.yaml` file or via `func secret` command, which should interactively suggest secrets available in the namespace on the cluster and then allow users to attach them to the function.

**func.yaml example:**
```yaml

name: my-func
namespace: "test"
runtime: node
...
volumes: 
   - secret: example-secret                              # mount secret as a Volume
     path: /etc/secret-volume
   - secret: example-secret2
     path: /etc/secret-volume2
   - configMap: example-cm                               # possibility to mount a ConfigMap
     path: /etc/cm-volume
envs:
   - name: EXAMPLE1                                      # ENV directly from a value 
     value: value1
   - name: EXAMPLE2                                      # ENV from the local ENV var 
     value: {{ env.MY_ENV }}
   - name: EXAMPLE3                                      # ENV from a key in secret
     value: {{ secret.example-secret.key1 }}      
   - value: {{ secret.example-secret1 }}                 # all ENVs from secret
```

embano1:
Thank you for the details @zroubalik. Two questions:

1) Can you provide an example with namespaces?
1) It looks like Example 3 above is incorrect (multiple value fields, incl. inconsistent use of array/non-array semantics)?

zroubalik:
> Thank you for the details @zroubalik. Two questions:
> 
>     1. Can you provide an example with namespaces?

Do you mean referencing a Secret from a different namespace? That's not possible currently and honestly I don't think this is a good practice wrt security. 

>     2. It looks like Example 3 above is incorrect (multiple value fields, incl. inconsistent use of array/non-array semantics)?
No it is correct, but those are 2 separate examples:
This sets `key1` from the secret as a env value `EXAMPLE3`:
```yaml
envs:
   - name: EXAMPLE3
     value: {{ secret.example-secret.key1 }}        
```

This sets all key=value pairs from the secret as env variables:
```yaml
envs:
   - value: {{ secret.example-secret1 }}         
```

embano1:
> Do you mean referencing a Secret from a different namespace? That's not possible currently and honestly I don't think this is a good practice wrt security. 

Ah sorry for the confusion here as I read the initial comment that this should be possible. I just reread and noticed my wrong assumption. I also agree that secrets should be assumed to live in the same namespace. All good then.

[...]

PR #391, opened 2021-06-16
feat: `func config volumes` - interactive prompt

zroubalik:

Interactive prompt for managing function configuration - this PR adds Volumes (Envs will be in a separate PR).

`func config` - starts interactive prompt, wizard that let's you select resource and the requested operation

`func config volumes` - list configured Volumes

`func config volumes add` - starts interactive prompt, wizard that let's you add Volume mount to the configuration. It does a lookup for existing Secrets and ConfigMaps in the cluster and namespace.

`func config volumes remove` - starts interactive prompt, wizard that let's you remove Volume mount from the configuration.

Relates: #362

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -1,0 +1,124 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/AlecAivazis/survey/v2/terminal"
+	"github.com/ory/viper"
+	"github.com/spf13/cobra"
+
+	bosonFunc "github.com/boson-project/func"
+)
+
+func init() {
+	root.AddCommand(configCmd)
+	configCmd.Flags().StringP("path", "p", cwd(), "Path to the project directory (Env: $FUNC_PATH)")
+}
+
+var configCmd = &cobra.Command{
+	Use:   "config",
+	Short: "Configure a function",
+	Long: `Configure a function
+
+Interactive propmt that allows configuration of Volume mounts and Environment variables for a function
+project present in the current directory or from the directory specified with --path.
+`,
+	SuggestFor: []string{"cfg", "cofnig"},
+	PreRunE:    bindEnv("path"),
+	RunE:       runConfigCmd,
+}
+
+func runConfigCmd(cmd *cobra.Command, args []string) (err error) {
+
+	function, err := initConfigCommand(args)
+	if err != nil {
+		return
+	}
+
+	var qs = []*survey.Question{
+		{
+			Name: "selectedConfig",
+			Prompt: &survey.Select{
+				Message: "What do you want to configure?",
+				Options: []string{"Environment values", "Volumes"},
+				Default: "Environment values",
+			},
+		},
+		{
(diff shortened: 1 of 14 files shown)


zroubalik:
https://user-images.githubusercontent.com/726523/122253424-71090980-cecc-11eb-8ebd-48958528225c.mp4

lkingland (approved):
Nice.  Looking forward to using this flow elsewhere.  Like Lance said, this is a strong example of how func doesn't fit entirely into a plugin box.

cmd/config.go
```diff
@@ -0,0 +1,124 @@
...
+// CLI Configuration (parameters)
+// ------------------------------
+
+type configCmdConfig struct {
+	Name    string
+	Path    string
+	Verbose bool
+}
+
+func newConfigCmdConfig(args []string) configCmdConfig {
```
lance, line 98: Dangerously close to `AbstractConfigServletInitializerConfig` :rofl:

cmd/config_volumes.go
```diff
@@ -0,0 +1,247 @@
...
+		return
+	},
+}
+
+var configVolumesAddCmd = &cobra.Command{
+	Use:   "add",
+	Short: "Add volume to the function configuration",
+	Long: `Add volume to the function configuration
+
+Interactive prompt that allows adding Secrets and ConfigMaps as Volume mounts to the function project
```
lance, line 52: Nit
```suggestion
Interactive prompt to add Secrets and ConfigMaps as Volume mounts to the function project
```

cmd/config_volumes.go
```diff
@@ -0,0 +1,247 @@
...
+		return runAddVolumesPrompt(cmd.Context(), function)
+	},
+}
+
+var configVolumesRemoveCmd = &cobra.Command{
+	Use:   "remove",
+	Short: "Remove volume from the function configuration",
+	Long: `Remove volume from the function configuration
+
+	Interactive prompt that allows removing Volume mounts from the configuration of the function project
```
lance, line 73:
```suggestion
	Interactive prompt to remove Volume mounts from the function project
```

`add` long description uses "as Volume mounts to the function project". But this says "removing Volume mounts from the configuration of the function project" which appears to have a different meaning and it's unclear if `add` would affect the cluster immediately or not. When we say "function project" we should always mean the configuration, so in this case I think using the words 'from the configuration" is probably redundant and could be confusing.

lance (approved):
I like it!

lance:
> Nice. Looking forward to using this flow elsewhere. Like Lance said, this is a strong example of how func doesn't fit entirely into a plugin box.

Yes! I would like to update https://github.com/boson-project/func/pull/373 to use this flow for labels.

cmd/config.go
zroubalik: Haha right 😅 I just wanted to follow the existing patterns for naming...

cmd/config_volumes.go
zroubalik: Much better, thx

PR #393, opened 2021-06-16
fix: use credsstore

matejvasek:

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -187,6 +187,12 @@ func credentialsProvider(ctx context.Context, registry string) (docker.Credentia
 		return result, nil
 	}
 
+	credentials, _ = docker.GetCredentialsFromCredsStore(registry)
+	if credentials != (containersTypes.DockerAuthConfig{}) {
+		result.Username, result.Password = credentials.Username, credentials.Password
+		return result, nil
+	}
+
 	fmt.Println("Please provide credentials for image registry.")
 	var qs = []*survey.Question{
 		{
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -203,7 +203,7 @@ func credentialsProvider(ctx context.Context, registry string) (docker.Credentia
 			Validate: survey.Required,
 		},
 		{
-			Name: "namespace",
+			Name: "password",
 			Prompt: &survey.Password{
 				Message: "Password:",
 			},
--- a/docker/credentials_helper.go
+++ b/docker/credentials_helper.go
@@ -1,0 +1,82 @@
+package docker
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/containers/image/v5/types"
+	"github.com/docker/docker-credential-helpers/client"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func GetCredentialsFromCredsStore(registry string) (types.DockerAuthConfig, error) {
+	result := types.DockerAuthConfig{}
+
+	dirname, err := os.UserHomeDir()
+	if err != nil {
+		return result, fmt.Errorf("failed to determine home directory: %w", err)
+	}
+
+	confFilePath := filepath.Join(dirname, ".docker/config.json")
+
+	f, err := os.Open(confFilePath)
+	if err != nil {
+		return result, fmt.Errorf("failed to open docker config file: %w", err)
+	}
+	defer f.Close()
+
+	conf := struct {
+		Store string `json:"credsStore"`
+	}{}
+
+	decoder := json.NewDecoder(f)
+
+	err = decoder.Decode(&conf)
+	if err != nil {
+		return result, fmt.Errorf("failed to deserialize docker config file: %w", err)
+	}
+
+	if conf.Store == "" {
+		return result, fmt.Errorf("no store configured")
+	}
+
+	helperName := fmt.Sprintf("docker-credential-%s", conf.Store)
+	p := client.NewShellProgramFunc(helperName)
+
+	credentialsMap, err := client.List(p)
+	if err != nil {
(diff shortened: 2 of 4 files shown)


cmd/deploy.go
```diff
@@ -194,6 +204,69 @@ func credentialsProvider(ctx context.Context, registry string) (docker.Credentia
...
+	}{}
+
+	decoder = json.NewDecoder(stdOut)
+	err = decoder.Decode(&outCreds)
+	if err != nil {
+		return creds, err
+	}
+	creds.Username, creds.Password = outCreds.Username, outCreds.Secret
+
+	fmt.Println(outCreds)
```
matejvasek, line 265: I should delete this.

matejvasek:
I tested this with gnome key-ring and on windows. It worked for me.
@lkingland @zroubalik @rhuss could you please test this on macOS?

zroubalik (commented):
This is still asking for username/password everytime I do `func deploy` :(

`.docker/config.json`
```json
{
	"auths": {
		"https://index.docker.io/v1/": {},
		"quay.io": {}
	},
	"credsStore": "desktop",
	"experimental": "disabled",
	"stackOrchestrator": "swarm"
}%
```

cmd/deploy.go
```diff
@@ -194,6 +204,69 @@ func credentialsProvider(ctx context.Context, registry string) (docker.Credentia
 	return result, nil
 }
 
+func getCredentialsFromCredsStore(registry string) (containersTypes.DockerAuthConfig, error) {
```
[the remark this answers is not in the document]
lkingland: One caviat being:  it prompts the user assuming stdin.  That's a bit of crossed wires.  I'd say separate out getting creds from credsstore, vs getting creds from the user interactively as separete functions.  One as a library, one here in the CLI.  But like Zbynek said... it's just a nitpick at this point.

lkingland:
Unfortunately this also does not work for me.  And while I have not had a chance to dig in and debug, one hint is that for me, the credentials are stored in the registry index `index.docker.io` rather than simply `docker.io` as I would have expected:
```
echo "index.docker.io" | docker-credential-desktop get
{"ServerURL":"index.docker.io","Username":"REDACTED","Secret":"REDACTED_BUT_POPULATED!"}
```

matejvasek:
@boson-project/core PTAL

zroubalik (commented):
Looking good, but as I mentioned before, could you please move docker/credentials related stuff out of `cmd/deploy.go`? Probably to `docker` package?

matejvasek:
For posterity:
I don't know why, but credentials for `docker.io` are not necessarily stored under key `docker.io`.
Sometimes they are stored as `https://index.docker.io/v1/` or `index.docker.io`.
To handle this I relaxed equivalence criterion on key (serverURL) in a way that two URL are considered equivalent if their 2nd level domain is the same (e.g. `https://index.docker.io/v1/` is equivalent to `docker.io`).

This can hypothetically cause issues: it is possible that `sub.example.com` is pointing to different repo with different credentials than `example.com`, also `example.com:1234` may point to different repo than `example.com:4321`.

Some observations:
It seems that pushes to `docker.io` are handled as a special case.
Push to `quay.io` results in following HTTP call to the docker deamon.
`POST /v1.41/images/quay.io/xvasek/lifecycle/push?tag=0.10.2 HTTP/1.1`,
then you might expect that push to `docker.io` would result into:
`POST /v1.41/images/docker.io/xvasek/lifecycle/push?tag=0.10.2 HTTP/1.1`
but that is NOT the case. Actual request looks like:
`POST /v1.41/images/xvasek/lifecycle/push?tag=0.10.2 HTTP/1.1`,
note the missing `docker.io` in the request URL.

Additionally in case of push to `docker.io` the `POST` above is preceded by another request
`GET /v1.41/info HTTP/1.1` the response contains JSON with member `"IndexServerAddress":"https://index.docker.io/v1/"`.

Also the `docker` CLI doesn't call `list` sub-comand of `docker-credential-*`. It calls only `get` sub-command.
So the `docker` CLI knows exactly what key should be used, it doesn't do some domain equivalence hack.

I just guess that proper way to handle this would be that in case of push to `docker.io` we would call the above `GET` and then used `IndexServerAddress` as credential key (serverURL). But that just a guess.

matejvasek:
@zroubalik PTAL

matejvasek:
https://stackoverflow.com/questions/49455033/docker-registry-name-resolution

docker/credentials_helper.go
```diff
@@ -0,0 +1,84 @@
...
+	p := client.NewShellProgramFunc(helperName)
+
+	credentialsMap, err := client.List(p)
+	if err != nil {
+		return result, fmt.Errorf("failed to list credentials: %w", err)
+	}
+
+	for serverUrl, username := range credentialsMap {
+		// TODO maybe use `username` if there are multiple credentials per url
+		_ = username
```
zroubalik, line 55: If I am not mistaken, you can do this, if you don't need `username`:
```suggestion
	for serverUrl := range credentialsMap {

```

zroubalik (commented):
Looking good, minor nit in `for range`

docker/credentials_helper.go
zroubalik: Could you please elaborate on the TODO there^ ?

matejvasek: I am not sure of it but I think that user can be logged under multiple logins.

In such a case we could decide based on username included in image name. But for now we don't need it. Also the namespace part of the image name/tag doesn't need to be  login, it could be name of organization in which case it's more complicated.

matejvasek:
@zroubalik PTAL, I rebased and + did small fix for prompt.

zroubalik (approved):
LGTM

PR #394, opened 2021-06-17
src: envs: use `:` instead of `.` as separators

zroubalik:

The syntax used for referencing secrets/configmap in `envs` section is not ideal: https://github.com/boson-project/func/blob/main/docs/guides/func_yaml.md#envs

eg. `value: '{{ secret.mysecret.key }}'` - but the secret name and key can contain `.` characters. Replace that with `:` , ie. `value: '{{ secret:mysecret:key }}'` as this should be safe.

Relates: #362


matejvasek:
@zroubalik I still bound some dot usages:
```
~> grep -RiP "\{\{\s*env\."                                                                                                                                  ST 15   envsFix 
docs/guides/func_yaml.md:  value: '{{ env.API_KEY }}'
knative/deployer.go://     value: {{ env.MY_ENV }}
docker/runner.go:// run command supports only ENV values in from FOO=bar or FOO={{ env.LOCAL_VALUE }}
docker/runner.go:// processEnvValue returns only value for ENV variable, that is defined in form FOO=bar or FOO={{ env.LOCAL_VALUE }}
```

zroubalik:
> @zroubalik I still found some dot usages:
> 
> ```
> ~> grep -RiP "\{\{\s*env\."                                                                                                                                  ST 15   envsFix 
> docs/guides/func_yaml.md:  value: '{{ env.API_KEY }}'
> knative/deployer.go://     value: {{ env.MY_ENV }}
> docker/runner.go:// run command supports only ENV values in from FOO=bar or FOO={{ env.LOCAL_VALUE }}
> docker/runner.go:// processEnvValue returns only value for ENV variable, that is defined in form FOO=bar or FOO={{ env.LOCAL_VALUE }}
> ```

Good catch! Thx. Fixed

lance (approved):
lgtm

PR #396, opened 2021-06-18
feat: `func config envs` - interactive prompt

zroubalik:

Interactive prompt for managing function configuration - this PR adds Environment variables

`func config envs` - list configured Environment variables

`func config envs add`- starts interactive prompt, wizard that let's you add Environment variable to the configuration. It does a lookup for existing Secrets and ConfigMaps in the cluster and namespace.

`func config envs remove`

 - starts interactive prompt, wizard that let's you remove Environment variable from the configuration.

Relates: #362

https://user-images.githubusercontent.com/726523/122539154-90687980-d027-11eb-97a8-0d794897ee0a.mp4

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -72,14 +72,20 @@ func runConfigCmd(cmd *cobra.Command, args []string) (err error) {
 	case "Add":
 		if answers.SelectedConfig == "Volumes" {
 			err = runAddVolumesPrompt(cmd.Context(), function)
+		} else if answers.SelectedConfig == "Environment values" {
+			err = runAddEnvsPrompt(cmd.Context(), function)
 		}
 	case "Remove":
 		if answers.SelectedConfig == "Volumes" {
 			err = runRemoveVolumesPrompt(function)
+		} else if answers.SelectedConfig == "Environment values" {
+			err = runRemoveEnvsPrompt(function)
 		}
 	case "List":
 		if answers.SelectedConfig == "Volumes" {
 			listVolumes(function)
+		} else if answers.SelectedConfig == "Environment values" {
+			listEnvs(function)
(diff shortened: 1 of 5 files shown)


lkingland (approved):
Looks good to me!
One thing I did not mention earlier, but may bear at least highlighting, is that with the addition of these interactive commands which hit the kubernetes API, our CLI is now coupled with the kubernetes API (rather than having that clean layer of func  in between).  But I am conflicted if this is really an existential problem, as it is coupling in order to ease the creation _of_ that clean intermediary layer.

Just worth mentioning I suppose, because it provides a good example of challenging my desire to maintin the "purity" of the CLI, but for a clearly user-assisting reason.

zroubalik:
> Looks good to me!
> One thing I did not mention earlier, but may bear at least highlighting, is that with the addition of these interactive commands which hit the kubernetes API, our CLI is now coupled with the kubernetes API (rather than having that clean layer of func in between). But I am conflicted if this is really an existential problem, as it is coupling in order to ease the creation _of_ that clean intermediary layer.
> 
> Just worth mentioning I suppose, because it provides a good example of challenging my desire to maintin the "purity" of the CLI, but for a clearly user-assisting reason.

You are right, I had the same concern, but I think that this is inevitable given the fact that we are going to part of Knative 🤷‍♂️

lance:
> You are right, I had the same concern, but I think that this is inevitable given the fact that we are going to part of Knative 

True - but it would be nice ultimately to have something like `grid` that proxies these API calls to Kube/Knative or some other platform. But for now going directly against the API is fine I suppose.

lance (approved):
I like it - sorry about the prompt impl Luke!

PR #397, opened 2021-06-18
src: refactor prompt to use `AlecAivazis/survey`

zroubalik:

I have migrated all prompts that used our `prompt` implementation to https://github.com/AlecAivazis/survey which is being used for the interactive `func config` command. So we have similar UX.

@lkingland I feel really bad about deleting the `prompt` package that you have written, I hope it is ok :)


cmd/create.go
```diff
@@ -131,30 +139,62 @@ func newCreateConfig(args []string) createConfig {
...
+				derivedName, _ := deriveNameAndAbsolutePathFromPath(val.(string))
+				return utils.ValidateFunctionName(derivedName)
+			},
+		},
+		{
+			Name: "runtime",
+			Prompt: &survey.Input{
+				Message: "Runtime:",
+				Default: c.Runtime,
+				// TODO add runtime suggestions: https://github.com/AlecAivazis/survey#suggestion-options
```
zroubalik, line 169: This framework has quite nice suggestions options: https://github.com/AlecAivazis/survey#suggestion-options

So we can then suggest available runtimes and templates in the future.

lkingland:
> Signed-off-by: Zbynek Roubalik [zroubali@redhat.com](mailto:zroubali@redhat.com)
> 
> I have migrated all prompts that used our `prompt` implementation to https://github.com/AlecAivazis/survey which is being used for the interactive `func config` command. So we have similar UX.
> 
> @lkingland I feel really bad about deleting the `prompt` package that you have written, I hope it is ok :)

No problem at all.  I always prefer to do a quick implementation early on, upgrading to a third-party dependency _only when it offers a compelling upgrade_.  I find that too often folks either 1) never try and end up with tons of dependencies they dont need or 2) try to do everything and end up being able to support nothing.  As usual; the Middle Way.

lkingland (approved):
LGTM

"one of my most productive days was throwing away 1000 lines"...

cmd/create.go
lance: Off topic, but I would like someday to be able to suggest runtimes and templates from remote repositories.

lance (approved):

PR #400, opened 2021-06-21
src: test temp dir helper standardization

lkingland:
Small PR updating the client unit tests to use the same test directory helper as the others.


lance (approved):
Nice refactoring

PR #401, opened 2021-06-21
src: rename bosonFunc fn

lkingland:
When we renamed `faas` to `func`, we lost the ability to simply import the package, as the name now conflicts with the builtin `func`.  The solution at the time was the somewhat verbose, though correct, `bosonFunc`.  At some point, I believe @lance used simply `fn` in a test somewhere.  I much prefer this shorter incarnation, and this PR proposes adopting `fn` as the import name throughout.


matejvasek (approved):

matejvasek:
This is only refactor with no effect on functionality, righ? @lkingland

lance (approved):
Touches a lot of files, but all just name changes. LGTM

lance:
> This is only refactor with no effect on functionality, righ? @lkingland

Yes - we discussed this in the call today. Just renaming.

lkingland:
> This is only refactor with no effect on functionality, righ? @lkingland 

That is correct.

lance:
I was worried this would happen with so many files touched. @lkingland can you resolve the conflicts?

lkingland:
Absolutely, no problem. I expected to take care of the merge once approved.

zroubalik (approved):

PR #402, opened 2021-06-21
chore: bump to buildpacks v0.8.3 for all versions

lance:
This is causing me to rethink using versions in these templates, and our
overall buildpack version/release strategy. But for now, we should land
this before 0.16.0.


go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.1.3
 	gopkg.in/yaml.v2 v2.4.0
+	gotest.tools/gotestsum v1.6.4 // indirect
```
matejvasek, line 22: we probably don't need this

dsimansk: Reruning `go mod tidy` should remove it. It was added through `go get` in Makefile right?

client_int_test.go
```diff
@@ -153,6 +153,8 @@ func TestRemove(t *testing.T) {
 // newClient creates an instance of the func client whose concrete impls
 // match those created by the kn func plugin CLI.
 func newClient(verbose bool) *boson.Client {
+	// TODO: Forcing verbose to false in order to pass integration tests
+	verbose = false
```
lkingland, line 157: This seems like an odd thing to have cause a failure.  Please let's chat after we're over the hump with this release; I'd like to fix the root cause

lkingland (approved):
LGTM!  Glad to see the E2Es are working now too.

client_int_test.go
matejvasek: It's indeed strange. I was able to run the test successfully on my machine with with and without verbose output.

lkingland: I was able to run it successfully only if running the test in isolation.  When in tandem with others: failure.

It looks like it has to do with the pack logger being flaky when passed a stdout-backed logger (such as in verbose mode) @matejvasek

lkingland: Furthermore, when working around that issu, it also appears to be having trouble when using the builtin logger as well.  it errors failing to connect to stderr.  @matejvasek

lkingland: I am not sure if this is a pack logger issue or what (thought it seems so), but it appears we can get around the issue by always using a bytes.Buffer-backed logger, and only printing when in verbose mode (or embedding it in any errors).  Refactoring a bit to clear things out, and submitting an update. @matejvasek

matejvasek: >I was able to run it successfully only if running the test in isolation. When in tandem with others: failure.

I run the failing test in isolation.

PR #405, opened 2021-06-23
Testable Commands

lkingland:
Refactors the create command to be testable, and adds a test which utilizes this structure.

Commands (cobra.Command instances) are created with a reference to a fn.Client creator (factory-ish pattern).  This reference is used in a closure for the RunE delegate.  In this way, the effective instance of fn.Client can be swapped out for testing.

If this looks acceptable to folks, I'll use the pattern to test the remaining CLI-centric features of the Extensible Templates FT.

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -15,26 +15,34 @@ import (
 )
 
 func init() {
-	root.AddCommand(createCmd)
-	createCmd.Flags().BoolP("confirm", "c", false, "Prompt to confirm all configuration options (Env: $FUNC_CONFIRM)")
-	createCmd.Flags().StringP("runtime", "l", fn.DefaultRuntime, "Function runtime language/framework. Available runtimes: "+buildpacks.Runtimes()+" (Env: $FUNC_RUNTIME)")
-	createCmd.Flags().StringP("repositories", "r", filepath.Join(configPath(), "repositories"), "Path to extended template repositories (Env: $FUNC_REPOSITORIES)")
-	createCmd.Flags().StringP("template", "t", fn.DefaultTemplate, "Function template. Available templates: 'http' and 'events' (Env: $FUNC_TEMPLATE)")
-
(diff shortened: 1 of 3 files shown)


cmd/create.go
```diff
@@ -15,26 +15,34 @@ import (
...
+	cmd := &cobra.Command{
+		Use:   "create [PATH]",
+		Short: "Create a function project",
+		Long: `Create a function project
 
 Creates a new function project in PATH, or in the current directory if no PATH is given. 
 The name of the project is determined by the directory name the project is created in.
 `,
-	Example: `
+		Example: `
```
lance, line 45: I wonder if your editor is auto formatting this perhaps?

cmd/create.go
```diff
@@ -46,23 +54,42 @@ kn func create --runtime quarkus myfunc
 
 # Create a function project that uses a CloudEvent based function signature
 kn func create --template events myfunc
-`,
-	SuggestFor: []string{"inti", "new"},
-	PreRunE:    bindEnv("runtime", "template", "repositories", "confirm"),
-	RunE:       runCreate,
-	// TODO: autocomplate or interactive prompt for runtime and template.
+	`,
+		SuggestFor: []string{"inti", "new"},
```
lance, line 58:
```suggestion
		SuggestFor: []string{"vreate", "creaet", "craete", "new"},
```
I think this is leftover from when the command was called `init`. Some better misspelling suggestions.

lance (approved):
lgtm - a couple of small suggestions

cmd/create.go
lkingland: Yes, this is `gofmt` fixing indentation to reflect that the struct is now embedded in a function.  Looks strange because of the string literals above, but it should be correct.

cmd/create.go
lkingland: Good catch @lance , I just glossed right over that.  Definitely a holdover from the switch from init to create

zroubalik (approved):

PR #406, opened 2021-06-23
fix: pack logger failing on verbose

lkingland:
The Pack logger was failing silently when passed a logger whose
output was set to stdout.  This required hard coding verbose to
false.  This works around the issue by always buffering output,
and only writing it to stdout when verbose loggin.


lkingland:
No longer needed

PR #409, opened 2021-06-23
fix: closing stdout

matejvasek:
If you pass `io.WriteCloser` as loggin output for `pack` it will close it, `os.Stdout` does implement `io.WriteCloser` so it ended up closed which is not really good. I added wrapping struct that implements only `io.Writer` to fix the issue.


matejvasek:
Note that this means that in `func v0.16.0` the `--verbose` flag doesn't work correctly.

zroubalik (approved):

lkingland (approved):

matejvasek:
https://github.com/buildpacks/pack/issues/1214

PR #412, opened 2021-06-25
Define handler in terms of CloudEvents rather than http/payload

jcrossley3:
Simplify the handler and its tests by using the latest release of the cloudevents-sdk, since they can be written in terms of Events rather than all the HTTP plumbing abstractions.


lance (approved):
As a Rust naif, I'll have to trust that this is correct and idiomatic - as I'm sure it is. This _looks_ much nicer anyway.

```rs
pub async fn handle(event: Event) -> Result<Event, actix_web::Error> {
```

zroubalik (approved):

PR #416, opened 2021-07-03
ci: merge local and CI integration scripts

lkingland:
CI is failing.  Nothing in our code changed.  It must be a change in the environment.

Going to hopefully take the opportunity to dogfood the localhost cluster setup and configuration scripts in CI.


lkingland:
finally!

lkingland:
This PR includes a move to using the same scripts in CI that are used to locally test.

Prior, we were doing _conceptually_ the same thing, but with two essentially competing implementations.

This allows us to have one implementation to test (good thing too, as the combination of GitHub Actions runners + Knative is apparently more brittle than I had hoped).  It also lays the groundwork for pulling these features out into a library/tool which we can use for CI/CD tooling in Function project repositories.

Note:  we'll definitely want to "squash and merge" this one, as it took quite a few debugging commits to get pounded into shape!

lkingland:
My best guess as to the root cause of the original failure was the startup of the "webhook" service too early in the process.   I also believe the system in general is too dependent on arbitrary wait times, and is generally brittle.   I hope that we can upgrade to using the knative/kubernetes client library to actually sample the active state of various services, rather than relying on these essentially arbitrary wait times.

lkingland:
Flaky brittle and a bit janky IMHO.  We need to replace these scripts with something more robust.  Such as a proper binary which loads the Knative/Kubernetes API to perform these operations.  I am open to suggestions.  But for now it is working again.

lance:
@lkingland I landed https://github.com/boson-project/func/pull/415 so this will probably need a rebase

lkingland:
This is superseded by the terse #419

PR #419, opened 2021-07-12
ci: move to inbuilt provider scripts for local and ci clusters

lkingland:
Fixes CI failures resulting from an unknown change in (presumably) the GitHub Actions operating environment (best guess: resource constriction).

As a byproduct of the fix, the cluster setup and configuration scripts have been restructured to be usable both for local testing as well as in CI.**  A few additional updates include:

-  Upgrades to Knative v0.23.0
-  Upgrades Eventing from v1beta1 API to v1
-  Adds binary installation script (linux) binaries.sh
-  Serializes Build, Int Test, e2e test steps to conserve resources and improve stability
-  Installs Broker and Channel to support, for example, Knative Quickstart tutorial
-  Runs a test echo server during installation to verify installation
-  Adds event-based waits in place of most blind time-based waits
-  Adds explicit Kubernetes version for underlying Kind cluster
-  Replaces a NodePort patched Kourier with an explicit ingress
-  Adds retries to particularly brittle areas such as activator webhook initial availability and echo service creation
-  Combines configure.sh into allocate.sh

Note that the underlying cluster allocated is showing signs of fragility and resource exhaustion.  It is my assumption that the GitHub Actions environment has _barely_ the resources necessary to achieve Knative Serving+Eventing operability.  I would prefer that we update these minimal scripts to a more robust Go-based solution, with retries and verification throughout, prior to offering this functionality in a reusable GitHub Action for Functions developers.

** To try this locally, assuming `kubectl`, `kind` and `yq` are installed already, local cluster setup can be accomplished with `hack/allocate.sh` .  The CI environment also runs `binaries.sh` to install the prerequisites and `ci.sh` to fully automate the process on the expected linux based system, and can be used for further reference.  Once started, `func` will use the local cluster for deployment, along with a locally running container registry.  All resources are released with `hack/delete.sh`.


hack/test.sh
```diff
@@ -0,0 +1,52 @@
...
+  apiVersion: serving.knative.dev/v1
+  kind: Service
+  metadata:
+    name: echo
+    namespace: func
+  spec:
+    template:
+      spec:
+        containers:
+          - image: docker.io/jmalloc/echo-server
```
matejvasek, line 32: Not sure if pull limit would be issue here. Maybe it would be better to mirror this in our quay? @jrangelramos knows more about the limitations of docker.io

hack/allocate.sh
```diff
@@ -22,33 +22,35 @@ set -o pipefail
...
 main() {
 
-  local serving_version=v0.22.0
-  local eventing_version=v0.22.0
-  local kourier_version=v0.22.0
+  local kubernetes_version=v1.21.1
+  local knative_version=v0.23.0
+  local kourier_version=v0.23.0
 
   local em=$(tput bold)$(tput setaf 2)
```
matejvasek, line 29: this doesn't seem to work in CI runner:
`tput: No value for $TERM and no -T specified`

hack/allocate.sh
```diff
@@ -61,41 +65,117 @@ containerdConfigPatches:
...
-  echo "${em}③ Knative Eventing${me}"
-  kubectl apply --filename https://github.com/knative/eventing/releases/download/$eventing_version/eventing-crds.yaml
-  sleep 2
-  curl -L -s https://github.com/knative/eventing/releases/download/$eventing_version/eventing-core.yaml | yq 'del(.spec.template.spec.containers[]?.resources)' -y | yq 'del(.metadata.annotations."knative.dev/example-checksum")' -y | kubectl apply -f -
-  curl -L -s https://github.com/knative/eventing/releases/download/$eventing_version/in-memory-channel.yaml | yq 'del(.spec.template.spec.containers[]?.resources)' -y | yq 'del(.metadata.annotations."knative.dev/example-checksum")' -y | kubectl apply -f -
-  curl -L -s https://github.com/knative/eventing/releases/download/$eventing_version/mt-channel-broker.yaml | yq 'del(.spec.template.spec.containers[]?.resources)' -y | yq 'del(.metadata.annotations."knative.dev/example-checksum")' -y | kubectl apply -f -
+dns() {
+  echo "${em}③ DNS${me}"
+  n=0
+  until [ $n -ge 10 ]; do
```
lance, line 93: Why are you looping here instead of looking for some kind of ready state? What happens when you try 10 times and they all fail. It doesn't seem like there is an explicit failure mode here.

hack/test.sh
```diff
@@ -0,0 +1,54 @@
...
+EOF
+    echo "Retrying..."
+    sleep 5
+  done
```
lance, line 38: As with the loop above, there is no error condition expressed if the loop executes 10 times unsuccessfully. I imagine the `kubectl wait...` below will ultimately time out, but perhaps we can fail fast here?

hack/allocate.sh
lkingland: Thanks, defaulted TERM to 'dumb' for CI which has none set.

hack/allocate.sh
lkingland: Thanks @lance and I share your general unease.  This is precisely what I was referring to in our call when I mentioned general flakiness.  This is one of the commands which I struggled with because it appears to have no condition upon which can be waited. The webhook process, which is supposedly ready (see wait from earlier step) fails with an IO timeout in low resource scenarios, but is considered ready.  Therefore, it seems there is simply a racing condition here and I have no choice but to simply try a few times and, as you say, let it fall through to failing on a later step if unavailable.

I agree this is ugly when compared with an explicit condition wait, and I am open to suggestions on how to improve it.  However, as you mention, if it fails 10 times (I never saw it go beyond 4), it will fail on the next kubectl call immediately.

hack/test.sh
lkingland: @lance You are correct that a fall-through will yield an error on the next call.

However, I agree that an explicit error when exceeding the loop retries is correct, for both loop situations, and I will add that.

An oversight I would not have made if not in bash! :)

hack/allocate.sh
matejvasek: Wouldn't `ansi` work too? github action do support ansi escape codes.

hack/test.sh
lkingland: This has been fixed.  Thanks @lance

hack/allocate.sh
lkingland: The two retry loops now fail-fast if `n` is reached.  Thanks @lance

zroubalik (approved):

hack/test.sh
lkingland: Good point.  I will keep an eye on this as a potential failure mode, and fork if/ when we hit that limit, or better yet replace this test entirely with a Function which does the same thing.

lkingland:
> New changes are detected. LGTM label has been removed.

Pushed to force re-run of checks as slash-command automation is not yet enabled.

lkingland:
The `cla/google` bot is now recognizing the repo (and that CLAs have been signed).  However the `tide` Prow bot is still pending setup.  I _presume_ we could merge but I refrain lest I be too.... presumptuous.

PR #420, opened 2021-07-13
fix: do not trust builder when using podman

matejvasek:


matejvasek:
@jrangelramos this also should fix the issues with podman but in a different way than https://github.com/containers/podman/pull/10905, this should disable trust to builder, just until fix is in official podman release.

lance (approved):

matejvasek:
@lance shall I merge this or wait for CI fix?

lance:
@matejvasek unit tests pass. I think it's okay

PR #425, opened 2021-07-15
Import hack, copy testing scripts from knative/client

evankanderson:
The important requirement is to have a `./test/presubmit-tests.sh` for Prow (see https://github.com/knative/test-infra/pull/2840). I did this by importing the current `knative/client` minimal-ish set of files, which includes `go test` and verifying cross-compilation.

/assign @lance @rhuss
/hold

Happy to re-arrange this some other way; I'd like both lance and roland to review. :grin:


rhuss:
Thanks !

Oops, that's quite a mouthful of a PR (and my browser has a hard time opening the files view :) 

The build script and CI integration are kind of unique how kn is built. It can definitely be reused for other projects and some plugins do. But we don't want to enforce that and as `func` comes already with its own way of building, I think we should integrate the Prow CI hook-scripts into the existing build system.

I'm also not clear whether there should be an interplay with the Github Actions based CI and Prow CI, and whether we should switch one of them or use both. Using the Prow CI has the advantage that it is integrated with the Prow workflow (`/ok-to-test`, `/retest` commands), Github actions have better support from the GitHub UI.

We should decide which road to go, but as mentioned I can't review this PR in a finite amount of time (because of my very sluggish Chrome that can't show 5000+ files in one view)

lkingland:
PR #419 contains a few updates to the existing CI of func, and is awaiting the CLA label but is otherwise ready to merge. It uses GitHub Actions so is a bit more lithe at a total of 450 lines.  Is that func-specific set of scripts usable, or could it be with some modification?

lance:
I don't understand what this PR is meant to do, and at 5000+ files I don't really know how to review it. As I mentioned in https://github.com/knative/test-infra/pull/2840#discussion_r670562725 we have done a lot of work to date to get our CI working, and I am concerned about dropping that simply to satisfy prow/tide. Rather than landing this, I think it would be great if we can identify what the goals are first, and then find a (hopefully less massive) way to achieve them.

evankanderson:
/close

We'll use the existing actions-based tooling for now.

PR #428, opened 2021-07-20
src: Makefile developer accessibility

lkingland:
# Changes

:broom: Restructures Makefile for better readability.
:gift: Adds `make help` with summary of primary targets
:gift: Lowers initial developer setup  for core development by creating a separate target for running language template unit tests
:wastebasket: removes unused targets

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,111 +1,150 @@
-REPO := quay.io/boson/func
-BIN  := func
+# ##
+#
+# Run 'make help' for a summary
+#
+# ##
 
-PKGER?=pkger
+# Binaries
+BIN         := func
+BIN_DARWIN  ?= $(BIN)_darwin_amd64
+BIN_LINUX   ?= $(BIN)_linux_amd64
+BIN_WINDOWS ?= $(BIN)_windows_amd64.exe
 
-DARWIN=$(BIN)_darwin_amd64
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,111 +1,150 @@
...
-DATE := $(shell date -u +"%Y%m%dT%H%M%SZ")
-HASH := $(shell git rev-parse --short HEAD 2>/dev/null)
-VTAG := $(shell git tag --points-at HEAD)
-# a VERS environment variable takes precedence over git tags
-# and is necessary with release-please-action which tags asynchronously
-# unless explicitly, synchronously tagging as is done in ci.yaml
-VERS ?= $(shell [ -z $(VTAG) ] && echo 'tip' || echo $(VTAG) )
-
-TEMPLATE_DIRS=$(shell find templates -type d)
-TEMPLATE_FILES=$(shell find templates -type f -name '*')
```
lkingland, line 20: @matejvasek I removed these conditions because the target was always running (I think because we remove caches before each run), and I was getting:
`make: *** No rule to make target `templates/python/http/__pycache__', needed by `pkged.go'.  Stop.git` when running after unit testing of templates was performed.  Please let me know if I am missing something imprtant.

zroubalik (approved):

PR #430, opened 2021-07-22
feat: Configure Rust functions

dejanb:
# Changes

- :gift: Add ability to add all custom function configuration in a separate module.

/kind enhancement

Usually for our functions we need to provide some configured resources, like database connections and such. We can configure that in the `main.rs` at the moment, but it would be much cleaner to provide a separate _configuration module_ where all such code lives.

How that would look in practice could be seen on this practical example: [Configuration](https://github.com/dejanb/drogue-influxdb-function/blob/main/src/config.rs#L16) and [Function](https://github.com/dejanb/drogue-influxdb-function/blob/main/src/handler.rs#L14)

**Release Note**

```release-note
* Add ability to configure Rust functions in a separate module.
```


dejanb:
@googlebot I signed it!

templates/rust/events/README.md
```diff
@@ -5,7 +5,9 @@ Welcome to your new Rust function project! The boilerplate
 [`src/main.rs`](./src/main.rs). It's configured to invoke the `handle`
 function in [`src/handler.rs`](./src/handler.rs) in response to a POST
 request containing a valid `CloudEvent`. You should put your desired
-behavior inside that `handle` function.
+behavior inside that `handle` function. In case you need to configure
+some resources for your function, you can do that in `configure` function
+in `src/config.rs`](./src/config.rs).
```
jcrossley3, line 10:
```suggestion
some resources for your function, you can do that in the [`configure` function](./src/config.rs).
```

templates/rust/events/src/config.rs
```diff
@@ -0,0 +1,26 @@
+use actix_web::web;
+
+/// Run custom configuration as part of the application building
+/// process.
+///
+/// This function should contain all custom configuration for your function application.
+///
+/// ```rust
+/// fn config(cfg: &mut web::ServiceConfig) {
```
jcrossley3, line 9:
```suggestion
/// fn configure(cfg: &mut web::ServiceConfig) {
```

templates/rust/http/README.md
```diff
@@ -5,7 +5,9 @@ Welcome to your new Rust function project! The boilerplate
 [`src/main.rs`](./src/main.rs). It's configured to invoke the `index`
 function in [`src/handler.rs`](./src/handler.rs) in response to both
 GET and POST requests. You should put your desired behavior inside
-that `index` function.
+that `index` function. In case you need to configure
+some resources for your function, you can do that in `configure` function
+in `src/config.rs`](./src/config.rs).
```
jcrossley3, line 10:
```suggestion
some resources for your function, you can do that in the [`configure` function](./src/config.rs).
```

templates/rust/http/src/config.rs
```diff
@@ -0,0 +1,26 @@
+use actix_web::web;
+
+/// Run custom configuration as part of the application building
+/// process.
+///
+/// This function should contain all custom configuration for your function application.
+///
+/// ```rust
+/// fn config(cfg: &mut web::ServiceConfig) {
```
jcrossley3, line 9:
```suggestion
/// fn configure(cfg: &mut web::ServiceConfig) {
```

templates/rust/http/src/config.rs
```diff
@@ -0,0 +1,26 @@
...
+/// ```rust
+/// pub async fn index(
+///     req: HttpRequest,
+///     driver: web::Data<DbDriver>,
+/// ) -> HttpResponse {
+///     HttpResponse::NoContent()
+/// }
+pub fn configure(_cfg: &mut web::ServiceConfig) {
+    log::info!("Configuring service");
+}
```
jcrossley3, line 26: Same as above

```suggestion
pub fn configure(cfg: &mut web::ServiceConfig) {
    log::info!("Configuring service from {:?}", cfg.config());
}
```

templates/rust/events/src/config.rs
```diff
@@ -0,0 +1,26 @@
...
+/// ```rust
+/// pub async fn handle(
+///     event: Event,
+///     driver: web::Data<DbDriver>,
+/// ) -> Result<Event, actix_web::Error> {
+///     Ok(Event::default())
+/// }
+pub fn configure(_cfg: &mut web::ServiceConfig) {
+    log::info!("Configuring service");
+}
```
jcrossley3, line 26: Might be nice to avoid the `_` prefix and use the arg in the body. I don't think `ServiceConfig` implements `Debug` but `AppConfig` should...

```suggestion
pub fn configure(cfg: &mut web::ServiceConfig) {
    log::info!("Configuring service from {:?}", cfg.config());
}
```

jcrossley3 (changes requested):
I dig it! :smile: 

Only minor nits/typos

jcrossley3:
Also, I could be wrong about this. (@lance @zroubalik @lkingland please correct me, if so)

You have to run `make` in order to generate the `pkged.go` file and include it in the PR. That's what gets your templates into the `func` binary, iirc.

lance:
> Also, I could be wrong about this. (@lance @zroubalik @lkingland please correct me, if so)
> 
> You have to run `make` in order to generate the `pkged.go` file and include it in the PR. That's what gets your templates into the `func` binary, iirc.

You are correct! If only this PR had landed a few days ago: https://github.com/knative-sandbox/kn-plugin-func/pull/432 ;)

dejanb:
@googlebot I fixed it.

templates/rust/events/src/config.rs
dejanb: I couldn't find a good example for this, service config doesn't expose any info that could be printed (`config()` function doesn't exists). Maybe adding 'health' route, but to me that should belong to the main. This

dejanb:
@lance @jcrossley3 Thanks for the pointers. I now incorporated most of the Jim's suggestions, ran make and squashed everything into a single commit. 

@jcrossley3 I think googlebot might be now complaining about your CLA as I added you as co-author. Can you check?

Let me know if you see anything else could be improved.

lkingland:
Regard ing `pkged.go`:  yes, any updates to the embedded templates requires that they be re-encoded into pkged.go in order for them to be included in the client library (and CLI).  An unfortunate, but seemingly unavoidable, step.

Thanks for the PR.  I am not familiar with Rust idioms, so minutiae such as underscores prefixing parameters is not something I can comment on.  However, conceptually the addition of an optional static `configure` method seems like a perfectly acceptable way to preconfigure the static state of a package prior to it being loaded into memory for execution.  If I understand correctly it's like a hook into a static initializer?  I would prefer that eventually we encourage the use of instanced functions with constructors, when those signatures are available, but this seems like a good way to get a large part of that benefit while still relying on static functions.  This might be a pattern worth exploring in the other language packs.

/lgtm

templates/rust/events/src/config.rs
jcrossley3: Hmm... I might've been looking at v4 of actix. Is there a `to_string()` or something?

dejanb: Maybe we can introduce a really simple configuration for our example function, like

https://github.com/dejanb/kn-plugin-func/commit/95294bf3fe0d53b9746d2ee78600d5ed2ae19a3f

I think that might be a good way to get people get started more easily.

If you think it's a good approach, I can follow up with something like this in a separate PR.

jcrossley3:
> @jcrossley3 I think googlebot might be now complaining about your CLA as I added you as co-author. Can you check?

Looks like they associate my CLA with jcrossley3@gmail.com. If that doesn't work, just remove it. :)

templates/rust/events/src/config.rs
jcrossley3: works for me. the simpler, the better!

jcrossley3:
@googlebot I consent. jeez, you're annoying!

jcrossley3 (approved):
/lgtm

PR #431, opened 2021-07-22
test: confirm config search path

lkingland:
# Changes

The default location of a user's Functions config location is $XDG_CONFIG_HOME/func.  This PR ensures that this is respected by adding a test which verifies the correct functioning of its only current usage:  the location of local non-core (extensible) template repositories.

:broom:  Cleans up current configPath derivation function
:broom:  Adds unit test confirming default value and respect of XDG_CONFIG_HOME

### DIFF
--- a/cmd/create_test.go
+++ b/cmd/create_test.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


lance (approved):
/lgtm

PR #433, opened 2021-07-23
Bump node dependencies

julz:
Runs `npm audit fix` in all directories with a `package-lock.json` (there were vulnerabilities in the normalize-url, glob-parent, ini, and axios deps that dependabot was complaining about).


lance (approved):
/lgtm 

Thanks!

PR #434, opened 2021-07-23
Bump cloudevents dep to 4.0.3

julz:
subj. This matches the version in the other templates, and is needed to get the latest, non-vulnerable version of the axios dependency.


lance (approved):
/lgtm

PR #436, opened 2021-07-26
feat: Add proper example of configuring Rust functions.

dejanb:
# Changes

-:gift: As discussed in #430, here's the concrete example on how to use function configuration.

/kind feature

It uses a simple configuration struct to provide users with easy to adapt example on how to provide configuration to their functions.


lance (commented):
Looks good to me. Though again, I'm not a rustafarian. Would love for @jcrossley3 to have a look.

templates/rust/events/src/handler.rs
```diff
@@ -1,16 +1,21 @@
...
+    config: web::Data<HandlerConfig>,
+) -> Result<Event, actix_web::Error> {
     info!("event: {}", event);
 
     let input = match event.data() {
         Some(Data::Binary(v)) => from_slice(v)?,
         Some(Data::String(v)) => from_str(v)?,
         Some(Data::Json(v)) => v.to_owned(),
-        None => json!({ "name": "world" }),
+        None => json!({ "name": config.name }),
```
jcrossley3, line 18: Very nice!

templates/rust/events/src/handler.rs
```diff
@@ -29,7 +34,7 @@ mod tests {
     async fn valid_input() {
         let mut input = Event::default();
         input.set_data("application/json", json!({"name": "bootsy"}));
-        let resp = handle(input).await;
+        let resp = handle(input, web::Data::new(HandlerConfig::default())).await;
```
jcrossley3, line 37: I really wish rust supported optional arguments. :)

Can we put this in a static constant local to the test module and call it `config` or something? Just so it won't look so gross?

templates/rust/events/src/handler.rs
```diff
@@ -39,7 +44,7 @@ mod tests {
 
     #[actix_rt::test]
     async fn no_input() {
-        let resp = handle(Event::default()).await;
+        let resp = handle(Event::default(), web::Data::new(HandlerConfig::default())).await;
```
jcrossley3, line 47: ibid

templates/rust/http/src/handler.rs
```diff
@@ -1,13 +1,14 @@
...
 // Implement your function's logic here
-pub async fn index(req: HttpRequest) -> HttpResponse {
+pub async fn index(req: HttpRequest, config: web::Data<HandlerConfig>) -> HttpResponse {
     info!("{:#?}", req);
     if req.method() == Method::GET {
-        HttpResponse::Ok().body("Hello!\n")
+        HttpResponse::Ok().body(format!("Hello {}!\n", config.name))
     } else {
-        HttpResponse::Ok().body("Thanks!\n")
+        HttpResponse::Ok().body(format!("Thanks {}!\n", config.name))
```
jcrossley3, line 11: This kinda makes me want to add a `name` argument, similar to the events example, in which the `config.name` is only used if the `name` isn't passed.

templates/rust/http/src/handler.rs
```diff
@@ -19,16 +20,24 @@ mod tests {
...
+            resp.body().as_ref().unwrap()
+        );
     }
 
     #[actix_rt::test]
     async fn post() {
         let req = test::TestRequest::post().to_http_request();
-        let resp = index(req).await;
+        let config = HandlerConfig::default();
+        let resp = index(req, web::Data::new(config.clone())).await;
```
jcrossley3, line 36: That `.clone()` call is making me question my wanting a static above. :)

jcrossley3 (commented):
Cool idea using config to un-hardcode the "world" string. Just a few questions...

templates/rust/http/src/handler.rs
dejanb: I'm all for improving this example and looked into it. However, I didn't find an easy/simple way to add it. We'd need to use extractors for both query and body (depending whether get or post is used) and possibly another struct. Maybe we should split the two handlers? We would also need to add more tests.
I would definitely evolve this, but maybe we should do it in a separate PR?

templates/rust/events/src/handler.rs
dejanb: I ended up adding `config()` function to tests that wraps this. This is the simplest thing I could find as making static traits complicates them for the example and I would like to keep those as simple and reusable as possible. Let me know how it looks.

templates/rust/http/src/handler.rs
jcrossley3: Yeah, I was thinking there was a `FromRequest` impl for `String` but that returns the whole body. I'm not sure introducing a `Query` param (or whatever) is worth it. Probably out of scope for an example. My bad!

templates/rust/events/src/handler.rs
jcrossley3: Much better, thanks!

lance:
Thanks for helping out with the review @jcrossley3 

/lgtm

lance:
@dejanb pkged.go is such a pita. Could you please rebase this PR?

dejanb:
> @dejanb pkged.go is such a pita. Could you please rebase this PR?

@lance sure, just did it. Tests look good now. Let me know if there's anything else needed.

PR #437, opened 2021-07-26
feat: remote template repositories

lkingland:
Adds support for remote Git template repositories.

Accessible using the `--repository` flag on the CLI, and `WithRepository` client Option.

Specifying this option takes precidence over both extensible (on-disk) template repositories and embedded.


lance (commented):
Overall looks great! I think we should probably doc this in docs/commands.md

cmd/create.go
```diff
@@ -63,7 +64,8 @@ kn func create --template events myfunc
 
 	cmd.Flags().BoolP("confirm", "c", false, "Prompt to confirm all configuration options (Env: $FUNC_CONFIRM)")
 	cmd.Flags().StringP("runtime", "l", fn.DefaultRuntime, "Function runtime language/framework. Available runtimes: "+buildpacks.Runtimes()+" (Env: $FUNC_RUNTIME)")
```
lance, line 66: This gets a bit tricky doesn't it? The languages available are part of the directory tree when `-r` or `-g` are used. For example, in [this example](https://github.com/lance/gcf-kn) repository the only runtime available is `ruby`. Is a user expected to know this? I'm not sure that we should print available runtimes by default given that extensible and remote language packs are probably going to become much more widely used with the move upstream.

This makes me wonder if we should be thinking a little bigger than a new flag. Just brainstorming here, so...

Should we consider that support for remote repositories also includes a new mechanism for managing the default repository used on all commands? For example:

```
❯ kn func config --global
? What do you want to configure?  [Use arrows to move, type to filter]
> Language Pack Repositories
```
Which would allow you to `List`, `Add` and `Remove`, and `Set Default`. `List` shows `Builtin`, and then lists top level directories in `$HOME/.config/func/repositories`. `Add` accepts a path to a local directory which is copied (symlinked?) into `$HOME/.config/func/repositories`, or it accepts a Git repository URI, and downloads/clones that repository into the repositories directory. `Remove` deletes the repository from the configuration directory, and `Set Default` writes to `$HOME/.config/func/config.yaml`.

I know this would  make the PR much bigger :fearful: 

But it might be a nice experience for both the user and vendors. And it would allow us to display the available runtimes for `-l` dynamically based on what is the user default. Initially those are the ones available builtin.  It also eliminates the need for the `-r` flag - or changes it to accept the name of any top level directory in `$HOME/.config/func/repositories`.

Sorry! This little detail had been bothering me, and I didn't realize it was going to such a long comment...

WDYT?

lance: I meant to point out something before I got carried away. If vendors bundle their language packs like this https://github.com/lance/gcf-kn where the buildpacks, stacks, etc. are also part of the repository, that makes listing all of the top level directories for the `-l` flag as I suggested above, a little problematic. Should we also propose in the documentation for this, that language pack providers/vendors should publish the repository with a specific structure? For example, we could say that a certain directory name is reserved for these kinds of files - i.e. `builder`. The `func` CLI will ignore this directory when listing available languages for a repository, so vendors could use it to contain buildpacks and other related artifacts.

cmd/create.go
```diff
@@ -63,7 +64,8 @@ kn func create --template events myfunc
...
-	cmd.Flags().StringP("repositories", "r", filepath.Join(configPath(), "repositories"), "Path to extended template repositories (Env: $FUNC_REPOSITORIES)")
+	cmd.Flags().StringP("repositories", "r", filepath.Join(configPath(), "repositories"), "Path to extended template repositories, or URL to a single git repository containing templates (Env: $FUNC_REPOSITORIES)")
+	cmd.Flags().StringP("repository", "g", "", "URI to a Git repository from which template can be pulled (causes builtin and extensible repositories to be ignored) (Env: $FUNC_REPOSITORY)")
```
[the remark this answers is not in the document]
lkingland: > Nit

I love nits.  Keep 'em comin

cmd/create.go
lkingland: Overall, I completely agree.  I do not, however, think the suggestions in any way conflict with the usefulness of this explicit "use this remote" flag, but rather build on it.  Perhaps the documentation will make that more clear, or prove me wrong when I try to express it therein.

Right now, it is a clear shortcoming that there is no feedback to the user which languages and templates are available within a repository (including the builtin), and thus we are dependent on good documentation.  This should be on our short list to improve.  Should this be part of this PR or a new one?

Additionally, creating a ligtweight repository management system (with requisite CLI tree where you can CRUD repo references), is indeed the general direction I was hoping we would go eventually!  That's why I went "local disk first".  What's the priority of this?  Should it even be part of this FT or a new one?

Makefile
```diff
@@ -30,7 +30,6 @@ PKGER           ?= pkger
 CODE := $(shell find . -name '*.go')
 
 all: build
-	# Run 'make help' for make target documentation.
```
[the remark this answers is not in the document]
lkingland: Let me know if there are any other things irking you like this that I overlooked elsewhere!

docs/guides/language-packs.md
```diff
@@ -0,0 +1,7 @@
+# Language Packs
+
+A Language Pack is a term referring to A Builder/Buildpack, the function signatures it recognizes, and the temlates which illustrate them.
```
lkingland, line 3:
```suggestion
A Language Pack is a term referring to a Builder(Buildpack), the function signatures it recognizes, and the templates which illustrate them when creating a new Function project.
```

docs/guides/templates.md
```diff
@@ -0,0 +1,41 @@
...
+For example, the Boson Project Templates repository contains an example "Hello World" Function implementation in each of the officially supported languages.  To use this template via the CLI, use the flags:
+
+func create <name> --template hello-world --repository https://github.com/boson-project/templates
+
+## Locally Installing Repositories
+
+Template Repositories can also be installed locally by placing them in the Functions configuration directory.  
+
+To install the Boson Project templates locally, for example, clone the repository as `~/.config/func/repositories/boson`
+
```
lkingland, line 22: Here's where we are proposing to expand with a lightweight repo management system .

cmd/create.go
lkingland: For future reference,  a little thread housekeeping:

The relevant section from [the Feature Track Specification](https://hackmd.io/@j8e4sRzyQ2O1Gc5V3z7YBQ/BkIXfXC6_#Remote-Repositories-backlog)  to this PR states:

> Remote Repositories
> 
> In addition to built-in and on-disk repositories, func should also support referring to a template in a remote repository. When specifying a template in a remote repository, the location of the remote is a URL provided for the repository location, which is expected to be an available Git repository. For example, to use a TypeScript template provided by Microsoft in a GitHub repository named knative-fn a user will type:
> 
> `func create -l typescript -r https://github.com/microsoft/knative-fn `

This is the feature added by this PR.

The brainstorm above is mostly regarding the previous section in the FT which talks about local repository installation:

> Installing a Repository Locally
> 
> Installing a repository simply means the creation of a conformant file structure. To install a future Boson Project extended templates repository could be accomplished as follows:
> 
> `git clone https://github.com/boson-project/templates ~/.config/func/repositories/boson`
> 

This was completed earlier, and thus the majority of the discussion is about enhancement requests outside the scope of the original FT.    These enhancements are now being tracked in https://github.com/orgs/knative-sandbox/projects/4 and will appear in future PRs.

lance:
I've added the following issues to address comments in this PR. @lkingland if I have missed any, shout out!

- https://github.com/knative-sandbox/kn-plugin-func/issues/451
- https://github.com/knative-sandbox/kn-plugin-func/issues/452
- https://github.com/knative-sandbox/kn-plugin-func/issues/453
- https://github.com/knative-sandbox/kn-plugin-func/issues/454

docs/guides/language-packs.md
```diff
@@ -0,0 +1,7 @@
+# Language Packs
+
+A Language Pack is a term referring to a Builder(Buildpack), the function signatures it recognizes, and the templates which illustrate them when creating a new Function project.
```
lance, line 3:
```suggestion
A Language Pack is the mechanism by which the Functions binary can be extended to support additional runtimes, function signatures, even operating systems and installed tooling for a function. A Language Pack includes
- a .builders.yaml file containing a reference to a builder OCI image reference, which conforms to the buildpack builder specification, and contains references to buildpacks supporting this Language Pack
- one or more template directories illustrating the Language Pack's recognized function signatures
- tests and documentation for the templates
```

docs/guides/language-packs.md
```diff
@@ -0,0 +1,7 @@
...
+
+Built in to the Functions library is a basic language pack for each supported language.
+
+For an example external language pack, see [https://github.com/lance/gcf-kn]
```
lance, line 7:
```suggestion
For an example external language pack, see [https://github.com/lance/gcf-kn/tree/main/ruby]
```
That directory is the language pack. The top level is the "language pack repository". At least that's how I've been thinking of it. If you do too, let's be consistent where we can. The terminology can get confusing!

docs/guides/templates.md
```diff
@@ -0,0 +1,41 @@
...
+When a Function is created, an example implementation and a Function metadata file are written into the new Function's working directory.  Together, these files are referred to as the Function's Template.  Included are the templates 'http' and 'events' for each supported language runtime.
+
+These embedded templates are minimal by design.  The Function contains a minimum of external dependencies, and the 'func.yaml' defines a final environment within which the Funciton will execute that is devoid of any extraneous packages or services.
+
+To make use of more complex inital Function implementions, or to define runtime environments with arbitrarily complex requirements, the templates system is fully pluggable.
+
+## External Git Repositories
+
+When creating a new Function, a Git repository can be specified as the source for the template files.  For example, the Boson Project maintains a set of example Functions at https://github.com/boson-project/templates which can be used during project creation.
```
lance, line 11: Do we? I'd prefer to state this as forward looking, or remove it altogether.

docs/guides/templates.md
```diff
@@ -0,0 +1,41 @@
...
+
+For example, the Boson Project Templates repository contains an example "Hello World" Function implementation in each of the officially supported languages.  To use this template via the CLI, use the flags:
```
lance, line 13: Can we just document a hypothetical instead? I'd prefer not to have an example that appears to be broken because we don't have the repo in place.

docs/guides/templates.md
```diff
@@ -0,0 +1,41 @@
...
+
+func create <name> --template hello-world --repository https://github.com/boson-project/templates
```
lance, line 15: Wouldn't this need the language specified as well? E.g.
```suggestion
func create <name> -l ruby --template hello-world --repository https://github.com/boson-project/templates
```
And it might be helpful to show what the repository's directory structure looks like.
```
├── README.md
└── ruby
    └── hello-world
        ├── func.rb
        └── Gemfile
```

docs/guides/templates.md
lance: Maybe show the actual git command?
```
git clone https://github.com/boson-project/templates ~/.config/func/repositories/boson
```

docs/guides/templates.md
```diff
@@ -0,0 +1,41 @@
...
+
+Once installed, the Boson Hello World template can be specified:
+
+func create <name> --template boson/hello-world
+
+## Language Packs
+
+In addition to example implementations, a template includes a `func.yaml` which includes metadata about the Function.  By default this is populated with things like the new Function's name.  It also includes a reference to the specific tooling which compiles and packages the Function into its deployable form.  This is called the Builder.  By customizing this metadata, it is more than just a template; it is referred to as a Language Pack. See [Project Configuration with func.yaml](func_yaml.md).
+
+A Language Pack can support additional function signatures and can fully cusstomize the environment of the final running Function.  For more information see the [Language Pack Guide](language-packs.md).
```
lance, line 31:
```suggestion
A Language Pack can support additional function signatures and can fully customize the environment of the final running Function.  For more information see the [Language Pack Guide](language-packs.md).
```

docs/guides/language-packs.md
lkingland: Ah, I see.  Thanks for taking the time to look over these (admittedly terse) docs; my understanding of the terminology we're adopting is still a WIP.

docs/guides/templates.md
lkingland: I just created that repository, with a "hello-world" template for each language as a placeholder.  Realistically it's going to be some time before there's anything actually there, but this paragraph does check out.

Would you prefer we omit this until such time as there's at least something _in_ the specified repo?   Pro: it illustrates the point.  Con: the illustration is an invisible sculpture.

docs/guides/templates.md
lkingland: The repo is no indeed in place, so it will work.  I also expect to be using it as soon/  However, a hypothetical may still be better than a template which, when used, says "coming soon"?

docs/guides/templates.md
lkingland: I left out the language because it has a default, and soon with `config --global` a user will be able to set a global default for their primary language.  Furthermore, there is a placeholder function in each language, so it won't matter which they choose.  Do you think this looks like an error or unintentional omission?  If so, I will commit your suggestion.

lkingland: The reason I left out any text about the internal structure of a repository is beacuse this document is from the perspective of the user of the repository, who may be better served by focusing just on the command structure.  A forthcoming document specifically targed at creating custom/external repositories would have this information in detail, however.  Do you still think this section would be improved by adding the directory structure?  Or perhaps should I add a link here: "Read more about custom repositories in [Creating a Repository](custom_repositories.md)"  and include in that document a full explanation?

lkingland:
> one or more commits were authored or co-authored by someone other than the pull request submitte

Dear Google-bot.  I rebased on `main` such that only myself and Lance exist in this PR.  Seeing as the commits I added from Lance were _suggested changes_, doesn't that imply consent?   @lance you may need to `@googlebot I consent.` until we find the "I" in "AI".

docs/guides/templates.md
lance: I think leaving it in is fine for now if it checks out. But it will probably change whenever we figure out what the real upstream situation is.

docs/guides/templates.md
lance: OK as I mentioned above - with the repo in place, all good

docs/guides/templates.md
lance: >  I left out the language because it has a default

What is the default language for remote repositories? And how does the user know what that default is?

> The reason I left out any text about the internal structure of a repository is because this document is from the perspective of the user of the repository, who may be better served by focusing on how to choose a template from a repository.

OK I see your point - it does make sense from that perspective to keep the commands simple. But maybe mentioning that there is a default, and what that default is would also be useful information?

> Or perhaps I should go all the way and create that document now, and add a link here such as: "Read more about how repositories are structured in [Creating a Repository](custom_repositories.md)"?

This would be great. From the upstream/task force perspective, I think it's important to really highlight how easy it is for vendors/providers to create, publish and support these repositories. So a document around that would really help this PR, in my opinion.

lance:
@googlebot I consent

docs/guides/templates.md
lkingland: You bet.  Working on it now and will drop a link here.

PR #438, opened 2021-07-26
fix: fix unit tests for Node.js event templates

lance:
# Changes

:bug: The bump of a major version of the Node.js dependencies changed
the format for the data received over HTTP (string vs. object) causing a regression.

This fixes that test to use JSON.stringify().

/kind bug


matejvasek:
lgtm

PR #449, opened 2021-07-30
fix: pkger runs only when necessary

lkingland:
# Changes

:bug: pkged.go generated only when ./templates change.
:broom: makefile cleanup

/kind bug

Removes python caches on template test.  When they existed, it caused a rebuild.

Adds pkger.go as an explicit entry in the CODE prerequisite var.  This ensures pkged.go is generated if it doesn't exist, and removes the need to explicitly enumerate it as a prerequisite to other targets.

Adds pkger.go to the clean target.  This allows a `make clean && make` to work as one might expect.  For example ensuring a rebuild if a template file is removed.

The notable conceptual change here is that make does not induce a build of pkged.go by explicitly enumerating it as a prerequisite (a difficult thing to get right, and prone to errors in the future), but rather directly enumerates ./templates as its prerequisite.


lance (approved):
/lgtm

PR #460, opened 2021-08-05
feat: generate json schema for func.yaml

### LINKED ISSUE #336: Add YAML schema for func.yaml
We should create a schem for `func.yaml`, this could help with the validation and easier editing of the file in IDEs (values suggestion).

https://developers.redhat.com/blog/2020/11/25/how-to-configure-yaml-schema-to-make-editing-files-easier/

zroubalik:

# Changes

- :gift: Generate json schema for func.yaml

/kind enhancement

Relates: #336 

We can generate json schema for `func.yaml`, the schema could be added to https://www.schemastore.org/json/ -> most of IDEs uses schemastore to enable code completion & validation  of certain files for users out of the box:

https://user-images.githubusercontent.com/726523/128372953-6f58bea0-d8f0-4407-a86e-180cf207be7a.mp4

I wasn't able to add validation to `Envs` and `Labels`, because they internally share the same type `Pairs`, but the validation of values is different for each type: https://github.com/knative-sandbox/kn-plugin-func/blob/818ec572aa52f12a35bb6b022c2e727ce6ebbfd3/config.go#L120-L123
A solution would be to re-introduce separate Go types for each property.

**Release Note**

```release-note
Generate json schema for func.yaml
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,7 +6,7 @@ on:
       - "main"
 jobs:
   test:
-    runs-on: 'ubuntu-latest'
+    runs-on: "ubuntu-latest"
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
@@ -39,6 +39,8 @@ jobs:
           PKGER: "./pkger"
       - name: Lint
         run: make check
+      - name: Check that 'func.yaml schema' is up-to-date
+        run: make schema-check
     outputs:
       pkger: ${{ steps.pkger-download-url.outputs.result }}
 
@@ -80,10 +82,10 @@ jobs:
         run: |
           curl -s -L -o pkger.tgz ${{ needs.test.outputs.pkger }}
           tar xvzf pkger.tgz
-  
+
       # Standard build tasks
       - name: Build
-        run:  make cross-platform
+        run: make cross-platform
         env:
           PKGER: "./pkger"
           # NOTE:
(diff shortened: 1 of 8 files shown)


lkingland (approved):
This is sure to be a much-appreciated addition by folks editing the func.yaml directly

👍🏻

zroubalik:
> I wasn't able to add validation to `Envs` and `Labels`, because they internally share the same type `Pairs`, but the validation of values is different for each type:
> 
> https://github.com/knative-sandbox/kn-plugin-func/blob/818ec572aa52f12a35bb6b022c2e727ce6ebbfd3/config.go#L120-L123
> 
> A solution would be to re-introduce separate Go types for each property.

@lance @matejvasek what do you think about this?

lance:
@zroubalik I guess I'm not really opposed to this. Of course, I think it's always best to have common/reusable code where possible. But validation _is_ really nice.

zroubalik:
> @zroubalik I guess I'm not really opposed to this. Of course, I think it's always best to have common/reusable code where possible. But validation _is_ really nice.

Yeah, I share the same concern. But in this case it makes sense, this makes me wonder, whether I should make to `labels` as well, currently they are `name`&`value`, but more correct syntax is imho `key` & `value`, that would be a breaking change though.

lkingland:
I vote for key, because `Name` implies a standalone value, whereas `Key` implies there is a value that goes with it.

PR #462, opened 2021-08-05
docs: add section about Node.js log levels

lance:
# Changes

- :book: adding some documentation around how to set logging levels in Node.js

Note that these documents will likely need to move out of this repo.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/328

### DIFF
--- a/docs/guides/nodejs.md
+++ b/docs/guides/nodejs.md
@@ -220,12 +220,15 @@ The function will log
 {"level":30,"time":1604511655265,"pid":3430203,"hostname":"localhost.localdomain","reqId":1,"msg":"Processing customer"}
 ```
 #### Setting the log level
-Developers can control the log level by setting
-the `FUNC_LOG_LEVEL` environment variable. The possible options for this
-adhere to the options available for [`pino`](https://getpino.io/#/docs/api?id=level-string),
+Developers can control the log level by setting the `logLevel` value in
+`func.yaml` The possible options for this adhere to the options available
+for [`pino`](https://getpino.io/#/docs/api?id=level-string),
 and may be one of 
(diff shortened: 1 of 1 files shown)


zroubalik:
This makes me wonder, are we able to unify the logging experience across all templates? So we can then configure logging in `config` and not via env variable (which is imho more user friendly).

lance:
> This makes me wonder, are we able to unify the logging experience across all templates? So we can then configure logging in `config` and not via env variable (which is imho more user friendly).

Agree that configuration in the yaml is nice. But how do you see this working across multiple languages/runtimes that may be idiomatically different?

zroubalik:
> > This makes me wonder, are we able to unify the logging experience across all templates? So we can then configure logging in `config` and not via env variable (which is imho more user friendly).
> 
> Agree that configuration in the yaml is nice. But how do you see this working across multiple languages/runtimes that may be idiomatically different?

Maybe the config could suggest appropriate logging levels for each language/runtime via `func config logging`? 🤷‍♂️ Maybe I am overthinking this.

lkingland:
> Maybe the config could suggest appropriate logging levels for each language/runtime via `func config logging`? 🤷‍♂️ Maybe I am overthinking this.

This would be preferable from my view as well.  It would be quite nice to have a generalized language for defining logging expectations.

lance:
I see, so we have the ability to set https://github.com/knative-sandbox/kn-plugin-func/pull/462/files#diff-b55c094f2e6794329700479dc202c7ea098e14eeec000cfe72ede2ab57bb00f8R222-R228 in func.yaml - and each buildpack can do what they want with that, or ignore it altogether. I like it.

lance:
@zroubalik @lkingland updated with `logLevel` in func.yaml. The schema should be updated obviously, and it would be nice to add this to the configure command.

/hold

PR #465, opened 2021-08-06
feat: introduce manifest.yaml for language pack providers

lance:
# Changes

## :gift: introduce a mechanism for language pack metadata

This commit introduces the concept of a `manifest.yaml` file which
can be added to each template directory to specify things such as health
endpoints for the function. There may be a case to be made for adding this
file at the language level instead of the template level to simplify and
eliminate redundancy. As written, a language pack repository such as the
one at https://github.com/lance/gcf-kn would include a `manifest.yaml`
for each template - e.g. `ruby/cloud-fn` and `ruby/http`. This might be
overkill.

See: https://github.com/lance/gcf-kn/blob/lance/paketo-builder/ruby/cloud-function/manifest.yaml

Also problematic is the fact that, because we don't yet have the concept
of a global configuration file, the only place to store this metadata is
in a project's `func.yaml` file. But there is no reason a function dev
should be exposed to things like health endpoints. This information is
determined by the language pack and integrated by the builder. The only
reason we need it is that it's necessary to determine these endpoints when
deploying. Otherwise, we would have to insist that all language packs
provide endpoints at `/health/readiness` and `/health/liveness`. In my
opinion, a better solution would be to have a global configuration which
stored this metadata for each language pack as it is discovered.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/463

/kind/enhancement

**Release Note**
```release-note
Provide a means for language pack providers to include metadata
```


client.go
```diff
@@ -409,6 +409,25 @@ func (c *Client) Create(cfg Function) (err error) {
...
+	languagePackFilePath := filepath.Join(f.Root, ".language-pack.yaml")
+	if funcConfig, err := ioutil.ReadFile(languagePackFilePath); err == nil {
+		cfg := make(map[string]map[string]string)
+		if err := yaml.Unmarshal(funcConfig, cfg); err == nil {
+			f.Builder = cfg["builders"]["default"]
+			if c.verbose {
+				fmt.Printf("Builder: %s\n", f.Builder)
+			}
+			f.BuilderMap = cfg["builders"]
+			f.HealthEndpoints = cfg["healtEndpoints"]
```
rhuss, line 422:
```suggestion
			f.HealthEndpoints = cfg["healthEndpoints"]
```

lance:
This latest push changes the `.builders.yaml` file to `mainfest.yaml` and modifies the structure just a bit.

```yaml
builders:
  default: quay.io/boson/faas-nodejs-builder:v0.8.4
buildpacks:
  - quay.io/boson/faas-nodejs-bp:v0.8.4
  - quay.io/boson/redhat-noop-buildpack:tip
healthEndpoints:
  liveness: /health/liveness
  readiness: /health/readiness
```

Note that the `buildpacks` array can be empty and if so, whatever buildpacks exist in the configured builder will be the only ones applied. I've included a noop buildpack in the typescript/event template for illustrative purposes and for playing around with. This should not ultimately be committed.

I still need to test the builders/buildpacks behavior.

As I go down this path, it seems more and more that there should be one or at most two built in templates that are extremely simple, and are built using an upstream builder and buildpack, and that the existing built in templates should ultimately move back to @boson-project. The builtin templates should use a generic builder for the language, and specify a buildpack in `manifest.yaml` that knows how to wrap a function up so that it can be run as a process. An example for Node.js using paketo builders/buildpacks.

```yaml
builders:
  default: gcr.io/paketobuildpacks/builder:full

buildpacks:
  - gcr.io/paketo-buildpacks/nodejs
  - gcr.io/knative-sandbox/func-base-bp
  - gcr.io/knative-sandbox/node.js-bp

healthEndpoints:
  liveness: /health/liveness
  readiness: /health/readiness
```

/cc @lkingland @matejvasek @zroubalik

client.go
```diff
@@ -391,23 +398,30 @@ func (c *Client) Create(cfg Function) (err error) {
...
+				if len(f.Buildpacks) > 0 {
+					fmt.Println("Buildpacks:")
+					for _, b := range f.Buildpacks {
+						fmt.Printf("           ... %s\n", b)
+					}
+				}
 			}
-			f.BuilderMap = builders
 		}
 		// Remove the builders.yaml file so the user is not confused by a
```
zroubalik, line 420:
```suggestion
		// Remove the manifest.yaml file so the user is not confused by a
```

zroubalik (commented):
Looking good.

+1 for the global config file. It is clear that we need one.

lkingland (approved):

Issue #454, opened 2021-07-30
Provide CLI based global management of language pack repositories


lance:
## Function developers should be able to easily manage language pack repositories
As a function developer, I want to manage the repositories that have been configured in `~/.config/kn/repositories` so that I can add, remove and list language pack repositories. For example, the initial prompt when running `kn func config --global` in any directory might look something like this.

```
❯ kn func config --global
? What do you want to configure?  [Use arrows to move, type to filter]
> Language Pack Repositories
```

When selecting `Language Pack Repositories`, the user may see a screen similar to this

```
? What operation do you want to perform?  [Use arrows to move, type to filter]
   Add
   Remove
> List
```

A function developer can add a language pack to their system by providing a git URI, for example https://github.com/lance/gcf-kn, or a path to a directory on disk. If a URI is provided, the CLI will clone the language pack repository into `~/.config/func/repositories/<repo-name>`. If a path to a directory is provided, the CLI will clone the language pack repository into `~/.config/func/repositories/<directory-name>`. The language pack is then available for use the next time `kn func create` is invoked.

When the user chooses `Remove`, the repository directory is deleted - perhaps with a warning prompt. 

`List` will, of course, list all installed language pack repositories, and perhaps display any metadata specified in a yet to be determined [`MANIFEST.yaml`](https://github.com/knative-sandbox/kn-plugin-func/issues/452). The list will also include the special value `Builtin` to indicate the default builtin templates. The first time a user runs this, `Builtin` is selected or indicated somehow as the default, but the user may use the arrow keys or type to filter the available templates from installed language packs. By selecting one of these templates, this becomes the default template created when a user executes `kn func create <project>`.

---

## Current behavior
Currently the built in templates are always the default, and a user must install or remove language pack repositories by either copying a directory or cloning a repository into `~/.config/func/repositories/`.

---

## Acceptance criteria
```
GIVEN a user has the functions plugin installed
WHEN they type `kn func config --global`
THEN they are able to install, remove and list language pack repositories
AND they can see what language packs are available in each repository
AND the can set the default language pack for use when creating new projects
```

PR #467, opened 2021-08-09
feat: repository management client api

lkingland:
# Changes

:gift: Client repository management API

Adds a Repositories member to Client which enables management of  extensible repositories via methods Add, Rename, Remove, and List. Prerequisite for #454.

/kind enhancement


repositories_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+	fn "knative.dev/kn-plugin-func"
+)
+
+const RepositoriesTestRepo = "repository-a"
+
+// TestRepositoriesList ensures the base case of an empty list
+// when no repositories are installed.
+func TestRepositoriesList(t *testing.T) {
+	root, rm := mktemp(t)
```
lance, line 20: I know it's a bit pedantic, but `mktemp()` does not return a URL. It just returns a path string (which is why, I assume, you have `ss := strings.Split(uri, string(os.PathSeparator))` in `repositories.go`). I wonder if the interface should actually use a `URL` type from `net/url` to ensure proper input. In that case, this might be:

```suggestion
	root, rm := "file://" + filepath.ToSlash(mktemp(t))
```

lance (commented):
Looks good. Just a couple of questions/suggestions.

repositories.go
```diff
@@ -0,0 +1,66 @@
...
+		if f.IsDir() && !strings.HasPrefix(f.Name(), ".") {
+			list = append(list, f.Name())
+		}
+	}
+	return
+}
+
+// repoNameFrom uri returns the last token with any .git suffix trimmed.
+func repoNameFrom(uri string) (name string) {
+	ss := strings.Split(uri, string(os.PathSeparator))
```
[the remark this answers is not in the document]
lkingland: Right, good question.  Yes, this is stable, as it is specifically just a way to uniquely reference a template when multiple repositories may contain one of the same name.  This isn't a file path or network address or anything like that.

repositories_test.go
lkingland: Not pedantic at all!  I am glad you brought it up, because this is probably more correct:

The tests which generate a local `file://` path for addition now do so using `filepath.ToSlash`.
The string passed in to `.Add` is now validated as a URL using `net/url.Parse`.

Thanks for the suggestion!  It's indeed better like this.

Regarding using URL as the input type:  I stopped short of _requiring_ input of type `net.URL` because, from what I have seen, the idiom has been to use strings in most all cases due to the notoriously loose definition of what constitutes a URL; it's requirement merely shifting the expectation of type shifting and validation to the library's user.   Instead sticking with strings for the API and doing the work ourselves of validating that no explicitly _invalid_ urls are passed, returning a standard error on invalid strings.  Thus the _user_ of the library gets the benefit of simple string acceptance and mutation without needing to import the net package, nor parse and error handle themselves.  I hope this is an acceptable compromise.

lance: Before I commented on this yesterday I did a bit of reading about URL vs URI and I agree, the whole thing is a mess, specified in a couple of archaic RFCs. Your approach here makes perfect sense in this context!

PR #468, opened 2021-08-09
feat: make func schema if config updated

lkingland:
# Changes

🎁 integrates new schema generator into core build targets

Invokes schema generation on `make` when:
- config.go modified
- after a `make clean`

Also includes an explicit target for the exact file as well via `make schema/func_yaml-schema.json`

/kind enhancement

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,8 +26,8 @@ LDFLAGS := "-X main.date=$(DATE) -X main.vers=$(VERS) -X main.hash=$(HASH)"
 # as Go objects and exposed at runtime as a filesystem.
 PKGER ?= pkger
 
-# Target prerequisites for the two stages of encoding and building.
-CODE := $(shell find . -name '*.go') pkged.go go.mod
+# All Code prerequisites, including generated files, etc.
+CODE := $(shell find . -name '*.go') pkged.go go.mod schema/func_yaml-schema.json
 TEMPLATES := $(shell find templates -name '*' -type f)
 
 # Default Targets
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -75,8 +75,9 @@ pkged.go: $(TEMPLATES)
 	# to insstall pkger: 'go get github.com/markbates/pkger/cmd/pkger'
 	$(PKGER)
 
-clean: ## Remove generated artifacts such as binaries
+clean: ## Remove generated artifacts such as binaries and schemas
 	rm -f $(BIN) $(BIN_WINDOWS) $(BIN_LINUX) $(BIN_DARWIN)
+	rm -f schema/func_yaml-schema.json
```
zroubalik, line 80: I am not sure if we want to remove the schema, it is going to be linked from the SchemaStore: https://github.com/SchemaStore/schemastore/pull/1787

zroubalik (commented):
That's much better, thanks!

Makefile
zroubalik: Btw. there are 2 ways how to include a schema in schemastore: add the schema directly to the web via PR or link it from an external repo, I've chosen the latter.

lkingland: Oh, that's perhaps worth clarifying:  that remove statement is just for the development loop:  it will never leave the repository. 

That remove statement is for the task `make clean`, which generally is used in conjunction with `make`.  So, it's a little bit like a hard-reset:  `make clean && make`.  So it is expected that a developer will run a few commands before submitting a pr:  `make clean && make`, `make test` and `make check`.   But during active, iterative development, they might just short-cut and run `make build` or even run `go build` directly.

So in no circumstance will the `func_yaml-schema.json` ever be actually removed from the repo:  just rebuilt and checked in.  Just like `pkged.go`:  it's a source-controlled artifact, even though it is generated. The difference here with `pkged.go` being that the schema file can be built in much more tightly than `pkged.go`, because it is part of the core (not templates), and doesn't require the installation of a third-party binary to build (you kindly built the generator right into the repo).

lkingland: >  add the schema directly to the web via PR or link it from an external repo, I've chosen the latter.
I like this too.  It's more likey to stay fresh.  Perhaps too fresh sometimes: there is of course the possibility that the reference will include updates which are not yet released.

zroubalik: Make sense! :)

lance (approved):
/lgtm

PR #472, opened 2021-08-11
feat!: use key&value for Labels

### LINKED ISSUE #336: Add YAML schema for func.yaml

zroubalik:

# Changes

- :broom: use `key&value` for Labels instead of `name&value` so we are aligned with Kuberentes https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
- :gift:  added `envs` and `labels` validation to the func.yaml schema

Fixes #336

### DIFF
--- a/config.go
+++ b/config.go
@@ -42,37 +42,59 @@ func (v Volume) String() string {
 	return ""
 }
 
-type Pairs []Pair
-type Pair struct {
-	Name  *string `yaml:"name,omitempty"`
+type Envs []Env
+type Env struct {
+	Name  *string `yaml:"name,omitempty" jsonschema:"pattern=^[-._a-zA-Z][-._a-zA-Z0-9]*$"`
 	Value *string `yaml:"value"`
 }
 
-func (e Pair) String() string {
+func (e Env) String() string {
 	if e.Name == nil && e.Value != nil {
 		match := regWholeSecret.FindStringSubmatch(*e.Value)
 		if len(match) == 2 {
-			return fmt.Sprintf("All key=value pairs from Secret \"%s\"", match[1])
+			return fmt.Sprintf("Env key=value pairs from Secret \"%s\"", match[1])
 		}
 		match = regWholeConfigMap.FindStringSubmatch(*e.Value)
 		if len(match) == 2 {
-			return fmt.Sprintf("All key=value pairs from ConfigMap \"%s\"", match[1])
+			return fmt.Sprintf("Env key=value pairs from ConfigMap \"%s\"", match[1])
 		}
 	} else if e.Name != nil && e.Value != nil {
 		match := regKeyFromSecret.FindStringSubmatch(*e.Value)
 		if len(match) == 3 {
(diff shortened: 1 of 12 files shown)


config.go
```diff
@@ -42,37 +42,59 @@ func (v Volume) String() string {
...
 	Value *string `yaml:"value"`
 }
 
-func (e Pair) String() string {
+func (e Env) String() string {
 	if e.Name == nil && e.Value != nil {
 		match := regWholeSecret.FindStringSubmatch(*e.Value)
 		if len(match) == 2 {
-			return fmt.Sprintf("All key=value pairs from Secret \"%s\"", match[1])
+			return fmt.Sprintf("Env key=value pairs from Secret \"%s\"", match[1])
```
lance, line 55: Shouldn't this remain `All`?

config.go
```diff
@@ -42,37 +42,59 @@ func (v Volume) String() string {
...
 		}
 		match = regWholeConfigMap.FindStringSubmatch(*e.Value)
 		if len(match) == 2 {
-			return fmt.Sprintf("All key=value pairs from ConfigMap \"%s\"", match[1])
+			return fmt.Sprintf("Env key=value pairs from ConfigMap \"%s\"", match[1])
```
lance, line 59: Same as above - shouldn't this be `All`?

config.go
```diff
@@ -42,37 +42,59 @@ func (v Volume) String() string {
...
+	}
+	return ""
+}
+
+type Labels []Label
+type Label struct {
+	// Key consist of optional prefix part (ended by '/') and name part
+	// Prefix part validation pattern: [a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*
+	// Name part validation pattern: ([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]
+	Key   *string `yaml:"key" jsonschema:"pattern=^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\\/)?([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$"`
```
lance, line 85: What I most love about RegEx in general is how easy they are to just take a quick glance at, and know exactly what they are matching. :sob: :rofl:

lance (commented):
Just one question about the console output.

config.go
zroubalik: Good catch! copypaste error 🤦

config.go
zroubalik: lol 🔥

config.go
zroubalik: Fixed

config.go
zroubalik: I copypasted those two regexes directly from Kubernetes validation code. I tested the combined regex, so it should be ok 🤞

PR #474, opened 2021-08-11
feat!: change `describe` command to `info`

lance:
# Changes

- :broom: Change the `describe` command to `info`

/kind enhancement

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/337
The describe command conflicts sematically with kubectl describe.

**Release Note**

```release-note
The `describe` command has been renamed to `info` to reduce conflict with the `kubectl` and `kn` commands of the same name. The behavior is otherwise identical.
```

**Docs**

Documentation changes are included in this PR.

### DIFF
--- a/knative/describer.go
+++ b/knative/describer.go
@@ -31,7 +31,7 @@ func NewDescriber(namespaceOverride string) (describer *Describer, err error) {
 // restricts to label-syntax, which is thus escaped. Therefore as a knative (kube) implementation
 // detal proper full names have to be escaped on the way in and unescaped on the way out. ex:
 // www.example-site.com -> www-example--site-com
-func (d *Describer) Describe(ctx context.Context, name string) (description fn.Description, err error) {
+func (d *Describer) Describe(ctx context.Context, name string) (description fn.Info, err error) {
 
 	servingClient, err := NewServingClient(d.namespace)
 	if err != nil {
--- a/client.go
+++ b/client.go
@@ -126,10 +126,10 @@ type ProgressListener interface {
 // Describer of Functions' remote deployed aspect.
 type Describer interface {
 	// Describe the running state of the service as reported by the underlyng platform.
-	Describe(ctx context.Context, name string) (description Description, err error)
+	Describe(ctx context.Context, name string) (description Info, err error)
 }
 
-type Description struct {
+type Info struct {
 	Name          string         `json:"name" yaml:"name"`
 	Image         string         `json:"image" yaml:"image"`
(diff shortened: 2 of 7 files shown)


knative/describer.go
```diff
@@ -31,7 +31,7 @@ func NewDescriber(namespaceOverride string) (describer *Describer, err error) {
 // restricts to label-syntax, which is thus escaped. Therefore as a knative (kube) implementation
 // detal proper full names have to be escaped on the way in and unescaped on the way out. ex:
 // www.example-site.com -> www-example--site-com
-func (d *Describer) Describe(ctx context.Context, name string) (description fn.Description, err error) {
+func (d *Describer) Describe(ctx context.Context, name string) (description fn.Info, err error) {
```
zroubalik, line 34:
```
func (d *Describer) Describe(ctx context.Context, name string) (description fn.Info, err error) {
```
Shouldn't this be changed as well?

lance: I thought about that and decided not to change it for a couple of reasons. I meant to put this in the PR description, but just forgot.

* `Infoer` is a terrible name! :rofl: 
* The `Describer` in the `knative` package is responsible for the verb "describe" which the client uses to surface the noun "info" and they don't need to necessarily evolve 1:1 :sunglasses: 

But now by the light of morning, those reasons seem pretty weak. Do you have any suggestions for naming? I don't particularly like this.

```go
type Infoer struct {
  // ...
}

func (i *Infoer) Info(ctx context.Context, name string) (description fn.Info, err error)
```

What do you think of this?

```go
type Info struct {
  // ...
}

func (i *Info) Fetch(ctx context.Context, name string) (description fn.Info, err error)
```

lkingland: Hm, seems like a tradeoff either way.  I agree that `Describer` seems a lot nicer than `Infoer`, and that it's fine if the CLI and client library aren't exactly mapped, if it means the client library is closer to what we think is "correct".  I could go either way on it, but leaving it Describe fits for me:  run the `info` command to get a description.

lkingland: Hm, seems like a tradeoff either way.  I agree that `Describer` seems a lot nicer than `Infoer`, and that it's fine if the CLI and client library aren't exactly mapped, if it means the client library is closer to what we think is "correct".  I could go either way on it, but leaving it Describe fits for me:  run the `info` command to get a description.

lance: > Not to mention, there is the possibility that `info` may display more than just the description from the describer in the future...

Thanks for mentioning this. I avoided saying it because it's easy to fall down the flexibility hole.

PR #476, opened 2021-08-12
chore: ignore IDE files

matejvasek:
- :broom: ignore IDE files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ __pycache__
 /bin
 
 .vscode
+.idea


lance (commented):
I don't feel terribly strongly about it, but personally I'd rather see this PR remove `.vscode` and not add `.idea` since these are not project specific files. IMO best practice is to have these kinds of files in a personal `~/.gitignore` file since there is no way this project will ever know all of the possible unrelated development tools contributors might be using which create temp files in the project directory. I'd rather that the project's `.gitignore` contain only project specific generated/temp files and directoreis.

zroubalik:
> I don't feel terribly strongly about it, but personally I'd rather see this PR remove `.vscode` and not add `.idea` since these are not project specific files. IMO best practice is to have these kinds of files in a personal `~/.gitignore` file since there is no way this project will ever know all of the possible unrelated development tools contributors might be using which create temp files in the project directory. I'd rather that the project's `.gitignore` contain only project specific generated/temp files and directoreis.

I personally have the opposite opinion :) I concur we are not able to cover all possibiliites, but we can at least cover the major stuff. Imho the project should try to make the development as easy as possible for developers and putting stuff into `.gitignore` is not something that bothers me. But that's just my opinion, no hard feelings so I am fine with any outcome :)

lance:
> But that's just my opinion, no hard feelings so I am fine with any outcome :)

Same here, tbh. Seems like @matejvasek is the only one who cares enough to actually submit a PR, so it seems like that should win the day! :)

lkingland:
> I don't feel terribly strongly about it, but personally I'd rather see this PR remove `.vscode` and not add `.idea` since these are not project specific files. IMO best practice is to have these kinds of files in a personal `~/.gitignore`

I also do not feel strongly about this, but completely agree that the project itself is the wrong place for this.  It is not the source code's responsibility to state which IDE's config files it should not source-controle.  The proper place for this is in a user's global git configuration.  It's not "correct", and as such my OCD flares up when I see it in there.

That said, there's lots of things in our world that are in the "wrong" place, and we seem to muddle along without too much trouble, so if it has a reasonable liklihood of helping someone, I will just swallow my indignation whenever I open that file and soldier on.

PR #481, opened 2021-08-17
Using systemctl to manage podman user owned service

cardil:
Using systemctl is a better idea because users can utilize other subcommands like `status` or `stop` to further manage the podman service.

# Changes

- :broom: Clean up commands to better follow standards and docs

/kind cleanup
/kind documentation


lkingland (approved):
A welcome couple of changes indeed

PR #485, opened 2021-08-18
feat!: move built in templates for Go, Node.js and TypeScript to Paketo builders

lance:
# Changes

- :gift: Move builtin templates for Go, Node.js and TypeScript to Paketo builders

/kind enhancement

This commit removes the Boson builders from the built in templates
for Go, TypeScript and Node.js, replacing them with paketo builders.
For Go, the builder is augmented with a simple buildpack that installs
the Go wrapper code and its dependencies. For TypeScript, the paketo
buildpacks oddly don't support an `npm build` step, so these templates
are also dependent on a small Boson buildpack. These buildpacks are
currently at https://github.com/lance/boson-buildpacks but should find
a home either in the boson-project organization, or the knative-sandbox
organization.

This change also slightly modifies how the Node.js and TypeScript
templates are structured, reducing the coupling between the buildpack
and a function project.

This commit includes the code in https://github.com/knative-sandbox/kn-plugin-func/pull/465
and is dependent on it in the use of manifest.yaml.


templates/go/events/manifest.yaml
```diff
@@ -0,0 +1,12 @@
+builders:
+  default: paketobuildpacks/builder:base
+  base: paketobuildpacks/builder:base
+  full: paketobuildpacks/builder:full
+
+buildpacks:
+  - paketo-buildpacks/go-dist
+  - docker.io/lanceball/boson-go # github.com/lance/buildpacks
```
lance, line 8: Before this PR lands, we should figure out where these should really live - both git org/repo, and container repo.

/hold

templates/go/events/manifest.yaml
```diff
@@ -0,0 +1,12 @@
+builders:
+  default: paketobuildpacks/builder:base
```
lance, line 2: Note how all of the templates use the same builder.

templates/node/events/manifest.yaml
```diff
@@ -0,0 +1,8 @@
+builders:
+  default: paketobuildpacks/builder:base
+  base: paketobuildpacks/builder:base
+  full: paketobuildpacks/builder:full
+
```
lance, line 5: Node functions don’t need a special buildpacks.

templates/typescript/events/manifest.yaml
```diff
@@ -0,0 +1,12 @@
+builders:
+  default: paketobuildpacks/builder:base
+  base: paketobuildpacks/builder:base
+  full: paketobuildpacks/builder:full
+
+buildpacks:
+  - paketo-buildpacks/nodejs
+  - docker.io/lanceball/boson-typescript # github.com/lance/buildpacks
```
lance, line 8: Use the paketo buildpacks to install Node.js, npm, and function dependencies. Then use the boson-typescript buildpack to do the TS>JS transpilation.

config.go
```diff
@@ -131,18 +131,20 @@ type ResourcesRequestsOptions struct {
...
-	Annotations map[string]string `yaml:"annotations"`
-	Options     Options           `yaml:"options"`
-	Labels      Labels            `yaml:"labels"`
+	Name            string            `yaml:"name" jsonschema:"pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"`
+	Namespace       string            `yaml:"namespace"`
+	Runtime         string            `yaml:"runtime"`
+	Image           string            `yaml:"image"`
+	ImageDigest     string            `yaml:"imageDigest"`
+	Builder         string            `yaml:"builder"`
+	BuilderMap      map[string]string `yaml:"builderMap"`
```
lkingland, line 140: Having both a Builder and Builders is perhaps marginally better than appending "Map" to the plural:  Not only is it a bit closer to natural language, it allows us to match `Config` closely with the `Manifest` (one step closer to being an embedded struct).

While there is of course the possibility of confusion, I consider plurality to be an important in establishing member uniqueness.

Throwing caution and potential confusion to the wind, the subtle plurality differentiator is already being embraced with the members `repository` vs `repositories` on the Client.  Perhaps its worth considering here as well.

knative/deployer.go
```diff
@@ -149,15 +149,25 @@ func probeFor(url string) *corev1.Probe {
...
 func generateNewService(f fn.Function) (*servingv1.Service, error) {
 	containers := []corev1.Container{
 		{
 			Image: f.ImageWithDigest(),
 		},
 	}
-
-	containers[0].LivenessProbe = probeFor("/health/liveness")
-	containers[0].ReadinessProbe = probeFor("/health/readiness")
+	setHealthEndpoints(f, &containers[0])
```
lkingland, line 170: Perhaps not the time for this, but all this indexing into a slice is nerve-wracking.  It's one of those things begging for a future error.

This should probably be refactored to be a single container, because at no point in this function is it used in any way other than a single container, until line 204, at which point would be more appropriate to set it as the only member of slice defined inline anonymously.  The complexity starting to compound as now the `setHealthEndpoints` function needs to pass a reference, making this indexing now pass through a function boundary.

Perhaps better done during a dedicated refactor pass, but thought I'd mention it now, since the function's being touched, that's usually the best time to spruce things up.

knative/deployer.go
```diff
@@ -219,6 +229,19 @@ func updateService(f fn.Function, newEnv []corev1.EnvVar, newEnvFrom []corev1.En
 		for k, v := range f.Annotations {
 			service.ObjectMeta.Annotations[k] = v
 		}
+		// I hate that we have to do this. Users should not see these values.
```
lkingland, line 232: Agreed... perhaps these values do not need to be populated and passed through for each Function.  We know that 99.9% of the time they will have the default values.

Perhaps the solution here is to make these health endpoint values default _unless_ they exist in the language pack.  That way we are sticking with the general principle:  most everything is invisible to the Function developer using sane defaults... unless one wants to pop the proverbial hood and override things.

By using defaults here in the Deployer, the Function developer would only see these values populated if the language pack chose to populate them... which would _also_ only need to happen if the language pack developer has a reason to deviate from the default.

knative/deployer.go
```diff
@@ -219,6 +229,19 @@ func updateService(f fn.Function, newEnv []corev1.EnvVar, newEnvFrom []corev1.En
...
+		// It is an implementation detail. These health endpoints should not be
+		// a part of func.yaml since the user can only mess things up by changing
+		// them. Ultimately, this information is determined by the language pack.
+		// Which is another reason to consider having a global config to store
+		// some metadata which is fairly static. For example, a .config/func/global.yaml
```
lkingland, line 237: Per my earlier comment, health endpoints may be better solved by providing sane defaults in `deployer.go` directly.  However, this is a good time to drop 2c on the global config.  In general I think this is a very good idea and a good way forward.

We can take the `git` configuration system as a guide here I think:

There are many settings which we default for the Function developer.   Making guesses on their behalf being preferred over prompting for everything.

There exist certain settings where the default chosen presents a nontrivial chance of the user desiring to modify that value.  This can be accomplished by the user modifying their local Function's `.func.yaml` either manually, via the CLI's `config` subcommand, or via the client library if this is an integration situation.

In the event this change to the default value is something the user would like to happen by default on all Functions created by the current user, they can make the change in their global `$XDG_CONFIG_HOME/func/config.yaml` (or via the CLI by providing the `--global` flag on config changes).

This seems like a sane approach; has worked well for git, and since we expect most users to be familiar with git already, this reuses their knowledge; requiring little cognitive load to assimilate.

knative/deployer_test.go
```diff
@@ -3,8 +3,31 @@ package knative
 import (
 	"os"
 	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	fn "knative.dev/kn-plugin-func"
 )
 
+func Test_setHealthEndpoints(t *testing.T) {
```
lkingland, line 11: Fleshing out the idea of my earlier comment,  should it strike your fancy:
Tests would look something like adding an additional test `Test_setHealthEndpointsDefaults` which ensures that a Function with no values in its `.HealthEndpoints` results in the LivenessProbe receiving the default values `DefaultLiveness` and `DeafaultReadiness`...  which are of course set with the standard Kube values.

templates/go/events/manifest.yaml
lkingland: That's a nice simplification!

templates/go/events/manifest.yaml
lkingland: It seems to me at least that git's boson-project is the right place for the code, and docker.io is still the most common public registry.

templates/go/events/manifest.yaml
```diff
@@ -0,0 +1,12 @@
...
+  base: paketobuildpacks/builder:base
+  full: paketobuildpacks/builder:full
+
+buildpacks:
+  - paketo-buildpacks/go-dist
+  - docker.io/lanceball/boson-go # github.com/lance/buildpacks
+
+healthEndpoints:
+  liveness: /health/liveness
+  readiness: /health/readiness
```
lkingland, line 12: If these defaults were instead defined as constants `DefaultLiveness` and `DefaultReadiness` in `deployer.go`, they could be omitted from the default embedded templates.. and from any Function project for that matter, unless the developers (of either Function or Language Pack) specifically need to override.

lkingland (commented):
This sure is nice to see coming together!

My only substantive comment (detailed below) is that in general all the default values should probably be yanked out of the `.yaml`s and instead be Go constants. 

In addition, a couple tests would need to be added to ensure that the defaults were indeed being applied.

/approve
(Dear robot: I want to APPROVE this PR as being sufficiently excellent to merge, but not actually TRIGGER a merge in case Lance wants to take some of my suggestions....  🤨 )

client.go
```diff
@@ -391,23 +398,30 @@ func (c *Client) Create(cfg Function) (err error) {
...
-		// TODO: A command line flag could be used to specify non-default builders
-		builders := make(map[string]string)
-		if err := yaml.Unmarshal(builderConfig, builders); err == nil {
-			f.Builder = builders["default"]
+	manifestFilePath := filepath.Join(f.Root, "manifest.yaml")
+	if manifestYaml, err := ioutil.ReadFile(manifestFilePath); err == nil {
+		// A manifest.yaml file was found. Read the default builder and set in the config file
+		manifest := Manifest{}
+		if err := yaml.Unmarshal(manifestYaml, &manifest); err == nil {
+			f.Builder = manifest.Builders["default"]
```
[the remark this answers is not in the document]
lance: Makes sense to me

config.go
lance: > While there is of course the possibility of confusion, I consider plurality to be sufficient for establishing uniqueness, despite the diminutive size of the poor little "s".

:rofl: I agree and will change it

templates/go/events/manifest.yaml
lance: The only issue with this is that it does not allow a language pack provider to forego supplying these endpoints. It places a requirement on all language packs to either provide these or simply specify `/` as the health check endpoint.

lance: See the commit message where this is addressed. I suggest we document that language packs wishing to avoid implementing these health checks at `/health/readiness` and `/health/liveness` could use the root path with query parameters. E.g. `/?health=readiness`.

lkingland (approved):
Lookin good

lance:
/hold
Updating the buildpack images and such

lance:
/hold cancel

Looks like this is fine to go ahead. We will need to update the two buildpack URIs when that has resolved in the boson-project/packs repo.

PR #487, opened 2021-08-19
style: create command output formatting for repository option

jrangelramos:
# Changes
- :broom: fix identation of create command output when specifying a repository

/kind cleanup


lance (approved):
/lgtm
I kept noticing that and then forgetting to fix it!

PR #507, opened 2021-09-03
chore: example message changes between standalone and plugin

### LINKED ISSUE #214: Make examples in usage message parameterizable
The examples that are being added require the command name, which is different whether used as standalone (`faas`) or as a plugin (`kn faas`).

Therefor the example message should be parameterized as a template on this value, which also requires to build up the command dynamically when it is known whether it is running in plugin context or not.

shinigambit:
# Changes
- The example message changes depending on the command being executed as standalone or as a plugin for the `kn` executable

/kind enhancement

Fixes #214

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -41,19 +41,19 @@ var root = &cobra.Command{
 	Long: `Serverless functions
 
 Create, build and deploy functions in serverless containers for multiple runtimes on Knative`,
-	Example: replaceNameInTemplate("func"),
+	Example: replaceNameInTemplate("func", "example"),
 }
 
-func replaceNameInTemplate(name string) string {
+func replaceNameInTemplate(name, template string) string {
 	var buffer bytes.Buffer
-	exampleTemplate.Execute(&buffer, name)
+	exampleTemplate.ExecuteTemplate(&buffer, template, name)
 	return buffer.String()
 }
 
 // NewRootCmd is used to initialize func as kn plugin
 func NewRootCmd() *cobra.Command {
 	root.Use = "kn func"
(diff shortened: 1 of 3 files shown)


lance (commented):
@shinigambit thanks for this! Because all of the individual commands also have similar examples, maybe `func replaceNameInTemplate(name string) string` could become `func replaceNameInTemplate(name, template string) string` so that other commands could be updated to take advantage of it. WDYT?

shinigambit:
Good idea @lance! I'll change it so it can be reused

cmd/root.go
```diff
@@ -27,21 +41,19 @@ var root = &cobra.Command{
...
-kn func deploy --registry <registry/user>
+	Example: replaceNameInTemplate("func", "example"),
+}
 
-# Curl the service with the service URL
-curl $(kn service describe myfunc -o url)
-`,
+func replaceNameInTemplate(name, template string) string {
+	var buffer bytes.Buffer
+	exampleTemplate.ExecuteTemplate(&buffer, template, name)
```
lance, line 49: CI is failing here with
> cmd/root.go:49:33: Error return value of `exampleTemplate.ExecuteTemplate` is not checked (errcheck)
	exampleTemplate.ExecuteTemplate(&buffer, template, name)
Can you check for the error condition here?

I suppose `replaceNameInTemplate()` should return `string, err` and I guess that means that `NewRootCmd()` should return `*cobra.Command, err` as well, which ultimately bubbles up to `plugin.go`.  :thinking:

lance (commented):
Thanks for the contribution @shinigambit !

/lgtm
/approve

PR #514, opened 2021-09-08
feat: repository management cli

### LINKED ISSUE #454: Provide CLI based global management of language pack repositories

lkingland:
# Changes

- :gift: repository management cli

/kind enhancement

Fixes #454


client.go
```diff
@@ -317,6 +316,16 @@ func WithEmitter(e Emitter) Option {
...
 
+// Repositories accessor
+func (c *Client) Repositories() *Repositories {
+	return c.repositories
+}
+
+// Templates accessor
+func (c *Client) Templates() *Templates {
+	return c.templates
+}
```
lkingland, line 327: Placing these behind an accessors allows them to be enumerated in an interface and tested.

client.go
```diff
@@ -8,9 +8,9 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
```
lkingland, line 11: `goimports` is run on save, which (among other things) alphabetically sorts imported packages

cmd/config.go
```diff
@@ -23,7 +23,7 @@ type functionLoaderSaver interface {
 	functionSaver
 }
 
-type standardLoaderSaver struct {}
+type standardLoaderSaver struct{}
```
lkingland, line 26: `gofmt` caught this

cmd/repository.go
```diff
@@ -0,0 +1,686 @@
...
+	// effective values to stdout.
+	fmt.Fprintf(os.Stdout, "Repositories path: %v\n", cfg.Repositories)
+	fmt.Fprintf(os.Stdout, "Verbose logging:   %v\n", cfg.Verbose)
+	return
+}
+
+// prompt returns a config with values populated from interactivly prompting
+// the user.
+func (c repositoryClientConfig) prompt() (repositoryClientConfig, error) {
+	if !interactiveTerminal() {
```
lance, line 289: I don't think this will ever be true, since `prompt()` is only ever called from `newRepositoryConfig()` which checks for this first.

cmd/repository.go
```diff
@@ -0,0 +1,686 @@
...
+		return
+	}
+
+	// List all repositories given a client instantiated about config.
+	rr, err := client.Repositories().All()
+	if err != nil {
+		return
+	}
+
+	// Print repository names, or name plus url if verbose
```
lance, line 392: My inclination is that the remote URL should always be shown, and that `-v` could provide some additional information. I'm not sure what that would be - perhaps what languages are available? Not sure.  Just a thought... at this point usability concerns are mostly speculation, but imo it would be nice to always see the URL.

lance (commented):
Looks great - small small wording suggestions and other nits

cmd/repository.go
```diff
@@ -0,0 +1,686 @@
...
+	the Boson Project Hello-World template without installing the template
+	repository locally, use the --repository (-r) flag on create:
+		$ func create -l go \
+			--template hello-world \
+			--repository https://github.com/boson-project/templates
+
+COMMANDS
+
+	With no arguments, this help text is shown.  To confirm repository management
+	command to run via an interactive prompt, use the use the --confirm (-c) flag.
```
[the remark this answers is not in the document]
lkingland: Definitely was having trouble with that awkward statement.

cmd/repository.go
lkingland: 👍🏻 I waffled about whether or not to leave this in there despite the redundancy, eventually opting to leave it for excessive clarity to the reader; but if you also notice this is strictly unnecessary, then that breaks my mental tie.  The check is now removed from `prompt`, presuming the caller will already have checked (even on future refactors)

lkingland:
> Looks great - small small wording suggestions and other nits

Thanks so much @lance for taking the time to read and point out the errors.  My eyes started to glaze over after too long on this PR... not useful when trying to self-proof.  All suggestions were made, along with their corollaries in commands.md

lkingland:
CI Fixed in #516 and will rebase this and re-request /lgtm when merged.

docs/guides/commands.md
```diff
@@ -220,3 +220,138 @@ Invokes interactive prompt to remove Volumes from the function configuration
 ```console
 func config volumes remove [-p <path>]
 ```
+
+## `repository`
```
zroubalik, line 224: Wouldn't it make more sense to move this under `config` instead of a top level command?

lkingland: `config` is specifically managing the config file, and is scoped to a single function.  The `repository` command is managing global settings, and is scoped to the system.
We could of course change this, but that would require a restructuring of config in addition to the creation of the `repository` command.
I am taking, as my reference, the venerable `git` command, and in particular `git remote`.  It  allows for mutation of the `gitconfig` using `git config`, and even a `--global` but is specific to values which appear in the gitconfig, which is similar to our func.yaml.

lkingland: This is probably a good time to mention that I do think we need a `func config --global` flag which mutates a ~/.config/func/config to provide global overrides to default values.

PR #523, opened 2021-09-13
chore: vendor modules

lance:
Use vendored dependencies


lance:
So, I think I am honing in on what is going on here. Something that we depend on has a dependency on `k8s.io/cli-runtime@v0.20.7` which has a dependency on `sigs.k8s.io/kustomize@v2.0.3+incompatible`. Although the error and versions are slightly different, here is an explanation provided by the kustomize developers: https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/depprobcheck/README.md#why. I'm not sure why exactly this is not a problem on my local machine, but I spun up a ubi8 instance and recreated the problem.

dsimansk:
```
        k8s.io/api v0.21.4
	k8s.io/apimachinery v0.21.4
	k8s.io/client-go v0.21.4
	knative.dev/client v0.25.1-0.20210913155632-82a21a5773be
	knative.dev/eventing v0.25.1-0.20210909163359-316e14d7fbc2
	knative.dev/pkg v0.0.0-20210909165259-d4505c660535
	knative.dev/serving v0.25.1-0.20210913112533-33aeffc6c9e2
```

So you're jumping directly to current `main` HEAD of Knative dependencies + bumping Kubernetes ones. So Knative already upgraded k8s to the above `0.21.4` that actually contains fix for "kustomize" issue, so IMO removing the `replace` directive from `go.mod` should fix the build for you.

lance:
Closing and trying again in a new PR. It looks like maybe github actions is caching some stuff. It's not entirely clear, but the action errors now say things like

> run 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory

PR #524, opened 2021-09-14
fix: prompt default options

lkingland:
- :bug: removal of repositories

Removes the default value from prompt options and adds a backup assertion on removal.

Works around a strange (bug?) behavior in the Prompt library where, if a default value for a set of Prompt options is set to the empty member, no value is permitted to pass through even though it is set as required _and_ the UI shows the first option as selected.  This bug is equivalent to a set of radio buttons showing in the UI the first element selected by default, but instead behaving as though no radios were selected on hitting submit (no initially selected radios).  This issue is an edge case and only seemed to rear its head when not interacting with the prompt in any way other than hitting enter for defaults.

The fix (workaround, really) is to remove the explicit setting of the default value to a member known to be empty.

/kind bug

### DIFF
--- a/cmd/repository.go
+++ b/cmd/repository.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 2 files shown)


lance (approved):
/lgtm
weird bug though :hankey:

Issue #512, opened 2021-09-07
`func` requires credentials on each local `func deploy` for Mac OS X  


omerbensaadon:
I literally have to type in the username and password to my personal machine each and every time

lkingland:
@matejvasek note that I am not seeing the same behavior.  My `~/.docker/config.json` is:
```
{
	"auths": {
		"https://index.docker.io/v1/": {},
		"localhost:5000": {
			"auth": "REDACTED_BUT_POPULATED"
		},
		"quay.io": {}
	},
	"credsStore": "desktop",
	"experimental": "enabled",
	"stackOrchestrator": "swarm",
	"currentContext": "default"
}
```
the registry which Omer is using when this issue is encountered is the `localhost:5000`.

matejvasek:
This shouldn't be `Mac OS X` exclusive. Users are prompted for credentials if they are not logged in via `docker login`.

omerbensaadon:
Ah, i see. I think including that help text when we prompt login will
satisfy me enough to close this issue out!

matejvasek:
I think we could save the entered credentials (or maybe first ask the user if it's OK to save it). We just need to decide where to put it, I would prefer to create our own auth file, and use `~/.docker/config.json` only as read-only.

PR #526, opened 2021-09-15
Better credentials handling

matejvasek:
# Changes

* If config file does contain credentials but they are incorrect then user will be prompted to enter correct ones.
* If user enters incorrect credentials then the user will be prompted again.

### DIFF
--- a/docker/pusher.go
+++ b/docker/pusher.go
@@ -5,15 +5,20 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"regexp"
 	"strings"
 
+	"github.com/docker/docker/errdefs"
+
+	"github.com/containers/image/v5/pkg/docker/config"
+	containersTypes "github.com/containers/image/v5/types"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
-	"github.com/pkg/errors"
+	dockerClient "github.com/docker/docker/client"
 
 	fn "knative.dev/kn-plugin-func"
 )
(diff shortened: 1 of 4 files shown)


docker/pusher.go
```diff
@@ -27,6 +31,92 @@ type Credentials struct {
...
+
+		for {
+			result, err = credentialsCallback(registry)
+			if err != nil {
+				return Credentials{}, err
+			}
+
+			err = verifyCredentials(ctx, result.Username, result.Password, registry)
+			if err == nil {
+				// TODO maybe save the credentials
```
matejvasek, line 107: @zroubalik right now we are only reading `~/.docker/config.json` and we are never writing it. I think we shouldn't modify it but rather create our own file e.g. `~/.config/func/auth.json`. Question is: if we will introduce such a file should we continue to read also `~/.docker/config.json`, or maybe just one time import it first time?

matejvasek:
relates to #512

docker/pusher.go
jrangelramos: Hello @matejvasek I think if the new file get introduced, func could still read `~/.docker/config.json` once users (along the path) may decidelogin or switch to different registries after first-time import, using docker commands or some specific auth input method depending the registry (i.e. gcr.io with gcloud auth). In this case auth creadentials for such registry can be combined with docker/podman login commands on CI systems.

jrangelramos: I think save is a good idea, otherwise it will keep asking for user/pw every time.

jrangelramos (commented):
Hello @matejvasek, one small suggestion is to r(maybe) estrict the prompt attempts. In case credentials are incorrect it requests for cred in a loop until dev enters a valid credential, and maybe throw a message on console (something as "bad or invalid credentials.") wdyt?

docker/pusher.go
lkingland: It might be nice to import the values from the docker config file at least on the first run, but that's not even necessary if we are caching the credentials.  It definitely makes sense for us not to start depending on / writing the config of other systems and instead use a config specific to func such as `~/.config/func/auth.json`

lkingland (approved):
Looks very nice!  So glad we have this sorted out now.

Let's talk about creating a `~/.config/func/auth.json` to store creds, and how that will fit into our overall config strategy, because what we decide will end up affecting a few things in here, such as how the tests are structured... but that can wait to a future PR!

lkingland:
Shucks, there I go triggering the bot again to merge, when I wanted to give Lance and Zbynek a chance to comment if they have time.

PR #529, opened 2021-09-16
Vendor modules

lkingland:
# Changes

Testing CI with Vendored Dependencies


lkingland:
Fix turns out to be a global .gitignore of all 'target' directories. I presume this was to ignore files created by running template tests, so specific (absolute pathed) ignores may need to be added individually.

PR #534, opened 2021-09-16
Setup knative/hack update-deps.sh machinery

markusthoemmes:
# Changes

This introduces `knative/hack` and the various helpers it provides. Most importantly, this allows plugging this repository into nightly dependency updates by adding `update-deps.sh`. This also bootstraps license checking and correct "license vendoring" inside the `third-party` directory.

`update-codegen.sh` is generally responsible for producing generated files. In this case, that would be `pkged.go` via `pkger` but due to having trouble getting a diff-less output, I dropped that from the Github Actions for now. Can be addressed seperately. 

I did include `update-pkger.sh` though, which uses the already vendored version of `pkger` to eliminate the unnecessary install steps in all the actions. A small caveat is that `pkger` really doesn't like the "hidden" tools.go file. That has been scripted around and should fix itself once we move to `//embed` anyway.


markusthoemmes:
/hold

To discuss the changes made with peeps a little more.

.github/workflows/pull_requests.yaml
```diff
@@ -10,66 +10,25 @@ jobs:
...
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Install Bash 4 on Mac
+        if: matrix.os == 'macos-latest'
+        run: |
+          /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+          brew update
+
+          brew install bash
+          brew install gnu-sed
```
lkingland, line 20: By requiring this update from the Mac-default BSD `sed` to `gnu-sed`, the `make` command will fail by default for mac users (developers) with a fairly cryptic message.  See the suggestion to the script below for an alternate solution that works on both platforms.

hack/update-pkger.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Hack: Remove the build tag before running pkger
+tools="${REPO_ROOT_DIR}/hack/tools.go"
+first="$(head -n1 "$tools")"
+sed -i '1d' "$tools"
+
+go run ./vendor/github.com/markbates/pkger/cmd/pkger
+
+# Hack: Restore build tag.
+echo -e "$first\n$(cat "$tools")" > "$tools"
```
lkingland, line 31: The `sed` command fails for developers with the default BSD `sed` on OSX with the somewhat cryptic error `extra characters at the end of d command`.  Perhaps we could use something which works on both platforms by default?
```suggestion
# Hack: touch a non-tagged file so that pkger doesn't complain
echo -e "package tools" > "$REPO_ROOT_DIR/hack/package.go"

go run ./vendor/github.com/markbates/pkger/cmd/pkger

# Hack: remove touched file.
rm "$REPO_ROOT_DIR/hack/package.go"
```

lkingland (commented):
Thanks for getting us up to speed with the hack infrastructure

hack/update-pkger.sh
lkingland: Alternately, we could just add the `package.go` file to the hack directory with an explanatory comment such as `// Required for pkger` or some such, and avoid the programmatic creation/removal above.  :shrug:

markusthoemmes: `gnu-sed` is required for Knative scripts anyway FWIW.

Sadly we can't just not have `tools.go` not have a build-constraint, because that then breaks the license checking tool :joy:

lkingland: We're not running the knative-X action scripts locally during development.  This proposed change does not, to my understanding, have any effect on tools.go nor the license checking tool because `tools.go` remains untouched.  Please enlighten me.

lkingland (approved):
Overall looks great.  Open question about how to not break local OSX developers' experience.  We can open that a separate discussion/pr this week, but I want to land this asap so we can move on.
/lgtm

hack/update-pkger.sh
markusthoemmes: The point with `sed`, `bash` and MacOSX is: you'll need gnu-sed and bash4+ for the Knative scripts anyway. Sure, you might somewhat be able to get around it by only using the `make` tooling, but as soon as you call `update-deps` (and you will have to do that locally), you'll need these anyway.

PR #535, opened 2021-09-16
fix: use full image names

matejvasek:
# Changes
Using full and non-docker.io names for `paketo` buildpack builders.
Useful when working with `podman` which dislike short names.


matejvasek:
@jrangelramos I don't know what are implications with regard to pull limit in `gcr.io`.

matejvasek:
This is not super important but it is more convenient for user because by default short names are forbidden in `podman`.

jrangelramos:
@matejvasek so far I have not seen pull limit issue using gcr.io. Let see how it goes along the way. /lgtm

lance (approved):
/lgtm

PR #537, opened 2021-09-17
feat: periodically update progress during build

lance:
# Changes

Some feedback from the user interviews indicated that people sometimes
get confused by how long it takes to  build a function the first time
and are tempted to kill the process. This change updates the progress
message every five seconds for twenty seconds, and if a build takes longer
than that the message remains "Still building".

Not sure if this is really needed, but it was a fun and easy fix. Happy to
use better/funnier/pithier text if you have suggestions.

/kind enhancement

### DIFF
--- a/client.go
+++ b/client.go
@@ -11,6 +11,7 @@ import (
 	"runtime"
 	"sort"
 	"sync"
+	"time"
 
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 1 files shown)


lkingland (approved):
This is a perpetual problem, and regular updates (that feel real) is a good solution.

/lgtm

PR #538, opened 2021-09-17
fix: hide a fmt.Println behind verbose flag

lance:
There was a `fmt.Println()` statement not wrapped by the verbose flag that resulted in deployments looking like this:

```
> kn func deploy
   Deploying function
🕕 Deploying function to the cluster
Function deployed at URL: http://tscript-default.lball-cluster-0c576f1
     Function deployed at URL: http://tscript-default.lball-cluster-0c576f1a70d464f092d8591997631748-0000.us-east.containers.appdomain.cloud
```


lkingland (approved):
/lgtm

The more the better!

PR #541, opened 2021-09-20
fix: update-pkger.sh sed error on osX

lkingland:
The script depends on GNU-sed command syntax, which is not installed on OS X by default (OSX uses BSD).  This update changes the solution to using a touched file rather than sed. 

A concern was raised that this solution may interfere with other automated tasks, such as the license checker, so we should watch carefully and either back it out, or propose a different solution, if that does indeed happen.

/kind bug


lance (approved):
This seems a better solution to me anyway
/approve

PR #544, opened 2021-09-20
fix: stop the progress ticker after build completes

lance:
Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/542

### DIFF
--- a/client.go
+++ b/client.go
@@ -516,27 +516,20 @@ func (c *Client) Build(ctx context.Context, path string) (err error) {
 		"This is taking a while",
 		"Still building"}
 	ticker := time.NewTicker(5 * time.Second)
-	quit := make(chan struct{})
+	defer ticker.Stop()
 	go func() {
 		for {
-			select {
-			case <-ticker.C:
-				if len(m) == 0 {
-					close(quit)
-					break
-				}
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -567,6 +567,7 @@ func (c *Client) Build(ctx context.Context, path string) (err error) {
 		message = fmt.Sprintf("Function image built: %v", f.Image)
 	}
 	c.progressListener.Increment(message)
+	close(quit)
```
zroubalik, line 570: Wouldn't it be better to use `defer` right after the definition?

lance:
Note that removing the line that prints "Deploying function" results in nicer output.

```
> kn func deploy
   🙌 Function image built: docker.io/lanceball/foobar:latest
   Function deployed at URL: http://foobar-default.lball-cluster-0c576f1a70d464f092d8591997631748-0000.us-east.containers.appdomain.cloud
>
```
vs.
```
> kn func deploy 
   Deploying function
   Function updated at URL: http://foobar-default.lball-cluster-0c576f1a70d464f092d8591997631748-0000.us-east.containers.appdomain.cloud
```

lkingland (approved):
/lgtm

PR #549, opened 2021-09-22
fix: registry URL comparison

matejvasek:
# Changes

Registries with different subdomain should be considered non-equal with exception of `docker.io` and it's subdomains.


docker/credentials_helper_test.go
```diff
@@ -2,26 +2,31 @@ package docker
...
+		{"non-matching host different sub-domain", "sub.quay.io", "quay.io", false},
+		{"non-matching host different sub-domain", "sub.quay.io", "sub3.quay.io", false},
+		{"localhost", "localhost", "localhost", true},
+		{"localhost with standard ports", "localhost:80", "localhost:443", true},
+		{"localhost with matching port", "https://localhost:1234", "http://localhost:1234", true},
+		{"localhost with match by default port", "http://localhost", "localhost:80", true},
+		{"localhost with match by default port", "https://localhost", "localhost:443", true},
+		{"localhost with mismatch by default port", "https://localhost", "http://localhost", true},
+		{"docker.io matching host", "https://docker.io", "docker.io", true},
+		{"docker.io matching host", "http://docker.io", "docker.io", true},
```
matejvasek, line 22: Are these good?

docker/credentials_helper_test.go
```diff
@@ -2,26 +2,31 @@ package docker
...
+		{"docker.io with path", "docker.io/v1/", "docker.io", true},
+		{"docker.io with protocol and path", "https://docker.io/v1/", "docker.io", true},
+		{"docker.io with subdomain index.", "https://index.docker.io/v1/", "docker.io", true},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := to2ndLevelDomain(tt.rawurl); got != tt.want {
+			if got := registryEquals(tt.urlA, tt.urlB); got != tt.want {
 				t.Errorf("to2ndLevelDomain() = %v, want %v", got, tt.want)
```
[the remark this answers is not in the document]
matejvasek: yes, the logic changed quite a bit, now it compares two urls.

docker/credentials_helper_test.go
```diff
@@ -2,26 +2,31 @@ package docker
...
-		{"localhost", "localhost", "localhost"},
-		{"2nd level with protocol", "https://docker.io", "docker.io"},
-		{"2nd level with path", "docker.io/v1/", "docker.io"},
-		{"2nd level with port", "docker.io:80", "docker.io"},
-		{"2nd level with protocol and path", "https://docker.io/v1/", "docker.io"},
-		{"3rd level with protocol and path", "https://index.docker.io/v1/", "docker.io"},
-		{"3rd level with protocol and path and port", "https://index.docker.io:80/v1/", "docker.io"},
-		{"localhost with protocol and path and port", "http://localhost:8080/v1/", "localhost"},
+		{"no port matching host", "quay.io", "quay.io", true},
+		{"non-matching host different sub-domain", "sub.quay.io", "quay.io", false},
```
lance, line 13: Really small nitpick here, but maybe make this descriptor slightly different than the one below where there are two different subdomains.
```suggestion
		{"non-matching host added sub-domain", "sub.quay.io", "quay.io", false},
```

docker/credentials_helper_test.go
lance: As above, a few very minor nitpicks to keep test case descriptions unique.
```suggestion
		{"localhost with standard ports", "localhost:80", "localhost:443", true},
		{"localhost with matching port", "https://localhost:1234", "http://localhost:1234", true},
		{"localhost with match by default port 80", "http://localhost", "localhost:80", true},
		{"localhost with match by default port 443", "https://localhost", "localhost:443", true},
		{"localhost with mismatch by default port", "https://localhost", "http://localhost", true},
		{"docker.io matching host https", "https://docker.io", "docker.io", true},
		{"docker.io matching host http", "http://docker.io", "docker.io", true},
```
As for whether or not 80 == 443, I guess it makes sense if the registry service upgrades/downgrades automatically. But I'm not completely sure of the intent here. Maybe some comments in the test code would be helpful.

docker/credentials_helper.go
```diff
@@ -72,18 +73,47 @@ func GetCredentialsFromCredsStore(registry string) (types.DockerAuthConfig, erro
...
+		port = u.Port()
+	} else if u.Scheme == "http" {
+		port = "80"
 	}
-	return parts[len(parts)-2] + "." + parts[len(parts)-1]
+	return
+}
+
+// checks whether registry matches in host and port
+// with exception where port 80 and 443 are considered equal
```
lance, line 102: I don't really understand why the two different ports are considered equal.

lance (commented):
Looks good other than a few minor nitpicks, and a question about HTTP/HTTPS equivalence.

docker/credentials_helper.go
matejvasek: I also not 100% sure about this. My idea here is that if you are running registry on `80` and also `443` then most likely the two sockets are serving the same registry. But if you run registry on `5000` and `5001` then most likely the two socket are serving two distinct registries.

matejvasek: but maybe it would make more sense to make `80` and `443` equivalent not to each other but to empty port `""`.

lance: > but maybe it would make more sense to make `80` and `443` equivalent not to each other but to empty port `""`.

This makes a bit more sense to me.

matejvasek:
@lance PTAL

lance (commented):
/lgtm

PR #551, opened 2021-09-22
avoid unintentional image/registry overwrite

### LINKED ISSUE #527: The `FUNC_REGISTRY` environment variable overrides image in `func.yaml`.
The `FUNC_REGISTRY` environment variable should be only used if image registry is not explicitly set in `func.yaml`.

Right now the envvar overrides image registry in `func.yaml` upon build. So if the envvar points to let say `quay.io` and user wants to change that in `func.yaml` to `docker.io`, it's not possible because the image will be overwritten.

shinigambit:
# Changes

- :bug: Apply registry change only when specified by command line flag (viper merges registry command line flag and env variable)

Fixes #527

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -131,6 +131,13 @@ func runBuild(cmd *cobra.Command, _ []string, clientFn buildClientFn) (err error
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


lance (commented):
/lgtm

PR #557, opened 2021-09-24
Update Quarkus version in templates

matejvasek:
# Changes

- :broom: updated Quarkus version in templates


matejvasek:
This requires update Java to 11, but:
1. User still can downgrade to 1.13 in `pom.xml`.
2. Buildpacks contains Java 11 so `func build` will work (only local build with `./mvn` might not).

lance (commented):
/lgtm

PR #560, opened 2021-10-01
Save password after user input

matejvasek:
# Changes

- :gift: Saving docker registry password


matejvasek:
I have no idea why the test is failing. It cannot be even build `[build failed]`. Any idea what can be causing that? It works locally on my machine.

matejvasek:
Oh CI don't use go1.16.

lance (approved):
/lgtm

lance:
@matejvasek do you want to unhold this now that it's passing?

matejvasek:
@lance @zroubalik @lkingland @omerbensaadon 
Is it acceptable that this is saving passwords in plaintext (`docker` on Linux does it that way by default), or do we need OS keyring all the time (or whenever possible)?

omerbensaadon:
I would assume if docker for Linux does it that way we are OK to copy that pattern... 

If we can avoid that for MacOS that would be a nice-to-have in my opinion. 

Paging @julz and @evankanderson from the Security WG!

matejvasek:
I could add support for docker helpers (binaries with name like: `docker-credential-*`), that should work by default on macOS, it's not difficult to implement. Or I could try create our own solution using some library like https://github.com/zalando/go-keyring, so we are not depended on docker helpers.

lance:
> Is it acceptable that this is saving passwords in plaintext
IMO no, it's not - even if Docker does it this way.

matejvasek:
> Agree, implementing it via helpers would be much better.

Should I depend on binary on PATH (user have to install it by themselves), or should I embed helpers code?

zroubalik:
> > Agree, implementing it via helpers would be much better.
> 
> Should I depend on binary on PATH (user have to install it by themselves), or should I embed helpers code?

Is it exposed as a golang dependency? Then embedding makes sense, we should avoid additional installation steps and configuration as much as possible.

matejvasek:
@zroubalik @lance @lkingland PTAL

matejvasek:
@jrangelramos could you please try this out on Windows?

cmd/deploy.go
```diff
@@ -216,6 +222,34 @@ func newCredentialsCallback() func(registry string) (docker.Credentials, error)
 	}
 }
 
+func newChooseHelperCallback() docker.ChooseCredentialHelperCallback {
+	return func(availableHelpers []string) (string, error) {
+		if len(availableHelpers) < 1 {
+			fmt.Fprintf(os.Stderr, `There is no credential helper available.
+As a result credentials won't be saved.
+To fix this consider installing docker credential helper https://github.com/docker/docker-credential-helpers.
```
lance, line 230: When does the user experience this? IIUC, the use case would be if they do not have auth credentials in a typical docker config location, so we need to interactively prompt for credentials. Is that right?

If that's right, maybe we can just state it as a recommendation vs. what almost sounds like a warning. 

```suggestion
			fmt.Fprintf(os.Stderr, `Credentials will not be saved.
If you would like to save your credentials in the future,
you can install docker credential helper https://github.com/docker/docker-credential-helpers.
```

cmd/deploy.go
```diff
@@ -216,6 +222,34 @@ func newCredentialsCallback() func(registry string) (docker.Credentials, error)
...
+		err := survey.AskOne(&survey.Select{
+			Message: "Choose credentials helper",
+			Options: append(availableHelpers, "None"),
+		}, &resp, survey.WithValidator(survey.Required))
+		if err != nil {
+			return "", err
+		}
+		if resp == "None" {
+			fmt.Fprintf(os.Stderr, `You haven't chosen credential helper.
+As a result credentials won't be saved.
```
lance, line 245:
```suggestion
			fmt.Fprintf(os.Stderr, "No helper selected. Credentials will not be saved.\n")
```

lance (commented):
Generally lgtm with a couple of small wording suggestions

docker/credentials_helper.go
```diff
@@ -4,53 +4,80 @@ import (
...
+	"io/ioutil"
 	"net"
 	"net/url"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
+	"github.com/docker/docker-credential-helpers/credentials"
+
```
zroubalik, line 16:
```suggestion
```

client.go
```diff
@@ -13,6 +13,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/mitchellh/go-homedir"
+
```
zroubalik, line 17:
```suggestion
```

lance:
@googlebot I consent

zroubalik:
@matejvasek nevermind, I'll build it and let you know later today.

docker/credentials_helper.go
```diff
@@ -70,7 +96,59 @@ func GetCredentialsFromCredsStore(registry string) (types.DockerAuthConfig, erro
...
+		if err != nil {
+			continue
+		}
+		for _, fi := range fss {
+			if fi.IsDir() {
+				continue
+			}
+			if !fi.Mode().IsRegular() {
+				continue
+			}
```
zroubalik, line 134: Working for me if we remove this, my docker-credentials helper is a symlink and thus it is not being resolved correctly.

PR #561, opened 2021-10-05
chore: remove dead code

lance:
:broom: The default builders configured in builder.go were not only wrong, but also not even used anymore. Removing them.

The rationale for their existence is so that language pack authors don't have to specify anything in manifest.yaml if they are OK with the defaults. But this is a stretch, in my opinion. First of all, what should the defaults be? I suggest that they should not be the boson builders. If we really think there should be default builders, I don't think there should be one for each runtime. I think there should just be a single default builder that is capable of building all kinds of projects.

For example, the paketo builder. However, the paketo builder on its own isn't actually enough for some runtimes. The go runtime, for example, has a small buildpack that contains the invocation code, overlaid on the function project at build time. My guess is that most compiled languages will have something similar. 


zroubalik (approved):

zroubalik:
e2e tests are failing

Shashankft9:
should the boson builder be removed from the trusted builders here? https://github.com/knative-sandbox/kn-plugin-func/blob/31a29a7963011f3fd7a6e5b6e623f229b096a7e9/buildpacks/builder.go#L103

lance:
> should the boson builder be removed from the trusted builders here?

Good catch! But I think probably not. We do still want boson builders to be considered trusted in the context of a buildpack build, even if we don't use them as the defaults.

PR #563, opened 2021-10-05
Fix deployment on k8s cloud providers

shinigambit:
# Changes

- :bug: fix auth deployment issue on k8s cloud providers

/kind dependencies

**Release Note**

```release-note
Deployments to GKE clusters failed due to an authorization issue, that required the presence of the auth plugin from
the client-go dependency to work properly, this plugin includes auth plugins for other providers like Azure and Openstack
so authorization against these services should be handled covered by this change too
```


lance (commented):
/lgtm

evankanderson:
You're right, I had an older version of the plugin when I made this change.

/close

Issue #497, opened 2021-08-30
Ease usage of private docker repository


matejvasek:
In order to use private repository like quay.io user has to set secrets to cluster I wish there was some easier way.

jrangelramos:
I think this can be configured directly on the cluster, for example on Openshift we can "teach" the cluster how to pull from private repos using secrets.

matejvasek:
@jrangelramos yes that's true, but we don't make easy for user, if user forgets to set the secret on the cluster then deployment fails and use has to search for the cause.

matejvasek:
I updated description and title.

shinigambit:
@matejvasek I wonder if you have any ideas in mind on how the usage of this would be. I was thinking:
- A new command in `func` like maybe `func config-registry`
- A flag in `func deploy` that would create/update and configure the pull secrets in the cluster
- A dynamic check at deploy time that detects that the pod's status is ImagePullBackoff and suggests the preferred way of configuring the secrets

There's probably more ways to achieve this but I'd like to know your thoughts.
Also I can see this being useful not only for functions. Maybe it's worth moving this to `kn` client?

matejvasek:
I don't have specific idea. For starer just some check that image is private and subsequent warning with link to documentation should be enough.

shinigambit:
Thanks @matejvasek , I think I can tackle this one.
/assign

PR #579, opened 2021-10-11
Detect and report unreachable image deployment

leetcope:
# Changes

- Copy feature from knative-sandbox/kn-plugin-func/pull/551 to avoid image repo overwrite at deployment
- Consolidate imports of "knative.dev/serving/pkg/apis/serving/v1" into a single import
- 🧹 Detect and surface private repository deployment issue #497

### DIFF
--- a/knative/deployer.go
+++ b/knative/deployer.go
@@ -142,7 +142,7 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (result fn.Deploym
 				}
 			}
 			if presumePrivate {
-				err := fmt.Errorf("it seems your docker registry is private, make sure you have set up pull secrets https://knative.dev/docs/developer/serving/deploying-from-private-registry")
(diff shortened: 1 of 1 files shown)


knative/deployer.go
```diff
@@ -80,7 +110,41 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (result fn.Deploym
...
+					if private {
+						presumePrivate = true
+						break main
+					}
+				case err = <-cherr:
+					break main
+				}
+			}
+			if presumePrivate {
+				err := fmt.Errorf("it seems your docker registry is private, make sure you have set up pull secrets https://knative.dev/docs/developer/serving/deploying-from-private-registry")
```
lance, line 145: I really like this PR, but I'm not sure if we should say this so concretely. It's really only an assumption on our part. Would something like this be better?
```suggestion
				err := fmt.Errorf("your function image is unreachable. It is possible that your docker registry is private. If so, make sure you have set up pull secrets https://knative.dev/docs/developer/serving/deploying-from-private-registry")
```

lance (commented):
Not sure about the wording, but generally lgtm

lance (commented):
/lgtm

PR #580, opened 2021-10-12
src: update codegen

lkingland:
# Changes

- :broom:  ./hack/update_codegen.sh


lkingland:
@markusthoemmes or @dsimansk , I have isolated the changes in the PR to somethinge more debuggable: nothing beyond running `update-codegen.sh` on HEAD.   I am sure this is user error. I would love to get your input on why this verify step is failing.  I am running `./hack/update-codgen.sh` locally with the following output:
```
❯ pwd
/Users/me/src/knative.dev/kn-plugin-func
❯ echo $GOPATH
/Users/me
❯ ./hack/update-codegen.sh
=== Update Deps for Golang
--- Go mod tidy and vendor
--- Removing unwanted vendor files
--- Updating licenses
--- Removing broken symlinks
❯ git status --porcelain
❯ 
```

markusthoemmes:
Can you share some more info on your Golang env? `go env` for example would be neat. You might also want to try to update your local version of github.com/google/go-licenses, if you have that preinstalled.

matejvasek:
I tried this on `fat` FS on Fedora and files were regenerated. Also wouldn't this also hit developers with Windows (if there are any on Knative project).

matejvasek:
@lkingland you could try mount some `ntfs` (it's case sensitive) formatted USB flash drive and try running `go-licences ...` there.

PR #581, opened 2021-10-12
Set name and email of git user

matejvasek:
# Changes

Configure name and email in GitHub Action.


markusthoemmes:
knative-sandbox is the canonical one.

PR #585, opened 2021-10-13
feat!: rename `emit` to `invoke` and default to local

lance:
# Changes

This commit renames 'func emit' command to 'func invoke' and makes the default behavior to send an event to localhost. The special '--sink' value 'local' is changed to 'cluster' to indicate that the function should be invoked on the cluster instead of locally. 

All other behavior remains the same.

BREAKING CHANGE

/kind enhancement

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/576
Partial fix: https://github.com/knative-sandbox/kn-plugin-func/issues/575

**Release Note**

```release-note
The 'emit' command has changed to 'invoke' and the default has changed to local invocation, with the special sink value 'local' changed to 'cluster' to invoke the function on the cluster.
```


cmd/invoke.go
```diff
@@ -14,11 +14,11 @@ import (
 )
 
 func init() {
-	root.AddCommand(NewEmitCmd(newEmitClient))
+	root.AddCommand(NewInvokeCmd(newInvokeClient))
 }
 
 // create a fn.Client with an instance of a
-func newEmitClient(cfg emitConfig) (*fn.Client, error) {
+func newInvokeClient(cfg emitConfig) (*fn.Client, error) {
```
zroubalik, line 21: Don't we want to rename `emitConfig -> invokeConfig` and the same for `emit-` properties etc? So there is not this mixed naming in this file.

lance:
I have moved `cloudevents/emitter[_test].go` to `invoker/cloudevents[_test].go` and have introduced an `invoker/http.go` with no implementation so far. The idea is to support function invocation with either events _or_ HTTP. But that has opened up some questions for me.

I would like to be smart about what the default behavior is. For example, if it's a CloudEvent function, it would be nice to invoke with an event, but if it's an HTTP function the default should be a simple HTTP POST. However, we don't currently store the template name in `func.yaml`. I think we used to, but there was no need for it to be there, so it was removed. I can't decide how I feel about reintroducing this configuration value. Any thoughts on this? 

If we don't want to reintroduce the `template` value in `func.yaml`, is it enough to simply default to HTTP (the default function with `create` is HTTP), and have a flag for events? `kn func invoke --event` Or vice versa, `kn func invoke --http`? The `invoke` command currently supports lots of flags for specifying event data. Maybe we default to HTTP unless any one of these flags is set? :shrug: 

@lkingland @matejvasek @zroubalik wdyt?

lance:
Handing this baton to @lkingland

PR #590, opened 2021-10-15
Enable gcp auth in kn func

evankanderson:
I was trying to use `kn func` with a GKE cluster, and got an error message that the `gcp` auth provider wasn't available. This is a simple library inclusion that registers the GCP auth provider.


shinigambit:
@evankanderson is your plugin up to date? It started working for me after the changes made in #563 for authentication in gcp and other providers

evankanderson:
You're right, I had an older version of the plugin when I made this change.

/close

PR #591, opened 2021-10-15
feat: add explicit probe definitions to manifest and func.yaml

evankanderson:
In building https://github.com/evankanderson/klr-buildpack, I ended up with an HTTP server that only wanted to respond to POST requests. This meant that it never became healthy with an `httpGet` readiness check, and deployment timed out.

This PR add/replaces the string-mapped health probes with a structured set of `corev1.Probe` definitions.

﻿Convert to sigs.k8s.io/yaml to be able to use json annotations on corev1.Probe so that empty Probe fields aren't serialized.


client.go
```diff
@@ -192,10 +193,16 @@ type Subscription struct {
...
 
 type Manifest struct {
-	Name            string            `yaml:"name"`
-	Buildpacks      []string          `yaml:"buildpacks"`
-	HealthEndpoints map[string]string `yaml:"healthEndpoints"`
-	Builders        map[string]string `yaml:"builders"`
+	Name            string            `json:"name"`
+	Buildpacks      []string          `json:"buildpacks"`
+	HealthEndpoints map[string]string `json:"healthEndpoints"`
+	Probes          *LifecycleProbes  `json:"lifecycle"`
```
matejvasek, line 199: How are `Probes` related to `HealthEndpoints`?

knative/deployer.go
```diff
@@ -224,6 +224,12 @@ func setHealthEndpoints(f fn.Function, c *corev1.Container) *corev1.Container {
 	c.ReadinessProbe = probeFor(READINESS_ENDPOINT)
 
 	// If specified in func.yaml, the provided values override the defaults
+	if f.Probes != nil {
+		// Prefer full probe specifications to HTTP URL specfications (more flexibility)
+		c.LivenessProbe = f.Probes.LivenessProbe
+		c.ReadinessProbe = f.Probes.ReadinessProbe
+		return c // Skip HealthEndpoints if Lifecycle status is defined.
+	}
 	if f.HealthEndpoints != nil {
```
lance, line 233: If a language pack has both `HealthEndpoints` and `Probes` set, then `HealthEndpoints` here will override the `Probes`.

config.go
```diff
@@ -27,9 +27,9 @@ var (
 
 type Volumes []Volume
 type Volume struct {
-	Secret    *string `yaml:"secret,omitempty" jsonschema:"oneof_required=secret"`
-	ConfigMap *string `yaml:"configMap,omitempty" jsonschema:"oneof_required=configmap"`
-	Path      *string `yaml:"path"`
+	Secret    *string `json:"secret,omitempty" jsonschema:"oneof_required=secret"`
```
zroubalik, line 30: It would be better if this change from `yaml` to `json`all across this PR would be done in a seperate PR, with some description and explanation. If I am not mistaken it is not relevant to probes.

client.go
evankanderson: I'd be happy to consolidate this with #558; I wasn't able to find an overarching design for what manifest / runtime files should look like, so I added this here but wasn't super-happy with it.

evankanderson: Let's land #558 and then I'll extend on that.

I'd sort of like to be able to include most of a PodTemplateSpec in the `runtime.yaml` and/or `manifest.yaml` so that runtimes could advise on things like Env vars (e.g. set `_HANDLER` to `lambda_function.lambda_handler` in https://github.com/evankanderson/klr-buildpack/blob/master/buildpack/bin/build#L31

config.go
zroubalik: That would be great!

PR #592, opened 2021-10-19
Documentation for buildEnv

pmgk07:
# Changes

Add documentation for buildEnv variable in func.yaml

### DIFF
--- a/docs/guides/func_yaml.md
+++ b/docs/guides/func_yaml.md
@@ -28,6 +28,26 @@ field will contain all of the available builders for a given runtime. Although
 it's typically unnecessary to modify the `builder` field, using values from
 `builders` is OK.
 
+### `buildEnvs`
+This field allows you to set environment variables available to the builder/buildpack that builds the function. This environment variable is NOT set at runtime, use [envs](#envs) instead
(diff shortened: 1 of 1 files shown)


docs/guides/func_yaml.md
```diff
@@ -28,6 +28,26 @@ field will contain all of the available builders for a given runtime. Although
...
+
+```yaml
+buildEnvs:
+- name: EXAMPLE1                            # (1) env variable directly from a value
+  value: value
+- name: EXAMPLE2                            # (2) env variable from a local environment value
+  value: '{{ env:LOCAL_ENV_VALUE }}'
+```
+
+For example, the below func.yaml snippet modifies the default golang buildpack to build source code with 1.15 compiler version. Refer to respective buildpack documentation to know more about environment variables that modify behavior of the `func build`.
```
zroubalik, line 44:
```suggestion
For example, the below `func.yaml` snippet modifies the default Golang buildpack to build source code with 1.15 compiler version. Refer to respective buildpack documentation to know more about environment variables that modify behavior of the `func build`.
```

zroubalik (commented):
Looking good, just a minor nit.

PR #594, opened 2021-10-19
Support ssh docker daemon connection

matejvasek:
# Changes

- :gift: Support connection to a remote docker daemon via SSH.
Users now can use docker host like `export DOCKER_HOST="ssh://user@host"`.


lkingland (approved):
/lgtm

PR #599, opened 2021-10-21
feat: use Paketo builders and Paketo community buildpack for rust template

Shashankft9:
# Changes
- Uses same paketo builder images as go, nodejs and typescript
- Uses a paketo composite rust buildpack: https://github.com/paketo-community/rust
- Adds a Procfile in template which is used by the buildpack to set the process. 

/kind enhancement

Fixes #
An attempt to make rust runtime vendor neutral


lance:
Thanks for the contribution @Shashankft9!

matejvasek:
@Shashankft9 I know it's been long time, but don't you remember why the procfile was needed? It seems buildpack can somehow determine process.

matejvasek:
> @Shashankft9 I know it's been long time, but don't you remember why the procfile was needed? It seems buildpack can somehow determine process.

I can only guess that the old version of BP did not know automatically determine process.

matejvasek:
I guess the need for Procfile was removed by https://github.com/paketo-community/cargo/pull/116/

Shashankft9:
@matejvasek it was because of this issue: https://github.com/paketo-community/rust/issues/115

> I guess the need for Procfile was removed by https://github.com/paketo-community/cargo/pull/116

interesting, I will try and see if it works - thanks.

PR #614, opened 2021-10-29
chore: adding goimports checks

jrangelramos:
# Changes

- :broom: Adding goimports check to Makefile

This is a suggestion to include goimports similar to what is executed on gitactions. This way contributors may save some time using it prior to push their PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,9 @@ PKGER ?= ./hack/update-pkger.sh
 CODE := $(shell find . -name '*.go') pkged.go go.mod schema/func_yaml-schema.json
 TEMPLATES := $(shell find templates -name '*' -type f)
 
+# Code for goimports (checkstyle)
+CODECHECK := $(shell find . -path './vendor' -prune -o -path './third_party' -prune -o -name '*.pb.go' -prune -o -name 'wire_gen.go' -prune -o -type f -name '*.go' -print)
+
 # Default Targets
 all: build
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -59,9 +62,15 @@ check: bin/golangci-lint ## Check code quality (lint)
 	./bin/golangci-lint run --timeout 300s
 	cd test/_e2e && ../../bin/golangci-lint run --timeout 300s
 
+checkstyle: tools/goimports ## Check code style (imports and formatting)
```
lkingland, line 65: There is already a directory for binaries necessary for testing: `./bin`.  Perhaps use that directory rather than a new `tools` directory?

Makefile
```diff
@@ -30,6 +30,9 @@ PKGER ?= ./hack/update-pkger.sh
 CODE := $(shell find . -name '*.go') pkged.go go.mod schema/func_yaml-schema.json
 TEMPLATES := $(shell find templates -name '*' -type f)
 
+# Code for goimports (checkstyle)
+CODECHECK := $(shell find . -path './vendor' -prune -o -path './third_party' -prune -o -name '*.pb.go' -prune -o -name 'wire_gen.go' -prune -o -type f -name '*.go' -print)
```
lkingland, line 34: Please help me understand:  Why is this set of files different than the set of files already used for checks (the CODE variable above)?

Perhaps the correct solution here is to use the CODE variable above, and modify it as necessary to include any of the additional `find` options that you see as necessary for running code checks.

After all, why would we "codechec" files that are not "code" used elsewhere?

Please let me know if you don't think this would make sense for any reason, but perhaps ti would be better to use the CODE files above, and augment that `find` command with the conditions of omitting the `vendor` directory, protobuffs and wire_gen?

Makefile
```diff
@@ -59,9 +62,15 @@ check: bin/golangci-lint ## Check code quality (lint)
...
 	cd test/_e2e && ../../bin/golangci-lint run --timeout 300s
 
+checkstyle: tools/goimports ## Check code style (imports and formatting)
+	@goimports -d $(CODECHECK)
+
 bin/golangci-lint:
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ./bin v1.40.1
 
+tools/goimports:
+	@go get golang.org/x/tools/cmd/goimports
```
lkingland, line 72: Does this `go get` command create `tools/goimports`?  If not, I am unclear how this works as a Make target.  The idea is that the make target (in this case `tools/goimports`) executes if `tools/goimports` does not exist.  The idea being that it only fires once per call to `make checkstyle`

lkingland (changes requested):

PR #622, opened 2021-11-03
fix: Environment values -> Environment variables

zroubalik:

# Changes

fix: Environment values -> Environment variables

I've just recently noticed that if you run `func config` it will give you option to configure `Environment values` not `Environment variables` as it should be and it is being used in the rest of the commands. 

So fixing this typo which happened I don't know why 🤷‍♂️


matejvasek:
/hold
until test pass

PR #623, opened 2021-11-03
fix: ConfigMap/Secret key validation

zroubalik:

# Changes

As per: https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-object
> Each key under the data or the binaryData field must consist of alphanumeric characters, -, _ or .

Our validation only allowed alphanumeric and underscore before, fixing this.


lance (commented):
/lgtm

PR #631, opened 2021-11-04
[WIP] SpringNative func template

### LINKED ISSUE #630: Support func Spring Native template
Add a template for Spring Boot with Spring Native to leverage GraalVM runtimes.

salaboy:
# Changes
- :gift:  Add the template for `cloudevents` using Spring Native and the packeto `paketobuildpacks/builder:tiny` builder. 

/kind enhancement 

Fixes #630


salaboy:
@lance  this is quite working but I am having trouble setting BuildEnvs in the manifest.yaml file -> https://github.com/knative-sandbox/kn-plugin-func/pull/631/files#diff-28190f6036cdaf54ab876572d9b9d6ef126ce167b164c127143ba9a6cce333ddR4

That variable is not propagated to `func.yaml` and that variable needs to be there for the builder to be found (i think) can you give me some pointers on where to look at? Is this the correct place to put those buildEnvs ?

templates/springnative/cloudevents/README.md
```diff
@@ -0,0 +1,26 @@
+# Java Native Image Sample Application
+
+## Building
+
+### With `pack`
+```bash
+pack build applications/native-image \
+  --builder paketobuildpacks/builder:tiny \
+  --env BP_NATIVE_IMAGE=true
+```
```
lance, line 10: This should probably be changed to show usage with `func`, shouldn't it?

templates/springnative/cloudevents/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+### With the Spring Boot Maven Plugin
+```
+./mvnw -Dmaven.test.skip=true spring-boot:build-image
+```
+
+## Running
+```bash
+docker run --rm --tty --publish 8080:8080 applications/native-image
```
lance, line 19: Why not `func run`?

templates/springnative/cloudevents/README.md
```diff
@@ -0,0 +1,26 @@
...
+```
+
+## Viewing
+
+```bash
+curl -s http://localhost:8080/actuator/health | jq .
```
lance, line 25: How about `func emit -k local` (which will soon be `func invoke`)?

templates/springnative/cloudevents/src/main/java/functions/SpringCloudEventsApplication.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private static final Logger LOGGER = Logger.getLogger(
+      SpringCloudEventsApplication.class.getName());
+
+  public static void main(String[] args) {
+    SpringApplication.run(SpringCloudEventsApplication.class, args);
+  }
+
+  @Bean
+  public Function<Message<Input>, Output> uppercase(CloudEventHeaderEnricher enricher) {
```
lance, line 31: I recommend adding a lot of comments here, and in the function body indicating where devs should add their code, and what the parameters to the function are, etc. If you don't @omerbensaadon will mention it. :wink:

templates/springnative/cloudevents/manifest.yaml
```diff
@@ -0,0 +1,5 @@
+builders:
```
lance, line 1: You don't need this file. These values are inherited from the language level `manifest.yaml`

lance (commented):
@salaboy oh this is a use case for `buildEnv` that we had not anticipated, but of course that 
 makes sense. Currently, those values are only settable by the end user in `func.yaml` and are not read from the `manifest.yaml` files on project creation. IIUC this env var will cause a native build. But without it, the build should work, right? Just not native? If that's the case, for the time being it might be OK to just  put that information in the README.md for the function template - at least temporarily. Currently the Quarkus templates can be built either natively or to JVM, and it's up to the user to decide which: https://github.com/knative-sandbox/kn-plugin-func/tree/main/templates/quarkus/cloudevents#building

If it works like this for Spring then you could have instructions in the README.md showing the user how to add the env var.

I do think we should add this capability to the binary though. Roughly, that would live here: https://github.com/knative-sandbox/kn-plugin-func/blob/f6a4ef21ac75ba1d536c771d1026773273445c38/templates.go#L141-L152

salaboy:
@lance a bit of change in the approach here.. I think that I will close this PR that adds a completely new template with 90% overlap with the spring boot and I will make a separate PR to make the spring-boot one use spring-native and document how to disable it if needed. do you have any objections to that plan?

lance:
Sounds good 👌

PR #632, opened 2021-11-05
Do not print header when `func list` does not find any functions

nak3:
`func list` prints the header when it gets `No functions found`.
This patch stops displaying the header.

ASIS
```
$ func list
No functions found
NAME  NAMESPACE  RUNTIME  URL  READY
```

The header `NAME  NAMESPACE  RUNTIME  URL  READY` should not be printed.

TOBE (by this patch)
```
$ func list
No functions found
```

/kind cleanup


lance (commented):
/lgtm
/approve

PR #639, opened 2021-11-08
MethodNotAllowed added for rust cloudevents template

Shashankft9:
# Changes
- Adds response `MethodNotAllowed` for a GET request on `/` 

/kind enhancement

Fixes #
Noticed that it has already been taken care of by go cloudevents sdk, but has to be done separately for actix in rust. Without this change, rust cloudevents func gives 404.


lance (approved):
/lgtm

PR #644, opened 2021-11-09
feat: allow build to be triggered from run when fn.Image is missing

### LINKED ISSUE #577: Func run should ask the user if they want to build
# Context
User interviews:
Users always used the `kn func run` command after the `kn func create` command, inevitably it returned an error message and forced the user to go do the 'kn func build" command first. This experience shows that it is adding an extra step to the users.

# Story 

**As Deandra the Developer**
**I want** to have `kn func run` build the func project for me
**So that** I don't have to go and run the `kn func build` command separately. 

# Acceptance Criteria

- [ ] Run command asks the users to do the func project build.
```
GIVEN I am running my function locally
WHEN I am using run command  `kn func run`
AND my function is not built yet
THEN  CLI ask me if I want to build my function . `Looks like the function has not been built yet, would you like to build it now?`
AND my function get built and then run showing me a message `Hello from func`
```

shinigambit:
# Changes

- :broom: A prompt will ask users if they want to build the current function when an image has not been found
- :broom: Build error messages are shown to the user

/kind enhancement

Fixes #577

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -121,6 +121,9 @@ type runConfig struct {
 
 	// Envs passed via cmd to removed
 	EnvToRemove []string
+
+	// Perform build if function hasn't been built yet
+	Build bool
 }
 
 func newRunConfig(cmd *cobra.Command) (c runConfig, err error) {
@@ -130,6 +133,7 @@ func newRunConfig(cmd *cobra.Command) (c runConfig, err error) {
 	}
 
 	return runConfig{
+		Build:       viper.GetBool("build"),
 		Path:        viper.GetString("path"),
 		Verbose:     viper.GetBool("verbose"), // defined on root
 		EnvToUpdate: envToUpdate,
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -19,6 +19,7 @@ func init() {
 }
 
 func newRunClient(cfg runConfig) *fn.Client {
+	bc := newBuildConfig()
 	runner := docker.NewRunner()
 	runner.Verbose = cfg.Verbose
 
@@ -30,10 +31,9 @@ func newRunClient(cfg runConfig) *fn.Client {
 	return fn.New(
 		fn.WithBuilder(builder),
 		fn.WithProgressListener(listener),
-		fn.WithRegistry(cfg.Registry),
+		fn.WithRegistry(bc.Registry),
 		fn.WithRunner(runner),
-		fn.WithVerbose(cfg.Verbose),
-	)
+		fn.WithVerbose(cfg.Verbose))
 }
 
 type runClientFn func(runConfig) *fn.Client
@@ -114,7 +114,9 @@ func runRun(cmd *cobra.Command, args []string, clientFn runClientFn) (err error)
 		}
 		err := survey.Ask([]*survey.Question{&qs}, &ans)
 		if err == nil && ans.Build {
-			client.Build(cmd.Context(), config.Path)
+			if err := client.Build(cmd.Context(), config.Path); err != nil {
+				return err
+			}
 		} else if err != nil {
 			fmt.Printf("Function will not be built: %v\n", err)
 		}
@@ -135,12 +137,6 @@ type runConfig struct {
 
 	// Envs passed via cmd to removed
 	EnvToRemove []string
-
-	// Required for Build to be triggered from Run
-	Registry string
-
-	// Required for Build to be triggered from Run
-	Builder string
 }
 
 func newRunConfig(cmd *cobra.Command) (c runConfig, err error) {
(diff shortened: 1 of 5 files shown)


shinigambit:
currently working on adding unit tests to this

cmd/run.go
```diff
@@ -87,6 +100,27 @@ func runRun(cmd *cobra.Command, args []string, clientFn runClientFn) (err error)
...
 	client := clientFn(config)
 
+	if !function.Built() {
+		ans := struct {
+			Build bool
+		}{}
+		qs := survey.Question{
+			Name: "Build",
+			Prompt: &survey.Confirm{
+				Message: "Looks like the function has not been built yet (building can take a few minutes), would you like to build it now?:",
```
zroubalik, line 110:
```suggestion
				Message: "Looks like the function has not been built yet (building can take a few minutes), would you like to build it now?",
```

Nit

lkingland (commented):
This looks good overall, but I do have one suggestion:

Commands should complete; with or without an error message by default.  When a command _blocks_ awaiting user input _by default_, it inhibits their ability to be scripted or to be chained into more complex patterns; a standard practice in *nix systems (and arguably a key reason the command line has stood the test of time.)

A command should, by default, either succeed or fail immediately, with an appropriate error (exit) code, and an informative message explaining how the user can avoid the failure in the future.  By blocking on user input by default, scripts will hang, combined commands (`cmdX && comY`) will not complete, and when not in an interactive terminal (if the command is ever used in a background task, in CI/CD, etc.) there will be no indication of why the command never compelted unless the sysadmin checks logs to find the command timed out pending input.   Command should fail or succeed by default.

User prompting (the confirmation prompts) is a useful upgrade to the user experience which provides a guided experience for more simplistic usage, and is therefore behind the `--confirm` (`-c`) flag in other commands.  This allows for a nice upgrade path for the user:

"Your command failed because of X.  Either fix X by providing flagy Y, or run with --confirm to be guided through the process".

I would prefer we followed the same pattern here, where the command "fails fast" with a helpful error message by default:
"Can not run because this function has not been built.  Use the `--build` flag to build and run.  Use the `--confirm` flag to complete this command using an interactive prompt"

This is the pattern in the other commands which support `--confirm`, and I think it would be appropriate to emulate that pattern here as well.

You can see an example of this paradigm in the `create` and `deploy` commands, for reference!

I defer to the hive mind, but that is my gentle suggestion.

cmd/run.go
```diff
@@ -51,6 +63,7 @@ kn func run
 			"You may provide this flag multiple times for setting multiple environment variables. "+
 			"To unset, specify the environment variable name followed by a \"-\" (e.g., NAME-).")
 	cmd.Flags().StringP("path", "p", cwd(), "Path to the project directory (Env: $FUNC_PATH)")
+	cmd.Flags().BoolP("build", "b", false, "Build the function only if the function has not been built before")
```
lkingland, line 66:
```suggestion
	cmd.Flags().BoolP("build", "b", false, "Build the Function if it has not already been built")
```

cmd/run.go
```diff
@@ -87,6 +100,16 @@ func runRun(cmd *cobra.Command, args []string, clientFn runClientFn) (err error)
...
+	if !function.Built() {
+		build, err := cmd.Flags().GetBool("build")
+		if err == nil && build {
+			if err := client.Build(cmd.Context(), config.Path); err != nil {
+				return err
+			}
+		} else if err != nil {
+			fmt.Printf("Function will not be built: %v\n", err)
+		}
+	}
```
lkingland, line 112:
```suggestion
	if config.Build && !function.Built() {
		if err = client.Build(cmd.Context(), config.Path); err != nil {
			return
		}
	}
```

cmd/run.go
lkingland: Since some of these commands have quite a few options, we have been bundling them up into a "config" object, and doing all the err checks, nil checks, and validation there.  That way, when it's time to actually use the values, we can just refer to `config.X`.  This seems to have cleaned things up a lot, especially in these more complex commands.

Therefore, it might be nice to add a `.Build` boolean member to the RunConfig object, populated when `newRunConfig(cmd)` is called on line 76

lkingland (commented):
This is coming along nicely!  I had a few suggestions on organization and style that might be helpful.  Not blockers.

cmd/run.go
lkingland: Since some of these commands have quite a few options, we have been bundling them up into a "config" object, and doing all the err checks, nil checks, and validation there.  That way, when it's time to actually use the values, we can just refer to `config.X`.  This seems to have cleaned things up a lot, especially in these more complex commands.

Therefore, it might be nice to add a `.Build` boolean member to the RunConfig object, populated when `newRunConfig(cmd)` is called on line 76

lance (commented):
/lgtm

shinigambit:
An e2e test seems to be failing due to a "Temporary failure in name resolution". Is it possible to rerun the test?

lance:
@shinigambit it seems that @matejvasek and @lkingland each had PRs that conflicted with the other resulting in an e2e test that was failing. Let's let them work this out and hopefully a rebase should resolve the problem.

lance (commented):
/lgtm

zroubalik (commented):
/approve
/lgtm

Issue #553, opened 2021-09-23
`func create` should require "-l" flag; be strict about list of supported runtimes


omerbensaadon:
# Context
Many users did not expect `func create` to automatically create a Node HTTP project, the easiest way to resolve this is to simply require the -l flag when running `func create`

There may be a follow-on story to this one which asks the user if they want to save the language they specify as the default language, but that behavior is out of scope for this implementation. 

# Story 

**As Deandra the Developer**
**I want** func create to require me to specify a language
**So that** func does not assume I am a Node developer 
**And so that** I can see what languages func supports 

# Acceptance Criteria
- [ ] Reject `func create` without a the -l flag 
```
GIVEN I type out "func create someName" without an "-l" flag 
WHEN I hit enter 
THEN func returns an error which tells me I need to specify the -l flag
AND THEN func returns the list of currently supported runtimes
```

- [x] Reject non-supported runtimes
```
GIVEN I type "func create someName -l {someRuntime}"
AND {someRuntime} is not currently supported by func
WHEN I hit enter 
THEN func returns an error which tells me the runtime I specified is not currently supported
AND THEN func returns the list of currently supported runtimes
```

lance:
AC #2 fixed in 
```
> ./func create -l asdf foo
Error: The language runtime 'asdf' is not recognized.
Available language runtimes are:
  go
  node
  nodejs
  python
  quarkus
  ruby
  rust
  springboot
  typescript
```

lkingland:
I fundamentally have concerns about having required flags.  The other name for a flag is an "option".  They are definitionally optional.

The correct solution to this problem is to either 1) change this flag to an argument:  `func create node myFunction`, in which case the argument is required.  Or to 2) very visibly notify the user of the chosen default, and instruct them how to change this default in the future:
```
$ func create myFunc
You are creating a "node" Function.  To create a Function of a different Language Runtime, use the --language (-l) flag.  To change the global default language used when creating Functions and suppress this message, use:

   func config set language node --global

Created node Function in [path]
```

#1 is easier to implement.  #2 seems a bit friendlier (and is the direction `git` has gone).  I am in the process of implementing a global config system, at which time we can implement #2 (and is the reason I set myself as assigned to this issue).  I am open for suggestions!

Shashankft9:
[...]

PR #647, opened 2021-11-10
making language runtime a required flag

Shashankft9:
# Changes

- :broom: add check to make language runtime a mandatory flag 

/kind enhancement

Fixes #
fixing AC1 of https://github.com/knative-sandbox/kn-plugin-func/issues/553


Shashankft9:
cc @lkingland

Shashankft9:
its ready for reviews now - thanks

cmd/create.go
```diff
@@ -315,6 +324,22 @@ func isValidTemplate(client *fn.Client, runtime, template string) bool {
 	return false
 }
 
+// noRuntimeError creates an error stating that the language flag
+// is required, and a verbose list of valid options.
+func noRuntimeError(client *fn.Client) error {
+	b := strings.Builder{}
+	fmt.Fprintf(&b, "Required flag \"language\" not set.\n")
+	fmt.Fprintln(&b, "Available language runtimes are:")
```
zroubalik, line 332: As I mentioned before, we should really try to offer available runtimes in a prompt and let user select and proceed.

lkingland (changes requested):
I have two concerns here.  I think we should leave the defaults in place at the library level.  As a user of the client library, defaulting as much as possible is preferable as it eases basic usage, tests etc. and the default is clearly visible in code/docs.    I would therefore consider leaving in the defaults, and tests, in the core client library and instead focus this change in the CLI.

Since this is a CLI, user-facing request, based on UX feedback, I would recommend we keep this as a CLI change: simply remove the default from the flag.  If in "confirm" mode they get a prompt, and if in non-confirm-mode they get an error stating language is required.

Secondly, I am a bit hesitant to create "required flags", because they are by definition optional.  If it's required, it should probably be an argument.

cmd/create.go
lance: I'm not opposed to that, but I think the solution that @lkingland and @Shashankft9 came up with in [this exchange](https://github.com/knative-sandbox/kn-plugin-func/issues/553#issuecomment-966952816) is good.

lance (commented):
/lgtm

cmd/create.go
zroubalik: OK, so then I am probably the only one who think that actually offering a runtime and proceed with the selection instead of an error msg is a better from UX (or more user friendly) 🤷‍♂️ I can live with that :)

lance: @zroubalik I see your point - your comment in 633 is convincing.

Shashankft9: so how do we plan to go forward with this? :)
I like what @zroubalik mentions, but I think we should go ahead with the way it is right now, with confirm and non-confirm modes kept separate, and since we are anyway going to come back to this particular issue as described in #553 by @lkingland after the work in `func config`, we can take another stab at this if there is an ask, what do you think?

lkingland: Well I think that _eventually_ we want to have a nice config option, but for now we can go ahead with this if everyone else agrees this is a good step in the right direction.

lkingland:
> I have two concerns here. I think we should leave the defaults in place at the library level. As a user of the client library, defaulting as much as possible is preferable as it eases basic usage, tests etc. and the default is clearly visible in code/docs. I would therefore consider leaving in the defaults, and tests, in the core client library and instead focus this change in the CLI.
> 
> Since this is a CLI, user-facing request, based on UX feedback, I would recommend we keep this as a CLI change: simply remove the default from the flag. If in "confirm" mode they get a prompt, and if in non-confirm-mode they get an error stating language is required.
> 
> Secondly, I am a bit hesitant to create "required flags", because they are by definition optional. If it's required, it should probably be an argument.

Just to follow up on this a bit.  Since we agree the config option is a good final destination, I am completely good with this being our path forward right now, if we can simply error on a missing required flag by default, but add it to the prompts when in "confirm" mode like the other commands.

lkingland (approved):
/lgtm

Per our lengthy discussion, this is a great step in the right direction.

PR #648, opened 2021-11-10
feat: automatically start podman service

matejvasek:
If following conditions are met:
* the DOCKER_HOST environment variable  is empty
* socket doesn't exist at default location
* `podman` is on `$PATH`
*  we are on Linux

Then temporary start `podman` service and `func` uses the service.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -1,18 +1,44 @@
 package docker
 
 import (
+	"context"
+	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"syscall"
+	"time"
 
 	"knative.dev/kn-plugin-func/ssh"
 
 	"github.com/docker/docker/client"
 )
 
-func NewDockerClient() (dockerClient client.CommonAPIClient, dockerHost string, err error) {
+func NewDockerClient(defaultDockerHost string) (dockerClient client.CommonAPIClient, dockerHost string, err error) {
+	var _url *url.URL
+
 	dockerHost = os.Getenv("DOCKER_HOST")
-	_url, err := url.Parse(dockerHost)
+
+	if dockerHost == "" && runtime.GOOS == "linux" && podmanPresent() {
+		_url, err = url.Parse(defaultDockerHost)
+		if err != nil {
+			return
+		}
+		_, err = os.Stat(_url.Path)
+		switch {
+		case err != nil && !os.IsNotExist(err):
+			return
+		case os.IsNotExist(err):
+			dockerClient, dockerHost, err = newDockerClientWithPodmanService()
+			return
+		}
+	}
+
+	_url, err = url.Parse(dockerHost)
 	isSSH := err == nil && _url.Scheme == "ssh"
 
 	if !isSSH {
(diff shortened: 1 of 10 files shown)


matejvasek:
@knative-sandbox/kn-plugin-func-approvers just an idea how to make user life easier.

docker/docker_client.go
```diff
@@ -1,18 +1,43 @@
...
+	"syscall"
+	"time"
 
 	"knative.dev/kn-plugin-func/ssh"
 
 	"github.com/docker/docker/client"
 )
 
-func NewDockerClient() (dockerClient client.CommonAPIClient, dockerHost string, err error) {
+func NewDockerClient(defaultDockerHost string) (dockerClient client.CommonAPIClient, dockerHost string, err error) {
```
lkingland, line 20: Minor nit, but since this is in the `docker` package, this could just be `docker.NewClient` and even `defaultHost` etc.

docker/docker_client.go
```diff
@@ -47,3 +72,53 @@ func NewDockerClient() (dockerClient client.CommonAPIClient, dockerHost string,
...
+
+	// give a time to podman to start
+	for i := 0; i < 40; i++ {
+		if _, e := dockerClient.Ping(context.Background()); e == nil {
+			break
+		}
+		time.Sleep(time.Millisecond * 250)
+	}
+
+	return
```
lkingland, line 112: Maybe I am misreading this, but if it pings 40 times and never breaks, the value of `err` is not returned because it's using the local variable `e` for ping errors, so it will exit having never been able to successfully ping the service... but also with no error.  Is that the intention?  Or should there be an error returned if not able to ping podman after the 40 tries?

docker/docker_client_linux_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+	"time"
+
+	"knative.dev/kn-plugin-func/docker"
+)
+
+func TestNewDockerClientWithAutomaticPodman(t *testing.T) {
```
lkingland, line 16: Please, if it's not to much trouble, can we please include at least a sentence comment on all tests?  it _really_ helps readers like me who are not familiar with the intention.  This is the "root" of this PR:  the statement of purpose.  Along with the test implementation itself this constitutes the expressed purpose of the whole feature, so having an English-natural-language description is not just helpful, it should really be the most basic expectation.

docker/docker_client_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/docker/docker/client"
+	"knative.dev/kn-plugin-func/docker"
+)
+
+func TestNewDockerClient(t *testing.T) {
```
lkingland, line 18: Just a little comment goes a long way!

lkingland (approved):
I think this is a very nice addition!

I only had one question about an error message, and a cordial request to state the purpose of the two tests in a comment.

matejvasek:
@lkingland please re-LGTM

lance (commented):
/lgtm

jrangelramos:
Pretty nice addition, and it works great!

PR #650, opened 2021-11-11
chore: remove redundant combined workflow

lkingland:
# Changes

- :broom: Removes redundant workflow

This removes the parent workflow which was recently separated into separate individuals:

- build.yaml
- check.yaml
- schema.yaml
- test-e2e.yaml
- test-integration.yaml
- test-podman.yaml
- test-unit.yaml

/kind chore


lance:
This explicitly removes release-please before we've adopted the knative release mechanisms. Shouldn't this be a part of that work?

lance:
> This explicitly removes release-please before we've adopted the knative release mechanisms. Shouldn't this be a part of that work?

That said - if release-please isn't even working at this point, why bother keeping it.

/lgtm
/hold for others to chime in on this

zroubalik (commented):
/lgtm

lkingland:
> > This explicitly removes release-please before we've adopted the knative release mechanisms. Shouldn't this be a part of that work?
> 
> That said - if release-please isn't even working at this point, why bother keeping it.
> 
> /lgtm /hold for others to chime in on this

I am pretty sure that release-please is over in `ci.yaml` and this only removes the workflows which run on PRs.  We should be GTG with that conciern there unless I am missing something?  I did only do a visual scan of the removed file to validate all functionality was preserved in the other (new) workflows.  I definitely could be missing it!

lkingland:
> That said - if release-please isn't even working at this point, why bother keeping it.

My earlier comment notwithstanding (I don't think this PR actually removes release-please), I would still prefer to keep release-please if we can.  With disciplined commit commentary, it is handy.   As we discussed in our recent call; perhaps there is a way to keep it _and_ generally adopt the knative tooling?

PR #653, opened 2021-11-11
chore: test helpers relocation

lkingland:
# Changes

- :broom: test helpers relocation

The knative automations run test with all tags set simultaneously.  This can cause collisions when using tags to define disjunct test sets.  This commit moves all helper functions in the `function_test` package into a file with no
build tags such that any combination of tags will work without collisions.

/kind chore


matejvasek:
Wouldn't it be better to have it in a separate package like `test_helpers`?
Then we could use it in packages other than `function` without possible cyclical references.

zroubalik:
> Wouldn't it be better to have it in a separate package like `test_helpers`? Then we could use it in packages other than `function` without possible cyclical references.

+1

lkingland:
> > Wouldn't it be better to have it in a separate package like `test_helpers`? Then we could use it in packages other than `function` without possible cyclical references.
> 
> +1
Sure thing.

My three comments are: first, it is generally preferable to do a small amount of copying rather than create a small dependency.  See the [Go Proverbs](https://go-proverbs.github.io/) which include _"A little copying is better than a little dependency."_ (and its [associated talk](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=568s)).  Secondly, it is usually a red flag when one creates a "util" or "helpers" package, as this is often an indication that things can be either restructured or simplified.  This suggestion creates what amounts to a "util" package.  And third, I am not sure how we could ever encounter "cyclical references" because the core would never import `function_test`, but maybe I am not understanding that concern.

Those three comments notwithstanding, this really is "right on the edge" in my mind of being a worthwhile package! It could really help ease the creation of Function tests, which may be in different packages themselves.  So, since both of you seem to think that is a good idea, I made that change.  Thanks for the suggestion.

According to Go Conventions, package names should ideally be single nouns, with no mixedCaps or-hyphens.  And since this is essentially an extension to the functions available to packages which are "testing" (all functions within this new package will include `t.Helper()`, I have named this new package `testing` and imported it into the importing package's namespace directly (`import .`) such that we don't have to prefix every call (`test_helpers.mkdir()`, `test_helpers.using()` ... etc.).

I always love when feedback causes me to change my stubborn mind.  Thank you for taking the time!  Please let me know if you have any more suggestions, or if the update is not what you were expecting.

matejvasek:
@lkingland  "A little copying is better than a little dependency."  -- I agree on this for external dependencies. But it's absolutely unnecessary to have `using()` function copied several times in our project.

OK, maybe "cyclical references"  won't happen, but when the functions are in `_test`  they wouldn't be usable from another package.

matejvasek:
`testing` is good name it's not util nor helper so it ok right :smiley:

PR #654, opened 2021-11-11
chore: build tag consistency

lkingland:
- :broom: adds missing build tag to some unit test files

The integration tests and unit tests are currently expected to be disjunct sets.  While we may change to being additive in the future, this change makes us consistent by adding the integration exclusion tag to a few non-integration test files where it was currently missing.


lance (commented):
/lgtm

PR #655, opened 2021-11-11
chore: prefer slices over plural types

lkingland:
- :broom: prefer slices over plural types unless type is needed for method receivers

Minor change to prefer slices of objects over a type to indicate plurality, as it is my understanding that this is a Go idiom, and is a helpful indicator to the developer that the type can be treated as a slice (for example, that `range` statements are available, `len` etc).  Therefore, in places where a type exists only to add plurality (there are no associated methods with the slice type as its receiver), this has been collapsed into the slice form.


lance (commented):
/lgtm
/hold since pkged.go should probably not be in this PR should it @lkingland ?

zroubalik (commented):
/lgtm

lkingland:
> /lgtm /hold since pkged.go should probably not be in this PR should it @lkingland ?

Well, not in principle no, but a workflow was complaining about pkger not adding the new build header to pkged.go.  So it's been added to shuush the test.

``` go

+ //go:build !skippkger
  // +build !skippkger

```

PR #656, opened 2021-11-11
chore: extract config validate

lkingland:
- :broom: extract config validation error text formatter

Minor cleanup to address config validation which is quite user friendly, but whose code was a little bit unwieldy.  
In particular the code now fails fast on errors reading func.yaml, extracts the unmarshaling error formatter, extracts logical validator, collapses validation string formatting logic a bit and switches to a string builder for clarity.

### DIFF
--- a/config.go
+++ b/config.go
@@ -167,86 +167,53 @@ func newConfig(root string) (c Config, err error) {
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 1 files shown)


zroubalik (commented):
/lgtm

nice improvement!

PR #660, opened 2021-11-15
feat: add POD_NAME as an environment variable

lance:
# Changes

:gift: This change adds POD_NAME as an environment variable for a deployed
function using the Kubernetes downward API. An alternative means of exposing
this information would be through the filesystem. For example, a file at
`/etc/func/pod_name` would contain the pod's name. I chose to use the env
as the place to surface this because it was easier to implement, and likely
easier to consume for a function developer.

The motivation for this change is that the Node.js and TypeScript invocation
runtime will be exposing telemetry/metrics through Prometheus at a `/metrics`
endpoint (see: https://github.com/boson-project/faas-js-runtime/pull/109).
The OpenTelemetry specification has some experimental labels for function
runtime environments which don't specifically mention pod name (the spec
is not unique to Kubernetes), but do imply the need for a unique identifier
such as this.

Ref: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/faas.md

/kind enhancement


lkingland (approved):
/lgtm

PR #663, opened 2021-11-16
chore: use buildEnvs instead of custom buildpack

lance:
# Changes

- :broom: Use Paketo build environment variables to replace custom TypeScript buildpack

The TypeScript templates have been using a small buildpack from the boson project. The sole purpose of the buildpack was to run the `npm run build` command so that a TypeScript project would be transpiled before deployment.

The Paketo buildpack for Node.js supports using environment variables at build time to enable additional scripts to be run, such as `build`. Prior to the addition of `buildEnvs` by @salaboy in
https://github.com/knative-sandbox/kn-plugin-func/pull/646 it was not possible to set this value for TypeScript projects. That new feature allows us to stop using the small buildpack from boson.

/kind techdebt


lkingland (approved):
/lgtm !

PR #664, opened 2021-11-17
feat: function version migrations

### LINKED ISSUE #674: Support pre-v0.19 Functions
Functions Created prior to v0.20 (current `main` HEAD) will yield the error "No initialized Function at <path>"

We are now using a Created timestamp to positively identify the point in time at which a new Function is initialized, rather than testing a side-effect.  This is an upgrade that has been in the works for a while, but we will need to support Functions which were created prior to this change going into effect.

lkingland:
- :gift: function version migrations

Introduces the concept of Function versions and associated migrations.  Includes a single migration which initializes version and creation fields for Functions <=v0.19.0.

PR #641 takes advantage of function creation timestamp to determine if a Function has been initialized.  Functions created prior to this stamp being introduced (forthcoming v0.20.0) should have their Creation stamp initialized to maintain backwards compatibility.

Migrations are applied when a Function is instantiated (read from func.yaml) in the order in which they are registered.

Currently the only migration is that which takes a Function from unversioned to versioned, and initialized a creation timestamp.  The version written to func.yaml is the latest migration applied, indicating migrations can be skipped on subsequent loads until a new migration is added to the codebase.

Fixes #674 

/kind enhancement


testdata/migrations/README.md
```diff
@@ -0,0 +1,4 @@
+# Migrations
+
+Contains Functions created with earlier versions to ensure backwards
+compatiblity and test migrations where applicable.
```
lance, line 4: The Knative reviewdog :dog: won't pass without correcting this nit.
```suggestion
compatibility and test migrations where applicable.
```

function_migrations.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// Migrate applies any necessary migrations, returning a new migrated
+// version of the Function.  It is the caller's responsibility to
+// .Write() the Function to persist to disk.
+func (f Function) Migrate() (migrated Function, err error) {
+	// Return immediately if the Function indicates it has already been
+	// migrated.
+	if f.Migrated() {
+		return f, nil
```
lance, line 16: Would it be worth testing this scenario with a `func.yaml` that already has a version?

function_migrations_unit_test.go
```diff
@@ -0,0 +1,88 @@
...
+	// codebase is considered fully migrated, for purposes of this version of func
+	v0 := semver.New(latestMigrationVersion())
+	v0.BumpMajor()
+	f = Function{Version: v0.String()}
+	if !f.Migrated() {
+		t.Fatalf("Function with version %v should be considered migrated when latest known by this codebase is %v", f.Version, latestMigrationVersion())
+	}
+}
+
+// TestMigrateVersions ensures that functions have migration apply the version
```
lance, line 46: Nit
```suggestion
// TestMigrate ensures that functions have migration apply the version
```

lance (commented):
This is a nice addition. Thanks!
/lgtm

function_migrations.go
lkingland: Yes this is probably a good idea.  After considering it a bit, this really boils down to .Migrate needing to be idempotent.
This boiling down can be bubbled up to asserting that .Write should also be idempotent.  Therefore I added a test which makes this expectation explicit, and it flushed out some needed tweaks to Function initial values in the process.  Thanks for the suggestion.

lkingland:
Tests have (thankfully, though a bit begrudgingly) uncovered a logical inconsistency in Function instantiation.  Marking this as draft until I can get it sorted out.

zroubalik (commented):
/lgtm

Looking good!

PR #670, opened 2021-11-18
fix: use specific version of paketo builder

matejvasek:
For some reason our node function is not working with latest base builder.


matejvasek:
@lance PTAL

lance:
/lgtm

But I want to hold until we figure out what's actually the problem
/hold

lance:
This is a temporary fix that will need to be reverted. But for now...
/unhold

PR #677, opened 2021-11-26
fix: schema-generate should point to `function.go`

zroubalik:

func.yaml config is now being generated from `function.go` and not `config.go` as it was before. Therefore we need to change the target for schema generation.


lkingland (approved):
Good catch.  I updated the schema generator itself but didn't see that make target!

/lgtm

PR #679, opened 2021-11-28
Clean up tests

evankanderson:
I started trying to forward-port #591 and got sidetracked by migrating to `sigs.k8s.io/yaml`, which was causing a test to fail. So I cleaned up the tests, and will send the `sigs.k8s.io/yaml` change next.

# Changes

- switch `reflect.DeepEqual` to `cmp.Diff` for better readability
- some `cmp.Diff` arguments were reversed(!)
- coverted one test into a table test
- cleaned up some misleading error messages
- preferred `testing.Error` to `testing.Fatal` to enable collecting more errors at once

/kind cleanup

### DIFF
--- a/function_migrations_unit_test.go
+++ b/function_migrations_unit_test.go
@@ -10,35 +10,38 @@ import (
 // TestMigrated ensures that the .Migrated() method returns whether or not the
 // migrations were applied based on its self-reported .Version member.
 func TestMigrated(t *testing.T) {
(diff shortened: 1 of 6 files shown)


lkingland (approved):
Thanks for the contribution!  The new dependency of the `cmp` package does look a bit nicer than the rather rough-edges of `reflect`

/lgtm

PR #684, opened 2021-11-29
Switch yaml library to `sigs.k8s.io/yaml`

evankanderson:
Convert to sigs.k8s.io/yaml, to make it easier to embed k8s objects in YAML.

This will enable a re-spin of #591, to allow template to specify all probe parameters.

This depends on #679 (merged)


matejvasek:
@evankanderson some tests are failing.

matejvasek:
@evankanderson speaking of yaml parser: 
Do you know if there is any way we could preserve comments?

PR #685, opened 2021-11-29
WithConfigPath option

lkingland:
## Changes

- :gift: Adds the WithConfigPath option.
- :broom: Passes config path into credentials providers, making global config path now configurable per client instance
- :broom: Refactors instantiation to be more clear about when the values are defaulted and then applied
- :broom: Creates the defined config and repositories directory separately, so they can be in different locations

Exploratory work using instance-based accessors for config and repositories paths rather than package static accessors.

### DIFF
--- a/client.go
+++ b/client.go
@@ -26,40 +26,7 @@ const (
 	// DefaultVersion is the initial value for string members whose implicit type
 	// is a semver.
 	DefaultVersion = "0.0.0"
-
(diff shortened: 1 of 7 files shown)


lkingland:
Closing in favor of a hybrid where Config is package static and Repositories is instance based with a static convenience accessor for the default state.

PR #690, opened 2021-11-30
fix: make registry validation work again

lance:
# Changes

In https://github.com/knative-sandbox/kn-plugin-func/commit/ccf00152be0ceba1794267f8e03a09cb32fee514
the behavior of `NewFunction()` changed slightly, such that it no longer actually
works without a provided and valid path. This was not caught in any of the
tests, because it's the CLI code that was making this call, and the end to end
tests that exercise the build command, use the `--registry` flag, causing
the repository validation code path to be slightly different.

:broom: This commit just adds the path so that `func build` without a `--registry` flag will work again.

EDIT: Converting to a draft, because really, any registry should be validated even if it's provided as a CLI flag. I will add that validation and a test.

/kind fix

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -92,11 +92,10 @@ func ValidNamespaceAndRegistry(path string) survey.Validator {
 			return errors.New("Value is required")
 		}
 
-		_, err := fn.DerivedImage("", val.(string)) //image can be derived without any error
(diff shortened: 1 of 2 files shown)


lkingland (approved):
Thanks for the fix!

/lgtm

PR #694, opened 2021-12-01
[WIP] Checking if build is true before prompting or initializing

### LINKED ISSUE #693: `func deploy` should only prompt build questions if `build` argument is set to `true`
Currently, even if you disable the build option for `func deploy` it still prompt questions related to the build phase. 

https://github.com/knative-sandbox/kn-plugin-func/blob/c2e1b769cc0d80082e6ef1be9ad2785ba8db4550/cmd/deploy.go#L114

I think this is just a bug.

salaboy:
# Changes
- :bug: Checks the build configuration argument before prompting build questions or trying to check if the func project has been initialized. 

/kind bug

Fixes #693

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -111,27 +111,32 @@ func runDeploy(cmd *cobra.Command, _ []string, clientFn deployClientFn) (err err
 		return err
 	}
 
-	config, err = config.Prompt()
-	if err != nil {
-		if err == terminal.InterruptErr {
-			return nil
-		}
-		return
-	}
-
 	function, err := functionWithOverrides(config.Path, functionOverrides{Namespace: config.Namespace, Image: config.Image})
 	if err != nil {
 		return
 	}
 
-	function.Envs, err = mergeEnvs(function.Envs, config.EnvToUpdate, config.EnvToRemove)
-	if err != nil {
-		return
+	if len(function.Envs) > 0 {
+		function.Envs, err = mergeEnvs(function.Envs, config.EnvToUpdate, config.EnvToRemove)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


salaboy:
@shinigambit This is just a way of doing this.. we need unit tests for this one .. and see what other people thinks about this change.

cmd/deploy.go
```diff
@@ -111,27 +111,32 @@ func runDeploy(cmd *cobra.Command, _ []string, clientFn deployClientFn) (err err
...
+		function.Envs, err = mergeEnvs(function.Envs, config.EnvToUpdate, config.EnvToRemove)
+		if err != nil {
+			return
+		}
 	}
 
-	// Check if the Function has been initialized
-	if !function.Initialized() {
-		return fmt.Errorf("the given path '%v' does not contain an initialized function. Please create one at this path before deploying", config.Path)
+	if config.Build {
```
lkingland, line 126: This is a good idea, but I think it should actually be part of the `Prompt` method.  For two reasons, first, this causes the user to never be prompted for other attributes, such as Namespace, which are pertinent whether or not build is enabled.  Second, the purpose of this `if` statement is to adjust how `Prompt` is implemented, so it should be part of Prompt's implementation.

The way this is can be handled is by splitting Prompt into multiple sets of queries (you can see this in action in `invoke.go` which only asks for `Data` if `File` was not specified.

lkingland (changes requested):
Perhaps consider placing the logic in `.Prompt()` to avoid skipping prompts for other attributes.

lkingland:
The referenced issue was fixed in #1071

PR #696, opened 2021-12-01
Check for image sha empty before deciding to push

### LINKED ISSUE #695: `func deploy` will push the image even if the SHA in the `func.yaml` file is present
Related to #676 

`func deploy` should have a clear behaviour for deciding when to push a container image to the registry. 
Currently is always pushing: https://github.com/knative-sandbox/kn-plugin-func/blob/db9ad07c7048361946a8c7d45c549323eee44a58/client.go#L597

I do understand the need to collapse push and deploy together in benefit to a better user experience, but for integration tasks, deploy should just do a deploy. This can be achieved by a consistent set of arguments or env vars.

salaboy:
# Changes

- :broom: only push image if there is no SHA in the `func.yaml` file. This avoid always pushing on `func deploy` which for integration use cases, requires access to the docker deamon. I will be happy with a different method to achieve the same, instead of checking the image sha in the `func.yaml` file, but we need a way to have a predictable behaviour for `func deploy`. 

/kind cleanup

Fixes #695

### DIFF
--- a/client.go
+++ b/client.go
@@ -593,16 +593,17 @@ func (c *Client) Deploy(ctx context.Context, path string) (err error) {
 		return ErrNotBuilt
 	}
 
-	// Push the image for the named service to the configured registry
-	imageDigest, err := c.pusher.Push(ctx, f)
-	if err != nil {
-		return
-	}
-
-	// Record the Image Digest pushed.
-	f.ImageDigest = imageDigest
-	if err = f.Write(); err != nil {
-		return
+	if f.ImageDigest == "" {
+		// Push the image for the named service to the configured registry
+		imageDigest, err := c.pusher.Push(ctx, f)
+		if err != nil {
+			return err
+		}
+		// Record the Image Digest pushed.
+		f.ImageDigest = imageDigest
+		if err = f.Write(); err != nil {
+			return err
+		}
 	}
 
 	// Deploy a new or Update the previously-deployed Function


salaboy:
@shinigambit This is just a way of doing this.. we need unit tests for this one as well.. and see what other people thinks about this check.

zroubalik:
@salaboy what will happen if you build and push -> sha will be added to func.yaml. And then user rebuilds the function again, so we should be pushing again, but it won't happen because the old sha will be already set in func.yaml.

zroubalik:
@salaboy we should anyway replace the docker deamon for pushing, we should use go-containerregistry imho. CC @matejvasek

shinigambit:
@salaboy I can see how this could be useful in a CI environment but as @zroubalik points out this looks like it can negatively impact local development. Would adding a flag like `--skip-push` to the `deploy` command achieve the same functionality you need here or is there something I'm overlooking?

embano1:
> Would adding a flag like --skip-push to the deploy command achieve the same functionality you need here or is there something I'm overlooking?

I commented on the initial issue and IMHO a simple flag would do the job, e.g. `--push=false` (like in `ko`)

zroubalik:
> > Would adding a flag like --skip-push to the deploy command achieve the same functionality you need here or is there something I'm overlooking?
> 
> I commented on the initial issue and IMHO a simple flag would do the job, e.g. `--push=false` (like in `ko`)
> ## Update
> 
> Due to #794 I was running an outdated version (though `brew` said it's latest). Installed from latest release and see that there is a `push` flag. So please ignore my use case here since it's resolved 🙏

Yeah, `push` flag is present for both `build` and `deploy` commands :)

PR #698, opened 2021-12-02
docs: `podman`: known issue + macOS doc

matejvasek:


docs/guides/podman.md
```diff
@@ -14,3 +14,29 @@ Then set the environment variable `DOCKER_HOST` to the socket so `func` knows wh
...
+❯ ssh-add -k $HOME/.ssh/podman-machine-default
+❯ export DOCKER_HOST=(podman system connection ls --format="{{.URI}}" | grep root)
+```
+
+## Known issues
+
+* In func version 0.19, some runtimes might be unable to build a function by using podman. You might see an error message similar to the following:
+
+    ```
+    ERROR: failed to image: error during connect: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.40/info": EOF
```
lance, line 35: Does it really say `failed to image`?

docs/guides/podman.md
```diff
@@ -14,3 +14,29 @@ Then set the environment variable `DOCKER_HOST` to the socket so `func` knows wh
...
+    ```
+  * The following workaround exists for this issue:
+    * Update the podman service by adding --time=0 to the service ExecStart definition.  The configuration files are: `/etc/systemd/[user|system]/podman.service`.You might need to copy default vendor configuration files from `/usr/lib/systemd/[user|system]/podman.service` first.
```
lance, line 38:
```suggestion
    * Update the podman service by adding --time=0 to the service ExecStart definition in the podman configuration file at `/etc/systemd/user/podman.service`.You might need to copy the default configuration file from `/usr/lib/systemd/user/podman.service` first.
```
Some minor wording tweaks. I wonder if it wouldn't be better to just refer to the  `user` config files, since that's probably the only thing they will need to modify.

docs/guides/podman.md
matejvasek: I copied that from osh doc.

lance: lol - that's probably a typo, but who knows!?

lance (commented):
/lgtm

docs/guides/podman.md
jrangelramos: yea.. it appears like that.. failed to image... lol

PR #701, opened 2021-12-06
fix: move integration port to unregistered range

### LINKED ISSUE #700: Integration Testing Port Conflict in MacOS
The recent addition of an AirPlay listener in MacOS Monteray utilizes local port 5000.  This conflicts with the port chosen for the local container registry used during integration testing.

We should probably move the port we use into the unregistered range.

lkingland:
:bug: default integraiton port conflict with Mac OS Monteray's new AirPlay listener

Moves the integration testing port into the unregistered range (above
49151) to minimize potential future conflicts.

/kind bug

Fixes #700


zroubalik (commented):
/approve

PR #702, opened 2021-12-06
chore: refactor CLI flags for consistency across commands

lance:
# Changes

:broom: refactor path and namespace CLI flags for consistency across commands

These flags are used by multiple commands, and explicitly setting them for
each command risks errors that may not be easily detected (see the deleted
flag text for `namespace` in deploy.go). The downside to this is that wording
is not sligtly tweaked for the command - see again some of the removed text
for `namespace` that was slightly different for each command, but usefully so.

I'm not 100% convinced this change is needed, since it does reduce the flexibility
of the user facing text. But if it had existed before we wouldn't have had
incorrect text for `deploy`.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/221

/kind chore


lkingland (approved):
I think it cleans things up a good bit!  We can always back it out later if it turns out to be too constrictive, or selectively as needed 👍🏻

PR #704, opened 2021-12-06
src: CheckAuth() calls registry directly

matejvasek:
CheckAuth() calls registry directly not using docker daemon as a middle-man.


zroubalik (commented):
/approve

Nice! I like that we are moving away from docker/cli

matejvasek:
@zroubalik would you /lgtm?

PR #705, opened 2021-12-06
feat!: invoke

lkingland:
- :gift: invoke local and remote Functions using test data

Renames `func emit` to `invoke`, expanding it to trigger both HTTP and CloudEvent Functions with detection of local vs remote targets.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/575
Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/576

/kind enhancement

```
$ func invoke --help

NAME
	func invoke - Invoke a Function.

SYNOPSIS
	func invoke [-t|--target] [-f|--format]
	             [--id] [--source] [--type] [--data] [--file] [--content-type]
	             [-s|--save] [-p|--path] [-c|--confirm] [-v|--verbose]
DESCRIPTION
	Invokes the Function by sending a test request to the currently running
	Function instance, either locally or remote.  If the Function is running
	both locally and remote, the local instance will be invoked.  This behavior
	can be manually overridden using the --target flag.
	Functions are invoked with a test data structure consisting of five values:
		id:            A unique identifier for the request.
		source:        A sender name for the request (sender).
		type:          A type for the request.
		data:          Data (content) for this request.
		content-type:  The MIME type of the value contained in 'data'.
	The values of these parameters can be individually altered from their defaults
	using their associated flags. Data can also be provided from a file using the
	--file flag.

	Invocation Target
	  The Function instance to invoke can be specified using the --target flag
	  which accepts the values "local", "remote", or <URL>.  By default the
	  local Function instance is chosen if running (see func run).
	  To explicitly target the remote (deployed) Function:
	    func invoke --target=remote
	  To target an arbitrary endpoint, provide a URL:
	    func invoke --target=https://myfunction.example.com

	Invocation Data
	  Providing a filename in the --file flag will base64 encode its contents
	  as the "data" parameter sent to the Function.  The value of --content-type
	  should be set to the type from the source file.  For example, the following
	  would send a JPEG base64 encoded in the "data" POST parameter:
	    func invoke --file=example.jpeg --content-type=image/jpeg

	Message Format
	  By default Functions are sent messages which match the invocation format
	  of the template they were created using; for example "http" or "cloudevent".
	  To override this behavior, use the --format (-f) flag.
	    func invoke -f=cloudevent -t=http://my-sink.my-cluster

EXAMPLES
	o Invoke the default (local or remote) running Function with default values
	  $ func invoke

	o Run the Function locally and then invoke it with a test request:
	  (run in two terminals or by running the first in the background)
	  $ func run
	  $ func invoke

	o Deploy and then invoke the remote Function:
	  $ func deploy
	  $ func invoke

	o Invoke a remote (deployed) Function when it is already running locally:
	  (overrides the default behavior of preferring locally running instances)
	  $ func invoke --target=remote

	o Specify the data to send to the Function as a flag
	  $ func invoke --data="Hello World!"

	o Send a JPEG to the Function
	  $ func invoke --file=example.jpeg --content-type=image/jpeg

	o Invoke an arbitrary endpoint (HTTP POST)
		$ func invoke --target="https://my-event-broker.example.com"

	o Invoke an arbitrary endpoint (CloudEvent)
		$ func invoke -f=cloudevent -t="https://my-event-broker.example.com"
```


docker/runner.go
```diff
@@ -30,122 +45,201 @@ func NewRunner() *Runner {
...
+			if err != nil {
+				return envs, err
+			}
+			if set {
+				envs = append(envs, *env.Name+"="+value)
+			}
+		}
+	}
+	if verbose {
+		envs = append(envs, "VERBOSE=true")
```
lance, line 236: Hmm - I wonder if we should consider some standard template behavior when this env is set.

docs/guides/commands.md
```diff
@@ -127,40 +127,43 @@ When run as a `kn` plugin.
...
-
-Emits a CloudEvent, sending it to the deployed function. The user may specify the event type, source and ID,
-and may provide event data on the command line or in a file on disk. By default, `event` works on the local
-directory, assuming that it is a function project. Alternatively the user may provide a path to a project
-directory using the `--path` flag, or send an event to an arbitrary endpoint using the `--sink` flag. The
-`--sink` flag also accepts the special value `local` to send an event to the function running locally, for
-example, when run via `func run`.
+## `invoke`
+
+Invokes a function. By default, a locally running instance will be preferred
```
lance, line 132: Maybe "Invokes a __running__ function."?

function.go
```diff
@@ -100,6 +100,22 @@ type Function struct {
 	// according to the client which is in charge of what constitutes being
 	// fully "Created" (aka initialized)
 	Created time.Time
+
+	// Invocation defines hints for use when invoking this function.
+	// See Client.Invoke for usage.
+	Invocation Invocation `yaml:"invocation,omitempty"`
```
lance, line 106: Ahh now I understand what "Created prior" means.

templates/go/cloudevents/manifest.yaml
```diff
@@ -0,0 +1,7 @@
+# optional.  Invocation defines hints for how Functions created using this
+# template can be invoked.  These settings can be updated on the resultant
+# Function as development progresses to ensure 'invoke' can always trigger the
+# execution of a running Function instance for testing and development.
+invocation:
```
lance, line 5: Shouldn't these defaults be added to all of the manifest.yaml files?

invoke.go
```diff
@@ -0,0 +1,178 @@
...
+	route, err := invocationRoute(ctx, c, f, target) // choose instance to invoke
+	if err != nil {
+		return err
+	}
+
+	// Format" either 'http' or 'cloudevent'
+	// TODO: discuss if providing a Format on Message should a) update the
+	// Function to use the new format if none is defined already (backwards
+	// compatibility fix) or b) always update the Function, even if it was already
+	// set. Once decided, codify in a test.
```
lance, line 61: We could ask the user if they want this value saved, iff func.yaml does not contain it or differs.

```
You provided the `http` as the format for this Function.
Would you like to save this for future invocations?
```

invoke.go
```diff
@@ -0,0 +1,178 @@
...
+			return "", err // unexpected error
+		}
+		return instance.Route, nil
+	} else { // treat an unrecognized target as an ad-hoc verbatim endpoint
+		return target, nil
+	}
+}
+
+// sendEvent to the route populated with data in the invoke message.
+// with the data from the invoke message.
```
lance, line 133: Nit
```suggestion
```

lance (commented):
Looks great! The changes as a whole will make the user experience so much better. A few nits and a question or two in comments.

client.go
```diff
@@ -548,34 +571,52 @@ func (c *Client) Create(cfg Function) (err error) {
...
+	if err := os.MkdirAll(filepath.Join(f.Root, RunDataDir), os.ModePerm); err != nil {
+		return err
+	}
+
+	gitignore := `
+# Functions use the .func directory for local runtime data which should
+# generally not be tracked in source control:
+/.func
+`
+	return os.WriteFile(filepath.Join(f.Root, ".gitignore"), []byte(gitignore), os.ModePerm)
```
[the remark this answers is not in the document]
lkingland: Thank you yes, this is correct.  I was hoping to jump straight to mutating an existing .gitignore if it exists, but this PR is big enough, so that can be offloaded as an enhancement and for now simply added to contentious files.

client.go
```diff
@@ -674,36 +715,40 @@ func (c *Client) Route(path string) (err error) {
...
+		return
+	}
 
-// List currently deployed Functions.
-func (c *Client) List(ctx context.Context) ([]ListItem, error) {
-	// delegate to concrete implementation of lister entirely.
-	return c.lister.List(ctx)
+	// Return to the caller the effective port, a function to call to trigger
+	// stop, and a channel on which can be received runtime errors.
+	return job.Port, job.Stop, job.Errors, nil
```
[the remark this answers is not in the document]
lkingland: I was on the fence about this, feeling at the time that returning three values may be cognitively less cumbersome than forcing a user of Client to internalize a whole new object.  But if you feel this way as well, that tips the balance.  I will return the Job itself.

client.go
```diff
@@ -747,13 +798,32 @@ func (c *Client) Remove(ctx context.Context, cfg Function) error {
...
+// which should be invoked.  This can be the literal names "local" or "remote",
+// or can be a URL to an arbitrary endpoint.  If not provided, a running local
+// instance is preferred, with the remote Function triggered if there is no
+// locally running instance.
+// Example:
+//  myClient.Invoke(myContext, myFunction, "local", NewInvokeMessage())
+// The message sent to the Function is defined by the invoke message.
+// See NewInvokeMessage for its defaults.
+// Functions are invoked in a manner consistent with the settings defined in
+// their metadata.  For example HTTP vs CloudEvent
```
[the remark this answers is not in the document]
lkingland: Thanks :)

cmd/create.go
```diff
@@ -56,7 +56,7 @@ NAME
 	{{.Prefix}}func create - Create a Function project.
 
 SYNOPSIS
-	func create [-l|--language] [-t|--template] [-r|--repository]
+	{{.Prefix}}func create [-l|--language] [-t|--template] [-r|--repository]
```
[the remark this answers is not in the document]
lkingland: You're right, and there are a couple of other minor modifications that really should have been split off in a separate PR.  Chief among these is the update to Run which changes it to asynchronous, and allows multiple instances.  Prerequisite to `invoke` but not strictly part of it.  I contemplated breaking this out, so you are right... but I actually already have compiled a list of 28 items which I consciously omitted inclusion in this PR to preserve terseness, and would like to just merge and be done with this thing already :)

docs/guides/commands.md
lkingland: Yes, that is better.  Modification made.

Tangential:  I think this entire document could perhaps be auto-generated by the --help text from the commands themselves.  After all, the two locations seem to be serving both the same purpose and same audience.

That would of course require some hackery, but it might be worth the effort.

templates/go/cloudevents/manifest.yaml
lkingland: Yep!

cmd/run.go
```diff
@@ -98,14 +100,34 @@ func runRun(cmd *cobra.Command, args []string, clientFn runClientFn) (err error)
...
-	return client.Run(cmd.Context(), config.Path)
+
+	// Run the Function at path
+	port, stop, runtimeErrCh, err := client.Run(cmd.Context(), config.Path)
+	if err != nil {
+		return
+	}
+	defer stop()
+
+	fmt.Fprintf(cmd.OutOrStderr(), "Function started on port %v\n", port)
```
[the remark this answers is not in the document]
lkingland: Thanks!   I am firmly of the belief that a running process should be "as quiet as pssible, but no quieter".  Echoing out the port on which the Function was started _is necessary information!_  My only qualm is that the wording of the output is not conducive to internationalization nor scropting.  I would personally prefer something like:
```
OK http://localhost:123445
```
Because that is a mere ` | awk '{print $2}'` away from being used in a larger command or script (the prefix `OK` a fairly portable ACK...) But I felt that might be going a bit too far?  After all, we seem to have by consensus chosen a generally more verbose, beginner-friendly route thus far.

cmd/invoke.go
```diff
@@ -0,0 +1,403 @@
...
+
+	Functions are invoked with a test data structure consisting of five values:
+		id:            A unique identifer for the request.
+		source:        A sender name for the request (sender).
+		type:          An application-level type for this request.
+		data:          Data (content) for this request.
+		content-type:  The MIME type of the value contained in 'data'.
+
+	The values of these parameters can be individually altered from their defaults
+	using their associated flags. Data can also be provided from file with --file.
```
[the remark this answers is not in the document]
lkingland: Interesting to consider! I have oft read "from file" as a fully formed sentence, akin to "can also be loaded from disk", but I agree this seems like domain specific language when I consider it compared directly to the complete "from a file" formulation.  Change made.

cmd/invoke.go
```diff
@@ -0,0 +1,403 @@
...
+DESCRIPTION
+	Invokes the Function by sending a test request to the currently running
+	Function instance, either locally or remote.  If the Function is running
+	both locally and remote, the local instance will be invoked.  This behavior
+	can be manually overridden using the --target flag.
+
+	Functions are invoked with a test data structure consisting of five values:
+		id:            A unique identifer for the request.
+		source:        A sender name for the request (sender).
+		type:          An application-level type for this request.
```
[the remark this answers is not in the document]
lkingland: I agree this is strange.  I was trying to clearly explain the value of this member is an application-level concern, not transport related (or any other level of stack).  Perhaps I will just remove it entirely.  The default value I think conveys fairly clearly this is arbitrary, and thus app-level.

cmd/invoke.go
```diff
@@ -0,0 +1,403 @@
...
+	o Send a JPEG to the Function
+	  $ {{.Prefix}}func invoke --file=example.jpeg --content-type=image/jpeg
+
+	o Invoke an arbitrary endpoint (HTTP POST)
+		$ {{.Prefix}}func invoke --target="https://my-event-broker.example.com"
+
+	o Invoke an arbitrary endpoint (CloudEvent)
+		$ {{.Prefix}}func invoke -f=cloudevent -t="https://my-event-broker.example.com"
+
+`,
```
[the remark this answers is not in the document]
lkingland: Thanks!

cmd/invoke.go
```diff
@@ -0,0 +1,403 @@
...
+		SuggestFor: []string{"emit", "emti", "send", "emit", "exec", "nivoke", "onvoke", "unvoke", "knvoke", "imvoke", "ihvoke", "ibvoke"},
+		PreRunE:    bindEnv("path", "format", "target", "id", "source", "type", "data", "content-type", "file", "confirm", "namespace"),
+	}
+
+	// Flags
+	cmd.Flags().StringP("path", "p", cwd(), "Path to the Function which should have its instance invoked (Env: $FUNC_PATH)")
+	cmd.Flags().StringP("format", "f", "", "Format of message to send, 'http' or 'cloudevent'.  Defaults is to auto-choose, but is required for direct invocation of remote CloudEvent Functions by URL, or for Functions created prior. (Env: $FUNC_FORMAT)")
```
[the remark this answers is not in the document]
lkingland: Prior to.... _oh look a kitty_

I got distracted looking up the version number and never finished my sentence.  I think that perhaps, since we are pre 1.0 to begin with, (and I added clear instructions for specifying message type in examples), I will just omit this caviat as it will perhaps cause more confusion than it will prevent!

docker/runner.go
lkingland: It would indeed be nice to know that all our various languages' templates treat the verbosity flag in roughly the same way we do here in the client library:  quiet, printing only the data necessary or errors, unless verbose is enabled in which case one gets a play-by-play.
Appropos is [Dave Cheney's "Let's talk about logging"](https://dave.cheney.net/2015/11/05/lets-talk-about-logging)

function.go
lkingland: Ay, that was _thoughtus interruptus_ indeed.  We do have an edge case which is backwards incompatible.

invoke.go
lkingland: I like the idea of prompting.
The conundrum for me was: what should we choose if the process is not being run in an interactive terminal, just update, making the choice by default "sticky"?  That seems like not a good idea, because side-effects.  My inital thought was that the most straightforward and documentation-friendly option might be to add a `--save` flag to `invoke` which would state "save the given format for future invocations".  This could perhaps be the value being prompted for by your suggestion (if not provided of course).

I will add this to my "punts" list now...

cmd/run.go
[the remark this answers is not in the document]
lance: Hmm - I can see your point, but I do think we're trying to optimize for beginner-friendly usability in most cases. Maybe down the road a `--quiet` or `--bare` flag could achieve a similar result?

function.go
lance: > We do have an edge case which is backwards incompatible.

We're still pre-1.0, we can do anything! bwahahahah! :wink:

invoke.go
```diff
@@ -0,0 +1,177 @@
...
+	return
+}
+
+// sendPost to the route populated with data in the invoke message.
+func sendPost(ctx context.Context, route string, m InvokeMessage, t http.RoundTripper) error {
+	client := http.Client{
+		Transport: t,
+		Timeout:   10 * time.Second,
+	}
+	resp, err := client.PostForm(route, url.Values{
```
matejvasek, line 162: This is not what I would expected here. To me this looks more like newly invented HTTP binding for CloudEvents than generic HTTP. I would expect that `data` is simply send as a body of the message and content-type is set as a http header.

matejvasek: @lkingland I know this is closed PR.

PR #707, opened 2021-12-08
Refactor credential provider

matejvasek:
# Changes

Allow to set custom credential loader.
This might be useful if the registry shares credentials with some other system, e.g. OpenShift builtin registry has the same credentials as the k8s cluster.


matejvasek:
@knative-sandbox/kn-plugin-func-approvers PTAL

matejvasek:
e2e are failing but that's not fault of this PR, the `main` is broken.

PR #711, opened 2021-12-08
feat: custom default HTTP transport

matejvasek:
# Changes

- :gift: use in cluster dialer by default:
  This sets http.DefaultTransport to our specialized RoundTripper.
  The RoundTripper first tries dial in standard way.
  If dial operation fails due to hostname resolution error the in cluster dialer will be used.

- :broom: dialer pod auto-exit:
  The pod should exit (be Completed) when func exits (even in case of SIGKILL).

### DIFF
--- a/cmd/func/main.go
+++ b/cmd/func/main.go
@@ -2,9 +2,12 @@ package main
 
 import (
 	"context"
+	"net/http"
 	"os"
 	"os/signal"
 	"syscall"
+
+	funcHttp "knative.dev/kn-plugin-func/http"
 
 	"knative.dev/kn-plugin-func/cmd"
 )
@@ -14,6 +17,11 @@ import (
 var date, vers, hash string
 
 func main() {
+
+	rt := funcHttp.NewRoundTripper()
+	http.DefaultTransport = rt
+	defer rt.Close()
+
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
--- a/http/transport.go
+++ b/http/transport.go
@@ -1,0 +1,92 @@
+package http
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net"
+	"net/http"
+	"sync"
+	"time"
+
+	"knative.dev/kn-plugin-func/k8s"
+)
+
+type RoundTripCloser interface {
+	http.RoundTripper
+	io.Closer
+}
+
+// NewRoundTripper returns new closable RoundTripper that first tries to dial connection in standard way,
+// if the dial operation fails due to hostname resolution the RoundTripper tries to dial from in cluster pod.
+//
+// This is useful for accessing cluster internal services (pushing a CloudEvent into Knative broker).
+func NewRoundTripper() RoundTripCloser {
+	d := &dialer{
+		netDialer: net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+		},
+	}
+	return &roundTripCloser{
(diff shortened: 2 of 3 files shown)


matejvasek:
@lkingland not this automatically enables things like:
`func emit --sink http://dialer-test-service.dialer-test-ns.svc`

matejvasek:
e2e are failing but that's not fault of this PR, the `main` is broken.

cmd/func/main.go
```diff
@@ -2,10 +2,13 @@ package main
 
 import (
 	"context"
+	"net/http"
 	"os"
 	"os/signal"
 	"syscall"
 
+	funcHttp "knative.dev/kn-plugin-func/http"
```
zroubalik, line 10: This imho doesn't follow Golang convetion, you can use:

```go
funchttp
fhttp
fnhttp
```

http/transport.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	return &roundTripCloser{
+		Transport: http.Transport{
+			Proxy:                 http.ProxyFromEnvironment,
+			DialContext:           d.DialContext,
+			ForceAttemptHTTP2:     true,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
```
zroubalik, line 39: Might be actually great to make those configurable. Not sure how though 😄

matejvasek: These are `Go` defaults, and they are not configurable so I think that's OK.

cmd/func/main.go
```diff
@@ -14,6 +17,11 @@ import (
 var date, vers, hash string
 
 func main() {
+
+	rt := funcHttp.NewRoundTripper()
+	http.DefaultTransport = rt
+	defer rt.Close()
```
lkingland, line 23: I am a little bit concerned with overriding the `http` package default transport, because this affects anything in the process using the `http` package's standard functions.  It also is a bit hidden, because code utilizing this feature is relying on something happening "out of band" (not immediately visible in the code making the http requests).  Finally, this is only going to affect code using the CLI library, meaning that anyone using the `func` library will not get this feature, and it is not obvious why.

I love the feature this provides, but I wonder if it's not more clear to instead explicitly define this custom transport on a client at the time it is being used.  In other workds, opt in rather than opt out.  Can't we just create a new http client with a custom RoundTripper from the `funchttp`?

lkingland (approved):
I really like the feature this provides!  My only concern is that it is being "forced" on the whole process (anyone using the default `http` pacakge commands) rather than the transport being explicitly utilized where needed.

I am interested in hearing what others think about this small point (see the more detailed question inline), but overall this is a great addition!  👍🏻

cmd/func/main.go
lkingland: /hold

lkingland: > /hold

Too late now I suppoes, the merge-bot beat me to the punch.
I am still interested to hear what others think about my concern above, but I am sure we can refactor it slightly if it ever becomes a problem.

zroubalik: @lkingland that's a really great point. You are right in my opinion.

lance: I tend to agree as well. I've added https://github.com/knative-sandbox/kn-plugin-func/issues/715 so we don't lose this.

matejvasek:
I'll prepare revert.

PR #712, opened 2021-12-08
fix: revert hostname alias

matejvasek:
The localhost alias was causing problems with new approach for registry access.
The new access is done directly not via daemon.
This renders any insecure registry exception in daemon config irrelevant.
The library facilitating registry access seems to allow insecure registry on localhost only.

### DIFF
--- a/client_int_test.go
+++ b/client_int_test.go
@@ -48,7 +48,7 @@ const (
 	// DefaultRegistry must contain both the registry host and
 	// registry namespace at this time.  This will likely be
 	// split and defaulted to the forthcoming in-cluster registry.
-	DefaultRegistry = "kind-registry:50000/func"
+	DefaultRegistry = "localhost:50000/func"
 
 	// DefaultNamespace for the underlying deployments.  Must be the same
 	// as is set up and configured (see hack/configure.sh)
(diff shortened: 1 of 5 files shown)


matejvasek:
>The library facilitating registry access seems to allow insecure registry on localhost only.

That's not entirely true, it can be somehow set but by default insecure works on localhost.

>it can be somehow set

but that would require new config option in ~/.config/func/*

matejvasek:
@lkingland PTAL

matejvasek:
>but that would require new config option in ~/.config/func/* where we put that registries are allowed to be accessed via http
then again this could lead to discrepancies with daemon :disappointed: 

Well, maybe not. It seems that docker API can tell you what registries are trusted and which are not.
The thing is that it seems that `podman` does not implement that, it returns `null`. I guess I have some work to do in `podman` upstream.

lkingland:
The change to consistently use the named registry was (apparently) necessary to enable moving off the default port 5000. This in turn is required to allow tests to run on MacOS (recent update from apple took port 5000 for AirPlay receive).

We can revert the named remote for now to get us past the problem with this new library while we look for a mutually compatible solution.

lkingland:
> This renders any insecure registry exception in daemon config irrelevant.

This helps reduce the setup complexity of i t tests as well

PR #713, opened 2021-12-09
Fix dialer test

matejvasek:
The test was flaky on KinD.

### DIFF
--- a/k8s/dialer_test.go
+++ b/k8s/dialer_test.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	"knative.dev/kn-plugin-func/k8s"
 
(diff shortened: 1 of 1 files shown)


matejvasek:
@lkingland PTAL

k8s/dialer_test.go
```diff
@@ -35,7 +37,7 @@ func TestDialInClusterService(t *testing.T) {
 
 	testingNS := &coreV1.Namespace{
 		ObjectMeta: metaV1.ObjectMeta{
-			Name: "dialer-test-ns",
+			Name: "dialer-test-ns-" + rand.String(5),
```
shinigambit, line 40: This looks good but I think maybe a `t.Cleanup` removing the namespace would be a better solution for people running the e2e suite locally otherwise they may end up with a bunch of `dialer-test-ns-***` namespaces without probably noticing

k8s/dialer_test.go
```diff
@@ -98,6 +100,8 @@ func TestDialInClusterService(t *testing.T) {
 	defer cliSet.CoreV1().Services(testingNS.Name).Delete(ctx, nginxService.Name, deleteOpts)
 	t.Log("created svc: ", nginxService.Name)
 
+	time.Sleep(time.Second * 10)
```
shinigambit, line 103: Probably could use a bit of context on why this wait is here, like `waiting for nginx service to be ready`

shinigambit (commented):
Thanks for fixing this problem, just a couple of nits

PR #717, opened 2021-12-10
feat!: use `function.knative.dev` for Functions related labels

zroubalik:

# Changes

Proposal:

We should change labels that are applied to Functions from `boson.dev` -> `function.knative.dev` (or `func.knative.dev`).

I am not sure which components are dependent on this, so we should be careful about the change.

### DIFF
--- a/knative/deployer.go
+++ b/knative/deployer.go
@@ -349,8 +349,10 @@ func updateService(f fn.Function, newEnv []corev1.EnvVar, newEnvFrom []corev1.En
 //     value: {{ env:MY_ENV }}
 func processLabels(f fn.Function) (map[string]string, error) {
 	labels := map[string]string{
-		"boson.dev/function": "true",
-		"boson.dev/runtime":  f.Runtime,
+		"boson.dev/function":           "true",
+		"boson.dev/runtime":            f.Runtime,
+		"function.knative.dev":         "true",
+		"function.knative.dev/runtime": f.Runtime,
 	}
 	for _, label := range f.Labels {
 		if label.Key != nil && label.Value != nil {
--- a/knative/lister.go
+++ b/knative/lister.go
@@ -12,8 +12,11 @@ import (
 )
 
 const (
-	labelKey   = "boson.dev/function"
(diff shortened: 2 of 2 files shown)


knative/deployer.go
```diff
@@ -349,8 +349,8 @@ func updateService(f fn.Function, newEnv []corev1.EnvVar, newEnvFrom []corev1.En
 //     value: {{ env:MY_ENV }}
 func processLabels(f fn.Function) (map[string]string, error) {
 	labels := map[string]string{
-		"boson.dev/function": "true",
-		"boson.dev/runtime":  f.Runtime,
+		"function.knative.dev":         "true",
+		"function.knative.dev/runtime": f.Runtime,
```
lance, line 353: Should these use `labelFunctionKey` and friends?

zroubalik: Yeah, we should have one place for labels. But I didn't want to do this big refactoring in this PR. Happy to do that in a follow up one, once we agree on this change.

knative/lister.go
```diff
@@ -12,8 +12,11 @@ import (
 )
 
 const (
-	labelKey   = "boson.dev/function"
-	labelValue = "true"
+	labelFunctionKey          = "function.knative.dev"
+	deprecatedFunctionKey     = "boson.dev/function"
```
lance, line 16: Minor nit - to be consistent with other naming.
```suggestion
	deprecatedLabelFunctionKey     = "boson.dev/function"
```

lance (commented):
Why is pkged.go updated? (I know that file is a nightmare)

knative/lister.go
zroubalik: good catch

zroubalik:
> Why is pkged.go updated? (I know that file is a nightmare)

I don't know 🤷‍♂️  I can omit it from the PR.

knative/lister.go
zroubalik: Fixed

lance (commented):
/approve
/lgtm

PR #718, opened 2021-12-13
feat: allow push to cluster internal registries

matejvasek:
- :gift: Allow push to cluster internal registries (registry that is not exposed via route).
   NOTE: this code doesn't take care of cluster custom certification authorities.
- :broom: Moved docker credential handling code to separate sub-package.


matejvasek:
@knative-sandbox/kn-plugin-func-approvers PTAL

docker/creds/credentials.go
```diff
@@ -0,0 +1,451 @@
...
+// This might be useful when credentials are shared with some other service.
+//
+// Example: OpenShift builtin registry shares credentials with the cluster (k8s) credentials.
+func WithAdditionalCredentialLoaders(loaders ...CredentialsCallback) Opt {
+	return func(opts *credentialsProvider) {
+		opts.credentialLoaders = append(opts.credentialLoaders, loaders...)
+	}
+}
+
+// NewCredentialsProvider returns new CredentialsProvider that tires to get credentials from docker/func config files.
```
zroubalik, line 139:
```suggestion
// NewCredentialsProvider returns new CredentialsProvider that tries to get credentials from docker/func config files.
```

docker/creds/credentials.go
```diff
@@ -0,0 +1,451 @@
...
+
+type CredentialsCallback func(registry string) (docker.Credentials, error)
+
+var ErrUnauthorized = errors.New("bad credentials")
+
+// VerifyCredentialsCallback checks if credentials are accepted by the registry.
+// If credentials are incorrect this callback shall return ErrUnauthorized.
+type VerifyCredentialsCallback func(ctx context.Context, registry string, credentials docker.Credentials) error
+
+func CheckAuth(ctx context.Context, registry string, credentials docker.Credentials, trans http.RoundTripper) error {
```
zroubalik, line 37: Do you mind adding doc for this function as well?

docker/creds/credentials.go
```diff
@@ -0,0 +1,451 @@
...
+
+	c.authFilePath = filepath.Join(fn.ConfigPath(), "auth.json")
+	sys := &containersTypes.SystemContext{
+		AuthFilePath: c.authFilePath,
+	}
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		panic(err)
+		//return result, fmt.Errorf("failed to determine home directory: %w", err)
```
zroubalik, line 183: ?

zroubalik (commented):
Great improvement! Just minor nits in comments.

Could you please make this behavior configurable (probaly in a separate PR)? So there's a global config where can user specify the preffered image registry.

lance (commented):
/approve

docker/creds/credentials.go
matejvasek: I must have forgotten this some times ago this is not new file.

matejvasek:
>It would be great if you could a little bit extend the code coverage, there are mutliple codecov annotations all of over the new code.

Those new uncovered lines are not really new. GIT just wasn't able figure code move. Few of them are really new lines but that files that are not unitested at all and never were. This PR increases coverage by 1.25%. The files in question have coverage around 80% which is IMO good enough and better than average of the project 45%.

matejvasek:
>NOTE: this code doesn't take care of cluster custom certification authorities.

In particular this means this PR doesn't resolve story of pushing into OpenShift internal registry because it uses custom CA.

PR #719, opened 2021-12-13
feat: add telemetry to Node.js and TypeScript function templates

lance:
# Changes

:gift: This commit pulls in the latest version of faas-js-runtime which supports metrics exposed at /metrics. Also, concidentally, adds CloudEvent batch support, so Node.js and TypeScript functions may be invoked with a set of more than one event simultaneously.

UPDATE:  This PR also includes some changes to `Makefile` in order to remove any stray go.sum files that end up in the go templates. And it removes a stray go.sum file that ended up in the go templates.

/kind enhancement

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/710


lance:
I'm not sure how best to address the code style issue with pkged.go

zroubalik (commented):
/approve

templates/typescript/cloudevents/src/index.ts
```diff
@@ -18,17 +18,17 @@ import { Context, Invokable } from 'faas-js-runtime';
  * See: https://github.com/knative-sandbox/kn-plugin-func/blob/main/docs/guides/nodejs.md#the-context-object
  * @param {CloudEvent} cloudevent the CloudEvent
  */
-export const handle: Invokable = function (
+const handle = function (
   context: Context,
-  cloudevent?: CloudEvent
+  cloudevent?: CloudEvent<string> | CloudEvent<Customer>
```
matejvasek, line 23: I don't know TypeScript, I am just curious: would `cloudevent?: CloudEvent<string | Customer>` work?

templates/typescript/cloudevents/test/unit.ts
```diff
@@ -13,20 +13,17 @@ test('Unit: handles a valid event', (t) => {
     customerId: '01234'
   };
   // A valid event includes id, type and source at a minimum.
-  const cloudevent: CloudEvent = new CloudEvent({
+  const cloudevent: CloudEvent<Customer> = new CloudEvent<Customer>({
     id: '01234',
     type: 'com.example.cloudevents.test',
     source: '/test',
     data
```
matejvasek, line 20: why this is not `data: data`?

templates/typescript/cloudevents/src/index.ts
lance: Actually it would. Nice catch.

PR #726, opened 2021-12-16
fix: ensure pkged.go is migrated

### LINKED ISSUE #725: Template updates fail CI gofmt check
Rebuilt `pkged.go` is missing a newly introduced build stanza

In Go1.17 a migration introducing build constraints was added.  This means that the `pkged.go` file generated when encoding templates is not fully in compliance with the latest format, as it needs the line `//go:build !skippkger`.  This causes the automated code format checks to fail in CI.

lkingland:
:bug: adds the missing build stanza to pkged.go by running gofmt which includes the migration in 1.17

Fixes #725


lance (commented):
/lgtm

PR #733, opened 2021-12-23
Adding support to ssh-agent on Windows using named pipes

### LINKED ISSUE #732: Unable to use ssh-agent thru named pipes on Windows
On Windows ssh-agent that is exposed thru a named pipe, for example `\\.\pipe\openssh-ssh-agent` is not used by `func` during build process. Even when ssh-agent is properly setup and keys added (and confirmed) it seems func does not make use of it. During it prompts user for password.

Steps to reproduce (On Windows powershell)

```
# Given there is a docker engine running on HOST with proper ssh key generated and configured
# Start SSH and add Key to remote server
Get-Service ssh-agent | Set-Service -StartupType Manual
Start-Service ssh-agent
ssh-add -k <keyfile>
ssh-add -l  # Confirm key is added

# Export Envs and build a sample func
$env:SSH_AUTH_SOCK = '\\.\pipe\openssh-ssh-agent'
$env:DOCKER_HOST = 'ssh://USER@HOST:/var/run/docker.sock'

func.exe create -l node myfunc
func.exe build -p myfunc
# At this point func prompts for user password (instead of retrieve from ssh-agent)
```

jrangelramos:
# Changes

- :bug: Enable use of ssh-agent on Windows thru named pipes on func build (when using a remote container engine thru ssh)

/kind enhancement

Fixes #732


ssh/ssh_dialer.go
```diff
@@ -271,7 +271,7 @@ func NewSSHClientConfig(url *urlPkg.URL, credentialsConfig Config) (*ssh.ClientC
 	if sock, found := os.LookupEnv("SSH_AUTH_SOCK"); found && sock != "" {
 		var agentSigners []ssh.Signer
 		var agentConn net.Conn
-		agentConn, err = net.Dial("unix", sock)
+		agentConn, err = dialSshAgentConnection(sock)
```
matejvasek, line 274: `SSH` should be all capitals.

matejvasek:
@jrangelramos please fix the CI errors.

matejvasek:
otherwise lgtm

Issue #620, opened 2021-11-02
On-cluster Builds


omerbensaadon:
WIP Design Doc: https://docs.google.com/document/d/1iQFVtfsrYb4pp616h3B_IXQB1Cck1oxt7ypZyU8ChzE/edit

Features:
Pushing code from local machine to an on-cluster build
Pushing code from Github to an on-cluster build

# Tasks required for MVP:
- [x] #748
- [x] #750
- [x] #751
- [x] #752
- [x] #753
- [x] #754
- [x] #847
- [x] #848
- [x] #765
- [x] #855
- [x] #1173
- [ ] #984
- [ ] #849
- [x] #856
- [x] #858
- [ ] #859
- [x] #852
- [x] #851
- [x] #857
- [x] #1192
- [ ] https://github.com/knative/func/issues/1858
- [ ] https://github.com/knative/func/issues/1859
# Tasks not required for MVP:
- [x] #989
- [ ] #850
- [ ] #854
- [ ] #853
- [x] #791
- [x] #952
- [x] https://github.com/knative/func/issues/1323

salaboy:
@zroubalik I would like to take these one: 
-  On-cluster build: Implement --git flag for func deploy

I don't seem to be able to convert it to an issue from here, can you?

omerbensaadon:
@salaboy converted for ya, check #765

lance:
@omerbensaadon welcome back! :)

senthilnathan:
If it is possible, can all the outstanding tasks be converted to issues by someone with the permission?

lance:
@zroubalik do you think this can be refined and/or closed given the move to PAC?

lkingland:
This epic is now in bug-fix/enhancement mode

PR #739, opened 2022-01-06
feat: add possibility to disable pushing of image in `deploy` command

zroubalik:

This is needed to have standalone `deploy` Tekton task.

Basically implementing the same functionality that has been done in https://github.com/knative-sandbox/kn-plugin-func/pull/681 for `build`.

# Changes
- :gift: add possibility to disable pushing of image in `deploy` command

/kind enhancement

Relates: #620

### DIFF
--- a/cmd/build_test.go
+++ b/cmd/build_test.go
@@ -170,7 +170,6 @@ created: 2009-11-10 23:00:00`,
 			if err != nil {
 				t.Fatalf("temp dir couldn't be created %v", err)
 			}
-			//	t.Log("tempDir created:", tempDir)
 			t.Cleanup(func() {
 				os.RemoveAll(tempDir)
 			})
--- a/cmd/build_test.go
+++ b/cmd/build_test.go
@@ -144,31 +144,58 @@ created: 2009-11-10 23:00:00`,
 
 func Test_newBuildClient(t *testing.T) {
 	tests := []struct {
-		name    string
-		cfg     buildConfig
-		succeed bool
+		name         string
+		cfg          buildConfig
+		fileContents string
+		succeed      bool
 	}{
 		{
 			name: "push flag set to false avoids pusher instanciation",
 			cfg: buildConfig{
 				Push: false,
 			},
+			fileContents: `name: test-func
+runtime: go
+image: registry.io/foo/bar:latest
+imageDigest: sha256:1111111111111111111111
+created: 2009-11-10 23:00:00`,
 			succeed: false,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+
+			tempDir, err := os.MkdirTemp("", "func-tests")
+			if err != nil {
+				t.Fatalf("temp dir couldn't be created %v", err)
+			}
+			//	t.Log("tempDir created:", tempDir)
+			t.Cleanup(func() {
+				os.RemoveAll(tempDir)
+			})
+
+			fullPath := tempDir + "/func.yaml"
+			tempFile, err := os.Create(fullPath)
+			if err != nil {
+				t.Fatalf("temp file couldn't be created %v", err)
+			}
+			_, err = tempFile.WriteString(tt.fileContents)
+			if err != nil {
+				t.Fatalf("file content was not written %v", err)
+			}
+
 			client, err := newBuildClient(tt.cfg)
 			if err != nil {
 				t.Error(err)
 			}
+
 			defer func() {
 				if r := recover(); r != nil && tt.succeed {
 					t.Errorf("expected function call to succeed %v, got actually %v", tt.succeed, r)
 				}
 			}()
-			err = client.Push(context.TODO(), &fn.Function{})
+			err = client.Push(context.TODO(), tempDir)
 			if err != nil {
 				t.Error(err)
 			}
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -27,19 +27,23 @@ func init() {
 
 func newDeployClient(cfg deployConfig) (*fn.Client, error) {
 	listener := progress.New()
-
 	builder := buildpacks.NewBuilder()
 
-	credentialsProvider := creds.NewCredentialsProvider(
-		creds.WithPromptForCredentials(newPromptForCredentials()),
-		creds.WithPromptForCredentialStore(newPromptForCredentialStore()),
-		creds.WithTransport(cfg.Transport))
-	pusher, err := docker.NewPusher(
-		docker.WithCredentialsProvider(credentialsProvider),
-		docker.WithProgressListener(listener),
-		docker.WithTransport(cfg.Transport))
-	if err != nil {
-		return nil, err
+	pusherOption := fn.WithPusher(nil)
+	if cfg.Push {
+		credentialsProvider := creds.NewCredentialsProvider(
+			creds.WithPromptForCredentials(newPromptForCredentials()),
+			creds.WithPromptForCredentialStore(newPromptForCredentialStore()),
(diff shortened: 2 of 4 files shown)


zroubalik:
This should be ready for review and merge.

cmd/deploy.go
```diff
@@ -100,6 +103,7 @@ kn func deploy --image quay.io/myuser/myfunc -n myns
 	cmd.Flags().StringP("image", "i", "", "Full image name in the form [registry]/[namespace]/[name]:[tag] (optional). This option takes precedence over --registry (Env: $FUNC_IMAGE)")
 	cmd.Flags().StringP("registry", "r", "", "Registry + namespace part of the image to build, ex 'quay.io/myuser'.  The full image name is automatically determined based on the local directory name. If not provided the registry will be taken from func.yaml (Env: $FUNC_REGISTRY)")
 	cmd.Flags().BoolP("build", "b", true, "Build the image before deploying (Env: $FUNC_BUILD)")
+	cmd.Flags().BoolP("push", "u", true, "Attempt to push the function image to registry before deploying (Env: $FUNC_PUSH)")
```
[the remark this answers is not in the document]
matejvasek: it's conflicting with `--path` right?

cmd/deploy.go
```diff
@@ -49,13 +53,12 @@ func newDeployClient(cfg deployConfig) (*fn.Client, error) {
...
 	listener.Verbose = cfg.Verbose
 	builder.Verbose = cfg.Verbose
-	pusher.Verbose = cfg.Verbose
 	deployer.Verbose = cfg.Verbose
 
 	return fn.New(
 		fn.WithProgressListener(listener),
 		fn.WithBuilder(builder),
-		fn.WithPusher(pusher),
+		pusherOption,
```
matejvasek, line 61: nit: wouldn't it be easier just to extract the `pusher` variable?

cmd/build_test.go
```diff
@@ -144,31 +144,58 @@ created: 2009-11-10 23:00:00`,
...
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+
+			tempDir, err := os.MkdirTemp("", "func-tests")
+			if err != nil {
+				t.Fatalf("temp dir couldn't be created %v", err)
+			}
+			//	t.Log("tempDir created:", tempDir)
```
matejvasek, line 173: please remove the comment

cmd/deploy.go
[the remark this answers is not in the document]
zroubalik: Yeah, and `-u` is already being used for build. So keeping it the same

cmd/deploy.go
zroubalik: I am not sure what exactly do you propose. But we need to cover 2 cases: 1. with push initialized and one basically with `nil` pusher.  The very same approach is being used for `build`, so I am keeping it consistent.
But if you have a better solution, I'd happy to use it. But I am not sure what do you propose 🤷‍♂️

matejvasek:
```diff
diff --git a/cmd/build.go b/cmd/build.go
index 16c897a5..b2f2f18a 100644
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -29,13 +29,16 @@ func newBuildClient(cfg buildConfig) (*fn.Client, error) {
        builder := buildpacks.NewBuilder()
        listener := progress.New()
 
-       pusherOption := fn.WithPusher(nil)
+       var (
+               pusher *docker.Pusher
+               err    error
+       )
        if cfg.Push {
                credentialsProvider := creds.NewCredentialsProvider(
                        creds.WithPromptForCredentials(newPromptForCredentials()),
                        creds.WithPromptForCredentialStore(newPromptForCredentialStore()),
                        creds.WithTransport(cfg.Transport))
-               pusher, err := docker.NewPusher(
+               pusher, err = docker.NewPusher(
                        docker.WithCredentialsProvider(credentialsProvider),
                        docker.WithProgressListener(listener),
                        docker.WithTransport(cfg.Transport))
@@ -44,14 +47,13 @@ func newBuildClient(cfg buildConfig) (*fn.Client, error) {
                        return nil, err
                }
                pusher.Verbose = cfg.Verbose
-               pusherOption = fn.WithPusher(pusher)
        }
        builder.Verbose = cfg.Verbose
        listener.Verbose = cfg.Verbose
 
        return fn.New(
                fn.WithBuilder(builder),
-               pusherOption,
+               fn.WithPusher(pusher),
                fn.WithProgressListener(listener),
                fn.WithRegistry(cfg.Registry), // for image name when --image not provided
                fn.WithVerbose(cfg.Verbose)), nil
```

matejvasek: why nil and not noop impl?

zroubalik: gotcha

matejvasek: or:
```diff
diff --git a/cmd/build.go b/cmd/build.go
index 16c897a5..4eae5aaa 100644
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -29,7 +29,13 @@ func newBuildClient(cfg buildConfig) (*fn.Client, error) {
        builder := buildpacks.NewBuilder()
        listener := progress.New()
 
-       pusherOption := fn.WithPusher(nil)
+       opts := []fn.Option{
+               fn.WithBuilder(builder),
+               fn.WithProgressListener(listener),
+               fn.WithRegistry(cfg.Registry), // for image name when --image not provided
+               fn.WithVerbose(cfg.Verbose),
+       }
+       
        if cfg.Push {
                credentialsProvider := creds.NewCredentialsProvider(
                        creds.WithPromptForCredentials(newPromptForCredentials()),
@@ -44,17 +50,12 @@ func newBuildClient(cfg buildConfig) (*fn.Client, error) {
                        return nil, err
                }
                pusher.Verbose = cfg.Verbose
-               pusherOption = fn.WithPusher(pusher)
+               opts = append(opts, fn.WithPusher(pusher))
        }
        builder.Verbose = cfg.Verbose
        listener.Verbose = cfg.Verbose
 
-       return fn.New(
-               fn.WithBuilder(builder),
-               pusherOption,
-               fn.WithProgressListener(listener),
-               fn.WithRegistry(cfg.Registry), // for image name when --image not provided
-               fn.WithVerbose(cfg.Verbose)), nil
+       return fn.New(opts...), nil
 }
 
 type buildClientFn func(buildConfig) (*fn.Client, error)
 ```
 this ^^^ should use noop impl if it's desired

matejvasek: also if the pusher is set to `nil` then push on client panics, shouldn't it rather return error?

zroubalik: Fixed, that's much better. Thanks!

matejvasek: >also if the pusher is set to nil then push on client panics, 

but that's only problem for test since in production code the `Push()` it's guarded.
well actually it's not problem for the test either since it's catching panic (which I find weird).

zroubalik: Yeah, I have had the same concern, but I think that it is okay this way. The panic is being tested here: https://github.com/knative-sandbox/kn-plugin-func/pull/739/files#diff-0b3e07894eabba354682863bda0d8a2e0b5e8166e81e20d13923695b720874c4R145

matejvasek: Well it tests that panics doesn't happen when it shouldn't, but it doesn't test that it does panic when it should.

matejvasek: IMHO we should rather return error than crash.

matejvasek: but that's Independent of this PR.

zroubalik: +1

PR #741, opened 2022-01-09
src: simplify go templates

### LINKED ISSUE #606: Simplify Go CE func template signature
I brought this up in one of the user interviews and couldn't find a note/issue in the docs here whether simplifying the CE handler template signature was already discussed (and why it was decided to keep the 12 signature options). So please excuse if this has been answered already.

Currently, the handler in the Go template for the `CloudEvents` type signature exposes (leaks) all 12 signature options from the underlying Go CE SDK: 

https://github.com/knative-sandbox/kn-plugin-func/blob/6385147e06bafc50107da49cdecd65bd152ab6fb/templates/go/cloudevents/handle.go#L13-L24

Picking the right option (signature) here is not only an additional step one has to make (and understand) but also increases the cognitive load to do the "right thing", especially for newcomers.

My feedback in the interview was to simplify the handler to `func(context.Context, event.Event) (*event.Event, error)` which captures all options and follows Go conventions, e.g. returning `nil,nil` if no response and/or error should be indicated. If I don't need any of the input parameters, I can simply `_` them.

Thoughts?

lkingland:
- :broom: Simplify Go templates in an attempt to reduce beginner confusion.

In user interviews it was recommended that we have all complexity in the templates pushed to the bottom, and a large "your code here" callout in the initial function to make the initial experience by developers better.  This was also brought up again in #606.  This PR proposes a format which, if acceptable, may serve as a template for the other embedded templates as well.  There is a more in-depth discussion there as to why this particular function signature was chosen as the default for Cloud Events that may be worth a read for those so inclined.

/kind enhancement

Fixes #606


zroubalik (commented):
/approve

Looking good

lance (commented):
/lgtm

PR #742, opened 2022-01-10
src: remove need for hosts entry for local clusters

lkingland:
- :broom: remove need for hosts entry for local clusters

We can now simply refer to localhost throughout.  This PR reflects that in the local allocation system.

/kind techdebt


lance (commented):
/lgtm

PR #756, opened 2022-01-12
src:  use common functionality for docker cred retrieval

### LINKED ISSUE #751: On-cluster build: Try to get registry credentials from a local developer machine

matejvasek:
resolves #751 #752

### DIFF
--- a/pipelines/tekton/pipeplines_provider.go
+++ b/pipelines/tekton/pipeplines_provider.go
@@ -3,6 +3,8 @@ package tekton
 import (
 	"context"
 	"fmt"
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
 	"sync"
 
 	"github.com/tektoncd/cli/pkg/pipelinerun"
@@ -110,6 +112,10 @@ func (pp *PipelinesProvider) Run(ctx context.Context, f fn.Function) error {
 			return err
 		}
 		pp.progressListener.Increment("Creating Pipeline resources")
+
+		if registry == name.DefaultRegistry {
+			registry = authn.DefaultAuthKey
+		}
 
 		err = k8s.CreateDockerRegistrySecret(ctx, getPipelineSecretName(f), pp.namespace, creds.Username, creds.Password, registry)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


pipelines/tekton/pipeplines_provider.go
```diff
@@ -97,16 +97,21 @@ func (pp *PipelinesProvider) Run(ctx context.Context, f fn.Function) error {
...
+	registry, err := docker.GetRegistry(f.Image)
+	if err != nil {
+		return err
+	}
+
 	_, err = k8s.GetSecret(ctx, getPipelineSecretName(f), pp.namespace)
 	if errors.IsNotFound(err) {
 		pp.progressListener.Stopping()
-		creds, err := pp.containerRegistryCredentialsCallback()
+		creds, err := pp.credentialsProvider(ctx, registry)
```
zroubalik, line 108: Have you actually test this? I am not sure that registry == server` in this context. For `docker.io` registry a server `https://index.docker.io/v1/` was needed.

matejvasek: I tested it with quay.io and it worked.

matejvasek: that logic is also used for local push

matejvasek: what happened when you used docker.io?

matejvasek: maybe it's some quirk in `lifecycle` binary? Locally just `docker.io` works fine.

matejvasek: in general `docker.io` is weird and treated as special case :disappointed:

matejvasek:
see https://github.com/google/go-containerregistry/blob/main/pkg/authn/keychain.go#L116-L122

pipelines/tekton/pipeplines_provider.go
matejvasek: see https://github.com/google/go-containerregistry/blob/main/pkg/authn/keychain.go#L116-L122

zroubalik: Yeah, it didn't work for `docker.io`.
Cool, let me try that.

zroubalik (commented):
/lgtm
/approve

Great, thanks!

PR #758, opened 2022-01-13
ensure build flag binds for run command

shinigambit:
# Changes

- :bug: Fix the build flag not properly binding to viper

/kind bug


lkingland (approved):
/lgtm

lkingland:
> ...  should I change the name of the flag? Just so it doesn't collide with the environment variable for `deploy`

This might be a good idea.  If it becomes a problem, we can make that change in a separate PR!

PR #761, opened 2022-01-13
Support git scp url format

matejvasek:


matejvasek:
@zroubalik does the tekton support scp syntax?

matejvasek:
> @zroubalik does the tekton support scp syntax?

it seems it does

zroubalik:
It should: https://hub.tekton.dev/tekton/task/git-clone

PR #763, opened 2022-01-17
feat: delete pipeline and resources with `func delete`

zroubalik:

# Changes

- 🎁 On cluster build: delete pipeline and all created resources with func delete

/kind enhancement

Fixes: #754

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -673,12 +673,100 @@ func TestClient_Remove_ByPath(t *testing.T) {
 		return nil
 	}
 
-	if err := client.Remove(context.Background(), fn.Function{Root: root}); err != nil {
+	if err := client.Remove(context.Background(), fn.Function{Root: root}, false); err != nil {
 		t.Fatal(err)
 	}
 
 	if !remover.RemoveInvoked {
 		t.Fatal("remover was not invoked")
+	}
+
+}
+
+// TestClient_Remove_DeleteAll ensures that the remover is invoked to remove
+// and that dependent resources are removed as well -> pipeline provider is invoked
+// the Function with the name of the function at the provided root.
+func TestClient_Remove_DeleteAll(t *testing.T) {
+	var (
+		root              = "testdata/example.com/testRemoveDeleteAll"
+		expectedName      = "testRemoveDeleteAll"
+		remover           = mock.NewRemover()
+		pipelinesProvider = mock.NewPipelinesProvider()
+		deleteAll         = true
+	)
+
+	defer Using(t, root)()
+
+	client := fn.New(
+		fn.WithRegistry(TestRegistry),
+		fn.WithRemover(remover),
+		fn.WithPipelinesProvider(pipelinesProvider))
+
+	if err := client.New(context.Background(), fn.Function{Runtime: TestRuntime, Root: root}); err != nil {
+		t.Fatal(err)
+	}
+
+	remover.RemoveFn = func(name string) error {
+		if name != expectedName {
+			t.Fatalf("Expected to remove '%v', got '%v'", expectedName, name)
+		}
+		return nil
+	}
+
+	if err := client.Remove(context.Background(), fn.Function{Root: root}, deleteAll); err != nil {
+		t.Fatal(err)
+	}
+
+	if !remover.RemoveInvoked {
+		t.Fatal("remover was not invoked")
+	}
+
+	if !remover.RemoveInvoked {
+		t.Fatal("pipelinesprovider was not invoked")
+	}
+
+}
+
+// TestClient_Remove_Dont_DeleteAll ensures that the remover is invoked to remove
+// and that dependent resources are not removed as well -> pipeline provider not is invoked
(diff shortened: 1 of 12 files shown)


cmd/delete.go
```diff
@@ -59,10 +72,11 @@ kn func delete -n apps myfunc
 `,
 		SuggestFor:        []string{"remove", "rm", "del"},
 		ValidArgsFunction: CompleteFunctionList,
-		PreRunE:           bindEnv("path", "confirm", "namespace"),
+		PreRunE:           bindEnv("path", "confirm", "namespace", "all"),
 	}
 
 	cmd.Flags().BoolP("confirm", "c", false, "Prompt to confirm all configuration options (Env: $FUNC_CONFIRM)")
+	cmd.Flags().StringP("all", "a", "true", "Delete all resources created for a function, eg. Pipelines, Secrets, etc. (Env: $FUNC_ALL) (allowed values: \"true\", \"false\")")
```
lkingland, line 79: Glad to see this is set as the default :)

cmd/delete.go
```diff
@@ -152,9 +168,37 @@ func (c deleteConfig) Prompt() (deleteConfig, error) {
...
+			Prompt: &survey.Input{
+				Message: "Function to remove:",
+				Default: deriveName(c.Name, c.Path)},
+			Validate: survey.Required,
+		},
+		{
+			Name: "all",
+			Prompt: &survey.Confirm{
+				Message: "Do you want to delete all resources?",
+				Default: true,
```
lkingland, line 183:
```
Default: c.DeleteAll,
```
This _question_ default could be set to the current value of the config object's default.  That way if the user passes in the flag `--all=false`, but also `--confirm`, the value of the question will be set to the value they specified on the command line first.

cmd/delete.go
```diff
@@ -152,9 +168,37 @@ func (c deleteConfig) Prompt() (deleteConfig, error) {
...
+	err := survey.Ask(qs, &answers)
+	if err != nil {
+		return dc, err
+	}
+
+	dc.Name = answers.Name
+	dc.DeleteAll = answers.All
+	dc.Namespace = c.Namespace
+	dc.Path = c.Path
+	dc.Verbose = c.Verbose
```
lkingland, line 201: I am pretty sure this mapping to a _new_ config object is not necessary and we could just use c directoy, if the original config `c` is treated as defaults (defaults which may have been mutated by command flags before they were parsed in its constructor).

knative/remover.go
```diff
@@ -33,8 +33,6 @@ func (remover *Remover) Remove(ctx context.Context, name string) (err error) {
 		return
 	}
 
-	fmt.Printf("Removing Knative Service: %v\n", name)
-
```
lkingland, line 37: ✺◟( ᐛ )◞✺

lkingland (approved):
Looks like a winner!

/hold so that others have a chance to take a look

client_test.go
```diff
@@ -683,6 +683,94 @@ func TestClient_Remove_ByPath(t *testing.T) {
...
+			t.Fatalf("Expected to remove '%v', got '%v'", expectedName, name)
+		}
+		return nil
+	}
+
+	if err := client.Remove(context.Background(), fn.Function{Root: root}, deleteAll); err != nil {
+		t.Fatal(err)
+	}
+
+	if !remover.RemoveInvoked {
```
matejvasek, line 720: Isn't this condition duplicated here?

client_test.go
```diff
@@ -683,6 +683,94 @@ func TestClient_Remove_ByPath(t *testing.T) {
...
```
matejvasek, line 764: Isn't this condition duplicated here?

client_test.go
```diff
@@ -683,6 +683,94 @@ func TestClient_Remove_ByPath(t *testing.T) {
 
 }
 
+// TestClient_Remove_DeleteAll ensures that the remover is invoked to remove
+// and that dependent resources are removed as well -> pipeline provider is invoked
+// the Function with the name of the function at the provided root.
+func TestClient_Remove_DeleteAll(t *testing.T) {
```
matejvasek, line 689: How does the two new tests differ?

mock/pipelines_provider.go
```diff
@@ -0,0 +1,31 @@
+package mock
+
+import (
+	"context"
+
+	fn "knative.dev/kn-plugin-func"
+)
+
+type PipelinesProvider struct {
+	RunInvoked    bool
```
matejvasek, line 10: I think that the callback are enough and boolean can be left out.
Users of the mock can set such bool values in theirs callbacks.

client_test.go
zroubalik: Bad copy&paste.

client_test.go
zroubalik: Related to the bad copy&paste

lance (commented):
/lgtm

PR #769, opened 2022-01-19
feat: improve help text for all commands

lance:
:gift: Standardize all help text to be filtered through a prefix/name handling function which can deal with `kn` vs `kn func` vs `kn-func`. Work in progress.

Improve this PR to add
- [ ] code documentation
- [ ] standard text width for all command help text (I prefer 70)
- [ ] move additional functions in cmd/root.go to cmd/help-helper.go or similar
- [ ] implications with lack of fully formed formatting capabilities (e.g.) for features in cobra we don't use
- [ ] address kn-func as a plugin
- [ ] address kn-func as a standalone binary

/kind enhancement

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/216

### DIFF
--- a/cmd/repository.go
+++ b/cmd/repository.go
@@ -45,19 +45,17 @@ func newRepositoryClient(args []string) (repositoryConfig, RepositoryClient, err
 
 func NewRepositoryCmd(clientFn repositoryClientFn) *cobra.Command {
 	cmd := &cobra.Command{
-		Short:   "Manage installed template repositories",
+		Short:   "Manage template repositories",
 		Use:     "repository",
 		Aliases: []string{"repo", "repositories"},
-		Long: `
-NAME
-	func-repository - Manage set of installed repositories.
+		Long: `Manage the installed set of template repositories.
 
 SYNOPSIS
-	func repo [-c|--confirm] [-v|--verbose]
-	func repo list [-r|--repositories] [-c|--confirm] [-v|--verbose]
-	func repo add <name> <url>[-r|--repositories] [-c|--confirm] [-v|--verbose]
-	func repo rename <old> <new> [-r|--repositories] [-c|--confirm] [-v|--verbose]
-	func repo remove <name> [-r|--repositories] [-c|--confirm] [-v|--verbose]
+	{{.Prefix}}func repo [-c|--confirm] [-v|--verbose]
+	{{.Prefix}}func repo list [-r|--repositories] [-c|--confirm] [-v|--verbose]
+	{{.Prefix}}func repo add <name> <url>[-r|--repositories] [-c|--confirm] [-v|--verbose]
+	{{.Prefix}}func repo rename <old> <new> [-r|--repositories] [-c|--confirm] [-v|--verbose]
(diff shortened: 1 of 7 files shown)


cmd/repository.go
```diff
@@ -120,49 +118,49 @@ COMMANDS
...
-	o Remove an installed repository
-	  $ func repository list
-	  default
-	  boson-examples
-	  $ func repository remove boson-examples
-	  $ func repository list
-	  default
+	    $ {{.Prefix}}func repository remove <name>
+`,
+		Example: `
```
lkingland, line 131: Not a big deal, but I was trying to follow `git` output as closely as possible (since it is the one other CLI tool a developer is likely to be familiar with), even so far as tabs and spacing:
```
NAME
→func-<subcommand> - <short description>

SYNOPSIS
→func subcommand [-c|--confirm] [-v|--verbose]
<... repeated for each usage >

DESCRIPTION
→Primary Description
→→$ <example command embedded in text>

COMMANDS

→With no arguments, this help text is shown.  To <do the thing> with
→an interactive prompt, use the use the --confirm (-c) flag.
→  $ func <subcommand> -c

→<optional sub-sub command>
→  Sub-descriptions are indented with spaces rather than more tabs

→  An example sub-item
→    $ An example command

EXAMPLES
→o Example Command Description preceded by an empty-circle as bullet
→  $ func subcommand example indented with spaces
<repeated as necessary>
```
Notice how the "EXAMPLES" section immediately follows the high-level explanation of any subcommands; a section which may be omitted if there are none (in which case examples immediately follow the description).

Unfortunately Cobra has it's own idea about how the command help should be formatted, and using the "Examples" member of the struct defers to that choice.

I personally prefer the `git` format due to both clarity and likely developer familiarity, and to have the entire help text in a single variable/member for easier ongoing maintenance.  However, this is not something I think is particularly important if we'd rather use the Cobra style; just thought you'd like to know there was method to the madness!

cmd/repository.go
```diff
@@ -171,6 +169,14 @@ EXAMPLES
 	cmd.Flags().BoolP("confirm", "c", false, "Prompt to confirm all options interactively (Env: $FUNC_CONFIRM)")
 	cmd.Flags().StringP("repositories", "r", fn.RepositoriesPath(), "Path to language pack repositories (Env: $FUNC_REPOSITORIES)")
 
+	cmd.SetHelpFunc(func(cmd *cobra.Command, args []string) {
+		runCommandHelp(cmd, "repositories")
+	})
+	cmd.SetUsageFunc(func(cmd *cobra.Command) error {
+		return runCommandUsage(cmd, `{{.Prefix}}func repository [flags]
+  {{.Prefix}}func repository [commands]`)
+	})
```
lkingland, line 178: Ah, it is nice to have this formatted.

lkingland:
The command help text preprocessing ('func' vs 'kn func' template) aspect of this was handled separately, and once [the PR](https://github.com/knative-sandbox/kn-plugin-func/pull/875) is accepted, I can merge it (and the other nontrivial updates that have happened to the `cmd` package recently) in, leaving this PR to provide the updates to the help text itself.

lance:
I think this needs to be redone from the start, or perhaps isn't even necessary any longer. Closing

PR #774, opened 2022-01-21
[DO NOT MERGE] Midstream patch

matejvasek:
This will be patch for midstream repo.
[DO NOT MERGE]

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -5,8 +5,6 @@ import (
 	"fmt"
 	"net/http"
 
-	fnhttp "knative.dev/kn-plugin-func/http"
-
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/AlecAivazis/survey/v2/terminal"
 	"github.com/ory/viper"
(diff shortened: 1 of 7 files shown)


lance:
@matejvasek this should be opened on openshift-knative/kn-plugin-func shouldn't it?

zroubalik:
> @matejvasek this should be opened on openshift-knative/kn-plugin-func shouldn't it?

@lance @matejvasek IMHO It would be great if we can keep a part of this change in the upstream repo (eg. `openshift` package) and in openshift-knative/kn-plugin-func just have functional changes (eg. setting the default registry: https://github.com/knative-sandbox/kn-plugin-func/pull/774/files#diff-5794c40aa4c9301aaf9c39eacf7d3c8511f2511cb24bdd95e0a08704ba583a5fR99 and similar stuff).

lance:
@zroubalik I spoke with @matejvasek about this today and I can see how this makes sense to have here in some form. I might change the subject so that it's not referencing "midstream" but instead is more like "feat: use known openshift endpoints for registry if available". While this change does improve Red Hat's integration capabilities, it seems that the best approach is to design this in a way that allows other vendors to easily plug-in their own defaults, if desired.

PR #777, opened 2022-01-24
test: fix URL

matejvasek:
On CI machines the image registry is present and reachable.
That causes the actual registry, not mock server, is accessed by the test.


matejvasek:

lkingland (approved):
/lgtm

PR #779, opened 2022-01-24
fix: local cluster name collision

### LINKED ISSUE #778: default cluster name collision
The test cluster used for local development, integration testing, and e2e testing in CI uses the default cluster name `kind`, which of course conflicts when the user is also using `kind` for any other reason.  This should be switched to `func` such that any single node can be used as a single-node Functions-capable provider (`kind` not to be used in production caviat of course in mind).

lkingland:
- :bug: fix default cluster name collision

local Kind cluster now named "func"

/kind bug

Fixes #778

### DIFF
--- a/hack/allocate.sh
+++ b/hack/allocate.sh
@@ -47,7 +47,7 @@ main() {
 
 kubernetes() {
   echo "${em}① Kubernetes${me}"
-  cat <<EOF | kind create cluster --wait=60s --config=-
+  cat <<EOF | kind create cluster --name=func --wait=60s --config=-
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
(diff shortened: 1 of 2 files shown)


hack/allocate.sh
```diff
@@ -179,8 +179,8 @@ registry() {
   # see https://kind.sigs.k8s.io/docs/user/local-registry/
 
   echo "${em}⑥ Registry${me}"
-  docker run -d --restart=always -p "127.0.0.1:50000:5000" --name "kind-registry" registry:2
-  docker network connect "kind" "kind-registry"
+  docker run -d --restart=always -p "127.0.0.1:50000:5000" --name "func-registry" registry:2
+  docker network connect "kind" "func-registry"
```
lkingland, line 183: The network created by kind is always named `kind` and does not collide (one per node regardless of the number of running clusters)

hack/delete.sh
```diff
@@ -14,12 +14,11 @@ main() {
   local red=$(tput bold)$(tput setaf 2)
   local reset=$(tput sgr0)
 
-  echo "${green}Deleting Cluster, Registry and Network${reset}"
-  kind delete cluster --name "kind"
-  docker stop kind-registry && docker rm kind-registry
-  docker network rm kind
```
lkingland, line 20: Managed automatically by Kind

PR #786, opened 2022-01-26
fix: error type-check condition

matejvasek:


matejvasek:
@lkingland @lance @zroubalik PTAL

PR #789, opened 2022-01-26
fix: use domain qualified registry for springboot

lance:
:bug: The builder image name should include the registry name. Also it seems that the tiny builder doesn't actually work. We need base.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/775


matejvasek:
@lance you might need to regenerate `pkged.go`

matejvasek:
> @lance you might need to regenerate `pkged.go`

but maybe it's fault of CI and the `pkged.go` is alright.

lance:
@matejvasek I think it's just because I'm using go 1.16 - confirmed. Regenerated with 1.17

PR #792, opened 2022-01-27
fix: use paketo base image for springboot

lance:
:bug: The build actually seems to require `base` vs `tiny`.

/kind fix


jrangelramos:
Yea `base` builds fine!!!

PR #802, opened 2022-01-31
feat: completion --build flag of deploy sub-cmd

matejvasek:


cmd/completion_util.go
```diff
@@ -125,3 +125,9 @@ func CompleteBuilderList(cmd *cobra.Command, args []string, complete string) (st
 	directive = cobra.ShellCompDirectiveDefault
 	return
 }
+
+func CompleteDeployBuildType(cmd *cobra.Command, args []string, complete string) (buildTypes []string, directive cobra.ShellCompDirective) {
+	buildTypes = []string{fn.BuildTypeLocal, fn.BuildTypeGit, fn.BuildTypeDisabled}
```
zroubalik, line 130: Nit: wouldn't it be better to add a function here: https://github.com/knative-sandbox/kn-plugin-func/blob/cb719ff56421352665abbb6902a80dc03c1588a0/function_buildtype.go

So once we add a new type, we don't have to change it on multiple places

matejvasek: @zroubalik ^^^

zroubalik: @matejvasek yeah I meant just a func that return all enum values, not the whole `CompleteDeployBuildType()`

zroubalik (approved):
/hold until tests pass

PR #815, opened 2022-02-05
Setting the default for language selector

### LINKED ISSUE #814: 'func create -c' fails when go is chosen as the language runtime.
RELEASE:
    The issue is observed in HEAD.

Steps to replicate the issue:
  Run 'func create -c' and to select go as the language just press enter without using the arrow keys to move up and down.
```
$func create -c
? Function Path: testfun
? Language Runtime:
? Template: [tab for suggestions] (http)
Error: Required flag "language" not set.
Available language runtimes are:
  go
  node
  python
  quarkus
  rust
  springboot
  typescript
```

senthilnathan:
# Changes

- :bug: Setting the default for language selector correctly

/kind bug

Fixes #814 

**Release Note**

```release-note
Fixes a bug during func create when confirm option is used with the go language runtime
```

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -255,7 +255,6 @@ func newCreateConfig(args []string, clientFn createClientFn) (cfg createConfig, 
 		Confirm:      viper.GetBool("confirm"),
 		Verbose:      viper.GetBool("verbose"),
 	}
-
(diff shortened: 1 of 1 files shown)


lkingland (approved):
Thank you @senthilnathan for uncovering and fixing the bug.

PR #816, opened 2022-02-06
printing helpful shorthand command

### LINKED ISSUE #768: Make `func create` more new-user-friendly
Via @lkingland 

> The `-c` flag is intended for first-time users or beginners.
It would be nice if we thought of it as "training wheels" to proper command-line usage.

Consider modifying the output from interactive mode to always include, at the end, a message indicating how the process could have been completed using a set of flags.

/kind feat

Related: https://github.com/knative-sandbox/kn-plugin-func/issues/635

/cc @omerbensaadon

senthilnathan:
# Changes

- :gift: Printing helpful shorthand command when confirm option is chosen

/kind enhancement

Fixes #768

**Release Note**

```release-note
printing helpful shorthand command when confirm option is chosen
```

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -287,7 +287,7 @@ func newCreateConfig(cmd *cobra.Command, args []string, newClient ClientFactory)
 // singleCommand that could be used by the current user to minimally recreate the current state.
 func singleCommand(cmd *cobra.Command, args []string, cfg createConfig) string {
 	var b strings.Builder
-	b.WriteString(os.Args[0])           // process executable
+	b.WriteString(cmd.Root().Name())    // process executable
 	b.WriteString(" -l " + cfg.Runtime) // language runtime is required
 	if cmd.Flags().Lookup("template").Changed {
 		b.WriteString(" -t " + cfg.Template)
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -467,6 +467,14 @@ func (c createConfig) prompt(client *fn.Client) (createConfig, error) {
 		return c, err
 	}
 
+	executable, err := os.Executable()
+	if err != nil {
+		fmt.Printf("Unable to get the executable path: %v", err)
+		executable = "func"
+	}
+	fmt.Printf("The command below is a shorthand method to create the same function!\n")
+	fmt.Printf("    %v -l %v -t %v %v\n", executable, c.Runtime, c.Template, c.Path)
+
 	return c, nil
 }
 
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -113,7 +113,7 @@ func runCreate(cmd *cobra.Command, args []string, newClient ClientFactory) (err 
 	// Config
 	// Create a config based on args.  Also uses the newClient to create a
 	// temporary client for completing options such as available runtimes.
-	cfg, err := newCreateConfig(args, newClient)
+	cfg, err := newCreateConfig(cmd, args, clientFn)
 	if err != nil {
 		return
 	}
@@ -159,7 +159,7 @@ func runCreateHelp(cmd *cobra.Command, args []string, newClient ClientFactory) {
 
 	tpl := createHelpTemplate(cmd)
 
-	cfg, err := newCreateConfig(args, newClient)
+	cfg, err := newCreateConfig(cmd, args, clientFn)
 	failSoft(err)
 
 	client := newClient(createConfigToClientOptions(cfg))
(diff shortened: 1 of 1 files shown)


cmd/create.go
```diff
@@ -475,6 +475,14 @@ func (c createConfig) prompt(client *fn.Client) (createConfig, error) {
 		return c, err
 	}
 
+	executable, err := os.Executable()
+	if err != nil {
+		fmt.Printf("Unable to get the executable path: %v", err)
+		executable = "func"
+	}
+	fmt.Printf("The command below is a shorthand method to create the same function!\n")
```
lance, line 483: Nit
```suggestion
	fmt.Printf("The command below is a shorthand method to create the same function\n")
```

lance (commented):
Looks good in general. However, I am wondering how it should work if `--repository` is specified.

cmd/create.go
```diff
@@ -475,6 +475,14 @@ func (c createConfig) prompt(client *fn.Client) (createConfig, error) {
 		return c, err
 	}
 
+	executable, err := os.Executable()
+	if err != nil {
```
lkingland, line 479: This will return an absolute path, and can even include `../` if the user is running the command outside of PATH (such as when developing).  Since we want to tell the current user exactly what command they can run again, presuming their current environment, you can simply use the first argument: `os.Args[0]`.

cmd/create.go
```diff
@@ -475,6 +475,14 @@ func (c createConfig) prompt(client *fn.Client) (createConfig, error) {
...
+		fmt.Printf("Unable to get the executable path: %v", err)
+		executable = "func"
+	}
+	fmt.Printf("The command below is a shorthand method to create the same function!\n")
+	fmt.Printf("    %v -l %v -t %v %v\n", executable, c.Runtime, c.Template, c.Path)
```
lkingland, line 484: Looks good!  Of course, this is a bit more complicated to do correctly, so we might want to add a little bit more logic here to handle other flags, and omit those that were not changed from their defaults.

This is opening the way to a deep rabbit hole, but perhaps we could solve it like this:  Append the flags to the printed command IFF the user provided a value that is different from the flag static default. 

For example, passing `cmd` through `newCreateConfig` you could add a print [right after prompting](https://github.com/knative-sandbox/kn-plugin-func/pull/816/files#diff-e52f1f4fa0c533e3287cbe2ef0436582b5ca471a3fddcdf1f49fecd1ec0c7d13R269-R271) that builds a short command based on the current state of the system post-prompting:
```go
  // IN confirm mode.  If also in an interactive terminal, run prompts.
  if interactiveTerminal() {
    cfg.prompt(client)
    fmt.Println("Command:")
    fmt.Println(singleCommand(cmd, args, cfg))
  }
```
where `singleCommand` only prints flags which were meaningfully altered (warning, minimally tested!)
```go
// singleCommand prints the single command that could be used by the current
// user to minimally recreate the state of the current execution state.
func singleCommand(cmd *cobra.Command, args []string, cfg createConfig) string {
  var b strings.Builder
  b.WriteString(os.Args[0])           // process executable
  b.WriteString(" -l " + cfg.Runtime) // language runtime is required
  if cmd.Flags().Lookup("template").Changed {
    b.WriteString(" -t " + cfg.Template)
  }
  if cmd.Flags().Lookup("repository").Changed {
    b.WriteString("-r " + cfg.Repository)
  }
  if cmd.Flags().Lookup("verbose").Changed {
    b.WriteString(fmt.Sprintf(" -v %v", cfg.Verbose))
  }
  if len(args) > 0 {
    b.WriteString(" " + os.Args[0]) // optional trailing <path> argument
  }
  return b.String()
}
```

cmd/create.go
matejvasek: or maybe use `cmd.Root().Name()`

cmd/create.go
senthilnathan: Thanks for sharing the code segment.
I have changed:
```
if len(args) > 0 {
    b.WriteString(" " + os.Args[0]) // optional trailing <path> argument
  }
```
  to
```
if len(args) > 0 {
    b.WriteString(" " + cfg.Path) // optional trailing <path> argument
}
```
Hope this is fine.

cmd/create.go
```diff
@@ -278,6 +284,26 @@ func newCreateConfig(args []string, newClient ClientFactory) (cfg createConfig,
 	return
 }
 
+// singleCommand that could be used by the current user to minimally recreate the current state.
+func singleCommand(cmd *cobra.Command, args []string, cfg createConfig) string {
+	var b strings.Builder
+	b.WriteString(os.Args[0])           // process executable
```
matejvasek, line 290: `b.WriteString(cmd.Root().Name())` would be better I think.

matejvasek:
Looks good.

cmd/create.go
senthilnathan: Done.

matejvasek:
@lkingland I am going to merge this. It might cause merge issue for your current PRs.

PR #818, opened 2022-02-07
src: error-less ctors

matejvasek:
Constructors do only necessary initialization with no possible error.
Errors are deferred until some actual operation is being done.

This allows to initialize all parts of `fn.Client` at once with no error.
We need this to be able to create a shared factory function of `fn.Client` for multiple commands.
Without this such a factory is not possible because the function could fail while initializing something that won't be actually used.
For example: `func create` doesn't need `k8s` configuration set up, but the command would still fail because of it.

### DIFF
--- a/client_int_test.go
+++ b/client_int_test.go
@@ -206,10 +206,7 @@ func newClient(verbose bool) *fn.Client {
 	builder := buildpacks.NewBuilder()
 	builder.Verbose = verbose
 
-	pusher, err := docker.NewPusher()
-	if err != nil {
-		panic(err)
-	}
+	pusher := docker.NewPusher()
 	pusher.Verbose = verbose
 
 	deployer := knative.NewDeployer(DefaultNamespace)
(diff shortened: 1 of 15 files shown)


matejvasek:
@zroubalik I updated description.

matejvasek:
>This allows to initialize all parts of fn.Client at once with no error.
We need this to be able to create a shared factory function of fn.Client for multiple commands.

meaning instead of `newCreateClient` `newBuildClient` `newDeployClient` `newDeleteClient` `newInfoClient` `newListClient` `newInvokeClient` `newRepositoryClient` `newRunClient` which are partially initializing `fn.Client`,
there would be just single factory function.

lkingland (approved):
Much improved!  It is quite nice to keep constructors dumb and validate integrity of members at time of use and/or in an explicit validation step.  Not always possible of course, but sure does clean things up when it is.

Looking forward to seeing if the shared fn.Client constructor works out too.

PR #821, opened 2022-02-07
fix: defaults for prompt from flags then options

### LINKED ISSUE #819: Specifying `--confirm` and `--repository` results in error for `func create`
When specifying both `--confirm` and `--repository` in a call to `func create` an error is thrown in spite of valid input.

```
❯ kn func create -c -r https://github.com/boson-project/test-language-pack
? Function Path: foo
? Language Runtime:
? Template: http
Error: Required flag "language" not set.
Available language runtimes are:
  go

Error: exit status 1

```

### LINKED ISSUE #820: handle missing and invalid values when prompting
When a user provides an invalid option for a flag, a prompt selection default is not actually selected despite its UX showing selected.

When a flag or environment variable being used as a prompt selection is empty (has no default value), the selection default is not actually selected despite its UX showing selected.

In situations where a Flag is being used to directly set the default value for a survey.Select prompt,  that default should only be set when the value provided by the user both exists and is one of the options.

lkingland:
- :bug: select prompt option based on flag or default

Fixes #819
Fixes #820 

Related issue #814

/kind bug

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -446,7 +446,7 @@ func (c createConfig) prompt(client *fn.Client) (createConfig, error) {
 			Prompt: &survey.Select{
 				Message: "Language Runtime:",
 				Options: runtimes,
-				Default: runtimes[0],
+				Default: surveySelectDefault(c.Runtime, runtimes),
 			},
 		}}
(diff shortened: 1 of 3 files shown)


lance (commented):
/lgtm

PR #823, opened 2022-02-08
fix: don't fail info/invoke if eventing missing

lance:
If eventing is not installed, the user should still be able to get info and use
`kn invoke`. Don't fail, if `describe` finds eventing missing.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/822

/kind fix


lkingland (approved):
Good call!

PR #825, opened 2022-02-09
Custom improvements for OpenShift

matejvasek:
:gift: add initial framework for cluster flavor detection/actions

```release-note
Detects when deploying to OpenShift and use internal registry
```

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -1,13 +1,13 @@
 package cmd
 
 import (
-	"net/http"
-
 	fn "knative.dev/kn-plugin-func"
 	"knative.dev/kn-plugin-func/buildpacks"
 	"knative.dev/kn-plugin-func/docker"
 	"knative.dev/kn-plugin-func/docker/creds"
+	fnhttp "knative.dev/kn-plugin-func/http"
 	"knative.dev/kn-plugin-func/knative"
+	"knative.dev/kn-plugin-func/openshift"
 	"knative.dev/kn-plugin-func/pipelines/tekton"
 	"knative.dev/kn-plugin-func/progress"
 )
@@ -22,8 +22,36 @@ type ClientOptions struct {
 
 type ClientFactory func(opts ClientOptions) *fn.Client
 
-func NewClientFactory(transport http.RoundTripper) ClientFactory {
-	return func(clientOptions ClientOptions) *fn.Client {
+// NewDefaultClientFactory returns a function that creates instances of function.Client and a cleanup routine.
+//
+// This function may allocate resources that are used by produced instances of function.Client.
+//
+// To free these resources (after instances of function.Client are no longer in use)
+// caller of this function has to invoke the cleanup routine.
+//
+// Usage:
+//   newClient, cleanUp := NewDefaultClientFactory()
+//   defer cleanUp()
+//   fnClient := newClient()
+//   // use your fnClient here...
+func NewDefaultClientFactory() (newClient ClientFactory, cleanUp func() error) {
+
+	var transportOpts []fnhttp.Option
+	var additionalCredLoaders []creds.CredentialsCallback
+
+	switch {
+	case openshift.IsOpenShift():
+		transportOpts = append(transportOpts, openshift.WithOpenShiftServiceCA())
+		additionalCredLoaders = openshift.GetDockerCredentialLoaders()
+	default:
+	}
+
+	transport := fnhttp.NewRoundTripper(transportOpts...)
+	cleanUp = func() error {
+		return transport.Close()
+	}
+
+	newClient = func(clientOptions ClientOptions) *fn.Client {
 		builder := buildpacks.NewBuilder()
 		builder.Verbose = clientOptions.Verbose
 
@@ -33,7 +61,8 @@ func NewClientFactory(transport http.RoundTripper) ClientFactory {
 		credentialsProvider := creds.NewCredentialsProvider(
 			creds.WithPromptForCredentials(newPromptForCredentials()),
 			creds.WithPromptForCredentialStore(newPromptForCredentialStore()),
-			creds.WithTransport(transport))
+			creds.WithTransport(transport),
+			creds.WithAdditionalCredentialLoaders(additionalCredLoaders...))
 
 		pusher := docker.NewPusher(
 			docker.WithCredentialsProvider(credentialsProvider),
@@ -83,5 +112,16 @@ func NewClientFactory(transport http.RoundTripper) ClientFactory {
 		}
 
 		return fn.New(opts...)
+	}
+
+	return newClient, cleanUp
+}
+
+func GetDefaultRegistry() string {
+	switch {
+	case openshift.IsOpenShift():
+		return openshift.GetDefaultRegistry()
+	default:
+		return ""
 	}
 }
--- a/openshift/openshift.go
+++ b/openshift/openshift.go
@@ -122,10 +122,12 @@ func GetDockerCredentialLoaders() []creds.CredentialsCallback {
 	cc := rawConf.Contexts[rawConf.CurrentContext]
 	authInfo := rawConf.AuthInfos[cc.AuthInfo]
 
-	user := "user"
+	user := "<UNKNOWN>"
 	parts := strings.SplitN(cc.AuthInfo, "/", 2)
 	if len(parts) >= 1 {
 		user = parts[0]
+	} else {
+		return nil
 	}
 	credentials := docker.Credentials{
 		Username: user,
--- a/openshift/openshift.go
+++ b/openshift/openshift.go
@@ -1,0 +1,164 @@
+package openshift
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+	"strings"
+	"sync"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	"knative.dev/kn-plugin-func/docker"
+	"knative.dev/kn-plugin-func/docker/creds"
+	fnhttp "knative.dev/kn-plugin-func/http"
+	"knative.dev/kn-plugin-func/k8s"
+)
+
+const (
+	registryHost     = "image-registry.openshift-image-registry.svc"
+	registryHostPort = registryHost + ":5000"
+)
+
+func GetServiceCA(ctx context.Context) (*x509.Certificate, error) {
+	client, ns, err := k8s.NewClientAndResolvedNamespace("")
+	if err != nil {
+		return nil, err
+	}
+
+	cfgMapName := "service-ca-config-" + rand.String(5)
+
+	cfgMap := &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        cfgMapName,
+			Annotations: map[string]string{"service.beta.openshift.io/inject-cabundle": "true"},
+		},
+	}
+
+	configMaps := client.CoreV1().ConfigMaps(ns)
+
+	nameSelector := fields.OneTermEqualSelector("metadata.name", cfgMapName).String()
+	listOpts := metav1.ListOptions{
+		Watch:         true,
+		FieldSelector: nameSelector,
+	}
+
+	watch, err := configMaps.Watch(ctx, listOpts)
+	if err != nil {
+		return nil, err
+	}
+	defer watch.Stop()
+
+	crtChan := make(chan string)
+	go func() {
+		for event := range watch.ResultChan() {
+			cm, ok := event.Object.(*v1.ConfigMap)
+			if !ok {
+				continue
+			}
+			if crt, ok := cm.Data["service-ca.crt"]; ok {
+				crtChan <- crt
+				close(crtChan)
+				break
+			}
+		}
+	}()
+
+	_, err = configMaps.Create(ctx, cfgMap, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		_ = configMaps.Delete(ctx, cfgMapName, metav1.DeleteOptions{})
+	}()
+
+	select {
+	case crt := <-crtChan:
+		blk, _ := pem.Decode([]byte(crt))
+		return x509.ParseCertificate(blk.Bytes)
+	case <-time.After(time.Second * 5):
+		return nil, errors.New("failed to get OpenShift's service CA in time")
+	}
+}
+
+// WithOpenShiftServiceCA enables trust to OpenShift's service CA for internal image registry
+func WithOpenShiftServiceCA() fnhttp.Option {
+	var selectCA func(ctx context.Context, serverName string) (*x509.Certificate, error)
+	ca, err := GetServiceCA(context.TODO())
+	if err == nil {
+		selectCA = func(ctx context.Context, serverName string) (*x509.Certificate, error) {
+			if strings.HasPrefix(serverName, registryHost) {
+				return ca, nil
+			}
+			return nil, nil
+		}
+	}
+	return fnhttp.WithSelectCA(selectCA)
+}
+
+func GetDefaultRegistry() string {
+	ns, _ := k8s.GetNamespace("")
+	if ns == "" {
+		ns = "default"
+	}
+
+	return registryHostPort + "/" + ns
+}
+
+func GetDockerCredentialLoaders() []creds.CredentialsCallback {
+	conf := k8s.GetClientConfig()
+
+	rawConf, err := conf.RawConfig()
+	if err != nil {
+		return nil
+	}
+
+	cc := rawConf.Contexts[rawConf.CurrentContext]
+	authInfo := rawConf.AuthInfos[cc.AuthInfo]
+
+	user := "user"
+	parts := strings.SplitN(cc.AuthInfo, "/", 2)
+	if len(parts) >= 1 {
+		user = parts[0]
+	}
+	credentials := docker.Credentials{
+		Username: user,
+		Password: authInfo.Token,
+	}
+
+	return []creds.CredentialsCallback{
+		func(registry string) (docker.Credentials, error) {
+			if registry == registryHostPort {
+				return credentials, nil
+			}
+			return docker.Credentials{}, nil
+		},
+	}
+
+}
+
+var isOpenShift bool
+var checkOpenShiftOnce sync.Once
+
+func IsOpenShift() bool {
+	checkOpenShiftOnce.Do(func() {
+		client, err := k8s.NewKubernetesClientset()
+		if err != nil {
+			isOpenShift = false
+			return
+		}
+		_, err = client.CoreV1().Services("openshift-image-registry").Get(context.TODO(), "image-registry", metav1.GetOptions{})
+		if k8sErrors.IsNotFound(err) {
+			isOpenShift = false
+			return
+		}
+		isOpenShift = true
+	})
+	return isOpenShift
+}
--- a/openshift/openshift.go
+++ b/openshift/openshift.go
@@ -122,7 +122,7 @@ func GetDockerCredentialLoaders() []creds.CredentialsCallback {
 	cc := rawConf.Contexts[rawConf.CurrentContext]
 	authInfo := rawConf.AuthInfos[cc.AuthInfo]
 
-	user := "<UNKNOWN>"
+	var user string
 	parts := strings.SplitN(cc.AuthInfo, "/", 2)
 	if len(parts) >= 1 {
 		user = parts[0]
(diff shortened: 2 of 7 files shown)


matejvasek:
@salaboy @shinigambit would you be OK with this?

matejvasek:
Just draft for now.

matejvasek:
@lance @lkingland with this we could just patch tags in Makefile in midstream.

evankanderson:
Would it make sense to contribute this to https://github.com/google/go-containerregistry/tree/main/pkg/authn, and leverage that library? That library seems to be used by a lot of components, including the `pack` CLI.

evankanderson:
In particular, there's a [credential helper protocol](https://github.com/google/go-containerregistry/tree/main/pkg/authn#credential-helpers) that would allow you to specify an external binary helper and authorization config for `image-registry.openshift-image-registry.svc` like so:

```json
{
	"credHelpers": {
		"image-registry.openshift-image-registry.svc": "os",
	}
}
```

This would also work with other commands like `crane`, `dive` and `docker`, assuming you had `docker-credential-os` in your path.

matejvasek:
@evankanderson I am not sure what you mean specifically. We are using `go-containerregistry` and `credsHelpers` internally for credentials handling. 

Here I am just adding very specialized loader `func GetDockerCredentialLoaders() []creds.CredentialsCallback {}` reason:
Internal OpenShift registry has the same credentials as the `k8s`. So this specialized call pulls credentials from `~/.kube/config` instead form `~/.config/func/auth.json`.

evankanderson:
I'm suggesting: why not have a `docker-credential-openshift` or `docker-credential-k8s` that extracts those credentials, and can use the standard `credsHelper` mechanism?

matejvasek:
>This would also work with other commands like crane, dive and docker, assuming you had docker-credential-os in your path.

We are deliberately using `~/.docker/config.json` as read-only, so if somebody is already logged into registry via `docker` CLI, the `func` should use that. Opposite is not true: if you are logged via `func` the `docker` CLI won't see the credentials.

openshift/openshift.go
```diff
@@ -0,0 +1,175 @@
+package openshift
```
salaboy, line 1: @matejvasek it will be great not to have this inside `func`.. is there any way to keep this out and add things like this as extensions? When I proposed profiles, I was thinking about: 
- registry details
- credentials / CA
These can be created in a generic and non Openshift exclusive way. 

I am not sure about the roundtripper.. maybe we can create a generic roundtripper (first I need to understand what it is doing exactly)

openshift/openshift.go
```diff
@@ -0,0 +1,175 @@
...
+	if ns == "" {
+		ns = "default"
+	}
+
+	return registryHostPort + "/" + ns
+}
+
+func GetDockerCredentialLoaders() []creds.CredentialsCallback {
+	if !IsOpenShift() {
+		return nil
```
salaboy, line 123: @matejvasek can this be used for non-openshift environments? what is Openshift specific from this code?

cmd/client_openshift.go
```diff
@@ -0,0 +1,23 @@
+//go:build openshift
```
salaboy, line 1: @matejvasek sorry for my newbie question.. is this only going to be built if it is an "openshift" build?

openshift/openshift.go
matejvasek: The OpenShift's internal registry uses the same credentials that `k8s`/`kubectl`, I think that is not generally true for other registries.

openshift/openshift.go
matejvasek: The `http.RoundTripper` here is our specialized implementation that can dial access services in cluster.
See: https://github.com/knative-sandbox/kn-plugin-func/blob/main/k8s/dialer_test.go#L26

matejvasek: > is there any way to keep this out

This code is called only if `-tags openshift` is used. So upstream build wouldn't use this.

matejvasek: >For OpenShift we just need to add the CA to the RoundTripper.

Important: this code trust the OpenShift CA only for limited set of hosts (right now exactly one host -- the registry).

matejvasek: wrt profiles:
You mean there would be some custom configuration on filesystem (e.g. in `~/.config/func/`)?

That might be possible, but it implies additional setup that would have to be done user.
(or maybe done by package manager if we had `kn` rpm)

I really would like to make this work without any additional setup.

matejvasek:
@lance @lkingland @zroubalik  If I understand it correctly @salaboy and @evankanderson suggest not to do this by direct integration in binary.

Instead this behaviour could be customized trough some local files.

What would it mean for us (RH):
We would need to create some (post)installation script that would create such a files.

Such a script might need to:
* install custom credential helper binary, and put it into path via append to shell env/rc files
* predefine credentials in `~/.config/func/auth.json`
* pull the OSh service CA add save it on FS
* create some configuration file that would say the OSh service CA is trusted for some specific hosts
* set export FUNC_REGISTRY envvar via append to shell env/rc files

evankanderson:
I'm not sure what tools users are already using; I see that there's [already an `oc` CLI](https://docs.openshift.com/container-platform/4.7/cli_reference/openshift_cli/getting-started-cli.html#cli-installing-cli-on-macos_cli-developer-commands) that suggests installing a binary and moving in onto your PATH. The same will need to be true for the `func` CLI -- at some point, you might want to have a script that copies things into PATH, but having `oc` and `docker-credential-oc` seems like it would enable using the full OCI ecosystem with OpenShift.

I'm _not_ blocking this submit, just trying to suggest ways to make the overall OpenShift experience align with the rest of the ecosystem, which I think helps us all.

evankanderson:
Depending on the space in the `oc` binary, you could [follow the `gcloud` example, and have a command to update the dockerconfig](https://cloud.google.com/sdk/gcloud/reference/auth/configure-docker).

lance:
> @lance @lkingland @zroubalik If I understand it correctly @salaboy and @evankanderson suggest not to do this by direct integration in binary.

Why not do this in openshift-knative/kn-plugin-func?

matejvasek:
> > @lance @lkingland @zroubalik If I understand it correctly @salaboy and @evankanderson suggest not to do this by direct integration in binary.
> 
> Why not do this in openshift-knative/kn-plugin-func?

yes we can do it by `git am patch`

matejvasek:
> > > @lance @lkingland @zroubalik If I understand it correctly @salaboy and @evankanderson suggest not to do this by direct integration in binary.
> > 
> > 
> > Why not do this in openshift-knative/kn-plugin-func?
> 
> yes we can do it by `git am patch`

But I strongly prefer not to have custom patches in openshift-knative/kn-plugin-func, in long term it will complicate maintenance.

lance:
> But I strongly prefer not to have custom patches in openshift-knative/kn-plugin-func, in long term it will complicate maintenance.

I agree with you there. This is the more desirable way for RHT, to be sure. Would other vendor offered clusters benefit from something similar?

salaboy:
@lance I think this will happen for sure.. other vendors will need their own specific hooks and checks, so I will tend to agree with @evankanderson in the sense of let's make sure that we make `func` and other libraries better so we can support multiple Kubernetes flavours. 

I wouldn't like it either if you need to do patches to create your specific binaries, so my suggestion would be to try to create the mechanisms to check if you are running on Openshift as generic as possible. 

All these check can conform a profile, for example if check A, B and C are true, you are in Openshift, if only check A and C are true, you are in another vendor. If we follow the approach implemented in this PR, then each vendor will need to have it's own set of checks and we will not be promoting people to reuse those mechanisms.

openshift/openshift.go
lkingland: >  is there any way to keep this out and add things like this as extensions?

I believe this is already the case.  The core `func` repo is structured as dependency-free, with third-party dependencies supported through packages which implement their interfaces.  So this is indeed essentially "kept out" and added as an "extension".

This is a common method of structuring repositories to accomplish the goal you are suggesting. 👍🏻

lkingland:
This is working as expected.  The core `func` library is dependency-free, purely logic.  Third-party-specifics are held behind interfaces.  `main` is responsible for introspecting the environment, either automatically or explicitly by the user, to load the core library (logic) with the correct implementation.  For example, there may be a different "deployer" if someone is deploying to a local Kind cluster, an EC2 cluster, an OpenShift cluster etc.  These nuanced implementations reside in packages `kind`, `openshift`, `ec2`, etc.

cmd/client_vanilla.go
```diff
@@ -0,0 +1,22 @@
+//go:build !openshift
+// +build !openshift
```
lkingland, line 2: I am happy to see that this is where it should be: `cmd` with a reference to an implementation-specific package.  This is the same as can be used for all future cases where flavor-specific code is necessary, not just for OpenShift.

My concern with this approach is that it results in a binary that is OpenShift specific.  This is not necessary, because the core `func` logic is agnostic if cluster flavor.

Is there not a way we can detect the cluster flavor at runtime?

evankanderson:
VMware actually has a slightly more complicated detection logic, because "Tanzu Application Platform" is a layer which can be applied to different types of clusters, including Tanzu Kubernetes Grid (VMware), EKS (Amazon), or even OpenShift on AWS (turtles all the way down).

Should this check be conditioned on the cluster, the registry, or some combination?

We might be able to use the auth preflight check with the `Server` header to give the different auth plugins a chance to update the client with auth credentials.

lkingland:
> Should this check be conditioned on the cluster, the registry, or some combination?
> ...
> We might be able to use the auth preflight check with the Server header

If we could use that same request to find out if it is Tanzu Application Platform, OpenShift, or vanilla; that seems like it could work well.  The analogy perhaps being this information is a bit like operating system, where Function is to Kubernetes Flavor as Binary is to Operating System?

Using the Server header in the auth preflight is indeed runtime, and done for every cluster, so it would accomplish the goal of allowing `func` to swap out any of the concrete implementations used by Client as necessary to take advantage of flavor-specific advantages.  It also accomplishes the goal of having a single `func` binary capable of interacting with any cluster which meets the minimum requirements (Knative installed), but with flavor-specific benefits.

Sounds like a good topic for this week's call.

matejvasek:
@lkingland just technical detail: In this PR we use the same `function.Deployer` implementation, no change in it. In this PR we are just setting up custom `http.RoundTripper` -- which is trusting the OSh CA, and custom credentials loader.

matejvasek:
>Is there not a way we can detect the cluster flavor at runtime?

@lkingland 
We can detect OpenShift and we do that [here](https://github.com/knative-sandbox/kn-plugin-func/pull/825/files#diff-148f58928fdea9d17884df74c0b92f348214ffe38353765f29198a9179b0665fR160). If the check returns false then everything works as before. I used build tags only to keep vanilla upstream clean.

matejvasek:
Even without build tag separation this shouldn't negatively affect anything.

matejvasek:
> We can detect OpenShift and we do that [here](https://github.com/knative-sandbox/kn-plugin-func/pull/825/files#diff-148f58928fdea9d17884df74c0b92f348214ffe38353765f29198a9179b0665fR160).

It's not like that we always assumed OpenShift cluster when `-tags openshift`.

cmd/client_vanilla.go
salaboy: +1 here.. I would be more in favour of runtime detection with "standard" mechanisms, that we can reuse for other environments.

cmd/client.go
```diff
@@ -22,8 +22,36 @@ type ClientOptions struct {
...
+//
+// To free these resources (after instances of function.Client are no longer in use)
+// caller of this function has to invoke the cleanup routine.
+//
+// Usage:
+//   newClient, cleanUp := NewDefaultClientFactory()
+//   defer cleanUp()
+//   fnClient := newClient()
+//   // use your fnClient here...
+func NewDefaultClientFactory() (newClient ClientFactory, cleanUp func() error) {
```
matejvasek, line 37: @salaboy the `function.Client` is created here. Additional conditional initialization should be done here.

matejvasek:
@salaboy PTAL

zroubalik (approved):
Looking good

openshift/openshift.go
lkingland: I agree you can just remove the `if !IsOpenShift() { return }` statement. There is no need to be defensive here, and the function already exists in the `openshift` package.  It is expected the caller either is using it as expected `get cred loaders for my OpenShift Clutser`, or they know they are bending things `get cred loaders for my cluster which happens to be configured the same way so this will work`, or they get an error later.  I don't see any need to prematurely limit the implementation to only completing `if IsOpenShift`.

cmd/client.go
```diff
@@ -22,8 +22,36 @@ type ClientOptions struct {
...
+
+	var transportOpts []fnhttp.Option
+	var additionalCredLoaders []creds.CredentialsCallback
+
+	switch {
+	case openshift.IsOpenShift():
```
lkingland, line 43: Just a comment:
I can see a future where we have a set of supported platforms, which are each given the opportunity to self-identify and add to the features offered by the Function Client.  This would work a bit like a connection upgrade attempt.  This keeps the core library clean of any third-party dependencies, but allows for our Functions library to offer the ability to seamelssly switch between cluster flavors.

lkingland (approved):
Looking good!
I am sure the exact way that we implement support for various Kubernetes flavors will evolve over time.  For now, it is nice to see that a single `func` will work seamlessly whether it is connected to a vanilla K8S cluster or an OpenShift cluster, transparently making use of any upgrades/preconfigurations available.

salaboy:
This looks like a good first iteration. I do feel that we haven't answered @evankanderson question yet: `Should this check be conditioned on the cluster, the registry, or some combination?` Right now these checks are performed by default, am I right?

matejvasek:
>Should this check be conditioned on the cluster, the registry, or some combination?

We are conditioning on cluster type (by checking that the cluster contains something that is own to OpenShift).

openshift/openshift.go
matejvasek: Maybe unnecessary but shouldn't hurt either.

matejvasek:
@lance @lkingland @salaboy @evankanderson are we OK with merging this into main?

lance (commented):
This in general looks good to me, @matejvasek and I am OK merging it. However, since I have not actively been engaged in the dialog I'd like at least @salaboy and @lkingland or @zroubalik to also give the OK. So, I'm not removing `hold` just yet.

/lgtm

lkingland:
> @lance @lkingland @salaboy @evankanderson are we OK with merging this into main?

Yes @matejvasek I think this is fine to merge.

openshift/openshift.go
```diff
@@ -0,0 +1,176 @@
...
+			return nil, nil
+		}
+	}
+	return fnhttp.WithSelectCA(selectCA)
+}
+
+func GetDefaultRegistry() string {
+	if !IsOpenShift() {
+		return ""
+	}
```
evankanderson, line 112: I notice both this and `WithOpenShiftServiceCA` have this conditional, and we _also_ have the conditional on line 43 of `cmd/client.go`. It feels like this should be centralized and the other locations shouldn't need to also do this check.

openshift/openshift.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil, err
+	}
+
+	cfgMapName := "service-ca-config-" + rand.String(5)
+
+	cfgMap := &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        cfgMapName,
+			Annotations: map[string]string{"service.beta.openshift.io/inject-cabundle": "true"},
+		},
```
evankanderson, line 41: Why not use `generateName` here?

This also feels like maybe this function should have a comment explaining what it does and why this is a safe way to get the CA cert.

evankanderson: (I think you're avoiding `generateName` because you want to simplify deleting this shortly after creation)

openshift/openshift.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+}
+
+// WithOpenShiftServiceCA enables trust to OpenShift's service CA for internal image registry
+func WithOpenShiftServiceCA() fnhttp.Option {
+	if !IsOpenShift() {
+		return nil
+	}
+
+	var selectCA func(ctx context.Context, serverName string) (*x509.Certificate, error)
```
evankanderson, line 96: This seems like it's finding the specific cert, not necessarily the CA for the cert (i.e. if you have a root CA that signs the registry cert, this will find the registry cert, not the root CA). That's okay, but we might prefer to call this `registryCert`.

openshift/openshift.go
evankanderson: I'm +1 on removing this as well -- between being scoped to the `openshift` package and operating only on local state (unlike `GetServiceCAs`), it seems simpler to be able to run this regardless of cluster detection.

openshift/openshift.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	cc := rawConf.Contexts[rawConf.CurrentContext]
+	authInfo := rawConf.AuthInfos[cc.AuthInfo]
+
+	user := "user"
+	parts := strings.SplitN(cc.AuthInfo, "/", 2)
+	if len(parts) >= 1 {
+		user = parts[0]
+	}
```
evankanderson, line 141: What is supposed to happen if `len(parts) == 1`? Is the user supposed to be `"user"`?

openshift/openshift.go
```diff
@@ -0,0 +1,176 @@
...
+	credentials := docker.Credentials{
+		Username: user,
+		Password: authInfo.Token,
```
evankanderson, line 144: This assumes that the token exchange has already happened. That seems reasonable, but it also seems worth documenting in case this is used in other circumstances.

openshift/openshift.go
```diff
@@ -0,0 +1,176 @@
...
+var checkOpenShiftOnce sync.Once
+
+func IsOpenShift() bool {
+	checkOpenShiftOnce.Do(func() {
+		client, err := k8s.NewKubernetesClientset()
+		if err != nil {
+			isOpenShift = false
+			return
+		}
+		_, err = client.CoreV1().Services("openshift-image-registry").Get(context.TODO(), "image-registry", metav1.GetOptions{})
```
evankanderson, line 168: Hmm -- this will return `false` if the user doesn't have `services` `get` permission on the `openshift-image-registry` namespace.

I'm not sure if there's another / better way to detect this, either from server headers or by probing the registry and looking at its header / challenge. I think `go-containerregistry` uses a `GET /v2/ HTTP /1.1` probe of the registry to figure out what and whether auth is required.

evankanderson (commented):
A few code review comments.

I'd like to see somewhat better coverage of the new package, but I understand that we'd like to bake this in rather than fooling around with credential helpers.

openshift/openshift.go
evankanderson: https://github.com/google/go-containerregistry/tree/main/pkg/authn#the-registry suggests looking at the `Www-Authenticate` header for clues as to the authentication mechanism.

openshift/openshift.go
matejvasek: I am using generated name so I can set up `Watch` before creation. Is that all right @evankanderson ?

openshift/openshift.go
matejvasek: It's CA for multiple services provided by OpenShift, but here for now are are enabling trust only for registry. I don't feel confident enabling the CA for any and all connections.

openshift/openshift.go
matejvasek: If token exchange did't happen (unlikely) then user will be prompted to enter credentials as usual.

openshift/openshift.go
matejvasek: I would avoid doing HTTP request to internal service here -- it requires some setup I don't want to do here.

matejvasek: >Hmm -- this will return false if the user doesn't have services get permission on the openshift-image-registry namespace.

@evankanderson  This does work with non-privileged user. Non-privilieged user will get different errors for non-existent service and for service that is not just permitted.

openshift/openshift.go
matejvasek: https://docs.openshift.com/container-platform/4.6/security/certificate_types_descriptions/service-ca-certificates.html

matejvasek: @evankanderson it should find self-signed CA -- authority not just certificate.

openshift/openshift.go
matejvasek: Also without the token user won't get this far anyway.

openshift/openshift.go
matejvasek: I added `else return nil`. But even without it, if the auth failed another auth method would kick in (e.g. CLI prompt).

cmd/client_openshift.go
matejvasek: Tags are no longer in use in latest version.

matejvasek:
> I'd like to see somewhat better coverage of the new package

Unfortunately good integration test here require OpenShift cluster that in not running in GH action.

matejvasek:
@lance @lkingland please lgtm

PR #828, opened 2022-02-11
Update guide for podman

matejvasek:
In the newer versions of `func` there is no need for starting the service and setting the `DOCKER_HOST` envvar.
It should be handled automatically.

### DIFF
--- a/docs/guides/podman.md
+++ b/docs/guides/podman.md
@@ -2,7 +2,7 @@
 
 Under Linux, it is possible to use [podman](https://podman.io/) instead of [docker](https://www.docker.com/). With Functions, you'll need `podman` version `v3.3` or better for this to work properly.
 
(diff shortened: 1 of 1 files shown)


matejvasek:
cc @abrennan89 @rh-max

docs/guides/podman.md
```diff
@@ -2,6 +2,11 @@
 
 Under Linux, it is possible to use [podman](https://podman.io/) instead of [docker](https://www.docker.com/). With Functions, you'll need `podman` version `v3.3` or better for this to work properly.
 
+For `func` version `v0.20.0` and latter no further setup should be needed,
```
lance, line 5: Nit:
```suggestion
For `func` version `v0.20.0` and later no further setup is needed,
```

PR #829, opened 2022-02-11
Fix --verbose mode for the version sub-command

matejvasek:


lkingland (approved):
Thanks for putting in the test!

Issue #809, opened 2022-02-02
On cluster build not working for a golang function


Shashankft9:
I am following this doc: https://github.com/knative-sandbox/kn-plugin-func/blob/main/docs/guides/on-cluster-build.md and have compiled func cli from main branch. 

I created function in go language. 
When I am doing `func deploy`, it fails with:
```
Error: function pipeline run has failed, please inspect logs of Tekton PipelineRun "my-function-git-pipeline-run-d92n4"
```
Upon taking a look at `step-create`, it says:
```
Paketo Go Build Buildpack 0.6.0
  Executing build process
    Running 'go build -o /layers/paketo-buildpacks_go-build/targets/bin -buildmode pie -trimpath .'
      Failed after 94ms
        go build: no main packages to build
failed to execute 'go build': exit status 1
ERROR: failed to build: exit status 1

```

I wonder if its because go functions require some scaffolding with boson go buildpack and that might not be included in the `create` step of `buildpacks` task here: https://raw.githubusercontent.com/tektoncd/catalog/master/task/buildpacks/0.3/buildpacks.yaml

zroubalik:
@Shashankft9 you are right, that's the problem.

@lance did this PR: https://github.com/boson-project/packs/pull/5

We just need to set the proper builder for Golang based func projects.

Shashankft9:
sure, related question - something like the PR you linked has to be done for rust runtime as well I believe? Since it uses one paketo community buildpack which is not included in the the paketo buildpacks builder.

Shashankft9:
> Do you think you can handle Rust as well? Thanks!

Sure! but we will need to host the rust builder somewhere, just like the go builder. I don't know how I can do that, any pointers? 
Note: the rust builder also deletes the source files, so the PR #830 will unblock that. 

> This is a good question, we need to tackle this? Maybe the global configuration could solve this? Users could overwrite default buildpacks there? WDYT?

I think this might not be an immediate issue, but might come up for users later who are already familiar with buildpacks and want to use their custom ones in on cluster builds, maybe the global configuration work will address some of it - I will open a separate issue for this

lance:
@jrangelramos do you think we can get an e2e test for this? I think it should work when using `func deploy --builder=host`.

jrangelramos:
Yea. I think so. I can manage that.

PR #830, opened 2022-02-14
configuring pipeline to avoid changing buildpacks

### LINKED ISSUE #827: Handling source files deletion in on-cluster-builds
While doing on cluster builds for go runtime, and making use of the builder `ghcr.io/boson-project/go-function-builder:tip` compiled from: https://github.com/boson-project/packs, I noticed that the custom buildpack in this builder is attempting to remove source files so that they dont appear in the final runnable image, and this won't really cause a problem when using the pack cli or its go client to build images - the problem happens if we use this builder in on-cluster-builds, where the source code is present in a pvc backed workspace, because now the removal of source files will cause mutation of the source code, and hence the tekton tasks after the build task that are either reading or in general using the source files _will fail_, specifically the `image-digest` and `deploy` tasks.
In case of usual `pack build`, the source files are uploaded to a new container from local machine fs, and so mutations are done to the source files in that container and not the local machine. 

As far as I could understand from the buildpacks slack thread here: https://buildpacks.slack.com/archives/C94UJCNV6/p1644416841067679, there is no standardization around the source file removal aspect in a buildpack currently, some languages do it and some don't. For example, for nodejs, the paketo builder wont remove the source files, but for rust, the paketo community rust builder will do the removal, same is the case for the boson go builder rn. 

Two ways this issue can be handled:

1. configuring the builder/buildpack to avoid removing the source files, which I feel is a overhead and can have certain implications around security, size of the final runnable image etc
for boson go builder, here is a PR attempting this: https://github.com/boson-project/packs/pull/8

2. changing the tekton pipeline - to backup the source files in a different path in the workspace before running the build task, and then restoring the files back to the original path for post build tasks. 
I have created a sample pipeline for this, which is all green without accommodating the PR changes in point 1 - I am cleaning it up, and will create a PR for it soon. 

ref: https://knative.slack.com/archives/CUU6LEMRT/p1643893973744699

Shashankft9:
# Changes

- adds two tekton tasks: `backup-source` and `restore-source` to avoid configuring buildpack/builder to _not_ remove the source files
- using boson go builder

/kind bug

Fixes #827  #809


zroubalik (commented):
@Shashankft9 do we really need to do this? I am sorry, I haven't had a time to follow the recent stuff. But I'd be definitely more inclined towards changing the builders in case they are removing anything from the source code. Adding new tekton tasks brings a lot of additional complexity.

If these kind of chnages are necessary, then we should include them into a build task, so the source is in a correct state after running the build step.

zroubalik:
> @zroubalik hey, regarding changing pipeline vs changing buildpacks - I have tried to capture the gist from the slack threads here: #827, wdyt about that?
> 
> And yeah, I agree, it would be better to anyway add these 2 tasks as steps in `build` task.

Thanks, I will check that issue, I missed it somehow, sorry.

Shashankft9:
@zroubalik thoughts on this? should I try the change of putting the new 2 tasks as steps instead - if this is how we would like to proceed considering the issue mentioned above.

Shashankft9:
@zroubalik added the build task yaml file in the pipeline resources of the kn-plugin-func repo - with 2 steps, backup and restore of source code. 
because of this, the documentation will change slightly for the apply of buildpack task, but I can take care of that in another PR, let me know.

zroubalik:
> @zroubalik added the build task yaml file in the pipeline resources of the kn-plugin-func repo - with 2 additional steps, backup and restore of source code. because of this, the documentation will change slightly for the apply of buildpack task, but I can take care of that in another PR, let me know.

Sorry for the delay, though I'd prefer if we change the Go and Rust builders to not delete the function source code and config, so we have consistent experience with all runtimes. WDYT?

Shashankft9:
> so we have consistent experience with all runtimes

thats actually my concern, in the slack thread of buildpacks I have linked in the original issue, i got to learn that there really is no standard in buildpacks today to handle source file deletion, so some do and some don't. So I see two choices, either to change buildpacks that do the source file deletion, maybe in future there will be more OR handle it like this. 
I do see your point though, in case if we are going to maintain buildpacks, then it makes sense to do the changes at buildpacks level, but the pipeline change still looks more simpler to me - although, I havent really worked much with CICD systems so I wouldn't know how hard it would be to do the same for systems other than tekton etc. 
and then there are security concerns like https://github.com/knative-sandbox/kn-plugin-func/issues/357 that @matejvasek briefly talked about in the slack thread.

matejvasek:
BTW I just noticed that paketo java BP probably does delete sources.

lance:
@Shashankft9 I am going to go ahead and close this, since I think the problem has been addressed with a less resource intensive workaround in this PR https://github.com/knative-sandbox/kn-plugin-func/pull/1070. Thanks for your contribution in any case!

PR #833, opened 2022-02-16
Update Knative dependencies to v1.0 (0.27)

dsimansk:
# Changes
- :broom: Update Knative dependencies to v1.0 (0.27)
- Split into two commits to separate all `vendor/` dir updates
- NOTE: `tektoncd/*` is also updated to align with matching `k8s.io/` version

/cc @lance


dsimansk:
> @dsimansk is `pkged.go` meant to be a part of this PR? Has something changed in `/templates`?

Hmm strangely it was re-generated when I tried to run `make test` as a smoke test. I don't think it should be updated though.

matejvasek:
> > @dsimansk is `pkged.go` meant to be a part of this PR? Has something changed in `/templates`?
> 
> Hmm strangely it was re-generated when I tried to run `make test` as a smoke test. I don't think it should be updated though.

The pkger also tracks mod time of files so even if you didn't changed content of files locally they still may have different mod time stamp. It's annoying.

PR #835, opened 2022-02-16
Use custom zip file embedding not pkger

matejvasek:
Getting rid of `pkger`.
* Added test that checks the embedded filesystem has the same content as `./templates`.
* Builds are now reproducible since the mod time is not set.
* Generation is much faster (30s vs. 400ms on my machine)


Makefile
```diff
@@ -64,22 +59,6 @@ check: bin/golangci-lint ## Check code quality (lint)
 bin/golangci-lint:
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ./bin v1.43.0
 
-pkged.go: $(TEMPLATES)
-	@rm -rf templates/node/cloudevents/node_modules
```
matejvasek, line 68: We probably still should do the clean up.

filesystem.go
```diff
@@ -18,85 +20,99 @@ import (
...
-	Open(path string) (file, error)
-	ReadDir(path string) ([]os.FileInfo, error)
+	fs.ReadDirFS
+	fs.ReadFileFS
 }
 
-type file interface {
-	io.Reader
-	io.Closer
+type IFS interface {
```
matejvasek, line 27: Unused interface.

filesystem.go
```diff
@@ -1,13 +1,15 @@
 package function
 
 import (
+	"embed"
```
matejvasek, line 4: Duplicate import.

lkingland:
Embed would solve half the equation by embedding files into the binary, but it requires that the user be building a binary, as it is evaluated as a build-time instruction.   We want source-level embedding.  For instance, if we used go:embed, anyone importing the Func library would not have access to the embedded templates.
There are two places code or instructions can be added to the development proces:  Pre-source code or Post-source code.  Pre-source includes metaprogramming in the form of code generation (such as how our func.yaml schema is generated).  Post-source includes instructions on how to use the source code to alter the resulting artifcates (such as how our Makefile builds binaries for alternative operating systems and architectures).
For embedded templates, we want Pre-source code generation, so that the contents of the templates are encoded as Go source.
Go:embed is a Post-source model, and as such takes place at the wrong time for our uses.

lkingland:
Thanks to the new fs package, it is quite straightforward to use an embedded tar or zip as a Filesystem at runtime (maybe zip because it retains file permissions?).  We would just need to tar up the templates directory as single (or more likely better a set of) .go source files with the tar encoded and embedded static member.  These files would be generated by make templates when changes are detected to ./templates; just as it is now.

lkingland:
You can see how go:embed would have worked here:
https://github.com/knative-sandbox/kn-plugin-func/pull/350
(mainly generate.go and the tarfs package therein, which is in turn used by templates.go)

lkingland (changes requested):
To replace `pkger`, we should encode `templates` as source code (currently `pkged.go`), and present a `fs.FS`.  This way our templates can continue to be used when our library is imported.

matejvasek:
@lkingland I trying to change the approach here. I am using `go generate` to generate byte array containing a zip.
It works on unix machines for now.

With Windows there are issues since some code assumes just `/` all the way, while some other parts of code assume OS depended separator.

matejvasek:
@lkingland check this out https://github.com/knative-sandbox/kn-plugin-func/pull/835/commits/21a457f61ea71b2a1720e3f7f3263a25c2393c22

lkingland:
> @lkingland check this out [21a457f](https://github.com/knative-sandbox/kn-plugin-func/commit/21a457f61ea71b2a1720e3f7f3263a25c2393c22)

Oh yes, I think this is definitely possibly a workable solution.  There's a lot of code in there so I will do my best to dig in deeper tomorrow, but so far looks good!

filesystem_test.go
```diff
@@ -0,0 +1,240 @@
...
+			fileSystem: initGitFS(t),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			templatesFS := tt.fileSystem
+
+			if templatesFS == nil && runtime.GOOS == "windows" {
+				t.Skip("FS == nil")
```
matejvasek, line 47: Interestingly this issues doesn't happen when I force `tmp` files not to be created on `C:` drive but on `Z:` drive -- tested on `wine`.

matejvasek:
Ignore the failing Win tests, it's not fault of this PR.

matejvasek:
I did local testing with embedded and git templates using `wine`. It seems to work.

matejvasek:
@lkingland give it another look.

matejvasek:
@jrangelramos run our internal Windows tests and it seems to pass.

filesystem_test.go
lkingland: That is indeed strange

lkingland (approved):
I am happy with this, broadly.  I am sure that we may make tweaks as time goes on, but getting rid of pkger (a temporary POC solution from the beginning), is worth almost any future pain working through any emergent bugs.  Looking good!

matejvasek:
>That is indeed strange

@lkingland this is probably similar error to the one we saw in the other PR.
I have no idea why that's happening in CI. I kinda works in `wine`.

matejvasek:
> That is indeed strange

I thought that it could be caused by some `file://` URI type issue.
But even when I used http git server it didn't work.

matejvasek:

zroubalik (approved):
/lgtm

Great change!

/hold for others to review

matejvasek:
There is still issue with generating the zip on Windows: the windows FS doesn't have exec permissions.
Maybe we could use file permissions from `git`? WDYT?

generate/templates/main.go
```diff
@@ -0,0 +1,122 @@
...
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"log"
+	"os"
+	"path/filepath"
+)
+
```
lance, line 15: I think it would be good to add some comments to this file, with an overall description of its purpose here at the beginning.

generate/templates/main.go
```diff
@@ -0,0 +1,122 @@
...
+			}
+		}
+
+		_, err = fmt.Fprintf(srcOut, "\t")
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		for i, b := range buff[:n] {
+			_, err = fmt.Fprintf(srcOut, "0x%02x,", b)
```
lance, line 99: This is actually what prompted me to ask for some comments. Do I understand correctly that this writes each byte to the `srcOut` writer in hex, with a `0x` prefix? It took me a minute to mentally parse all of the x's and 0's.

Makefile
```diff
@@ -143,17 +135,17 @@ cross-platform: darwin linux windows ## Build all distributable (cross-platform)
 
 darwin: $(BIN_DARWIN) ## Build for Darwin (macOS)
 
-$(BIN_DARWIN): pkged.go
+$(BIN_DARWIN): zz_filesystem_generated.go
```
lance, line 138: Definitely prefer this name change.

lance (commented):
I have mainly a couple of small questions. Overall, this looks like a very nice change to how we package templates.

/lgtm
/hold in case anyone else would like to review

generate/templates/main.go
matejvasek: yes, %02x -- means hex format with width padding

Makefile
matejvasek: @lance  Like it's ok, or you would rename it?
`zz_` is just prefix to push it down in diffs.

filesystem.go
```diff
@@ -1,92 +1,148 @@
...
+func (z zipFS) Stat(name string) (fs.FileInfo, error) {
+	f, err := z.archive.Open(name)
 	if err != nil {
 		return nil, err
 	}
-	return f.Readdir(-1)
+	return f.Stat()
 }
 
+//go:generate go run ./generate/templates/main.go
```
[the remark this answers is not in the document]
matejvasek: It generates variable `templatesZip` that is used here, just few lines below.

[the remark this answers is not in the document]
matejvasek: >//go:generate go run ./generate/templates/main.go

Could be probably in any file but I put it here since the product of the generation (the byte array) is used here.

Makefile
lance: It's OK - I like it better than `pkged.go` for that reason.

> @lance Like it's ok, or you would rename it? `zz_` is just prefix to push it down in diffs.

lance (commented):
/lgtm

lkingland:
> There is still issue with generating the zip on Windows

I am in favor of expecting contributors to the project (as distinct from users) to use Windows Subsystem for Linux.

matejvasek:
If there are no objections I'll merge this.

matejvasek:
I need to rebase.

matejvasek:
@lance @zroubalik @lkingland I need re-lgtm because of rebase.

PR #837, opened 2022-02-17
Fix checkum for github.com/tektoncd/cli

dsimansk:
# Changes

- :bug: Fix checkum for github.com/tektoncd/cli

/cc @lance @matejvasek


matejvasek:
@dsimansk thanks
/lgtm

PR #838, opened 2022-02-17
Updating to latest spring cloud release and refactoring code

### LINKED ISSUE #832: Updating Spring Boot templates to use latest versions of Spring Cloud Functions and other minor updates
The Spring Boot template has some strange things and relies on an old version of Spring Cloud Functions that has some issues that were already fixed in newer versions

- Update to Spring Cloud Functions 3.2.2
- Use Spring Boot Actuators instead of @Bean for health checks
  - Change the health endpoints in `func.yaml`
- Remote company reference in example
- Improve Function Signature to use Message for both Input and Output and remove CloudEvents enricher, that only works for very limited use cases.

salaboy:
# Changes

Fixes #832  

/kind cleanup

:broom: Cleanups
- Change package name to match the function example, also avoid using special package name that gets scanned by spring cloud functions
- Use actuators instead of special bean for healthchecks, hence new health endpoints
- Implement simple echo function to match other templates
- Updating to latest Spring Cloud release 3.2.2 where several fixes were introduced to improve compatibility with CloudEvents SDK
- Disable native build by default, still can be enabled with a flag


salaboy:
@lkingland @lance based on what we discussed today this aligns with go and typescript now

templates/springboot/cloudevents/pom.xml
```diff
@@ -18,8 +18,8 @@
 
   <properties>
     <java.version>11</java.version>
-    <spring-cloud.version>2020.0.4</spring-cloud.version>
-    <spring-native.version>0.10.3</spring-native.version>
+    <spring-cloud.version>2021.0.1</spring-cloud.version>
+    <spring-native.version>0.11.3-SNAPSHOT</spring-native.version>
     <compiler-plugin.version>3.8.1</compiler-plugin.version>
     <log4j2.version>2.16.0</log4j2.version>
```
trisberg, line 24: this is no longer needed after boot upgrade, 2.6.3 pulls in `2.17.1`

templates/springboot/cloudevents/pom.xml
```diff
@@ -18,8 +18,8 @@
 
   <properties>
     <java.version>11</java.version>
-    <spring-cloud.version>2020.0.4</spring-cloud.version>
-    <spring-native.version>0.10.3</spring-native.version>
+    <spring-cloud.version>2021.0.1</spring-cloud.version>
+    <spring-native.version>0.11.3-SNAPSHOT</spring-native.version>
```
trisberg, line 22: this `0.11.3` version is now released

trisberg:
Should we make the Spring Boot version changes for the http template as well, or is that a separate PR?

matejvasek:
@salaboy since this modifies `./templates` you will have to rebase this on `main` are regenerate `zz_filesystem_generated.go`

templates/springboot/cloudevents/manifest.yaml
```diff
@@ -1,7 +1,9 @@
-# optional.  Invocation defines hints for how Functions created using this
-# template can be invoked.  These settings can be updated on the resultant
-# Function as development progresses to ensure 'invoke' can always trigger the
-# execution of a running Function instance for testing and development.
-invocation:
-  # Invocations of Functions from this template is via basic HTTP CloudEvent
-  format: "cloudevent"
```
lance, line 7: Why remove these lines from the yaml?

salaboy: good catch!

salaboy:
`zz_filesystem_generated.go` added, thanks @matejvasek for pointing that out!

salaboy:
@matejvasek @lance can we approve this one?

templates/springboot/cloudevents/manifest.yaml
```diff
@@ -1,3 +1,11 @@
+builders:
+  default: gcr.io/paketo-buildpacks/builder:base
+buildEnvs:
+  - name: BP_NATIVE_IMAGE
+    value: "false"
```
matejvasek, line 5: @salaboy  shouldn't this be false also for spring level manifest that is used for `HTTP` template?

matejvasek: also the health endpoint, shouldn't they be set for http template too.

matejvasek:
@salaboy other than the comment above this looks good to me.

templates/springboot/cloudevents/manifest.yaml
lance: > also the health endpoint, shouldn't they be set for http template too.

This implies that the `buildEnv` properties should be set a directory above in `templates/springboot/manifest.yaml`

lance (commented):
/lgtm other than the small change to `buildEnvs`.
/hold for possible changes there

templates/springboot/cloudevents/manifest.yaml
salaboy: @lance I will send another PR for the HTTP template

PR #842, opened 2022-02-21
feat: builder flag with s2i option

lkingland:
# Changes

- :gift: can specify the S2I builder with --builder flag

The default builder for Functions is Pack, with the option to specify S2I using `--builder s2i`.

WIP:
Currently only `--language node` is supported
The ability to use custom implementations of Builder in the Build command is temporarily unavailable while we consolidate the Client creation process into a single default constructor in #824 and #825.

Update 2/28:  The aforementioned are now merged, so this is now dependent on #865 #868 and #869, and probably a few more.

Update 3/16:  The aforementioned are now merged or closed, so this is dependent on #883

/kind enhancement


buildpacks/builder.go
```diff
@@ -21,6 +20,12 @@ import (
 	"github.com/buildpacks/pack/pkg/logging"
 )
 
+// DefaultBuilderImages for Pack builders indexed by Runtime Language
+var DefaultBuilderImages = map[string]string{
+	"node": "gcr.io/paketo-buildpacks/builder:base",
+	"go":   "gcr.io/paketo-buildpacks/builder:base",
```
lance, line 26: Will this eventually have a key for all runtimes supported by paketo?

lkingland: Yes, and more specifically for all runtimes we support.  For users providing completely new runtimes/templates that we are not aware of and map here, they will need to set that value in their custom language pack's `manifest.yaml`

lkingland:
Closed in favor of #923

PR #860, opened 2022-02-22
Windows unit test fixes

matejvasek:
* force unit test run by using `.PHONY` in `Makefile`
* fixed issues related to using back slash in path part of URI
* disabled few tests on Windows since I am not able to debug it

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,8 @@ LDFLAGS := "-X main.date=$(DATE) -X main.vers=$(VERS) -X main.hash=$(HASH)"
 CODE := $(shell find . -name '*.go') pkged.go go.mod schema/func_yaml-schema.json
 TEMPLATES := $(shell find templates -name '*' -type f)
 
+.PHONY: test
+
 # Default Targets
 all: build
 
--- a/docker/creds/credentials.go
+++ b/docker/creds/credentials.go
@@ -395,7 +395,7 @@ func listCredentialHelpers() []string {
 			}
 			if runtime.GOOS == "windows" {
 				ext := filepath.Ext(fi.Name())
-				if ext != "exe" && ext != "bat" {
+				if ext != ".exe" && ext != ".bat" {
 					continue
 				}
 			}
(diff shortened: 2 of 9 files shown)


matejvasek:
The unit test must have been broken for Windows for some time now.
For some reason the test were not run in CI,
the make command though the `test` recipe is satisfied (`mingw32-make: 'test' is up to date.`).

matejvasek:
Even after few fixes this still fails. I have absolutely no idea why.
The Win tests pass all right on my machine: `fedora`+`wine`.

matejvasek:
Could anybody with Windows machine try this?

matejvasek:
@lkingland @lance PTAL

lkingland:
With so many tests disabled in-code, I am leaning towards perhaps a different solution:

Could we just duisable Windows from the CI task entirely, and open a _new_ Issue for fixing Wndows the unit tests?  That would require merely [removing `windows-latest` from the CI task](https://github.com/knative-sandbox/kn-plugin-func/blob/main/.github/workflows/test-unit.yaml#L12), and be quite obvious what needs to be fixed via the issue.

Anyone else have an opinion?

matejvasek:
@lkingland better to run some test than none. Right know there are only issues with custom git repositories.

matejvasek:
Note: At first I suspected there is some issue with `file://` URLs on windows, but even when I switched to `http://` it didn't work.

matejvasek:
@lkingland @lance PTAL I think we should merge this as it is.

lkingland:
> @lkingland @lance PTAL I think we should merge this as it is.

That sounds good to me too.  We need to keep the momentum towards our upcoming release.  Let's just make sure we circle back to this when we have time.  These would be a great task to ask others with windows-specific experience to help debug.

/lgtm

lkingland (approved):

matejvasek:
https://github.com/knative-sandbox/kn-plugin-func/issues/864

PR #862, opened 2022-02-23
fix: minor typos in docs

helio-frota:
Just some typo fix


matejvasek:
Thanks for the contribution @helio-frota !

PR #865, opened 2022-02-28
src: update root and version structure and help text

lkingland:
# Changes

The root and version commands were far out of sync with more recent command structures.  This minor PR updates them to the more recent format, updates their help text, and explicity references verbosity level in the version command.

- :broom: Update root command structure to match updated command structure
- :broom: Update root help text to match recent structure and deemphasize explicit Function name
- :broom: Update Version command to match updated command structure
- :broom: Adds an explicit check for verbosity in Version to fix what appears to be a bug
- :broom: Cleans up Version verbosity logic to be more clear about defaults

/kind cleanup

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -131,11 +131,6 @@ func TestRoot_CMDParameterized(t *testing.T) {
 		t.Fatalf("default command use should be \"func\".")
 	}
 
-	usageExample, err := replaceNameInTemplate("func", "example")
-	if root.Example != usageExample || err != nil {
-		t.Fatalf("default command example should assume \"func\" as executable name. error: %v", err)
-	}
-
 	rootConfig = RootCommandConfig{
 		Name: "kn func",
 	}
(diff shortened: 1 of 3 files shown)


zroubalik (approved):
/lgtm

/hold for others to review

cmd/root_test.go
```diff
@@ -131,11 +131,6 @@ func TestRoot_CMDParameterized(t *testing.T) {
 		t.Fatalf("default command use should be \"func\".")
 	}
 
-	usageExample, err := replaceNameInTemplate("func", "example")
```
lance, line 134: It's too bad that we lose these tests - could `runRootHelp` perhaps be made more easily testable by setting the command's output stream?

lance (commented):
/lgtm

cmd/root_test.go
lkingland: Yes @lance that's do-able and a good idea.  I have now refactored that test to confirm actual command output contains usage text with a properly substituted command name string.

lance (commented):
/lgtm

matejvasek:
I've just restarted the tests in case it's just a flake.

lance (approved):
/unhold

PR #868, opened 2022-02-28
Lkingland/openshift int test mode

### LINKED ISSUE #867: Tests failing locally
Running `go test ./...` fails with:

```
--- FAIL: TestRoundTripper (5.01s)
    openshift_test.go:27: Get "https://image-registry.openshift-image-registry.svc.cluster.local:5000/v2/": Post "https://127.0.0.1:54405/api/v1/namespaces/default/pods": dial tcp 127.0.0.1:54405: connect: connection refused
```

This appears to be because of the implementation of the new `IsOpenShift` function which assumes any error other than `NotFound` means the cluster is OpenShift:

```
func IsOpenShift() bool {
  checkOpenShiftOnce.Do(func() {
    client, err := k8s.NewKubernetesClientset()
    if err != nil {
      fmt.Print("error in NewKubernetesClientset")
      isOpenShift = false
      return
    }
    _, err = client.CoreV1().Services("openshift-image-registry").Get(context.TODO(), "image-registry", metav1.GetOptions{})
    if k8sErrors.IsNotFound(err) {
      isOpenShift = false
      return
    }
    // FIXME(lkingland): This seems to ignoire legitimate errors other than NotFound
    isOpenShift = true
  })
  return isOpenShift
}
```

As Evan mentioned in the PR [in which this check was added](https://github.com/knative-sandbox/kn-plugin-func/pull/825), there might be a better way to detect this:

> Hmm -- this will return `false` if the user doesn't have `services` `get` permission on the `openshift-image-registry` namespace.
> 
> I'm not sure if there's another / better way to detect this, either from server headers or by probing the registry and looking at its header / challenge. I think `go-containerregistry` uses a `GET /v2/ HTTP /1.1` probe of the registry to figure out what and whether auth is required.

_Originally posted by @evankanderson in https://github.com/knative-sandbox/kn-plugin-func/pull/825#discussion_r811520395_

The error appears to be a simple connection refused (no cluster is running locally, as one is not needed to run unit tests).

If this is expected behavior, and the test depends on a cluster being available, perhaps tagging it as an integration test would to the trick?
```
//go:build integration
// +build integration
```

lkingland:
# Changes

The OpenShift integration test was running in unit test mode; simply needing the tag.

- :bug: OpenShift integration test running in unit test mode

/kind bug

Fixes #867

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -16,18 +15,6 @@ import (
 	fn "knative.dev/kn-plugin-func"
 )
 
-var exampleTemplate = template.Must(template.New("example").Parse(`
-# Create a node function called "node-sample" and enter the directory
-{{.}} create myfunc && cd myfunc
-
-# Build the container image, push it to a registry and deploy it to the connected Knative cluster
-# (replace <registry/user> with something like quay.io/user with an account that have you access to)
-{{.}} deploy --registry <registry/user>
-
-# Curl the service with the service URL
-curl $(kn service describe myfunc -o url)
-`))
-
 type RootCommandConfig struct {
 	Name      string // usually `func` or `kn func`
 	Date      string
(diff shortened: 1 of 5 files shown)


matejvasek:
this contains a lot fo changes that does not seem to be related

matejvasek:
Closing in favor of https://github.com/knative-sandbox/kn-plugin-func/pull/870.

PR #869, opened 2022-02-28
src: promote verbosity to constructor arg

lkingland:
# Changes

**tl;dr** Probably best to use _either_ constructor args _or_ struct member initialization and not mix.  We have chosen constructors for package public API and struct members for package internal implementations.  This enforces that choice with regards to the oft-ignored verbosity flag.

- :broom: Moves verbosity flag to constructor arg

This rectifies a long-standing inconsistency where, early in the system's evolution all subsystems were initialized as simple structs.  When we moved to using proper constructors, the Verbosity flag was left as a struct member.  This is a bit confusing and awkward because it mixes constructor-based configuration with the direct setting of struct members.  The former is often best for external package APIs, the latter is often best for internal.  This situation straddles the line, because these packages are used by their parent package, so either way could be a defensible position.

Explicitly requiring the verbosity to be provided in the constructor is not awe-inspiring elegance, but I suppose mixing constructor-based initialization with struct members much less so.  This is evidenced by how much `main`'s default Client constructor is simplified by requiring this flag in the constructor.

/kind cleanup

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -41,7 +41,6 @@ DESCRIPTION
 	cmd.Run = func(cmd *cobra.Command, args []string) {
 		runVersion(cmd, args, version)
 	}
-	cmd.SetHelpFunc(defaultTemplatedHelp)
 
 	return cmd
 }
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"text/template"
 
 	"github.com/ory/viper"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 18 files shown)


cmd/version.go
```diff
@@ -41,7 +41,6 @@ DESCRIPTION
 	cmd.Run = func(cmd *cobra.Command, args []string) {
 		runVersion(cmd, args, version)
 	}
-	cmd.SetHelpFunc(defaultTemplatedHelp)
```
lance, line 44: Why is this line removed?

lance (commented):
/lgtm

cmd/version.go
lkingland: Hm, this looks like a casualty of merge.  Thanks for noticing.

lkingland: I went through and checked all the other commands for a similar merge error. Should be GTG now.

zroubalik (approved):
/lgtm

PR #873, opened 2022-03-01
fix: use full root name for cmd help prefixes

### LINKED ISSUE #872: Subcommand help text incorrect when plugin
The [recent update](https://github.com/knative-sandbox/kn-plugin-func/commit/1b8c2c3cf72e84d612fa75532d7cf9b2e877faf4) to replace the command help text prefix environment variable with a call to the root command's `.Name()` is unfortunately not fully functional.  It does work correctly on the root command but not on the subcommands:

When name is a multi-token value (such as when in plugin mode, and the value is set to `kn func`), the value is truncated to the first token.  For example, when run as `kn func invoke` (plugin mode), output is:
```
$ kn func invoke --help

NAME
	kn invoke - Invoke a Function.
...
```
note the missing `func` subcommand.  This is due to the Cobra package truncating `cmd.Root().Name()` to only the first token.   The quick fix for this would probably be to use `cmd.Root.Use` (which can be an arbitrary string) to carry the full Name, rather than revert to the old environment variable approach which is good to have refactored away.

lkingland:
# Changes

Small change to use `cmd.Root().Use` for subcommand help text because `.Name()` is truncated, leading to incorrect values when run in plugin mode (see #872)

- :bug: fix help text subcommand prefixes when a plugin

/kind bug

Fixes #872

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -172,7 +172,7 @@ func runCreateHelp(cmd *cobra.Command, args []string, newClient ClientFactory) {
 		Name    string
 	}{
 		Options: options,
-		Name:    cmd.Root().Name(),
+		Name:    cmd.Root().Use,
 	}
 	if err := tpl.Execute(cmd.OutOrStdout(), data); err != nil {
 		fmt.Fprintf(cmd.ErrOrStderr(), "unable to display help text: %v", err)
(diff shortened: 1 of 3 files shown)


zroubalik (approved):
/lgtm

PR #878, opened 2022-03-03
Relax permission check only to owner exec perm

matejvasek:
Make test more permissive since only that really matters is that whether the file is executable for owner.


lance (commented):
/lgtm

PR #879, opened 2022-03-04
Generate templates zip unconditionally

matejvasek:
Changes:
* Generate templates zip unconditionally (i.e. no `make` prerequisites).
  Rationale:
  * The way we create prerequisites (the `find` command) we cannot detect deleted files.
  * The generation is fast (around 500ms) so we can afford that.
* Some tiny performance tweaks.
* Test embedded templates content on push to `main`.
   The test has to run prior to `make test` because the `test` recipe enforces generation of the templates zip.
   It would prevent test from being run against git committed template.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,6 +15,8 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: '11'
+      - name: Check embedded templates content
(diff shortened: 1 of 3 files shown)


matejvasek:

lkingland (approved):
/lgtm

Completely agree!

PR #881, opened 2022-03-04
Make `func invoke` print any response it receives.

lance:
# Changes

- :gift: Make `func invoke` print any response it receives

/kind enhancement

**Release Note**

```release-note
`func invoke` now prints any response it receives from the invoked function
```


lkingland (approved):
Good addition;  It should have been like this in the first place!
Also, all the templates should, by default echo the request received both to the response channel and to local logs

/lgtm

PR #882, opened 2022-03-07
Fix return value of goByteArrayWriter.Write()

matejvasek:
The `goByteArrayWriter.Write()` shall return number of dumped bytes (hex literals) not number of bytes written to underlying writer (which is approximately 5x more).


matejvasek:
@lkingland PTAL

lkingland (approved):
/lgtm

PR #883, opened 2022-03-08
src: exported client constructor with options

lkingland:
# Changes

- :broom: simplified client constructor and direct usage of fn.Option

A prerequisite for #842, this PR adds back in the ability to provide custom Client concrete implementations of subsystems, while retaining the new centralized default Client instantiation (which is in place of the previous Client Factories).

Other than some minor cleanup throughout `cmd`, changes worth explaining a bit include:

### `NewClient` Constructor
  The recent inclusion of the `verbose` flag as a constructor argument
  allowed for further simplification of the `Client` instantiation block; now
  extracted to be simply `NewClient()`

  Furthermore, the ClientFactory itself can be directly referenced in the method signature of the NewClient constructor,
  which adds flexibility while retaining the ability to fully mock for tests.

### WithVerbose Options
  Subsystems which use Functional Options have been updated to include
  a WithVerbose Option rather than a standalone constructor argument.

### ClientOptions Struct
  The ClientOptions struct was renamed to ClientConfig, added to the signature are actual Client Options (`...fn.Option`)
  such that individual commands can directly utilize the ClientFactory, and provide their own fn.Options as needed.
  This includes removing `XConfigToClientOptions` methods because the number of members would have been merely
  two (namespace and verbose), and the ClientFactory is no longer different from the method signature of NewClient itself.
 This assistance method could of course be added back if we move back to having a larger number of Client construction
  arguments.

### ClientFactory
  Now even further simplified, using a standard `NewClient` constructor,
  since the deferred instantiation is not necessary now that Client Options
  are provided using actual client options (`[]fn.Option`) in stead of a
  rather than a separate struct.

### Command Args
  To avoid edge cases where some commands were failing in IDEs which precomile
  tests, we are now explicitly indicating that arguments provided to the test command
  are not in turn propagated to command objects created within tests by ensuring
  `cmd.SetArgs([]string{})` is used throughout.

### Version
  Now using the Version struct in the root command directly rather than mapping through
  an interstitial config struct.

### IsOpenShift
  Places where OpenShift status results in a different implementation have been
  extracted into separate, dedicated functions for clarity, including `newTransport()` and `newCredentialsProvider(transport)`.

### Repositories Tests in `cmd`
  Now runs from a temp directory defined using XDG_CONFIG_HOME rather than
  an in-memory-only mock and removed several chuncks of complexity that was
 supporting the former methodology.

### Flags Tests and Bugs
  Added and upgraded flags tests, including fixing a few missing flag mappings.

/kind cleanup


matejvasek:
@lkingland I still believe the factory should be settable from top level to `NewRootCmd()` for sake of overall wiring testing.

Lately we had issue where refactor broken `--verbose` flag wring.
That could have been easily detcted by unittest that could look like:
```go
func TestVerboseFlagPropagation(t *testing.T) {

	tests := []struct {
		name string
		args []string
	}{
		{
			name: "verbose as sub-commands flag",
			args: []string{"build", "-v"},
		},
		{
			name: "verbose as roots flag",
			args: []string{"-v", "deploy"},
		},
		{
			name: "verbose as sub-sub-command flag",
			args: []string{"repositories", "list", "-v"},
		},
	}

	for _, tt := range tests {
			var err error
			defer viper.Reset()
			verboseSetByCommand := false
			cmd, err := NewRootCmd(RootCommandConfig{
				Name:    "func",
				Date:    "",
				Version: "0.0.0",
				Hash:    "",
				NewClient: func(opts ClientOptions) *fn.Client {
					verboseSetByCommand = opts.Verbose
					return fn.New() // noop client
				},
			})
			if err != nil {
				t.Fatal(err)
			}
			cmd.SetArgs(tt.args)
			err = cmd.Execute()
			if err != nil {
				t.Fatal(err)
			}
			if !verboseSetByCommand {
				t.Error("expected verbose to be set to true by the sub-command")
			}
		})

	}

}
```

lkingland:
I added back in a root-level ClientFactory and added the following test for both the global flags, `namespace` and `verbose`. Thanks @matejvasek PTAL

```go
func TestRoot_PersistentFlags(t *testing.T) {
  tests := []struct {
    name          string
    args          []string
    skipNamespace bool
  }{
    {
      name: "provided as root flags",
      args: []string{"--verbose", "--namespace=namespace", "list"},
    },
    {
      name: "provided as sub-command flags",
      args: []string{"list", "--verbose", "--namespace=namespace"},
    },
    {
      name:          "provided as sub-sub-command flags",
      args:          []string{"repositories", "list", "--verbose"},
      skipNamespace: true, // NOTE: no sub-sub commands yet use namespace, so it is not checked.
    },
  }

  for _, tt := range tests {
      defer fromTempDir(t)()                    // from a temp dir, deferred cleanup
      cmd := NewCreateCmd(NewClient)            // Create a new Function
      cmd.SetArgs([]string{"--language", "go"}) // providing language (the only required param)
      if err := cmd.Execute(); err != nil {     // fail on any errors
        t.Fatal(err)
      }

      // Assert the persistent variables were propagated to the Client constructor
      // when the command is actually invoked.
      cmd = NewRootCmd("func", Version{}, func(cfg ClientConfig, _ ...fn.Option) (*fn.Client, func()) {
        if cfg.Namespace != "namespace" && !tt.skipNamespace {
          t.Fatal("namespace not propagated")
        }
        if cfg.Verbose != true {
          t.Fatal("verbose not propagated")
        }
        return fn.New(), func() {}
      })
      cmd.SetArgs(tt.args)
        t.Fatal(err)
      }
    })
  }
}
```

cmd/repository.go
```diff
@@ -12,29 +12,10 @@ import (
...
-	client := repositoryClient{fn.New(
-		fn.WithRepositories(cfg.Repositories),
-		fn.WithVerbose(cfg.Verbose))}
-	return cfg, client, nil
-}
-
 // command constructors
 // --------------------
 
-func NewRepositoryCmd(clientFn repositoryClientFn) *cobra.Command {
```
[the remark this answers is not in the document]
zroubalik: +100

zroubalik (approved):
/lgtm

this is a great change!

cmd/root.go
```diff
@@ -471,3 +480,49 @@ func defaultTemplatedHelp(cmd *cobra.Command, args []string) {
 		fmt.Fprintf(cmd.ErrOrStderr(), "unable to display help text: %v", err)
 	}
 }
+
+// Create and return a path to a temp dir prefixed 'func'.  Does not change
```
matejvasek, line 484: @lkingland consider using `t.TempDir()` for temporary directories creation.

matejvasek:
I see something like:
```go
os.Setenv("XDG_CONFIG_HOME", newTempDir())
```
at several places.

I might be good idea set the envvar to previous value after test.
I believe there is helper function in the `testing` package for that.

cmd/root.go
```diff
@@ -104,18 +117,19 @@ EXAMPLES
 	cmd.AddCommand(NewInfoCmd(newClient))
 	cmd.AddCommand(NewListCmd(newClient))
 	cmd.AddCommand(NewInvokeCmd(newClient))
-	cmd.AddCommand(NewRepositoryCmd(newRepositoryClient))
-	cmd.AddCommand(NewRunCmd(newRunClient))
+	cmd.AddCommand(NewRepositoryCmd(newClient))
+	cmd.AddCommand(NewRunCmd(newClient))
```
matejvasek, line 121: Nice. I forgot about these two during my refactor.

cmd/deploy_test.go
```diff
@@ -110,13 +110,19 @@ created: 2009-11-10 23:00:00`,
...
+			cmd := NewDeployCmd(NewClientFactory(func() *fn.Client {
 				return fn.New(
 					fn.WithPipelinesProvider(pipeline),
-					fn.WithDeployer(deployer),
-				)
-			})
+					fn.WithDeployer(deployer))
+			}))
 
+			// FIXME: the below viper.SetDefault calls mutate static data, meaning it
```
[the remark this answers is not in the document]
lkingland: Yes, but I am pretty sure that is a bit more of a workaround to the root problem: the deploy test should not be mutating package-static flag members as that affects other users of the package to change flag values.

Furthermore, I am not sure this test should be changing default values of flags to test _value_ of flags, but rather use actual flag arguments to the test invocation like we're doing in other tests.

cmd/create_test.go
```diff
@@ -33,9 +28,7 @@ func TestCreate_Execute(t *testing.T) {
 func TestCreate_NoRuntime(t *testing.T) {
 	defer fromTempDir(t)()
 
-	// command with a client factory which yields a fully default client.
-	cmd := NewCreateCmd(func(ClientOptions) *fn.Client { return fn.New() })
-	cmd.SetArgs([]string{})
```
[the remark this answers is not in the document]
lkingland: Args are only passed through if the `--` delimiter is used, which is the user of the test package explicitly asking to pass flags through to the test, which is useful for test debugging without needing to make a code change.   Just normal args will not interfere (test flags).

cmd/root.go
lkingland: Yes, thanks!  I changed to using this in most places

cmd/create_test.go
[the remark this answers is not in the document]
matejvasek: >Just normal args will not interfere (test flags)

Really? In my experience I had issues when executing specific test using `go test -run "REGEXHERE"`.

lkingland:
> It might be good idea set the envvar to previous value after test.
> I believe there is helper function in the `testing` package for that.

`t.Setenv` (defined on the `testing.TB` interface) looks like a winner.

It's very handy to know about that method now being available as of Go1.17.  The Go issue #41260 is a good read about the back-story of why it's important to use.

cmd/create_test.go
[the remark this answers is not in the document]
matejvasek: For instance if I remove the same thing from `TestDelete_ByProject` then test fails:
```
=== RUN   TestDelete_ByProject
    delete_test.go:76: expected name 'bar', got '^\QTestDelete_ByProject\E$'
```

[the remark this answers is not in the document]
matejvasek: I believe is really should be explicitly set to empty slice.

cmd/deploy_test.go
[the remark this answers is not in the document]
matejvasek: The `viper.Reset()` is needed even for such a trivial test as `TestVerbose`.

matejvasek:
> > It might be good idea set the envvar to previous value after test.
> > I believe there is helper function in the `testing` package for that.
> 
> Yes that is correct; these calls should be `t.Setenv` (defined on the `testing.TB` interface).

I meant our own testing package, but if it has been added to Go, that even better.

cmd/root.go
lkingland: Everything good takes a few passes!

cmd/create_test.go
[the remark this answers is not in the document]
matejvasek: @lkingland ^^^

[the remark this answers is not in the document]
matejvasek: But maybe it's given by how tests are run in Intellij Idea.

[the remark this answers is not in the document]
matejvasek: Idea first compiles tests and then it runs it, and that's when issues happens.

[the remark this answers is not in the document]
matejvasek:
```
GOROOT=/home/mvasek/go17 #gosetup
GOPATH=/home/mvasek/go #gosetup
/home/mvasek/go17/bin/go test -c -o /tmp/GoLand/___TestDelete_ByProject_in_knative_dev_kn_plugin_func_cmd.test knative.dev/kn-plugin-func/cmd #gosetup
/home/mvasek/go17/bin/go tool test2json -t /tmp/GoLand/___TestDelete_ByProject_in_knative_dev_kn_plugin_func_cmd.test -test.v -test.paniconexit0 -test.run ^\QTestDelete_ByProject\E$
```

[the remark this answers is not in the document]
lkingland: Interesting; this is not my experience running `go test` directly.  Maybe it is indeed an IDE difference.  I will run a few more tests locally to get more information.

cmd/root.go
```diff
@@ -471,3 +480,49 @@ func defaultTemplatedHelp(cmd *cobra.Command, args []string) {
 		fmt.Fprintf(cmd.ErrOrStderr(), "unable to display help text: %v", err)
 	}
 }
+
```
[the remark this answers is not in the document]
lkingland: Ah, I see you may be referring to these other helper functions in addition to the one I referred to above.

Yes, I will move these into our `testing` package and reference as necessary. That's why we made it!

cmd/create_test.go
```diff
@@ -17,11 +15,8 @@ import (
 func TestCreate_Execute(t *testing.T) {
 	defer fromTempDir(t)()
 
-	// command with a client factory which yields a fully default client.
-	cmd := NewCreateCmd(func(ClientOptions) *fn.Client { return fn.New() })
-	cmd.SetArgs([]string{
-		fmt.Sprintf("--language=%s", "go"),
```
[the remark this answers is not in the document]
lkingland: 🤣

matejvasek:
@lkingland this needs rebase/merge.

lkingland:
> I meant our own testing package, but if it has been added to Go, that even better.

There is one in the Go stdlib now, but we'll need to upgrade to go1.17 before we can use it, unfortunately.

The rebase was quite divergent, so I merged main this time.  Thanks @matejvasek

cmd/create_test.go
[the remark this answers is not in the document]
lkingland: I think you are right that this might be an issue with the IDE or environment.  I was able to confirm locally that only arguments after the `--` delimiter are actually passed through to the code being tested.  Also , first compiling the tests before running them does not recreate the issue for me:
```
# cd cmd
# go test -c
# go test -test.run=TestDelete_ByProject
PASS
```

cmd/create.go
```diff
@@ -89,20 +81,18 @@ EXAMPLES
...
 
 	// Run Action
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
 		return runCreate(cmd, args, newClient)
 	}
 
-	// Tab Completion
-	if err := cmd.RegisterFlagCompletionFunc("language", newRuntimeCompletionFunc(newClient)); err != nil {
+	// Tab completion
+	if err := cmd.RegisterFlagCompletionFunc("language", newRuntimeCompletionFunc()); err != nil {
```
matejvasek, line 92: Here and some other places there is removed propagation of `newClient`, instead the `cmd.NewClient()` is called directly.
I believe this shouldn't be done this way. It affect potential testability.

lkingland: Thanks @matejvasek !

"potential testability" is not actual tests, and as such I see no reason to pass an argument which is not being used.

lkingland: The ClientFactory is now passed through in all situations

matejvasek: I think you might have misses on test file.

matejvasek: I had `TestCreateConfig_RepositoriesPath` specifically in mind here. Because of no propagation it had to be rewritten --accessing/testing internal details (`newCreateConfig`). With propagation it can be reverted I think.

matejvasek: >I had TestCreateConfig_RepositoriesPath specifically in mind here. Because of no propagation it had to be rewritten -- accessing/testing internal details (newCreateConfig). With propagation it can be reverted I think.

or maybe not.

cmd/create_test.go
```diff
@@ -100,78 +82,26 @@ func TestCreate_ValidatesName(t *testing.T) {
...
+	// The expected full path is XDG_CONFIG_HOME/func/repositories
 	expected := filepath.Join(xdgConfigHome, "func", "repositories")
 
-	// Create command takes a function which will be invoked with the final
-	// state of the createConfig, usually used to do fn.Client instantiation
-	// after flags, environment variables, etc. are calculated.  In this case it
-	// will validate the test condition:  that config reflects the value of
-	// XDG_CONFIG_HOME, and secondarily the path suffix `func/repositories`.
-	cmd := NewCreateCmd(func(cfg ClientOptions) *fn.Client {
-		if cfg.RepositoriesPath != expected {
```
matejvasek, line 125: Inability to do this test is exactly why I don't like using `fn.Options` which is totally opaque.
But I guess I can live with it.

matejvasek:
>Inability to do this test is exactly why I don't like using fn.Options which is totally opaque.

I this forces us to test private functions in test which is not good IMHO.
Also it make tests more of a integration test, unittesting is difficult.

cmd/repository_test.go
```diff
@@ -238,3 +188,16 @@ func piped(t *testing.T) func() string {
...
 	}
 }
+
+// TEST REPO URI:  Return URI to repo in ./testdata of matching name.
+// Suitable as URI for repository override. returns in form file://
+// Must be called prior to mktemp in tests which changes current
+// working directory as it depends on a relative path.
+// Repo uri:  file://$(pwd)/testdata/repository.git (unix-like)
+//            file: //$(pwd)\testdata\repository.git (windows)
+func testRepoURI(name string, t *testing.T) string {
```
matejvasek, line 198: I thought I refactored something like this to use HTTP.

matejvasek:
```go
func TestRepoURI(name string, t *testing.T) string {
	t.Helper()
	wd, err := os.Getwd()
	if err != nil {
		t.Fatal(err)
	}
	gitRoot := filepath.Join(wd, "testdata")
	hostPort := RunGitServer(t, gitRoot)
	return fmt.Sprintf(`http://%s/%s.git`, hostPort, name)
}
```

matejvasek: You copy-pasted unfixed version.

lkingland: That was prior to the fix, as this PR has been open for quite some time now...

cmd/create_test.go
lkingland: I understand the frustration, and indeed sympathize with it!  However this is actually as indended:  XDG_CONFIG_HOME is an implementation detail of the `func` (core) package, and is therefore opaque to the command package.  Therefore tests can only verify it by side-effect to avoid coupling `cmd` and `func` packages.  This is the same reason why tests in the `func` package are in actually in the `func_test` package.

It would actually be better to refactor the system such that environment variables are never used from anywhere outside of `main`, with their values plumbed through as needed.  This would both improve the determinism of the system as well as make it the responsibility of the `cmd` (`main`) package to test.

matejvasek: >XDG_CONFIG_HOME is an implementation detail of the func (core) package, and is therefore opaque to the command package.

But then handing of `XDG_CONFIG_HOME` should be tested (and probably is) in core package and not here.

IMHO `cmd` pacakge can be free of dependency on `fn` core pacakge.
At least test should use only mock not actual fn.Client.

cmd/create.go
lkingland: > I think you might have misses on test file.

It has been un-missed

lkingland: > accessing/testing internal details (`newCreateConfig`)

Yes, I think this is probably just fine, because the use of XDG_CONFIG_HOME to derive the config home directory is an implementation detail of the `func` package's config, which is tested in `config_test.go`.  The `cmd` package is only utilizing this feature in order to pre-populate a flag which ends up in the command's config object.  The responsibility of the cmd package is to ensure this is plumbed through from flag to config.  So each is testing _their part_ of the puzzle; not each-others :)

cmd/create_test.go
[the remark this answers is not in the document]
lkingland: Quick update, this actually fails when not using the `=`.  This is due to a nuance of Cobra, which uses its own flag parsing package.  To avoid this in the future I did add back the []string{} args, _and_ added it to all other commands which do not explicitly set args.

PR #889, opened 2022-03-09
Update wrt switch from pkger to custom ZIP pkging

matejvasek:
Update some docs and scripts WRT switch from `pkger` to custom ZIP template serialization.


matejvasek:
@zroubalik @lkingland @lance PTAL

lkingland (approved):

PR #893, opened 2022-03-14
:bug: allow adding environment variables without a cluster connection

### LINKED ISSUE #706: Users are not allowed to add environment variables without a cluster connection
Users are not allowed to use `func config envs` to ...

- add environment variables with specified value
- add environment variables from with value sourced from local environment variable

... when user is not connected to a cluster. The below error is thrown

```
$ func config envs add
Error: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```

I think at least these 2 operations should not require a real cluster connection. User should be able to do some local development and experiments with func without a real cluster.

zroubalik:

# Changes

- :bug: allow adding environment variables without a cluster connection

/kind bug

Fixes #706

**Release Note**

```release-note
allow adding environment variables without a cluster connection
```


k8s/configmaps.go
```diff
@@ -16,7 +18,23 @@ func GetConfigMap(ctx context.Context, name, namespaceOverride string) (*corev1.
...
 }
 
-func ListConfigMapsNames(ctx context.Context, namespaceOverride string) (names []string, err error) {
+// ListConfigMapsNamesIfConnected lists names of ConfigMaps present and the current k8s context
+// returns empty list, if not connected to any cluster
+func ListConfigMapsNamesIfConnected(ctx context.Context, namespaceOverride string) (names []string, err error) {
+	names, err = listConfigMapsNames(ctx, namespaceOverride)
+	if err != nil {
+		var dnsErr *net.DNSError
+		errors.As(err, &dnsErr)
```
matejvasek, line 27: Shouldn't we check return value?

matejvasek:
@zroubalik It seems that your regenerated `zz_filesystem_generated.go` file contains some eclipse files in springboot templates.

k8s/configmaps.go
```diff
@@ -16,7 +18,23 @@ func GetConfigMap(ctx context.Context, name, namespaceOverride string) (*corev1.
 	return client.CoreV1().ConfigMaps(namespace).Get(ctx, name, metav1.GetOptions{})
 }
 
-func ListConfigMapsNames(ctx context.Context, namespaceOverride string) (names []string, err error) {
+// ListConfigMapsNamesIfConnected lists names of ConfigMaps present and the current k8s context
+// returns empty list, if not connected to any cluster
+func ListConfigMapsNamesIfConnected(ctx context.Context, namespaceOverride string) (names []string, err error) {
+	names, err = listConfigMapsNames(ctx, namespaceOverride)
+	if err != nil {
+		var dnsErr *net.DNSError
```
matejvasek, line 26: Why only DNS error? What about when we are not logged in at all?

zroubalik: Right, fixing.

k8s/configmaps.go
zroubalik: fixed

matejvasek:
@zroubalik  Could we add some tests?
Something like:
```go
func TestListSecrete(t *testing.T) {

	defer WithEnvVar(t, "KUBECONFIG", "/tmp/non-existent.config")()
	// tests there
}
```

k8s/configmaps.go
```diff
@@ -36,6 +38,16 @@ func ListConfigMapsNamesIfConnected(ctx context.Context, namespaceOverride strin
...
 			}
 		}
+
+		// connection refused
+		if errors.Is(err, syscall.ECONNREFUSED) {
+			return []string{}, nil
+		}
+
+		// invalid configuration: no configuration has been provided
+		if strings.HasPrefix(err.Error(), "invalid configuration") {
```
matejvasek, line 48: I believe there some `somepkg.Is*()` checks for this

zroubalik: I tried to find it, but hasn't been successful.

zroubalik: Okay, this one might be it: https://pkg.go.dev/k8s.io/client-go/tools/clientcmd#IsEmptyConfig

matejvasek:
you added tests and coverage decreased :rofl:

zroubalik:
> you added tests and coverage decreased 🤣

🤦‍♂️

PR #895, opened 2022-03-14
docs: parameter `-P` is not needed in grep in pre-commit hook

zroubalik:


matejvasek:
@zroubalik I think you can remove the `P` directly from the script.

zroubalik:
@matejvasek is it working for you as well? without the `-P`?

matejvasek:
> @matejvasek is it working for you as well? without the `-P`?

I think it should.

zroubalik:
@matejvasek fixed PTAL

PR #897, opened 2022-03-15
Add flag and command for custom pipelines

### LINKED ISSUE #791: On-cluster build: allow pipeline override or pipeline export for a func project
Currently, pipelines are defined inside the CLI and cannot be overridden. allowing to export the pipeline into a YAML file that can be updated and replaced on the cluster for that specific function project will enable other tasks to be performed as part of the function deploy or build process.

shinigambit:
# Changes

- Allow custom pipeline to be applied on remote builds
- Allow currently applied pipeline to be exported to a file

/kind <kind>

Fixes #791

**Release Note**

```release-note
Added flag to specify a custom tekton pipeline to be applied from a file and command to export cluster pipeline to a file
```

**Docs**

```docs

```

### DIFF
--- a/client.go
+++ b/client.go
@@ -184,8 +184,9 @@ type DNSProvider interface {
 
 // PipelinesProvider manages lifecyle of CI/CD pipelines used by a Function
 type PipelinesProvider interface {
-	Run(context.Context, Function) error
+	Run(context.Context, Function, bool) error
 	Remove(context.Context, Function) error
+	Export(context.Context, Function, string) error
 }
 
 // New client for Function management.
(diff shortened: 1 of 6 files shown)


zroubalik:
As mentioned on #791, do you think it would make a sense to think about the solution that will mimic the approach that is being used for language templates? To add posibility to define custom pipelines with the very similar workflow.

matejvasek:
I think @zroubalik knows most about pipelines.

client.go
```diff
@@ -184,8 +184,9 @@ type DNSProvider interface {
 
 // PipelinesProvider manages lifecyle of CI/CD pipelines used by a Function
 type PipelinesProvider interface {
-	Run(context.Context, Function) error
+	Run(context.Context, Function, bool) error
```
matejvasek, line 187: I think @lkingland will not like this.
Maybe custom pipeline can exist rather that new implementer of the interface?

zroubalik:
As mentioned on the team call, I think that it might be beneficial to investigate, whether using pipelines templates is not a better one. By templates I mean a similar approach that we have taken with Extensible Language Packs -> https://github.com/knative-sandbox/kn-plugin-func/blob/main/docs/language-pack-providers/language-pack-contract.md#extensible-language-packs

lance:
@shinigambit ping - any status updates on this?

zroubalik:
I have an idea how to refactore the whole Tekton integration around the Pipeline as a Code tooling. This should bring us this functionality as well.

PR #900, opened 2022-03-15
src: differentiate repository from repositories

### LINKED ISSUE #896: Better Differentiation between Repository and Repositories
The system allows two orthogonal features:
1)  Create a new Function from a template in a repository at a specific URI
2)  Customize the location on-disk where locally-installed template repositories are kept

These are referred to, respectively as `repository` and `repositories`.   This overlap on the root "repository" causes confusion.

This is a valid concern, so we should remove (or at least reduce) the similarity.  Though, we likely need to retain the features these two settings offer.  First, the modal nature is a required/expressly requested feature, so some amount of irreducible complexity is necessary to differentiate between creating a new Function from a locally installed repository vs. overriding the use of the installed repositories and insted specifying a one-off remote repository.  Additionally, customizing the location on disk where repositories are manged is a required setting for situations such as scripted automation where the user has no home directory, and during testing.  Therefore these two settings can not be either merged or either removed.

To alleviate the confusion, I intend to submit the following changes:
1)  Only the former setting, specifying a single remote repository, should be available as a flag.  I.E. `--repository` takes precedence as the highlighted setting.
2)  The location on disk where repositories are managed can be configured only using an environment variable, since it is not expected to be oft used. 
3)  The name of the variable which indicates the path to repositories on disk will be qualified with a "Path" suffix.  I.e. `FUNC_REPOSITORIES_PATH` is available but deemphasized.

I hope this is a reasonable solution.   The forthcoming Global Config system will likely make the difference that much more intuitive as well

lkingland:
:broom: Rename Repositories to RepositoriesPath
:broom: Relegates RepositoriesPath to an environment-varible-only setting

/kind cleanup

Fixes #896

### DIFF
--- a/client.go
+++ b/client.go
@@ -246,7 +246,7 @@ func ConfigPath() (path string) {
 
 // RepositoriesPath accesses the currently effective repositories path,
 // which defaults to [ConfigPath]/repositories but can be set explicitly using
-// the WithRepositories option when creating the client..
+// the WithRepositoriesPath option when creating the client..
 // The path will be created if it does not already exist.
 func (c *Client) RepositoriesPath() (path string) {
 	path = c.repositories.Path()
@@ -256,7 +256,7 @@ func (c *Client) RepositoriesPath() (path string) {
 
 // RepositoriesPath is a convenience method for accessing the default path to
 // repositories that will be used by new instances of a Client unless options
-// such as WithRepositories are used to override.
+// such as WithRepositoriesPath are used to override.
 // The path will be created if it does not already exist.
 func RepositoriesPath() string {
 	return New().RepositoriesPath()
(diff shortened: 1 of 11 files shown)


matejvasek:
Nice, `RepositoriesPath` make is more clear to me (maybe `RepositoriesDir` would be even better).

lkingland:
> Nice, `RepositoriesPath` make is more clear to me (maybe `RepositoriesDir` would be even better).

I usually reserve the `Dir` suffix for an actual Directory object, and `Path` for a path _to_ a location.

matejvasek:
>But I can be convinced to change if you feel strongly about that.

I don't have strong opinions on it. Others, what do you thing?

lance:
> > But I can be convinced to change if you feel strongly about that.
> 
> I don't have strong opinions on it. Others, what do you think?

I prefer `RepositoriesPath` in this case.

lance:
Since it's all the same to you, @matejvasek I'm going to land this and if the name needs to change, we can revert it.

/lgtm

PR #907, opened 2022-03-16
doc: add documentation for using podman remotely

lance:
- 📖   This commit adds more documentation for using podman, specifically in this case, connecting to and using a remote podman service from a Mac computer.

/kind documentation


matejvasek:
I find the name `registry` confusing there, it suggest relation to image registry in my mind.
Maybe rename it to `my-remote-machine`.

lance:
> I find the name `registry` confusing there, it suggest relation to image registry in my mind. Maybe rename it to `my-remote-machine`.

@matejvasek I renamed to `podman-remote` - wdyt?

matejvasek:
I think that's fine.

PR #914, opened 2022-03-21
Update springboot templates

### LINKED ISSUE #912: Update springboot http template to match the cloudevent template
The cloudevent template was updated in #832, we should make similar changes for the http template.

trisberg:
# Changes

- http should match cloudevents template
- update Spring Boot version
- update README for cloudevents template

/kind enhancement

Fixes #912

**Release Note**

```release-note

```

**Docs**

```docs

```


matejvasek:
the e2e test failure is not caused by this PR probably

matejvasek:
@salaboy PTAL

PR #916, opened 2022-03-22
Improve error handling

matejvasek:
* replaced some plain `==` by `error.Is()`
* wrapping some more errors using `fmt.Errorf()`

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -1112,7 +1112,7 @@ func TestClient_Invoke_HTTP(t *testing.T) {
 	}
 	s := http.Server{Handler: handler}
 	go func() {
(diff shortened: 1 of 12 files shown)


client_test.go
```diff
@@ -1112,7 +1112,7 @@ func TestClient_Invoke_HTTP(t *testing.T) {
 	}
 	s := http.Server{Handler: handler}
 	go func() {
-		if err = s.Serve(l); err != nil && err != http.ErrServerClosed {
+		if err = s.Serve(l); err != nil && !errors.Is(err, http.ErrServerClosed) {
```
lkingland, line 1115: Good good.  We should keep our eyes open to make sure we always use this method of comparison in the future!

lkingland (approved):

PR #917, opened 2022-03-22
Improve template ZIP generation

matejvasek:
Use only 755 and 644 mod permission, not exact mod from local FS.
The exec bit is the only one that we really care about.

Also other bits may be set differently on developer machines depending
on its configuration, this would affect generation reproducibility.
For instance on some developer's machines, regular files have mod 664 rather that usual 644 for some reason.


matejvasek:
@trisberg please try this on your macOS machine. Hopefuly is should produce identical `zz_filesystem_generated.go` file there.

matejvasek:
AFAIK `git` also knows  755 and 644 only.

trisberg:
I tried this on my Ubuntu system with umask set to 0002 and also 0022 - both runs worked fine now - no difference in the generated file.

lkingland (commented):
This is a good change.  It should really help insulate against differences in developer machines permissions settings.

Would you mind adding an explanatory comment above, because this is a bit unobvious why it is required?

matejvasek:
@lkingland @zroubalik @lance PTAL

PR #923, opened 2022-03-30
feat: s2i builder with preliminary node support

### LINKED ISSUE #908: Add support for s2i builds
The CLI is structured such that the builder is an interface. Currently there is a buildpack implementation, and a mock implementation for testing. We should additionally support s2i as a build strategy.

https://github.com/knative-sandbox/kn-plugin-func/blob/cad0a04276cbd0fe15a83afb68dc5fd680f47d48/client.go#L66-L70

lkingland:
# Changes

- :gift: Source-to-Image Builder implementation which can be specified with `--builder=s2i`
- :gift: Adds preliminary support for building Node.js Functions with the s2i builder.

/kind enhancement

First step towards accomplishing #908


buildpacks/builder.go
```diff
@@ -21,6 +20,12 @@ import (
 	"github.com/buildpacks/pack/pkg/logging"
 )
 
+// DefaultBuilderImages for Pack builders indexed by Runtime Language
+var DefaultBuilderImages = map[string]string{
+	"node": "gcr.io/paketo-buildpacks/builder:base",
+	"go":   "gcr.io/paketo-buildpacks/builder:base",
+}
```
lkingland, line 27: This is a lazily-poplated list of defaults as they are given s2i support (and thus can not rely on `manifest.yaml` only). 
This list will eventually expand to include all Paketo builders for runtimes we include.
For language packs which provide entirely new runtimes, the default will need to be provided in a `manifest.yaml` bu the language pack developer.

matejvasek:
@lkingland why the changes to the node/go templates?

lkingland:
> @lkingland why the changes to the node/go templates?

Builders are no longer defined in a template's manifest, because a Function can be built with either a Buildpack or an S2I builder, and thus the default is dependant on the builder chosen at build time.

These defaults should have been in the code from the beginning, and this is the forcing function.

matejvasek:
>These defaults should have been in the code from the beginning, and this is the forcing function.

I think we had defaults in code and then removed it.

matejvasek:
@lkingland  `s2i` build is not working with `podman`.
Probably ok for now, but we should put it into documentation.

https://github.com/openshift/source-to-image/issues/1082

matejvasek:
https://github.com/containers/podman/issues/13770

s2i/builder.go
```diff
@@ -0,0 +1,115 @@
...
+	ErrRuntimeRequired = errors.New("runtime is required to build")
+
+	// ErrRuntimeNotSupported indicates the given runtime is not (yet) supported
+	// by this builder.
+	ErrRuntimeNotSupported = errors.New("runtime not supported")
+)
+
+// DefaultBuilderImages for s2i builders indexed by Runtime Language
+var DefaultBuilderImages = map[string]string{
+	"node": "registry.access.redhat.com/ubi8/nodejs-16", // TODO: finalize choice and include version
```
lance, line 33: Right here is one of the downsides of using s2i instead of buildpacks. Buildpacks done properly allow the runtime version to be specified at build time. What happens if a user wants/needs to build a Node.js 14 application instead, because they have a dependency that depends on a deprecated Node.js API that was removed in 16. It's a hypothetical - something we should think about. But in all truth the problem lies with s2i architecture, and not with this line of code, IMO.

s2i/builder.go
```diff
@@ -0,0 +1,115 @@
...
+		Endpoint: endpoint,
+	}
+	if errs := validation.ValidateConfig(cfg); len(errs) > 0 {
+		for _, e := range errs {
+			fmt.Fprintf(os.Stderr, "ERROR: %s\n", e)
+		}
+		return errors.New("Unable to build via the s2i builder.")
+	}
+
+	builder, _, err := strategies.Strategy(client, cfg, build.Overrides{})
```
lance, line 86: Am I correct to understand this is the "docker" build strategy? One of the other build strategies that s2i uses is "binary" which we may consider leveraging for push-to-cluster-my-function-and-build-it-there, since that's exactly what it does.

test/_e2e/cmd_build_test.go
```diff
@@ -24,3 +31,20 @@ func Build(t *testing.T, knFunc *TestShellCmdRunner, project *FunctionTestProjec
...
+func TestBuild_S2I(t *testing.T) {
+	var (
+		root        = "testdata/e2e/testbuild"
+		bin, prefix = bin()
+		cleanup     = Within(t, root) // TODO: replace with Fromtemp?
+		cwd, _      = os.Getwd()
+	)
+	defer cleanup()
+
+	run(t, bin, prefix, "create", "-v", "--language=node", cwd)
```
lance, line 45: Did you test with typescript? I don't know if s2i can seamlessly do that or not, but it sure would be nice

s2i/builder.go
```diff
@@ -0,0 +1,115 @@
...
+	// Build Config
+	cfg := &api.Config{}
+	cfg.Quiet = !b.verbose
+	cfg.Tag = f.Image
+	cfg.Source = &git.URL{URL: url.URL{Path: f.Root}, Type: git.URLTypeLocal}
+	cfg.BuilderImage = f.Builder
+	cfg.BuilderPullPolicy = api.DefaultBuilderPullPolicy
+	cfg.PreviousImagePullPolicy = api.DefaultPreviousImagePullPolicy
+	cfg.RuntimeImagePullPolicy = api.DefaultRuntimeImagePullPolicy
+	cfg.DockerConfig = &api.DockerConfig{
```
matejvasek, line 76: There is `GetDefaultDockerConfig()` function that can default init this.

s2i/builder.go
```diff
@@ -0,0 +1,115 @@
...
+func (b *Builder) Build(ctx context.Context, f fn.Function) (err error) {
+	// Ensure the Function has a builder specified
+	if f.Builder == "" {
+		f.Builder, err = defaultBuilderImage(f)
+		if err != nil {
+			return
+		}
+	}
+
+	client, endpoint, err := docker.NewClient(dockerClient.DefaultDockerHost)
```
matejvasek, line 55: The returned endpoint is docker host from running container PoV (as needed by Buildpack). I believe it shouldn't be used in this context here.

lance (commented):
Looks like a great first step into s2i builds. There are some things I think we'll want to consider in the long run, such as how we deal with chained builds (important to avoid having 500MB application images), and perhaps other things of a similar nature. A few questions within, but overall lgtm.

s2i/builder.go
matejvasek: Also is that `Endpoint` really used or only some validation is complaining? We are passing docker client so that should be used.

cmd/build.go
```diff
@@ -40,17 +43,18 @@ and the image name is stored in the configuration file.
 {{.Name}} build --builder cnbs/sample-builder:bionic
 `,
 		SuggestFor: []string{"biuld", "buidl", "built"},
-		PreRunE:    bindEnv("image", "path", "builder", "registry", "confirm", "push"),
+		PreRunE:    bindEnv("image", "path", "builder", "registry", "confirm", "push", "builder-image"),
 	}
 
-	cmd.Flags().StringP("builder", "b", "", "Buildpack builder, either an as a an image name or a mapping name.\nSpecified value is stored in func.yaml for subsequent builds.")
+	cmd.Flags().StringP("builder", "b", "pack", "builder to use when creating the underlying image. Currently supported builders are 'pack' and 's2i'.")
+	cmd.Flags().StringP("builder-image", "", "", "builder image, either an as a an image name or a mapping name.\nSpecified value is stored in func.yaml for subsequent builds. ($FUNC_BUILDER_IMAGE)")
```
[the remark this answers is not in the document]
zroubalik: I agree. We should try to consolidate this a little bit. I wouldn't block this PR on this though. But it is something we should try to tackle as soon as possilbe. To provide consistency and more clarity for users.

zroubalik (commented):
/lgtm

buildpacks/builder.go
```diff
@@ -21,6 +20,12 @@ import (
 	"github.com/buildpacks/pack/pkg/logging"
 )
 
+// DefaultBuilderImages for Pack builders indexed by Runtime Language
+var DefaultBuilderImages = map[string]string{
+	"node": "gcr.io/paketo-buildpacks/builder:base",
+	"go":   "gcr.io/paketo-buildpacks/builder:base",
```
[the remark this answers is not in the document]
lkingland: It is my understanding that the base is the same, but the buildpack builder will still apply any buildpacks defined, which include `ghcr.io/boson-project/go-function-buildpack:tip`, which has the scaffolding.  The S2I builder, when updated to support Go, will of course ignore that and use its own.

cmd/build.go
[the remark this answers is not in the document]
lkingland: In order to not block this PR, I have opened a new issue to discuss better flag naming: #942

go.mod
```diff
@@ -45,6 +46,9 @@ require (
 )
 
 replace (
+	// update docker to be compatible with version used by pack and removes invalid pseudo-version
+	github.com/openshift/source-to-image => github.com/boson-project/source-to-image v1.3.2
```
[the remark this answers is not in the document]
lkingland: Agreed.  I will do so as soon as I have confirmation from our team that this PR is the direction we want to go.  Not that I have any better alternatives handy.

s2i/builder.go
lkingland: I can't think of anything other than a brute-force approach where the latest node version is used as we do here, but there exists a mapping of major revisions to their builder image, and we accept an optional parameter where one can specify a Node version.

Note that the `--builder-image` flag is still available, so if the user is savvy about builder images, they can always provide that on the command line and it will take precedence over this default.

s2i/builder.go
lkingland: Yes, it is my understanding this is the Docker method.  I am not familiar with the binary strategy, but I will look into it.

test/_e2e/cmd_build_test.go
lkingland: I have not run a typescript test yet, but I am hopeful as well.

s2i/builder.go
lkingland: The first line of `docker.NewClient` is `os.Getenv("DOCKER_HOST")`.  How is that not the host we want to use?  I am confused and that function contains no documentation.

s2i/builder.go
lkingland: `GetDefaultDockerConfig` seems to be a better choice, now that I know it exists.  I have replaced this manual config creation with that call, and therefore ignoring the endpoint returned from the docker client constructor.

s2i/builder.go
matejvasek: It reads the `DOCKER_HOST` envvar but it may or may not return it.
* For local connection (unix socket and windows named pipe) it returns the `DOCKER_HOST` directly.
* For `ssh` connection it reads the `DOCKER_HOST` from the `ssh` remote.
* For `TCP` connections it returns `""` so it defaults in the remote (and no you cannot use `client.DefaultDockerHost`), this is needed beaus of `TCP+tls` connections.

matejvasek: >The returned endpoint is docker host from running container PoV (as needed by Buildpack). 

This is not accurate. It's docker connection string as seen by the machine running the daemon.
So there is some special handling for `ssh` and `TCP` connections.

matejvasek: > and no you cannot use `client.DefaultDockerHost`
Because the value is different for various platforms: if you run `func` on Windows and connected via `TCP` to Linux there would be mismatch ("//./pipe/docker_engine" vs "unix:///var/run/docker.sock").

s2i/builder.go
matejvasek: I think that the config is not used anyway. It could be used to construct an instance of `dockerClient.CommonAPIClient`, but we are passing an instance of the interface to the API so there is no need to use that I believe.

buildpacks/builder.go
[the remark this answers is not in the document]
lance: > It is my understanding that the base is the same, but the buildpack builder will still apply any buildpacks defined, which include `ghcr.io/boson-project/go-function-buildpack:tip`, which has the scaffolding.

Yes, this is the case, if it is explicitly provided, for example in a `func.yaml` file in the `buildpacks` field. Is that what you are expecting? Without that, the scaffold isn't there.

s2i/builder.go
lance: > Note that the `--builder-image` flag is still available, so if the user is savvy about builder images, they can always provide that and it will take precedence over this default.

Yes and I think this also removes the need for an additional parameter. I was just pointing out the flaw in s2i.

PR #925, opened 2022-03-30
Update Knative dependencies to v1.1

dsimansk:
# Changes

- :broom: Update Knative dependencies to v1.1

/cc @lance


go.mod
```diff
@@ -11,37 +11,39 @@ require (
...
+	github.com/docker/docker v20.10.12+incompatible
 	github.com/docker/docker-credential-helpers v0.6.4
 	github.com/docker/go-connections v0.4.0
 	github.com/go-git/go-billy/v5 v5.3.1
 	github.com/go-git/go-git/v5 v5.4.2
-	github.com/google/go-cmp v0.5.6
-	github.com/google/go-containerregistry v0.6.0
+	github.com/go-logr/logr v1.2.2 // indirect
+	github.com/google/go-cmp v0.5.7
+	github.com/google/go-containerregistry v0.8.1-0.20220219142810-1571d7fdc46e
```
dsimansk, line 22: I'll remove this bit for now.

docker/pusher.go
```diff
@@ -40,6 +40,7 @@ type PusherDockerClient interface {
 	ImageLoad(context.Context, io.Reader, bool) (types.ImageLoadResponse, error)
 	ImageTag(context.Context, string, string) error
 	ImagePush(ctx context.Context, ref string, options types.ImagePushOptions) (io.ReadCloser, error)
+	ImageInspectWithRaw(context.Context, string) (types.ImageInspect, []byte, error)
```
lance, line 43: Sorry for my ignorance. Is this just needed because of the bump in the docker dependency version?

matejvasek: Probably dep to go-containers lib.

vendor/github.com/google/go-containerregistry/pkg/v1/daemon/options.go
```diff
@@ -99,4 +99,5 @@ type Client interface {
 	ImageSave(context.Context, []string) (io.ReadCloser, error)
 	ImageLoad(context.Context, io.Reader, bool) (types.ImageLoadResponse, error)
 	ImageTag(context.Context, string, string) error
+	ImageInspectWithRaw(context.Context, string) (types.ImageInspect, []byte, error)
```
matejvasek, line 102: Here @lance .

docker/pusher.go
dsimansk: Yep, that's an updated interface from go-containers. Should I refactor it to Matej's suggestion? It's definitely cleaner.

PR #938, opened 2022-04-04
Use "openshift" uname for OpenShift image registry

matejvasek:
The user name from the config is invalid sometimes.
We can use `openshift` user name with user's token (it could be almost any other non-empty uname without colons in it).

```release-note
Fix issues with pushing to OpenShift's integrated registry when uname `kube:admin`
```


lance (commented):
/lgtm

PR #944, opened 2022-04-05
feat: invoke verbose metadata

### LINKED ISSUE #922: Provide an option to display headers with `func invoke`
Currently the `invoke` command nicely echoes the response data from a function. But there is no way to see the headers. The `-v` flag is accepted, but has no effect. This should be changed so that `--verbose` or `-v` displays all received headers as well.

lkingland:
# Changes

- :gift: `invoke -v` prints metadata

HTTP headers or CloudEvent fields are printed on invoke with verbose enabled.

```shell
$ func invoke --verbose
"Content-Type":text/plain; charset=utf-8
"Date":Tue, 05 Apr 2022 14:02:10 GMT
"Content-Length":3

OK
```

/kind enhancement

Fixes #922

### DIFF
--- a/cmd/invoke.go
+++ b/cmd/invoke.go
@@ -162,12 +162,12 @@ func runInvoke(cmd *cobra.Command, args []string, newClient ClientFactory) (err 
 		return err
 	}
 
-	// Print metadata (headers for HTTP, event fields for CloudEvents) if in
-	// verbose mode.
+	// Print metadata (headers for HTTP, CloudEvents include metadat in their
+	// default stringification) if in
 	if cfg.Verbose {
 		for k, vv := range metadata {
 			values := strings.Join(vv, ";")
-			fmt.Fprintf(cmd.OutOrStdout(), "\"%v\":%v\n", k, values)
+			fmt.Fprintf(cmd.OutOrStdout(), "%v: %v\n", k, values)
 		}
 		if len(metadata) > 0 {
 			fmt.Fprintln(cmd.OutOrStdout())
--- a/cmd/invoke.go
+++ b/cmd/invoke.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/google/uuid"
@@ -156,12 +157,24 @@ func runInvoke(cmd *cobra.Command, args []string, newClient ClientFactory) (err 
 	}
 
 	// Invoke
-	s, err := client.Invoke(cmd.Context(), cfg.Path, cfg.Target, m)
+	metadata, body, err := client.Invoke(cmd.Context(), cfg.Path, cfg.Target, m)
 	if err != nil {
 		return err
 	}
 
-	fmt.Fprintln(cmd.OutOrStderr(), s)
+	// Print metadata (headers for HTTP, event fields for CloudEvents) if in
+	// verbose mode.
+	if cfg.Verbose {
+		for k, vv := range metadata {
+			values := strings.Join(vv, ";")
+			fmt.Fprintf(cmd.OutOrStdout(), "\"%v\":%v\n", k, values)
+		}
+		if len(metadata) > 0 {
+			fmt.Fprintln(cmd.OutOrStdout())
+		}
+	}
+
+	fmt.Fprintln(cmd.OutOrStdout(), body)
 	return
 }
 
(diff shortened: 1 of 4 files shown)


invoke.go
```diff
@@ -154,11 +156,20 @@ func sendEvent(ctx context.Context, route string, m InvokeMessage, t http.RoundT
...
 
+	// return metdata about the CloudEvent that just so happens to be the same
+	// data structure as HTTP headers such that they are interchangeable at the
+	// higher level of the Invoke... invocation.
+	meta = make(map[string][]string)
+	meta["id"] = []string{evt.ID()}
+	meta["type"] = []string{evt.Type()}
+	meta["source"] = []string{evt.Source()}
+	meta["subject"] = []string{evt.Subject()}
+
```
lance, line 167: I think here we should probably just return an empty map, or one that includes all headers not prefixed with `ce-` (the prefix used for the event attributes in the header). The reason I had not already contributed this fix is because it was not immediately obvious to me how to get the headers when receiving an event with the golang library.

The string representation of a CloudEvent is already nicely formatted and displays all the event attributes. For example:

```
❯ func invoke -v
Context Attributes,
  specversion: 1.0
  type: echo
  source: function.eventViewer
  id: b9a722da-9dd7-4754-9eb7-f0cb1651433b
  time: 2022-04-05T19:47:18.518Z
  datacontenttype: application/json; charset=utf-8
Data,
  Hello World
```

Whereas with this addition, it ends up looking like this, which is difficult to read, incomplete in the first section, and duplicative.

```
~/demo/viewer is 📦 v0.1.0 via  v16.14.0
❯ func invoke -v
"source":function.eventViewer
"subject":
"id":40ee7efd-a6eb-44e6-9df7-f3d1fb56a430
"type":echo

Context Attributes,
  specversion: 1.0
  type: echo
  source: function.eventViewer
  id: 40ee7efd-a6eb-44e6-9df7-f3d1fb56a430
  time: 2022-04-05T20:00:50.653Z
  datacontenttype: application/json; charset=utf-8
Data,
  Hello World
```

cmd/invoke.go
```diff
@@ -156,12 +157,24 @@ func runInvoke(cmd *cobra.Command, args []string, newClient ClientFactory) (err
...
 		return err
 	}
 
-	fmt.Fprintln(cmd.OutOrStderr(), s)
+	// Print metadata (headers for HTTP, event fields for CloudEvents) if in
+	// verbose mode.
+	if cfg.Verbose {
+		for k, vv := range metadata {
+			values := strings.Join(vv, ";")
+			fmt.Fprintf(cmd.OutOrStdout(), "\"%v\":%v\n", k, values)
```
lance, line 170: It's just a style thing, but I don't think surrounding the header name with quotes is necessary. To me, it muddies the display. And I think there should be a space after the separating `:`. Compare:

```
❯ func invoke -v
"Access-Control-Allow-Origin":*
"Content-Length":132
"Date":Tue, 05 Apr 2022 20:09:10 GMT
"Connection":keep-alive
"Keep-Alive":timeout=5
"Content-Type":application/json; charset=utf8
"Access-Control-Allow-Methods":OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH

{"ContentType":"text/plain","Data":"Hello World","ID":"42fd8604-75db-46d3-bdc1-39b3c213595e","Source":"/boson/fn","Type":"boson.fn"}
```
With
```
❯ func invoke -v
Access-Control-Allow-Origin: *
Content-Length: 132
Date: Tue, 05 Apr 2022 20:09:10 GMT
Connection: keep-alive
Keep-Alive: timeout=5
Content-Type: application/json; charset=utf8
Access-Control-Allow-Methods: OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH

{"ContentType":"text/plain","Data":"Hello World","ID":"42fd8604-75db-46d3-bdc1-39b3c213595e","Source":"/boson/fn","Type":"boson.fn"}
```

I know as programmers, it's nice to see if there are invisible characters at the end of the string or other small debugging concerns. It's just my opinion that in most cases, a cleaner and simpler view is better.

lance (commented):
Overall looks great, but I recommend some changes to the output for event invocation, and I've made a small suggestion for the output formatting.

cmd/invoke.go
lkingland: I have removed the quotes around the keys, added a space after the colon, and removed metadat from the CloudEvents invocation.

PR #957, opened 2022-04-11
feat: s2i builder typescript support

lkingland:
# Changes

- :gift: s2i builder typescript support

The node-16 builder does indeed appear to support building typescript Functions; though I do see the following concerning warnings in the verbose build logs:

```
$ func build --builder=s2i -v
...
/opt/app-root/src/.npm is not a mountpoint
---> Cleaning the npm cache /opt/app-root/src/.npm
/tmp is not a mountpoint
---> Cleaning the /tmp/npm-*
🙌 Function image built: localhost:50000/func/tmp:latest
```

/kind enhancement

Additional work towards #908


lance:
@lkingland thanks for pointing out the warnings. I have opened an issue for that https://github.com/knative-sandbox/kn-plugin-func/issues/960

PR #966, opened 2022-04-11
chore: add version.txt to .gitattributes for style

lance:
🧹  The Knative style check uses linguist to determine if a given file is a generated file. If so, it excludes that file from the check. This change should prevent Knative's style check from complaining about no newline on version.txt

/kind chore


matejvasek:
@lance while at it would you mind also adding `zz_filesystem_generated.go` as generated there?

PR #969, opened 2022-04-12
WIP

matejvasek:
WIP

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -482,7 +482,7 @@ func (r *Repository) Write(path string) (err error) {
 		}
 		fs = billyFilesystem{fs: wt.Filesystem}
 	}
-	return copy(".", path, fs)
+	return copyFromFS(".", path, fs)
 }
 
 // URL attempts to read the remote git origin URL of the repository.  Best
--- a/repository.go
+++ b/repository.go
@@ -244,7 +244,7 @@ func repositoryRuntimes(r Repository) (runtimes []Runtime, err error) {
 
 	// Load runtimes
 	if r.TemplatesPath == "" {
-		r.TemplatesPath = "."
+		r.TemplatesPath = DefaultTemplatesPath
 	}
 
 	fis, err := r.FS.ReadDir(r.TemplatesPath)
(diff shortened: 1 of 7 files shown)


repository.go
```diff
@@ -39,75 +39,55 @@ const (
...
 // Repository represents a collection of runtimes, each containing templates.
 type Repository struct {
-	// Name of the repository.  Naming things and placing them in a hierarchy is
-	// the responsibility of the filesystem; metadata the responsibility of the
-	// files within this structure. Therefore the name is not part of the repo.
-	// This is the same reason a git repository has its name nowhere in .git and
-	// does not need a manifest of its contents:  the filesystem itself maintains
-	// this information.  This name is the denormalized view of the filesystem,
-	// which defines the name as the directory name, and supports being defaulted
-	// to the value in the .yaml on initial add, which is stored as DefaultName.
```
lkingland, line 49: Why do my comments always end up getting deleted?  I put a lot of work into them, and find them quite helpful later when I come back to the system months or years later.  I presume others may find them useful as well.  "Self-documenting-code" has never worked for me, but "literate programming" often does.  Furthermore, it is Go bet practices that any exported member have a comment included, with the first token matching the exported member:
```
// Name ...
Name string
```

matejvasek: I kept comments where I am sure they are valid. With this one I am not so sure. I mean: name doesn't have to be directory name from filesystem -- repository can be named arbitrarily. Repository can have its origin in filesystem, but it very well may be some network service or it can be created in memory somehow.

I don't want Repository to equate with filesystem/directory.

matejvasek: Right now Repository==Filesystem is ingrained in code, that what I want to change.

matejvasek: Maybe I could add short comment:
// Name typically can be: name of containing directory, information from manifest.yaml, or last part of git URI.

matejvasek: Imagine we created Repository that generates projects via `https://code.quarkus.io/`, what's the name there? There is no source directory nor manifest.yaml.

matejvasek: @lkingland does this make sense?

PR #970, opened 2022-04-13
CI enhancements (e2e tests) and Go template fix

jrangelramos:
# Changes

- :gift: including workflow to run e2e lifecycle tests for remaining runtimes (go, python, quarkus, springboot, typescript)
- :broom: adding timeout and additional logs on e2e http test
- :broom: fix go cloudevents default template (returning error 500) 

/kind cleanup

Note: Here is a proposal to include a new workflow to execute e2e lifecycle tests for all supported language/runtimes (existing e2e only checks node). The goal is to mitigate issues with language/runtimes (and buildpacks).

In additional, the PR also has some proposal changes to cloudevents default template for GO, in order to fix server 500 error on the deployed function.


matejvasek:
@jrangelramos will this fix the flakes?

jrangelramos:
@matejvasek well.. not really (maybe w/ explicity timeout). but it will at least give us more info next time it happens

matejvasek:
Nit: set the timeout in `{}`.

jrangelramos:
fixes #971

.github/workflows/test-e2e-runtime.yaml
```diff
@@ -0,0 +1,30 @@
+name: Func E2E Lifecycle Test
```
matejvasek, line 1: Why "Lifecycle" in name here @jrangelramos ?

matejvasek:
@jrangelramos you need to rebase/regenerate zz_* file.

.github/workflows/test-e2e-runtime.yaml
jrangelramos: Well.. it is because there is 2 sets of e2e tests: the lifecycle tests (e2e_lifecycle_tests.sh) and extended tests (e2e_extended_tests.sh)

The lifecycle one tests a function lifecycle such as, creation, deployment, read, update (change function and redeploy) and deletion.. the CRUD. against a specific language runtime and template.

The extended tests runs other sort of tests such as invoke, config labels, volumes, etc.

The workflow in this case will run the lifecycles tests for all runtimes.

templates/go/cloudevents/handle.go
```diff
@@ -20,24 +20,28 @@ func Handle(ctx context.Context, event cloudevents.Event) (*event.Event, error)
...
 
 	fmt.Printf("Outgoing Event: %v\n", outputEvent)
 
 	return &outputEvent, nil
 }
 
+type Echo struct {
+	Message string `json:"message"`
+}
+
```
lkingland, line 44: I am gong to approve this for now, but my general opinion is that the default "echo" handlers should be two lines:
1) print the incoming event to log
2) return the received event.  That is an Echo.

I am grateful for the improvement, but I feel compelled to ask the question: why is this handler, which should be the epitome of simplicity, continueing to grow?  This is perhaps a topic for a future discussion, but an "Echo" is literally "reflect, as in a mirror, what was received".  Why can this handler not simply be:
```
func Handle(ctx context.Context, event cloudevents.Event) (*event.Event, error) {
  fmt.Println(event)
  return &event, nil
}
```

test/_e2e/trigger_http_test.go
```diff
@@ -37,7 +38,7 @@ func (f FunctionHttpResponsivenessValidator) Validate(t *testing.T, project Func
 	if f.contentType != "" {
 		req.Header.Add("Content-Type", f.contentType)
 	}
-	client := &http.Client{}
+	client := &http.Client{Timeout: time.Second * 15}
```
lkingland, line 41: Interesting addition.  Does your debugging suggest that the reason we see tests flaking is (as we guessed) a timeout?

lkingland (approved):
Looks good to me!  My only suggestion is that the Go CloudEvents update _might_ be better accomplished by just making the Echo implementation closer to an actual echo.  YMMV.

/lgtm
/hold for others to comment, or for @jrangelramos ; you to pull the merge trigger

templates/go/cloudevents/handle.go
jrangelramos: Thanks for your feedback @lkingland. Personally I prefer a template as simple as possible, as it was on previous commits. Looking to the history and the direction the template took I realized  that apply a fix would be more appropriate (instead of rolling back), hence my initial goal in this PR was to get E2E tests covered without changing (at least change much on the templates). But I'd be more than welcome to makethe template simpler (my preference) and leave the sample implementation (with json parsing) in  a different place (let's say on 'func-tastic') as long as @salaboy is ok with that!

test/_e2e/trigger_http_test.go
jrangelramos: Well, actually I was suspecting it was using the previous version (despite the new one report as ready). So changed the mechanistic to verify the new revision availability
https://github.com/knative-sandbox/kn-plugin-func/pull/970/files#diff-231b1bc48c9920dbdf67aae9fa6582460b37c59b89c1b226bc97a715137522d3R42

Here it was about to let it a explicit a reasonable time we expect a function to respond (even those I think that number is little high for a function).

templates/go/cloudevents/handle.go
```diff
@@ -20,24 +20,28 @@ func Handle(ctx context.Context, event cloudevents.Event) (*event.Event, error)
...
-	payload := ""
+	var payload Echo
 	err := json.Unmarshal(event.Data(), &payload)
 	if err != nil {
 		fmt.Printf("%v\n", err)
 		return nil, err
 	}
 
-	payload = "echo " + payload
+	payload.Message = "echo " + payload.Message
```
zroubalik, line 30: Do we need to add the `echo` here?

PR #973, opened 2022-04-14
Refactor: clean Repository and Runtimes structs

matejvasek:
* moved `HealthEndpoints` and `BuildConfig` structs to `function.go`
* moved `Template` to `repository.go` so `Repository`, `Runtime` and `Template` are together.
* removed fields that were leaking details from `Repository` and `Runtime`

### DIFF
--- a/template.go
+++ b/template.go
@@ -7,11 +7,11 @@ import (
 
 // template
 type template struct {
-	name        string
-	runtime     string
-	repository  string
-	fs          Filesystem
-	templConfig templateConfig
+	name       string
+	runtime    string
+	repository string
+	fs         Filesystem
+	config     templateConfig
 }
 
 func (t template) Name() string {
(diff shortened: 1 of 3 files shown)


template.go
```diff
@@ -5,49 +5,13 @@ import (
...
-	name       string
-	runtime    string
-	repository string
-	fs         Filesystem
-	manifest   templateConfig
+	name        string
+	runtime     string
+	repository  string
+	fs          Filesystem
+	templConfig templateConfig
```
lance, line 14: This is really a tiny nit, but couldn't this just be `config`. Since the struct is `template`, a `templateConfig` field name feels a little redundant.

lance (commented):
A nitpicky suggestion, but otherwise, /lgtm

template.go
matejvasek: Renamed.

matejvasek:
@lance @lkingland please re-LGTM

PR #985, opened 2022-04-20
feat: use Paketo buildpack for Quarkus

matejvasek:
- 🎁  Use paketo-buildpack for Quarkus.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/978

```release-notes
Changes Quarkus to use Paketo buildpacks
```


templates/quarkus/cloudevents/README.md
```diff
@@ -30,8 +30,23 @@ echo "export FUNC_REGISTRY=docker.io/johndoe" >> ~/.bashrc
...
+By default, JVM build is used.
+To enable native build set following environment variables to `func.yaml`:
+```yaml
+buildEnvs:
+  - name: BP_NATIVE_IMAGE
+    value: "true"
+  - name: BP_MAVEN_BUILT_ARTIFACT
+    value: target/native-sources/*
+  - name: BP_MAVEN_BUILD_ARGUMENTS
+    value: package -DskipTests=true -Dmaven.javadoc.skip=true -Dquarkus.package.type=native-sources
```
lance, line 45: Weird that you need to have both this and `BP_NATIVE_IMAGE: true`. It would be nice if we could have all of this in the func.yaml from the start, and it's ignored if `BP_NATIVE_IMAGE: false`. But I guess it's not going to be that simple.

lance (commented):
/lgtm
/hold for tests

templates/quarkus/cloudevents/README.md
matejvasek: It would be nice if could have it in `func.yaml` as comment, but we do not support comments preservation.

PR #991, opened 2022-04-25
feat: s2i builder env var interpolation

### LINKED ISSUE #990: S2I builds ignore environment
When building with buildpacks, a user's environment variables, as set in func.yaml are set in the resulting image. See here:

https://github.com/knative-sandbox/kn-plugin-func/blob/bbfb3e9f989b4f1d5fad1d38514055f188e3af18/buildpacks/builder.go#L179-L197

This does not happen for s2i builds.

/kind bug

lkingland:
- :gift: S2I Builder Env Interpolation

More specifically:

* Extracts env var interpolation logic from buildpack builder to shared package static `fn.Interpolate`
* Refactors implementation slightly to pass-through unmatched values (previously a bit ambiguous, presuming permissive).
* Adds interpolation tests, including explicit tests for asserting both unmatched strings pass through and nil value pointers indicate env var should be omitted from the final interpolated map.
* Integrates interpolated build envs into S2I builder.

/kind enhancement

Fixes #990

### DIFF
--- a/buildpacks/builder.go
+++ b/buildpacks/builder.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"regexp"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


lance (commented):
Nice consolidation - ty
/lgtm

PR #996, opened 2022-04-27
Fix read pwd from non-tty input

matejvasek:
# Changes

- :bug: docker credentials can be now read from non-tty stdin (e.g. from piped input).

/kind bug

**Release Note**

```release-note
fix: docker credentials can be now read from non-tty stdin (e.g. from piped input)
```

### DIFF
--- a/cmd/prompt_test.go
+++ b/cmd/prompt_test.go
@@ -1,0 +1,72 @@
+//go:build linux
+// +build linux
+
+package cmd
+
+import (
+	"io"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/hinshun/vt10x"
+
+	"knative.dev/kn-plugin-func/docker"
+)
+
+func Test_newPromptForCredentials(t *testing.T) {
+	expectedCreds := docker.Credentials{
+		Username: "testuser",
+		Password: "testpwd",
+	}
+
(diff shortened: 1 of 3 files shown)


lance (commented):
Could a test be added for this?

matejvasek:
@lance test added.

cmd/prompt_test.go
```diff
@@ -0,0 +1,72 @@
...
+		t.Fatal(err)
+	}
+	defer console.Close()
+
+	go func() {
+		_, _ = console.ExpectEOF()
+	}()
+
+	go func() {
+		chars := expectedCreds.Username + enter + expectedCreds.Password + enter
```
lance, line 34: @matejvasek sorry for my ignorance here, but where does `enter` come from? I only see it twice in this file and it appears to be uninitalized, but I know that can't be the case or it would not compile.

cmd/prompt_test.go
```diff
@@ -0,0 +1,72 @@
...
+		for _, ch := range chars {
+			_, _ = console.Send(string(ch))
+			time.Sleep(time.Millisecond * 50)
```
lance, line 37: Why do you need to sleep here?

lance (commented):
Just a couple of questions. Overall looks good. Thanks for adding the test!

/lgtm

PR #999, opened 2022-04-28
Update spring boot cloud and native versions

trisberg:
# Changes

- :broom: Update springboot templates with current versions of spring-boot, spring-cloud and spring-native

/kind enhancement

**Release Note**

```release-note
Updated the springboot templates with the latest versions of spring-boot, spring-cloud and spring-native.
```

**Docs**

```docs

```


salaboy:
@trisberg I think you need to run `make` into the project so the `zz_filesystem_generated.go` is generated with the changes in the template

templates/springboot/http/pom.xml
```diff
@@ -18,9 +18,8 @@
 
   <properties>
     <java.version>11</java.version>
-    <spring-cloud.version>2021.0.1</spring-cloud.version>
-    <spring-cloud-function.version>3.2.3</spring-cloud-function.version>
-    <spring-native.version>0.11.3</spring-native.version>
+    <spring-cloud.version>2021.0.2</spring-cloud.version>
+    <spring-native.version>0.11.4</spring-native.version>
```
matejvasek, line 22: Why was setting of `spring-cloud-function.version` removed?

salaboy: I am guessing that is coming from spring cloud now..

salaboy: so we can probably remove the version on that artifact, @trisberg can you confirm?

trisberg: Yes, the new spring-cloud 2021.0.2 BOM pulls in spring-cloud-function version 3.2.4

trisberg: Plus, setting `spring-cloud-function.version` property doesn't override what is in the BOM for spring-cloud 2021.0.1 since this isn't part of the spring-boot BOM, you would actually have to define the dependency explicitly like

```
			<dependency>
				<groupId>org.springframework.cloud</groupId>
				<artifactId>spring-cloud-function-dependencies</artifactId>
				<version>3.2.3</version>
				<type>pom</type>
				<scope>import</scope>
			</dependency>
```

trisberg: Looks like I missed some dependency settings, added another commit to fix this

trisberg:
Rebased and squashed he changes

PR #1001, opened 2022-04-29
chore: use paketo builders for all runtimes

lance:
- 🧹  Use paketo for all builders

All runtimes were actually using paketo builders already, but the builders were still being explicitly set in manifest.yaml instead of existing as a default in builder.go like all of the others.


matejvasek:
@lance you need to regenerate zz_* file since there was another change to spring boot templates.

lkingland:
Thanks @lance, I was making this change on-language-at-a-time as I added the defaults to S2I; but there's no harm in taking care of them all at once!
/lgtm

PR #1010, opened 2022-05-09
fix: add paketo builder by default to func.yaml

lance:
# Changes

Removes the default builder image lookup map (which resulted in the same
returned value for all supported runtimes), and instead uses that as the
default when creating a new project from the default builtin templates.
This means that we will not default to using the paketo builder for
external language packs, but instead rely on the language pack authors
to specify this value explicitly.

This change is necessary to support on cluster builds using Tekton
Pipelines.

Tangentially, this commit also removes the `builders` array from
func.yaml. Now that I am writing this commit message, I am wondering if
that means that I should also write a migration for it. Hmmm... The
array was removed because it is no longer needed/relevant/used.

Also, unrelated to the original intent of this commit is the removal of
the health endpoints from the top level manifest.yaml in the default
templates. These defaults are already handled in repository.go (where a
copy and paste error was also addressed).

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/1005

/kind fix

```release-note
Makes paketo the explicitly default builder for all built in templates.
Removes `builders` from func.yaml
```


lance:
Note: migration appears to be additive - no obvious mechanism for removing fields (or presumably, renaming). To remove fields, we either need to use `Unmarshal` instead of `UnmarshalStrict`, or we need to accept that the user will see an error and add to the release notes what is up.

matejvasek:
cc @lkingland

repository.go
```diff
@@ -141,9 +140,10 @@ func NewRepository(name, uri string) (r Repository, err error) {
 		funcDefaults: funcDefaults{
 			HealthEndpoints: HealthEndpoints{
 				Liveness:  DefaultLivenessEndpoint,
-				Readiness: DefaultLivenessEndpoint,
```
lkingland, line 144: OOps!  Good catch

lkingland (commented):
@lance here is a summary of our synchronous discussion:
If we were to keep the structure named `builders` and as a map, we could repurpose the key to be the builder name (`buildpack` or `s2i`).  Then write this structure out to func.yaml when a Function is created using the in-code defaults.  When a Function is loaded, this would take precidence over the defaults, and of course a value provided manually (such as via the `builder-image` flag) would take precidence and update the func.yaml when provided.
This would solve a few problems:  able to use the func.yaml in the on-cluster builds when Func is not available, not require a migration, address our need to differentiate builders in func.yaml for downstream

lkingland:
One trick to this is that we would have to break or bend the abstraction a bit, because the core would need have both types of builders loaded at all times, rather than loading one implementation or the other depending.

lance:
Closing - after discussing in person, we think there is a better way to achieve this.

PR #1014, opened 2022-05-11
Fix S2I build with podman

matejvasek:
Workaround a bug in podman:
https://github.com/containers/podman/issues/13770

# Changes

- :bug: Fix s2i build with podman

```release-note
fix: s2i build with podman
```

### DIFF
--- a/s2i/docker_client_wrapper.go
+++ b/s2i/docker_client_wrapper.go
@@ -1,0 +1,62 @@
+package s2i
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+)
+
+// Wrapper to workaround https://github.com/containers/podman/issues/13770
+type podmanDockerClient struct {
+	client.CommonAPIClient
+}
+
+func (p podmanDockerClient) ContainerCommit(ctx context.Context, nameOrID string, opts types.ContainerCommitOptions) (types.IDResponse, error) {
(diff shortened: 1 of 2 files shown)


lkingland (commented):
Very glad to see.  Two questions
1.  I see there is a generated file, but no indication of how it was generated or how to update the generation should the API change.  Can we have (ideally) a generator included in the codebase or (at least) a comment on how to recreate this generation in the future?
2.  The generated file includes a mapping of ever method from the parent to the implementation to approximate a subclass.  Could we just use Go's [Struct Embedding](https://go.dev/doc/effective_go#embedding) to avoid having to generate the file in the first place?

s2i/docker_client_wrapper.go
```diff
@@ -0,0 +1,62 @@
...
+	"strings"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+)
+
+// Wrapper to workaround https://github.com/containers/podman/issues/13770
+type podmanDockerClient struct {
+	pimpl client.CommonAPIClient
+}
```
lkingland, line 16:
```suggestion
type podmanDockerClient struct {
	client.CommonAPIClient
}
```
This will map all fields from the embedded struct to the parent.  Presuming you match pointer receivers if necessary, this should remove the need to manually map every method on the parent to the child, while still allowing an "override" for `ContainerCommit` which exists on the parent.

lkingland (approved):
Looks like a winner to me
/lgtm

PR #1016, opened 2022-05-12
Use struct embedding instead of generated code

matejvasek:


lkingland (approved):
esoteric language features FTW!

/lgtm
/hold for tests

PR #1019, opened 2022-05-16
fix: ignore node_modules for s2i builds

### LINKED ISSUE #1018: S2I Node build fails when using dependencies
When a dependency is added to a Node project, the resultant Function fails to run.

The following Function:
```
$ func create --language node --template cloudevents myfunc
$ cd myfunc
$ npm install --save @google-cloud/translate
```
Yields the following error on `run`:
```
$ func build --builder=s2i
$ func run
...
> faas-js-runtime ./index.js

node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module '../'
Require stack:
- /opt/app-root/src/node_modules/.bin/faas-js-runtime
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/opt/app-root/src/node_modules/.bin/faas-js-runtime:4:29)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/opt/app-root/src/node_modules/.bin/faas-js-runtime' ]
}
npm timing command:run-script Completed in 217ms
npm timing npm Completed in 799ms
```
Thanks for the report @rhuss

lkingland:
- :gift: Exclude certain directories form S2I build tar know to be either unnecessary or error-prone
- :bug:  Error running Node Functions which unecessarily had node_modules sent to the build daemon

Exclude various files and cache directories from S2I builds.  This list may need to be expanded in the future.

/kind bug

Fixes #1018

### DIFF
--- a/s2i/builder.go
+++ b/s2i/builder.go
@@ -76,6 +76,13 @@ func (b *Builder) Build(ctx context.Context, f fn.Function) (err error) {
 	cfg.RuntimeImagePullPolicy = api.DefaultRuntimeImagePullPolicy
 	cfg.DockerConfig = s2idocker.GetDefaultDockerConfig()
 
+	// Excludes
+	// Do not include .git, .env, .func or any language-specific cache directories
+	// (node_modules, etc) in the tar file sent to the builder, as this both
+	// bloats the build process and can cause unexpected errors in the resultant
+	// Function.
+	cfg.ExcludeRegExp = "(^|/)\\.git|\\.env|\\.func|node_modules(/|$)"
+
 	// Environment variables
 	// Build Envs have local env var references interpolated then added to the
 	// config as an S2I EnvironmentList struct


s2i/builder.go
```diff
@@ -76,6 +76,13 @@ func (b *Builder) Build(ctx context.Context, f fn.Function) (err error) {
 	cfg.RuntimeImagePullPolicy = api.DefaultRuntimeImagePullPolicy
 	cfg.DockerConfig = s2idocker.GetDefaultDockerConfig()
 
+	// Excludes
+	// Do not include .git, .env, .func or any language-specific cache directories
+	// (node_modules, etc) in the tar file sent to the builder, as this both
+	// bloats the build process and can cause unexpected errors in the resultant
+	// Function.
+	cfg.ExcludeRegExp = "(^|/)\\.git|\\.env|\\.func|node_modules(/|$)"
```
matejvasek, line 84: Why is `(^|/)` and `(^|/)` is required?

matejvasek: Don't we also exclude the `func.yaml` file from the image since we do it in our buildpacks?

matejvasek:
I am afraid this won't merege because of the `cla/google`.

s2i/builder.go
lkingland: The first one is such that it catches them when the file is nested, but not a _part_ of a filename, the second question is that the func.yaml was requested to be included in the image by someone for introspection, becase it represents actual metadata about the Function which is pertinent.  Note how `.func` is not included, because that is expressly transient local-only metadata.  I would lean toward _not_ including func.yaml in the build, because I don't want any third-party systems to start relying on the exact format of func.yaml, but instead use `func` directly, or the client library where possible.  Perhaps we can pop the question to the community and ask if anyone has a good reason to keep it in

PR #1020, opened 2022-05-16
Make JS/TS Function templates async by default

matejvasek:
# Changes

- :broom: Make JS/TS Function template Functions async by default.
  Rationale: `async` is almost necessary for any non-trivial functionality.


templates/node/cloudevents/index.js
```diff
@@ -18,7 +18,7 @@ const { CloudEvent, HTTP } = require('cloudevents');
  * See: https://github.com/knative-sandbox/kn-plugin-func/blob/main/docs/guides/nodejs.md#the-context-object
  * @param {CloudEvent} event the CloudEvent
  */
-function handle(context, event) {
+const handle = async (context, event) => {
```
lance, line 21: Why not just add the `async` keyword as in `async function handle()`?

matejvasek: No particular reason, I just like arrow functions.

matejvasek: Also `function` binds `this` to something. Do we want that something expose?

zroubalik: I am not node expert, but I tend to agree with Lance, it is more obvious for me that way. But again, I am not expert in this area, so I am fine with any approach.

matejvasek: @lance are we OK with arrow functions or shall I revert to `function`.

matejvasek:
@lance PTAL

PR #1021, opened 2022-05-17
feat: expose default builder image logic for in-cluster builds

### LINKED ISSUE #1005: On-cluster build: builder image required in func.yaml
Currently, because we have defaults for the builder images, they are not set in `func.yaml` for functions created using the default builtin templates. However, the builder image name is required when running builds on the cluster. This causes friction with new projects, requiring the developer to manually set this value in the `func.yaml` configuration file. When it is not set, the following error occurs when the pipeline runs.
```
failed to create task run pod "viewer-git-pipeline-run-88t7b-build": missing field(s): steps[1].Image. Maybe missing or invalid Task default/func-buildpacks
```
/kind bug

lkingland:
# Changes

- :gift: use defaulting logic for on-cluster builder image deriviation

Tekton pipeline runs now use the same logic for determining default builder image as a local buildpack build.

In order to implement varying build strategies (s2i vs buildpacks), we were unable to rely on hard-coded builder images.  These defaults were moved in-code, with builders defined in func.yaml being an optional manual overrdide.  This PR updates the Tekton deployment process to rely on this defaulting logic, rather than require the func.yaml override (which would preclude a Function from being able to be built using other strategies).

/kind enhancement

Fixes #1005


matejvasek:
Linter is complaining.

zroubalik (approved):
/lgtm
/approve

Issue #1022, opened 2022-05-17
func.yaml per-builder settings


lkingland:
The func.yaml currently specifies a `Builder` and `Builders` which are in need of update.

While it is possible the naming may change, to make it consistent with our current logic, the Builder setting would be removed, and the key of the `builders` map should be set to either `s2i` or `buildpack`, whose value is the builder image the named builder should use.  

Currently, the value of Builder, if it exists, is used as the builder image regardless of which builder type is chosen.  This of course has the unintended effect of "pinning" any function which uses this field to one builder type or another.

[Docs](https://github.com/knative-sandbox/kn-plugin-func/blob/main/docs/reference/func_yaml.md) should be uddpated, along with the schema, and It would be ideal if this came with a func.yaml Migration as well.

PR #1024, opened 2022-05-24
feat: configurable s2i builder images

lkingland:
# Changes

- :gift: configurable s2i builder images
- :gift: refactor to enable fuller testing
- :gift: tests

S2I builder can now have its builder image defined in func.yaml (on the Function structure).   Previously it was a shared setting with Pack builders, which meant defining a builder image other than the default would peg the function to only work with one builder or the other.  also enables per-builder settings, which contributes to (but does not fully comple) #1022 

/kind enhancement


zroubalik (approved):
/lgtm
/approve

nice!

lance:
/override cla/google since this PR is already covered by easycla

s2i/builder.go
```diff
@@ -120,15 +134,41 @@ func (b *Builder) Build(ctx context.Context, f fn.Function) (err error) {
...
+// Function's Builders map.
+func builderImage(f fn.Function) (string, error) {
 	if f.Runtime == "" {
 		return "", ErrRuntimeRequired
 	}
-	v, ok := DefaultBuilderImages[f.Runtime]
-	if !ok {
-		return "", ErrRuntimeNotSupported
+
+	v, ok := f.BuilderImages["s2i"]
```
lance, line 148: @lkingland am I reading this correctly that if there is an s2i builder image defined, it will be used as the default?

PR #1028, opened 2022-05-25
feat: pack builder images individually configurable

lkingland:
# Changes

- :gift: configurable pack builder images via `builderImages` member
- :gift: refactor to enable fuller testing
- :gift: tests

Pack builder can now have its builder image defined in func.yaml (on the Function structure) alongside S2I without interfering with each other.   Previously it was a shared setting, which meant defining a builder image other than the default would peg the Function to only work with one builder or the other.  Also enables per-builder settings, which along with the update to the S2I builder and the [associated func.yaml migration](https://github.com/knative-sandbox/kn-plugin-func/pull/1033), will address #1022. 

/kind enhancement

### DIFF
--- a/buildpacks/builder.go
+++ b/buildpacks/builder.go
@@ -19,125 +19,197 @@ import (
 	"github.com/buildpacks/pack/pkg/logging"
 )
 
-// DefaultBuilderImages for Pack builders indexed by Runtime Language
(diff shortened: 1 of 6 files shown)


lance (commented):
/lgtm
glad to see this and https://github.com/knative-sandbox/kn-plugin-func/pull/1033 tidying things up

PR #1032, opened 2022-05-31
fix: increase timeout when testing int cluster availability

lkingland:
- :bug: fix frequent test timeouts

Of late the "Verify Configuration" step of the integrations tests has been failing.  This appears to be because the cluster made available in GitHub actions simply doesn't stand up the test `echo` service within the 30s default wait timeout.

Until such time as Kubernetes gets faster, or the GitHub VMs are more highly specced... this PR just increases the timeout from the default 30s to 120.  Also increased is the time between attempts to create the service, as it appears to always try 5 or more times before the endpoint is available.

/kind bug

### DIFF
--- a/hack/test.sh
+++ b/hack/test.sh
@@ -41,13 +41,17 @@ EOF
       exit 1
     fi
     echo "Retrying..."
-    sleep 5
+    sleep 10
   done
 
-  sleep 60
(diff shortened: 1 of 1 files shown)


zroubalik (approved):
/lgtm
/approve

PR #1033, opened 2022-06-01
feat: builder images map migration

### LINKED ISSUE #1022: func.yaml per-builder settings

lkingland:
:gift: Function auto-migration to `builderImages` map

func.yaml now includes a `builderImages` map which allows Functions to specify builder-specific images to use when building.  This deprecates the non-specific `builder` and `builders` attiributes.  This PR migrates Functions which were created prior to this deprecation by either removing the attributes if unused, or carrying them over if the user had customized the value.  This (along with #1024 and #1028) Fixes #1022

/kind enhancement

### DIFF
--- a/client.go
+++ b/client.go
@@ -557,18 +557,8 @@ func (c *Client) Create(cfg Function) (err error) {
 		return
 	}
 
-	// TODO: Create a status structure and return it for clients to use
-	// for output, such as from the CLI.
-	if c.verbose {
-		fmt.Printf("Builder:       %s\n", f.Builder)
-		if len(f.Buildpacks) > 0 {
-			fmt.Println("Buildpacks:")
-			for _, b := range f.Buildpacks {
-				fmt.Printf("           ... %s\n", b)
-			}
-		}
-		fmt.Println("Function project created")
-	}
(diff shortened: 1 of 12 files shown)


zroubalik:
@lkingland could you please rebase this PR so it includes only new changes? It's easier to do the review then

lkingland:
> @lkingland could you please rebase this PR so it includes only new changes? It's easier to do the review then

Done; I should have done that before removing this from draft status.  Thanks for taking a look @zroubalik

zroubalik (approved):
/lgtm
/approve

PR #1039, opened 2022-06-07
Plain JSON body in func invoke with HTTP

matejvasek:
# Changes

Using plain JSON body in func invoke with HTTP instead of form encoding.


lance:
/lgtm
However
/hold for test fixes

invoke.go
```diff
@@ -193,7 +194,20 @@ func sendPost(ctx context.Context, route string, m InvokeMessage, t http.RoundTr
...
 		}
 	}
-	resp, err := client.PostForm(route, values)
+
+	req, err := http.NewRequestWithContext(ctx, "POST", route, bytes.NewBufferString(m.Data))
+	if err != nil {
+		return nil, "", fmt.Errorf("failure to create request: %w", err)
+	}
+
+	ct := "application/json"
```
matejvasek, line 203: @lance seems the defaulting is handled somewhere else and the default value there is `text/plain` shall I change that?

matejvasek:
> I see how could I cause test breakage, but I don't see how could I have cause race.

I think the race must have been there all along and it only manifest when assert fails.

invoke.go
lance: > @lance seems the defaulting is handled somewhere else and the default value there is `text/plain` shall I change that?

Yes, I think so.

client_test.go
```diff
@@ -1054,20 +1055,30 @@ func TestClient_Invoke_HTTP(t *testing.T) {
 	handler := http.NewServeMux()
 	handler.HandleFunc("/", func(res http.ResponseWriter, req *http.Request) {
 		invoked = true
```
matejvasek, line 1057: How is this not a race?

lance (commented):
/lgtm
But we should address the race condition

client_test.go
matejvasek: To be clear this race is not reported by `-race`. What was reported and what I did fix is usage of `t.Fatal()`. It's not possible to run `t.Fatal()` from another thread.

matejvasek:
@lance ok I fixed potential race, plz final lgtm

matejvasek:
CI is flaking more than usual.

lkingland (approved):
I prefer this as well 👍🏻 

I am perplexed why the tests would throw a race error now and not earlier, however.

/lgtm

matejvasek:
@lkingland 
>I am perplexed why the tests would throw a race error now and not earlier, however.

You cannot call `t.Fatal()` outside the main test goroutine (`t.Error()` seems to be ok).

PR #1042, opened 2022-06-08
feat: use JSON object as default message for the invoke sub-command

matejvasek:
# Changes

Use JSON object as default message for the invoke sub-command

```release-note
Use JSON object (instead of plain text) as default message for the invoke sub-command
```


lkingland (approved):
/lgtm

We might need to update templates too

PR #1046, opened 2022-06-08
chore: print correct logs for failed PipelineRun

zroubalik:

# Changes

If Tekton Task consists of multiple Steps and there is a failure in one of the steps, the `ExitCode` on this step Step is set to `number != 0`.  `ExitCode == 1` is set on the last failed Step.

So if there are for example 3 steps and the 2nd step fails, this causes failure in the 3rd step as well. In this case the ExitCodes are set this way:
```
step1: ExitCode: 0
step2: ExitCode: 145 (and maybe random? number != 0)
step3: ExitCode: 1
```

The previous implementation printed logs for step3, but we are more interested in the step2 logs.  This PR fixes this problem.


lkingland (approved):
/lgtm

PR #1054, opened 2022-06-13
fix: use --as-dockerfile mode for s2i

matejvasek:
# Changes
* Use --as-dockerfile mode for `s2i` build.
  Rationale: daemon mode doesn't work with `podman` bacause of `podman`'s bugs.

/kind bug

```release-note
fix: s2i build with podman
```


jrangelramos:
lgtm
works fine w/ podman 4.0.0 and higher.

zroubalik (commented):
/lgtm

PR #1063, opened 2022-06-14
fix: s2i build as-dockerfile mode

matejvasek:
# Changes
* Use `"as dockerfile"` mode of `s2i` builds:
  Rationale: `podman` < 4.0 has faulty `/commit` endpoint so `"daemon mode"` is not working

/kind bug

```release-note
fix: s2i+podman<4.0 build error "CommitFailure: invalid change "LABEL" - must be formatted as KEY VALU"
```


matejvasek:
This version inspects appropriate label and sets s2i script url. This fixes the issue with openjdk builds.

matejvasek:
@lance @lkingland @zroubalik  PTAL

matejvasek:
Too bad this won't make it to midstream release. Maybe we could do at least upstream release?

zroubalik (approved):
/lgtm

/hold - merge as you like

lkingland:
Good workaround 
/lgtm as well

matejvasek:
To be honest I not super happy about this since this introduces unnecessary project copy but this is only way how to make this work on `podman v3.3`.

PR #1068, opened 2022-06-16
fix: minor improvements for s2i build

matejvasek:
# Changes
* Close opened file
* Better handling for symlinks in code


lance (commented):
/lgtm

PR #1074, opened 2022-06-20
Do not codecov ./templates

matejvasek:
# Changes

- :broom: Do not codecov `./templates`. Templates are not tested in a way that supports codecov.


lance (commented):
/lgtm

PR #1078, opened 2022-06-25
feat: introduce echo return for Python http template

### LINKED ISSUE #1077: Align python template to return an echo as other templates
The Python template is currently not aligned with other templates: 

https://github.com/knative-sandbox/kn-plugin-func/blob/main/templates/python/http/func.py

This should be refactored to follow the go template:

https://github.com/knative-sandbox/kn-plugin-func/blob/main/templates/go/http/handle.go#L20

mocdaniel:
# Changes

- :broom: Introduce echo return for Python http template

/kind enhancement

Fixes #1077 

```release-notes
Introduces echo return value fro Python HTTP template
```


matejvasek:
@mocdaniel the generated zz* file contains some files that shouldn't be there: `python/http/.func`, `python/http/.func/instances`.
https://github.com/knative-sandbox/kn-plugin-func/runs/7062209536?check_suite_focus=true#step:7:10

mocdaniel:
Ah, must've missed those. I will push another commit later today.

matejvasek:
Also you need to update e2e tests:
https://github.com/knative-sandbox/kn-plugin-func/runs/7062209687?check_suite_focus=true#step:8:15

matejvasek:
> Ah, must've missed those. I will push another commit later today.

It's really tricky, the test checks if zz-* matches your ./templates so it passes locally. It fails in CI action because the version in git differs.

matejvasek:
@mocdaniel look at this: https://github.com/knative-sandbox/kn-plugin-func/blob/main/docs/CONTRIBUTING.md#templates

lance (commented):
@mocdaniel thanks for the contribution! I just have a couple of questions about the actual echo behavior.

templates/python/http/func.py
```diff
@@ -9,5 +18,12 @@ def main(context: Context):
     """
 
     # Add your business logic here
+    print("Received request")
 
-    return { "message": "Howdy!" }, 200
+    if 'request' in context.keys():
+        ret = pretty_print(context.request)
+        print(ret)
+        return { "message": "hello" }, 200
```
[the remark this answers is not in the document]
mocdaniel: Hey lance, yes, you're probably right! I changed it back from an earlier commit after seeing the e2e-tests matching on fixed strings like `hello`, `howdy` or derivations of that, so I thought *printing* the response and *returning* a short answer was the way the templates were supposed to behave.

I'll gladly update the template so that it returns the actual response, probably some time tomorrow.

templates/python/http/func.py
```diff
@@ -1,4 +1,13 @@
 from parliament import Context
+from flask import Request
+
+
+def pretty_print(req: Request) -> str:
```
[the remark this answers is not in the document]
mocdaniel: Will do!

mocdaniel:
I updated `func.py` so that the response will always be a string of key-value pairs, no matter the request and content-type, like this:

```
{'message': 'hello'}
{'foo': 'bar', 'answer': '42'}
{}
```

I also adjusted the e2e-test and `test_func.py`.

templates/python/http/func.py
```diff
@@ -1,4 +1,46 @@
...
+
+
+# parse request body, json data or URL query parameters
+def payload_print(req: Request) -> str:
+    if req.method == "POST":
+        if req.is_json:
+            return str(req.json)
+        else:
+            # MultiDict needs some iteration
+            ret = "{"
```
matejvasek, line 12: Couldn't be this JSON crafting replaced by `json.dumps(req.{form,args})`?

mocdaniel: flask/werkzeug returns an `ImmutableMultiDict` for both, `form` and `args` - I haven't tried if `json.dumps()` transforms this data structure into the same representation. I went for the JSON crafting to make sure every valid request returns data formatted the exactly same way.

matejvasek: And does it returns actual JSON, I see it uses `'` not `"`.

matejvasek: It looks like it returns python dict literal.

matejvasek: For pretty print only it's OK I guess. But for date sent over write I would prefer JSON.

mocdaniel:
Hopefully the final update! I changed the output format for all requests, it looks like this now on the client's side:

```
curl "http://localhost:8080?data1=test&data2=123"                                                           
{"data1": "test", "data2": "123"}

curl -X POST -H "Content-Type: application/x-www-form-urlencoded" http://localhost:8080 -d "data1=test&data2=123"
{"data1": "test", "data2": "123"}

curl -X POST -H "Content-Type: application/json" http://localhost:8080 -d '{"data1":"test", "data2":"123"}'
{"data1": "test", "data2": "123"}
```

And like this on the server's side:

```
Received request
GET http://localhost:8080/?data1=test&data2=123 localhost:8080
  Host: localhost:8080
  User-Agent: curl/7.79.1
  Accept: */*
URL Query String:
  {"data1": "test", "data2": "123"}

Received request
POST http://localhost:8080/ localhost:8080
  Host: localhost:8080
  User-Agent: curl/7.79.1
  Accept: */*
  Content-Type: application/x-www-form-urlencoded
  Content-Length: 20
Request body:
  {"data1": "test", "data2": "123"}

Received request
POST http://localhost:8080/ localhost:8080
  Host: localhost:8080
  User-Agent: curl/7.79.1
  Accept: */*
  Content-Type: application/json
  Content-Length: 31
Request body:
  {"data1": "test", "data2": "123"}
```


Anything left that needs to be done/fixed?

matejvasek:
/lgtm
@lance PTAL

lance:
/unhold
Thanks for the contribution @mocdaniel !

mocdaniel:
> /unhold
> 
> Thanks for the contribution @mocdaniel !

Thank you for the positive discussions!

PR #1079, opened 2022-06-27
feat!: deploy autobuild and flag persistence

### LINKED ISSUE #857: On-cluster build: Validate all Git options
We should validate all options for `git`(`gitUrl`,`contextDir`, `revision`...), there are two scenarios:

1. if `build: git` is specified in `func.yaml`.
2. if `build` in `func.yaml` is not set to `git`, but `-b git` is specifed as a parameter for `func build|deploy`

Currently there is a problem if `build: local` is set in `func.yaml` and `-b git` is specified as a parameter and `gitUrl` is not set:

```bash
./func deploy -p test-func-b git 
🕛 Creating Pipeline resources
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x27eee7d]

goroutine 1 [running]:
knative.dev/kn-plugin-func/pipelines/tekton.generatePipeline({{0xc0005a0fa8, 0x6}, {0x7ff7bfeff5f1, 0x11}, {0xc0001292a8, 0x11}, {0x0, 0x0}, {0xc0005a0fd8, 0x2}, ...}, ...)
        /Users/zroubali/go/src/knative.dev/kn-plugin-func/pipelines/tekton/resources.go:40 +0x11d
knative.dev/kn-plugin-func/pipelines/tekton.(*PipelinesProvider).Run(_, {_, _}, {{0xc0005a0fa8, 0x6}, {0x7ff7bfeff5f1, 0x11}, {0xc0001292a8, 0x11}, {0x0, ...}, ...})
        /Users/zroubali/go/src/knative.dev/kn-plugin-func/pipelines/tekton/pipeplines_provider.go:94 +0x305
knative.dev/kn-plugin-func.(*Client).RunPipeline(0xc0005ad800, {0x30db938?, 0xc0009b1180}, {0x7ff7bfeff5f1, 0x11}, {0x0?, 0x0?, 0x0?})
        /Users/zroubali/go/src/knative.dev/kn-plugin-func/client.go:709 +0x271
knative.dev/kn-plugin-func/cmd.runDeploy(0xc0008b8a00, {0xc000c3fcb8?, 0xc000c3fc90?, 0x0?}, 0x2e80008)
        /Users/zroubali/go/src/knative.dev/kn-plugin-func/cmd/deploy.go:189 +0xbac
knative.dev/kn-plugin-func/cmd.NewDeployCmd.func1(0xc0008b8a00?, {0xc000b15280?, 0x4?, 0x4?})
        /Users/zroubali/go/src/knative.dev/kn-plugin-func/cmd/deploy.go:69 +0x25
github.com/spf13/cobra.(*Command).execute(0xc0008b8a00, {0xc000b15240, 0x4, 0x4})
        /Users/zroubali/go/src/knative.dev/kn-plugin-func/vendor/github.com/spf13/cobra/command.go:856 +0x67c
github.com/spf13/cobra.(*Command).ExecuteC(0xc000274000)
        /Users/zroubali/go/src/knative.dev/kn-plugin-func/vendor/github.com/spf13/cobra/command.go:974 +0x3b4
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/zroubali/go/src/knative.dev/kn-plugin-func/vendor/github.com/spf13/cobra/command.go:902
github.com/spf13/cobra.(*Command).ExecuteContext(...)
        /Users/zroubali/go/src/knative.dev/kn-plugin-func/vendor/github.com/spf13/cobra/command.go:895
main.main()
        /Users/zroubali/go/src/knative.dev/kn-plugin-func/cmd/func/main.go:37 +0x232
```

### LINKED ISSUE #1075: Enable auto builds on local deploy
## Summary

The `--build=auto` option recently added to the `run` subcommand should also be added to `deploy`.

In order to support this complexity, and two additional scenarios (discussed below) I would suggest we separate the current deploy subcommand's flags into two:  `--remote` indicating the deployment should not be done locally but should instead trigger a remote on-cluster deployment and `--build=[true|false|auto]` to match that which is currently on `run`.

## Explication

Currently the `--build` flag on `deploy` is used to both indicate if building should take place as well as whether the deployment should be triggered remotely on cluster rather than locally (`--build=[local|disabled|git]`).  We currently support building and deploying Functions 1) from the local system or 2) by triggering a build and deploy via Tekton pipelines.

With the introduction of cached builds (See #1031) and its `--build=auto|true|false`), the way these dimensions are expressed as flags becomes more complex.  Since we may soon add the third dimension of git repository vs source tarball (currently we only support git repository remotely and tarball locally), it is even more important to be clear before permutations start to expand.

For example, triggering a remote deploy but with a forced rebuild, or a remote deploy using a tarball of local source code, or a local deploy using source from a remote git repository are all scenarios that should in theory be possible but would be onerous when expressed as a single flag.  Allowing all permutations via a single `--build` flag worked until now, because we only had three possible scenarios, but now with the addition of `--build=auto|true|false`, it seems clear separation is in order.

The proposal is to simply separate into two flags:

`--remote`: whether or not we are deploying using local processing or triggering a remote deployment on-cluster.  Default is local, but will be configurable both on a per-Function basis, and globally as a user setting when the config system is introduced.

`--build`: whether a build should be performed, skipped or performed if the system detects it is necessary (`--build=auto`).

See #1031 for an example of how the `--build` flag is implemented on the `run` subcommand, which is intended as a trial run for how it would work on `deploy`.

### Examples:

```
func deploy
```
Perform a deployment using local context, only building if necessary (first build or when changes to the local filesystem have been detected)

```
func deploy --build
```
Perform a deployment using local context, forcing a build even if the source is unchanged (useful for example if the user manually removed the resultant container). 

```
func deploy --build=false
```
Perform a deployment using local context, skipping build even if filesystem change detected

```
func deploy --remote
```
Trigger a remote deployment of the local Function by uploading source to be built and deployed remotely.
(Until such time as the local source upload usecase is implemented, the first invocation will error that only --remote with a likewise remote --git-url for source code is supported).

Ideally, a future enhancement would further also only rebuild if the head of the git repository indicated (including branch) has advanced since the last deployment, but that is out of scope for this issue.

```
Func deploy --remote --build
```
Trigger a remote deployment.  Rebuilds the container even if the repository has no new updates.
Presuming the Function has already been configured with an assoicated remote --git-url, this is the way the system currently works (done).

```
func deploy --remote --build=false
```
Trigger a remote deployment.  Reuses the container from the last deployment if it exists.  Essentially equivalent to “redeploy”.

This is also a future enhancement, and could perhaps be implemented without the need of Tekton Pipelines, as it is effectively just re-deploying an existing, built Function.

### The `--git-url` flag

In addition to the –remote and –build dimensions, the –git-url dimension is worth a quick sidebar: we could allow local deploys to pull a git repository in lieu of using the local filesystem, and remote on-cluster deployments could forego using a git repository and instead receive a tarball of local source files.  This dimension is out of scope for this discussion, but it seems a useful thought experiment when deciding to separate the --build flag now rather than later.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/1125

### LINKED ISSUE #1115: cmd unit tests create func.yaml test file in source tree
Running command unit tests leave a `func.yaml` in the source tree.

This is caused by `Test_imageWithDigest`

This appears to be a new issue, and from a quick perusal of the source, it _should_ be writing this to a temp directory due to the use of `Fromtemp()`

### LINKED ISSUE #1229: On Cluster builds does not allow to override the namespace specified via CLI
On Cluster builds does not allow to override the namespace specified via CLI. It only uses the ns from `func.yaml`.

For exmaple:
I have a node js functions on GH with a default namespace https://github.com/jrangelramos/node-fn/blob/main/func.yaml#L3
When I invoke deploy commands using
``` 
$ func deploy -n test-ns
```
It does not use the new namespace, so I'm forced to update the namespace on func.yaml and push it to git.
```
rror: failed to run pipeline: function pipeline run has failed with message: 

Warning: Current namespace 'test-ns' does not match function which is deployed at 'default' namespace
⬆️  Deploying function to the cluster
Error: knative deployer failed to get the Knative Service: services.serving.knative.dev "node-fn" is forbidden: User "system:serviceaccount:test-ns:pipeline" cannot get resource "services" in API group "serving.knative.dev" in the namespace "default"
```

lkingland:
- :gift: Add autobuild to deploy
- :gift: Add --remote flag as distinct from --build
- :gift: Adds git-x flag persistence
- :gift: Command help update and expansion
- :gift: Improved tests and bug fixes to namespace
- :broom: Test update to use client API
- :broom: General test case cleanup

```release-notes
* Updates the `deploy` command with the `--build` logic from the `build` command which now supports `--build=auto`
* Adds a flag `--remote` to indicate if a deployment is to be triggered on-cluster
* Git flags for on-cluster builds persist by default
```

Fixes #1075  - Primary discussion of UX changes
Fixes #1115 
Fixes #857 
Fixes #1229 

/kind enhancement


cmd/deploy_test.go
```diff
@@ -1,182 +1,261 @@
...
-		expectFileURL        *string
-		expectFileBranch     *string
-		expectFileContextDir *string
-		expectCallURL        *string
-		expectCallBranch     *string
-		expectCallContextDir *string
-		errString            string
-	}{
-		{
-			name:      "Git arguments don't get saved to func.yaml but are used in the pipeline invocation",
```
lkingland, line 32: Part of this PR is proposing that we now default to persisting these flags

cmd/deploy_test.go
```diff
@@ -1,182 +1,261 @@
...
-			expectCallURL:        pointer.StringPtr("git@github.com:knative-sandbox/kn-plugin-func.git"),
-			expectCallBranch:     pointer.StringPtr("main"),
-			expectCallContextDir: pointer.StringPtr("func"),
-		},
-		{
-			name:      "Git arguments override func.yaml but don't get saved",
-			gitURL:    "git@github.com:knative-sandbox/kn-plugin-func.git",
-			gitBranch: "main",
-			gitDir:    "func",
-			funcFile: `name: test-func
```
lkingland, line 61: This PR updates tests to use the Client API rather than write func.yaml directly, as this is more likely to be backwards compatible (the exact format of func.yaml is an implementation detail of the client package not cmd, so tests relying on its structure should probably only exist there)

cmd/deploy.go
```diff
@@ -185,59 +201,133 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
...
+	// first serialized out to disk.  This interferes with the ability to have a
+	// --save option which only writes mutated values to disk on a successful
+	// completion.. i.e. we have to save to communicate the Functin state to the
+	// builder, potentially prematurely.  Therefore, in the forthcoming PR which
+	// persists all flags to func.yaml by default (option to --save=false), the
+	// Client API will need to change to accepting Function instances rather than
+	// paths on disk.  This has other knock-on benefits as well.
+	if err = f.Write(); err != nil {
+		return
+	}
```
lkingland, line 238: @zroubalik I believe this directly pertains to your question in the general discussion thread.

cmd/deploy_test.go
```diff
@@ -1,182 +1,421 @@
...
+	}
+	if f.Git.Revision != expectedBranch {
+		t.Errorf("expected git branch '%v' got '%v'", expectedBranch, f.Git.Revision)
+	}
+}
+
+// TODO: Test that an error is generated if --git-url includes a branch and
+// --git-branch is provided.
+
+// TODO:  --save
```
lkingland, line 421: @zroubalik here is the entrypoint for adding the --save flag, which will allow the user to opt-out of having the value of the other flags (`--git-url etc`) written to the func.yaml by default.

cmd/deploy.go
zroubalik: Yeah, I agree with the proposal for the Client API to work on Function instances, no matter if we go with the `--save` flag apporach.

lance (commented):
Overall this is a nice step forward, though quite large.
/lgtm
/hold for others

function.go
```diff
@@ -56,13 +56,8 @@ type Function struct {
 	// SHA256 hash of the latest image that has been built
 	ImageDigest string `yaml:"imageDigest"`
 
-	// BuildType represents the specified way of building the fuction
-	// ie. "local" or "git"
-	BuildType string `yaml:"build" jsonschema:"enum=local,enum=git"`
-
-	// Git stores information about remote git repository,
-	// in case build type "git" is being used
-	Git Git `yaml:"git"`
```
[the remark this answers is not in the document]
lkingland: I will work this into this PR 👍🏻

cmd/build_test.go
```diff
@@ -106,25 +92,6 @@ created: 2009-11-10 23:00:00`,
...
-		{
-			name:     "push flag with failing push",
-			pushFlag: true,
-			fileContents: `name: test-func
-runtime: go
-created: 2009-11-10 23:00:00`,
-			shouldBuild: true,
-			shouldPush:  true,
-			wantErr:     true,
-		},
```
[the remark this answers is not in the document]
lkingland: That is in error

cmd/deploy.go
```diff
@@ -540,42 +615,64 @@ func parseImageDigest(imageSplit []string, config deployConfig, cmd *cobra.Comma
...
+// effective given the current value and a given requested.
+func checkNamespaceDeploy(f fn.Function, config deployConfig, out io.Writer) string {
+	// --namespace provided
+	if config.Namespace != "" {
+		// If updating a function which is alredy presumably deployed (has the
+		// namespace member populated), that if we are requesting it be deployed
+		// to a different namespace, this results in an warning because it may
+		// result in an orphaned instance in another namespace.
+		if f.Namespace != "" && config.Namespace != f.Namespace {
+			fmt.Fprintf(out, "Warning: Function is in namespace '%v', but requested namespace is '%v'. Continuing with deployment to '%v'.\n", f.Namespace, config.Namespace, config.Namespace)
```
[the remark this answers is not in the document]
lkingland: This is an attempt to clean up the logic a bit in preparation for the Global Config updates, which should solidify this concept across the board.   The value is indeed persisted, just in a slightly different spot:

 The logic of this function just determines which value for namespace should take precidence in various situations:
  --namespace flag always wins
  Existing namespace if already deployed (and warning message if being changed)
  Active Kubernetes namespace if nothing provided and its the first deployment

The value returned is the final namespace to use, and the Function struct updated accordingly.
The deploy is  performed, and when successful, the Funciton is serialized to disk, updating the value of Namespace for subsequent (re)deployments.

buildpacks/builder.go
```diff
@@ -182,7 +182,7 @@ func BuilderImage(f fn.Function) (string, error) {
 		return "", ErrRuntimeRequired{}
 	}
 
-	v, ok := f.BuilderImages[fn.BuilderPack]
+	v, ok := f.BuilderImages["pack"]
```
zroubalik, line 185: Why this change?

cmd/build.go
```diff
@@ -101,96 +132,68 @@ func runBuild(cmd *cobra.Command, _ []string, newClient ClientFactory) (err erro
...
-	if function.Builder == "" || cmd.Flags().Changed("builder") {
-		function.Builder = config.Builder
-	} else {
-		config.Builder = function.Builder
-	}
-	if err = fn.ValidateBuilder(config.Builder); err != nil {
-		return err
-	}
-	if config.Builder == fn.BuilderPack {
+	if f.Builder == "pack" {
```
zroubalik, line 161:
```suggestion
	if f.Builder == fn.BuilderPack {
```

cmd/build.go
```diff
@@ -101,96 +132,68 @@ func runBuild(cmd *cobra.Command, _ []string, newClient ClientFactory) (err erro
...
 		if config.Platform != "" {
-			err = fmt.Errorf("the --platform flag works only with s2i build")
+			err = errors.New("the --platform flag works only with s2i builds")
 			return
 		}
 		builder = buildpacks.NewBuilder(buildpacks.WithVerbose(config.Verbose))
-	} else if config.Builder == fn.BuilderS2i {
+	} else if f.Builder == "s2i" {
```
zroubalik, line 167:
```suggestion
	} else if f.Builder == fn.BuilderS2i {
```

cmd/deploy.go
```diff
@@ -116,157 +174,187 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
...
-	if function.Builder == "" || cmd.Flags().Changed("builder") {
-		function.Builder = config.Builder
-	} else {
-		config.Builder = function.Builder
-	}
-	if err = fn.ValidateBuilder(config.Builder); err != nil {
-		return err
-	}
-	if config.Builder == fn.BuilderPack {
+	if config.Builder == "pack" {
```
zroubalik, line 219:
```suggestion
	if config.Builder == fn.BuilderPack {
```

cmd/deploy.go
```diff
@@ -116,157 +174,187 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
...
 		if config.Platform != "" {
-			err = fmt.Errorf("the --platform flag works only with s2i build")
+			err = errors.New("the --platform flag works only with s2i builds")
 			return
 		}
 		builder = buildpacks.NewBuilder(buildpacks.WithVerbose(config.Verbose))
-	} else if config.Builder == fn.BuilderS2i {
+	} else if config.Builder == "s2i" {
```
zroubalik, line 225:
```suggestion
	} else if config.Builder == fn.BuilderS2i {
```

s2i/builder.go
```diff
@@ -372,7 +372,7 @@ func builderImage(f fn.Function) (string, error) {
 		return "", ErrRuntimeRequired
 	}
 
-	v, ok := f.BuilderImages[fn.BuilderS2i]
+	v, ok := f.BuilderImages["s2i"]
```
zroubalik, line 375: I really prefer one place with a constants for this kind of fields, than some random strings spread across the codebase.

function_builder.go
```diff
@@ -1,33 +0,0 @@
-package function
```
zroubalik, line 1: I don't like the removal of this.

zroubalik (commented):
I have a concern in moving from constants defined in one place to strings. 

Also it is very hard to give a proper review on such a huge PR. Would be much better to actually split it into multiple smaller ones. And I know it is sometimes hard to achieve :)

lkingland:
> Would be much better to actually split it into multiple smaller ones. And I know it is sometimes hard to achieve :)

Yes, this was one of those situations where I pull on a string and the whole sweater unravels.  I can take the time to split out some of the sub-goals which ended up being done here for more isolated introspection.

s2i/builder.go
lkingland: Sometimes strings are better than constants.  I won't try to defend them in this case, and am happy to put them back in.  One thing worth ponderig: the core package has only interfaces, and knows nothing of the various implementations of the Builder interface which exist elsewhere.  Placing a constant there may cause confusion in the future if we are not diligent to keep the core clean of dependencies :)

function.go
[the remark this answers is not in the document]
lkingland: Closing this as resolved since it has its own issue.

function_builder.go
lkingland: This still exists, it was moved along with the constants.  I will extract into a separate PR

buildpacks/builder.go
lkingland: Extracted into #1180

cmd/prompt_test.go
```diff
@@ -33,8 +33,8 @@ func Test_newPromptForCredentials(t *testing.T) {
 	go func() {
 		chars := expectedCreds.Username + enter + expectedCreds.Password + enter
 		for _, ch := range chars {
+			time.Sleep(time.Millisecond * 100)
```
lkingland, line 36: This test was failing at times, with a value for username as `estuser`.  My assmption was that it sent the first character, `t`, ever so slightly before the mock terminal was ready to accept.

cmd/deploy.go
```diff
@@ -199,58 +251,95 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
...
+			// to be consistent, this should throw an error, with the registry
+			// prompting code placed within config.Prompt and triggered with --confirm
+			fmt.Println("A registry for function images is required. For example, 'docker.io/tigerteam'.")
+			if err = survey.AskOne(
+				&survey.Input{Message: "Registry for function images:"},
+				&config.Registry, survey.WithValidator(
+					NewRegistryValidator(config.Path))); err != nil {
+				return ErrRegistryRequired
+			}
+			fmt.Println("Note: building a function the first time will take longer than subsequent builds")
```
lance, line 264: Noting the same here as in my previous comment about this text. I also don't really love the little bits of duplicated code between build and deploy, but haven't looked too closely at that.

cmd/deploy.go
```diff
@@ -199,58 +251,95 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
...
+	// Perform the deployment either remote or local.
+	if config.Remote {
+		if f.Git.URL == "" {
+			return ErrURLRequired // Provides CLI-specific help text
 		}
-		if err := client.Build(cmd.Context(), config.Path); err != nil {
-			return err
+		// Invoke a remote build/push/deploy pipeline
+		// Returned is the function with fields like Registry and Image populated.
+		if f, err = client.RunPipeline(cmd.Context(), f); err != nil {
```
lance, line 277: 💯

lance (approved):

matejvasek:
@lance also linter is complaining.

lance:
> @lance also linter is complaining.

Frustrating because this is not as easy to rectify as the previous issue with the conflict. This file actually should not even be in the PR.

lkingland:
> > @lance also linter is complaining.
> 
> Frustrating because this is not as easy to rectify as the previous issue with the conflict. This file actually should not even be in the PR.

Thanks for noticing.  That must have been remoed after this PR was first started, and it snuck back in during a rebase!

I made a few minor wording changes, help text tweaks etc... and removed that file.

lance:
> > > @lance also linter is complaining.
> 
> > 
> 
> > Frustrating because this is not as easy to rectify as the previous issue with the conflict. This file actually should not even be in the PR.
> 
> 
> 
> Thanks for noticing.  Perhaps `commands.txt` has since been removed after this PR was first started, and it sneaked back in during a rebase.
> 
> 
> 
> I made a few minor wording changes, help text tweaks etc... and removed that file.

Awesome thanks @lkingland - truly appreciate you checking in on this while on your lengthy travels! We should be able to land this PR tomorrow. 🙏💪

lance:
/lgtm
Let's do this!

lkingland:
For some reason this PR is stuck in "Checking for Ability to merge automatially"

I first tried pushing a minor commit to restart the process (clearing the /lgtm status of course) to no avail.

I have now contacted GitHub support via ticket.  I don't expect a timely response, however, because I had no choice but to choose the option "Issue is transient, happening only once" 

GitHub's official status shows no outages nor degredation.  I am now in a "wait-and-see" holding pattern..

zroubalik (approved):
/lgtm

lkingland:
🥂

PR #1080, opened 2022-06-27
Improve error message for s2i unsupported runtimes

matejvasek:
# Changes
- :broom:  Improve error message for s2i unsupported runtimes


s2i/builder.go
```diff
@@ -386,5 +382,18 @@ func builderImage(f fn.Function) (string, error) {
...
+	var e ErrRuntimeNotSupported
+	return errors.As(err, &e)
+}
+
+type ErrRuntimeNotSupported struct {
+	Runtime string
+}
+
+func (e ErrRuntimeNotSupported) Error() string {
+	return fmt.Sprintf("the s2i builder has no default builder image for the %q language runtime", e.Runtime)
```
lance, line 398: Do you think it's useful to indicate here that buildpacks can be used if the runtime is one that is supported OOTB? Otherwise, lgtm

matejvasek: Changed. @lance PTAL

lance (commented):
/lgtm

PR #1081, opened 2022-06-27
Explicit define builder for Remote and Config e2e tests

jrangelramos:
# Changes

- :broom: Defines builder for e2e tests tied to GoLang functions such a Test Remote Repository, Config Envs, Labels and Volumes
kind <kind>

/kind cleanup

### DIFF
--- a/test/_e2e/cmd_config_labels_test.go
+++ b/test/_e2e/cmd_config_labels_test.go
@@ -57,6 +57,7 @@ func TestConfigLabel(t *testing.T) {
 	project.Template = "http"
 	project.FunctionName = "test-config-labels"
 	project.ProjectPath = filepath.Join(os.TempDir(), project.FunctionName)
+	project.Builder = "pack"
 
 	Create(t, knFunc.TestShell, project)
 	defer func() { _ = project.RemoveProjectFolder() }()
@@ -75,6 +76,7 @@ func TestConfigLabel(t *testing.T) {
 
 	// Deploy
 	knFunc.TestShell.WithEnv(testEnvName, testEnvValue)
+	Build(t, knFunc.TestShell, &project)
 	Deploy(t, knFunc.TestShell, &project)
 	defer Delete(t, knFunc.TestShell, &project)
 
--- a/test/_e2e/cmd_config_volumes_test.go
+++ b/test/_e2e/cmd_config_volumes_test.go
@@ -101,6 +101,7 @@ func TestConfigVolumes(t *testing.T) {
 	project.FunctionName = "test-config-volumes"
 	project.ProjectPath = filepath.Join(os.TempDir(), project.FunctionName)
 	project.RemoteRepository = "http://github.com/boson-project/test-templates.git"
+	project.Builder = "pack"
 
 	Create(t, knFunc.TestShell, project)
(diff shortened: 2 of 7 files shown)


test/_e2e/config.go
```diff
@@ -8,7 +8,7 @@ import (
 // Intended to provide setup configuration for E2E tests
 const (
 	defaultRegistry        = "localhost:50000/user"
-	testTemplateRepository = "http://github.com/lance/test-templates.git" //nolint:varcheck,deadcode
+	testTemplateRepository = "http://github.com/boson-project/test-templates.git" //nolint:varcheck,deadcode
```
lance, line 11: Woah - how long has that been there?! Oops :)

test/_e2e/cmd_config_volumes_test.go
```diff
@@ -139,6 +140,7 @@ func TestConfigVolumes(t *testing.T) {
 	configVolumesRemove("/bad-cm", enter)
 
 	// Deploy
+	Build(t, knFunc.TestShell, &project)
```
lance, line 143: Same comment here about the deploy command now accepting the `--builder` flag.

lance (commented):
/lgtm
/hold for others

test/_e2e/cmd_config_labels_test.go
```diff
@@ -75,6 +76,7 @@ func TestConfigLabel(t *testing.T) {
 
 	// Deploy
 	knFunc.TestShell.WithEnv(testEnvName, testEnvValue)
+	Build(t, knFunc.TestShell, &project)
```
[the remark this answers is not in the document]
jrangelramos: yea.. it will. and yes.. that check is for.. "in case" it is already built (by `func build`) the deploy command doesn't need to build it

test/_e2e/config.go
jrangelramos: for months.. lol.. just notice now as well...

zroubalik (commented):
Should we create a separate test case for `explicit builder` instead?  Ie. keep existing tests for testing of the default behaviour (builder not set explictily).

jrangelramos:
@zroubalik Yes, and we already have one specific for s2i builder. Just a little more background about this change. The tests being modifed here are meant to test the ones related to Remote repository and Config (as envs and volumes). These tests actually builds and deploys a pre-existing function from a template on boson-project. They are implemented in Go and have some logic that aids test validations. 

For example, the config envs test function adds envs (from config maps and secrets) https://github.com/knative-sandbox/kn-plugin-func/blob/main/test/_e2e/cmd_config_envs_test.go#L34-L54 and the validation expected the deployed function (https://github.com/boson-project/test-templates/blob/main/go/envs/handle.go) to read ENV vars and return them to the actual test, so it can compare and ensure the envs got properly loaded from maps/secreats.

These template test functions are all written in Go, so having the builder as "pack" ensure the tests will be executed properly on upstream and eventually on midstream projects, which in OSH case, uses s2i builder (but that builder currently does not support golang as runtime).  

The other solution would be change these test templates implementation to Node so the could be cross-builder (pack+s2i) and test would pass on both upstream and midstream.

zroubalik (approved):
/lgtm
/approve

PR #1089, opened 2022-06-29
docs: replace commands.md with generated text file

lance:
This commit removes commands.md document which tends to get out of date
pretty easily and replaces it with a generated commands.txt file. The
help text generation is fairly naive, but should be sufficient for the
time being.

/kind documentation


zroubalik:
Interesting approach. Though we will loose some information with this migration, for example information like this:
```
Function name must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character (e.g. 'my-name', or '123-abc', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?').
```

What if we put this kind of info directly into the source code and generate it then from there?

lance:
> What if we put this kind of info directly into the source code and generate it then from there?

Do you mean add this text to the help text in the commands?

lance:
So strange that I keep closing issues when typing a comment and hitting cmd-enter. Trying again to see if it's just my sloppy fingers or if the UX has changed.

lkingland (approved):
/lgtm

Definitely agree this is The Way.

/hold for others to offer feedback as well

lance:
This is an improvement. But I wonder if a better approach for the long term is to have external markdown documents that the CLI uses for help, and are also the web-based reference documentation. Inside the CLI we could use Glamour to render the markdown. https://github.com/charmbracelet/glamour

But I am stumbling a little bit, as I am not sure how to slurp up these markdown documents at compile time. It seems this would require something similar to the templates, where the help text documents are used to construct an internal file system which we then slurp the bits up and render with Glamour.

Does this make sense? Am I making it too complicated? /cc @lkingland @salaboy

lkingland:
This PR is a step in the right direction, and I wouldn't want to let the perfect be the enemy of better.  This at least gets us using one source:  the source code.

I like the idea of Markdown, I don't care for the idea of extracting help from the source code and being dependent on external non-source code.  Go works on source code alone, and (for example) non-go files are not included in vendoring nor when importing a library as a dependency.

Perhaps we can have the best of both worlds:  we can keep the source code the Source of Truth, use Markdown syntax for the help text, render the Markdown using a shiny library on `func [command] --help`, and a `go generate` command to export this help text (markdown) to external documentation files/wiki.

I would of course like to reiterate that this PR, as it stands, is an improvement, and perhaps this Markdown suggestion is a good "next step".  Just please, keep the .go files the actual canonical source...code

lance:
> Perhaps we can have the best of both worlds: we can keep the source code the Source of Truth, use Markdown syntax for the help text, render the Markdown using a shiny library on func [command] --help, and a go generate command to export this help text (markdown) to external documentation files/wiki.

This makes sense to me - thanks for the suggestion.

> I reiterate that this PR, as it stands, is an improvement, and perhaps worth merging while we investigate this Markdown suggestion as a good next step?

I tend to agree. Will continue to hold if others want to chime in.

lance:
@zroubalik are there other sentences you think should be moved into the source code? I'm happy to make this change.

lance:
@zroubalik @lkingland not sure where we stand on this PR. Happy to make changes, close it, or merge it, but want to do whatever with consensus.

lkingland:
> @zroubalik @lkingland not sure where we stand on this PR. Happy to make changes, close it, or merge it, but want to do whatever with consensus.

@lance my vote is that we merge it, but I am open to suggestions.  Reason being this seems a solid step forward, but that perhaps we also open a new issue to revisit the solution herein and consider some of the other suggestions that came up (most notably perhaps using markdown syntax in the mix)

lance:
> > @zroubalik @lkingland not sure where we stand on this PR. Happy to make changes, close it, or merge it, but want to do whatever with consensus.
> 
> @lance my vote is that we merge it, but I am open to suggestions. Reason being this seems a solid step forward, but that perhaps we also open a new issue to revisit the solution herein and consider some of the other suggestions that came up (most notably perhaps using markdown syntax in the mix)

@zroubalik yours is the only outstanding objection - wdyt?

zroubalik:
I agree with solution proposed by @lkingland, we should track this issue (that we are losing bunch of info with this migration).

/lgtm

lance:
@zroubalik I've pushed new changes to address recent changes in CLI help text. PTAL and /lgtm it again - ty

hack/build-docs.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+BIN=./func
```
zroubalik, line 3: One nit, shouldn't we run automatically? add Makefile target for this to be part of `build`

lance: Yes I suppose so - I will take a look into that.

lance:
@zroubalik added to Makefile ptal

zroubalik (approved):
/lgtm
/approve

PR #1090, opened 2022-06-29
feat: deploy success message also displays the namespace

### LINKED ISSUE #1083: The CLI should print the namespace a function was deployed to
When a user runs `func deploy`, if the deployment is successful, they are shown the URL endpoint for their function. This is nice and useful, but it would be good to also print the namespace that the function was deployed in. This is because our target user may not be very Kubernetes savvy and as a result may not really understand the implications of deploying to one namespace or another (and may not even be aware of the namespace context).

Here is the current output from `func deploy`.

```
❯ func deploy --build disabled
   Function updated at URL: http://viewer.default.127.0.0.1.sslip.io
```

The output should instead look like this or similar

```
❯ func deploy --build disabled
   Function updated at URL: http://viewer.default.127.0.0.1.sslip.io in the <user's namespace> namespace
```

/kind good-first-issue

vyasgun:
# Changes

- Added the function's namespace to the success message after deploy
- If the namespace in function obj is blank, display current context namespace

Fixes #1083

**Release Note**

```release-note
Prints the namespace when a function is deployed
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/mitchellh/go-homedir"
+	"knative.dev/kn-plugin-func/k8s"
 )
 
 const (
@@ -705,10 +706,21 @@ func (c *Client) Deploy(ctx context.Context, path string) (err error) {
 	// Deploy a new or Update the previously-deployed Function
 	c.progressListener.Increment("Deploying function to the cluster")
 	result, err := c.deployer.Deploy(ctx, f)
+
+	var namespace string
+	if f.Namespace == "" {
+		var err1 error
+		namespace, err1 = k8s.GetNamespace("")
+		if err1 != nil {
+			return err1
+		}
+	} else {
+		namespace = f.Namespace
+	}
 	if result.Status == Deployed {
-		c.progressListener.Increment(fmt.Sprintf("Function deployed at URL: %v", result.URL))
+		c.progressListener.Increment(fmt.Sprintf("Function deployed in namespace %q and exposed at URL: %v", namespace, result.URL))
 	} else if result.Status == Updated {
-		c.progressListener.Increment(fmt.Sprintf("Function updated at URL: %v", result.URL))
+		c.progressListener.Increment(fmt.Sprintf("Function updated in namespace %q and exposed at URL: %v", namespace, result.URL))
 	}
 
 	return err
--- a/client.go
+++ b/client.go
@@ -15,7 +15,6 @@ import (
 	"time"
 
 	"github.com/mitchellh/go-homedir"
-	"knative.dev/kn-plugin-func/k8s"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


client.go
```diff
@@ -705,10 +706,21 @@ func (c *Client) Deploy(ctx context.Context, path string) (err error) {
 	// Deploy a new or Update the previously-deployed Function
 	c.progressListener.Increment("Deploying function to the cluster")
 	result, err := c.deployer.Deploy(ctx, f)
+
+	var namespace string
```
zroubalik, line 710: The namespace is already resolved in `c.deployer.Deploy()`, might be better to set it in the `fn.DeploymentResult` and reuse here.

client.go
```diff
@@ -705,10 +706,21 @@ func (c *Client) Deploy(ctx context.Context, path string) (err error) {
...
+		namespace, err1 = k8s.GetNamespace("")
+		if err1 != nil {
+			return err1
+		}
+	} else {
+		namespace = f.Namespace
+	}
 	if result.Status == Deployed {
-		c.progressListener.Increment(fmt.Sprintf("Function deployed at URL: %v", result.URL))
+		c.progressListener.Increment(fmt.Sprintf("Function deployed in namespace %q and exposed at URL: %v", namespace, result.URL))
```
zroubalik, line 721:
```suggestion
		c.progressListener.Increment(fmt.Sprintf("Function deployed in namespace %q and exposed at URL: \n%v", namespace, result.URL))
```

See the comment: https://github.com/knative-sandbox/kn-plugin-func/issues/1083#issuecomment-1169739020

client.go
```diff
@@ -705,10 +706,21 @@ func (c *Client) Deploy(ctx context.Context, path string) (err error) {
...
 	} else if result.Status == Updated {
-		c.progressListener.Increment(fmt.Sprintf("Function updated at URL: %v", result.URL))
+		c.progressListener.Increment(fmt.Sprintf("Function updated in namespace %q and exposed at URL: %v", namespace, result.URL))
```
zroubalik, line 723:
```suggestion
		c.progressListener.Increment(fmt.Sprintf("Function updated in namespace %q and exposed at URL: \n%v", namespace, result.URL))
```
See the comment: https://github.com/knative-sandbox/kn-plugin-func/issues/1083#issuecomment-1169739020

zroubalik (commented):
Please update On cluster build related messages as well: 
https://github.com/knative-sandbox/kn-plugin-func/blob/81091863e7e02681b5951b211bbba69282d0a430/pipelines/tekton/pipeplines_provider.go#L157-L159

client.go
vyasgun: Just to clarify, should I add a field for namespace in the DeploymentResult type and populate it when it's resolved in `c.deployer.Deploy()`?

lkingland (approved):
/lgtm !

This is nice information to have very readily available on deploy.

/hold for others to take a look

zroubalik (approved):
/lgtm thanks for the contribution!

vyasgun:
@zroubalik @lkingland Thanks for the reviews. I am stuck on the codecov part. I have added changes to the Deploy function but I can't see a way to use mock clients for writing unit tests.

zroubalik:
@vyasgun I think that's okay, the test  that you have added is imho enough for this feature

PR #1091, opened 2022-06-29
Improve usability of `func` CLI for IDE plugins

matejvasek:
# Changes
- :gift: Better input handling for non terminal stdio. Useful for scripting with `func`.
- :gift: Make `not implemented` error of cred-helper non fatal error.

```release-note
feat: better input handling for non terminal stdio. Useful for scripting with `func`.
```


lkingland (approved):
/lgtm

PR #1096, opened 2022-07-05
[WIP] Initial dry-run using client NewServingClientGitOps

### LINKED ISSUE #1060: Exporting Knative Service when requested in the CLI using --dry-run
As discussed in this issue: https://github.com/knative-sandbox/kn-plugin-func/issues/817 
As discussed in the weekly meeting

The `func` CLI should: 
- Allow the user to export the Knative Service that is being created programmatically using --dry-run

This will allow users to save the resource and apply it manually if they want to override some configurations.

salaboy:
DO NOT MERGE this is a draft PR to discuss collaborations with the `client` project and their `--target` flag: 
https://github.com/knative/client/issues/1195

# Changes
Uses the NewServingClientGitOps client to write down YAML files instead of doing a deployment to a cluster. For this PR to work you need to create a directory called `out`, we can change that later, where all the files will be created. 

- :gift: Add new feature

/kind enhancement

Fixes #1060

### DIFF
--- a/client.go
+++ b/client.go
@@ -90,7 +90,7 @@ type Pusher interface {
 // Deployer of Function source to running status.
 type Deployer interface {
 	// Deploy a Function of given name, using given backing image.
-	Deploy(context.Context, Function) (DeploymentResult, error)
+	Deploy(context.Context, Function, bool) (DeploymentResult, error)
 }
 
 type DeploymentResult struct {
@@ -478,9 +478,9 @@ func (c *Client) New(ctx context.Context, cfg Function) (err error) {
 	}
 
 	// Deploy the initialized Function, returning its publicly
-	// addressible name for possible registration.
+	// addressable name for possible registration.
 	c.progressListener.Increment("Deploying Function to cluster")
-	if err = c.Deploy(ctx, f.Root); err != nil {
+	if err = c.Deploy(ctx, f.Root, false); err != nil {
 		return
 	}
 
@@ -685,7 +685,7 @@ func (c *Client) printBuildActivity(ctx context.Context) {
 
 // Deploy the Function at path. Errors if the Function has not been
 // initialized with an image tag.
-func (c *Client) Deploy(ctx context.Context, path string) (err error) {
+func (c *Client) Deploy(ctx context.Context, path string, dryrun bool) (err error) {
 	go func() {
 		<-ctx.Done()
 		c.progressListener.Stopping()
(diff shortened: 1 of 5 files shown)


client.go
```diff
@@ -703,12 +703,16 @@ func (c *Client) Deploy(ctx context.Context, path string) (err error) {
...
 
 	// Deploy a new or Update the previously-deployed Function
-	c.progressListener.Increment("Deploying function to the cluster")
-	result, err := c.deployer.Deploy(ctx, f)
-	if result.Status == Deployed {
-		c.progressListener.Increment(fmt.Sprintf("Function deployed at URL: %v", result.URL))
-	} else if result.Status == Updated {
-		c.progressListener.Increment(fmt.Sprintf("Function updated at URL: %v", result.URL))
+	if !dryrun {
+		c.progressListener.Increment("Deploying function to the cluster")
```
rhuss, line 707: why would you skip the progress message ? I would suggest to do something like for kubectl:

```
$ kubectl run  --image quay.io/minio/mc mc --dry-run=client
pod/mc created (dry run)
```

i.e. print out the progress message but also indicate that it's just a dry run.

knative/deployer.go
```diff
@@ -73,18 +73,55 @@ func (d *Deployer) isImageInPrivateRegistry(ctx context.Context, client clientse
...
+	if err != nil {
+		err = fmt.Errorf("knative deployer failed to generate the Knative Service: %v", err)
+		return fn.DeploymentResult{}, err
+	}
+	err = client.CreateService(ctx, service)
+	if err != nil {
+		err = fmt.Errorf("knative deployer failed to deploy the Knative Service: %v", err)
+		return fn.DeploymentResult{}, err
+	}
+	return fn.DeploymentResult{}, nil
```
rhuss, line 110: Why would you always return an empty `DeploymentResult{}` ? What is the purpose of this return value ?

client.go
```diff
@@ -1009,7 +1013,11 @@ func (n *noopPusher) Push(ctx context.Context, f Function) (string, error) { ret
 // Deployer
 type noopDeployer struct{ output io.Writer }
 
-func (n *noopDeployer) Deploy(ctx context.Context, _ Function) (DeploymentResult, error) {
+func (n *noopDeployer) Deploy(ctx context.Context, _ Function, _ bool) (DeploymentResult, error) {
+	return DeploymentResult{}, nil
+}
+
+func (n *noopDeployer) DeployDryRun(ctx context.Context, _ Function) (DeploymentResult, error) {
```
rhuss, line 1020: this seems to be never used ? (as you pass the dry-run thing via an option ?)

rhuss (commented):
I'm not sure what the GitOps client does, but I suspect it will print out generated resources to the standard (is this correct ?)

As mentioned elsewhere, I don't think that `dry run` capture this behavior, and it is very confusing (especially if looking at the semantics that other UNIX tools are doing it). Yes, helm does it wrong, too :).

A "dry run" means a "simulation of the action requested" with all steps, validation, console output, and server-side interaction like the _real_ action, but without side effects. It is used to (a) see what steps are happening (therefore, i.e., the console output, like in `make --dry-run` to find out which targets will be called) and to (b) minimize the probability that the actual action will fail (therefor server-side interaction to see whether authentication and validation would succeed).

It does not mean that the output should be reused in any context other than this simulation.

I suppose that you are looking for the generation of the YAML files that are targeted to be reused elsewhere (like storing it in git and applying it via kubectl in a later step). Using an `--output` or `--file` or anything else that reflects that use case better is IMO a much better choice than `--dry-run`

client.go
rhuss: btw, `kubectl run` follows my reasoning in the review comment: It does *not* print out the Pod manifest's yaml but the _steps_ it would perform (and includes validation and that it can reach the server).

PR #1098, opened 2022-07-07
feat: func deploy accepts image digest in --image

### LINKED ISSUE #1058: `--image` flag should allow specifying image digest with `func deploy`
For s2i on cluster builds we need to find a way how to relay infromation about produced image digest from s2i build task to func deploy task. With Buildpacks approach we are modifying `func.yaml`, we would need to modify each s2i Tekton task to achieve the same.

Instead we should allow specifying image sha in the `--image` flag -> if sha is not present we will keep the existing behavior, if sha is present we should use it as a source of truth when deploying an image.

so for example this:

`func deploy --image docker.io/foo/my-function:latest@sha356:412401928741029875012975102987501298712312`
will use the input sha during deployment.

gauron99:
# Changes
*changes for func deploy*
- --image allows image digest for func deploy as [registry]/[namespace]/[name]:[tag]@[digest]
- if digest is given, push and build have to be disabled and image with digest is saved in func.yaml file
- few tests in deploy_test.go included

/kind enhancement

Fixes #1058

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -63,7 +63,7 @@ that is pushed to an image registry, and finally the function's Knative service 
 	// Flags shared with Build specifically related to building:
 	cmd.Flags().StringP("builder", "", "pack", "build strategy to use when creating the underlying image. Currently supported build strategies are 'pack' and 's2i'.")
 	cmd.Flags().StringP("builder-image", "", "", "builder image, either an as a an image name or a mapping name.\nSpecified value is stored in func.yaml (as 'builder' field) for subsequent builds. ($FUNC_BUILDER_IMAGE)")
-	cmd.Flags().StringP("image", "i", "", "Full image name in the form [registry]/[namespace]/[name]:[tag]@[digest] (optional). This option takes precedence over --registry (Env: $FUNC_IMAGE). If digest is given, build and push have to be disabled.")
+	cmd.Flags().StringP("image", "i", "", "Full image name in the form [registry]/[namespace]/[name]:[tag]@[digest] (optional). This option takes precedence over --registry (Env: $FUNC_IMAGE). If digest is given, build and push are disabled.")
 	cmd.Flags().StringP("registry", "r", GetDefaultRegistry(), "Registry + namespace part of the image to build, ex 'quay.io/myuser'.  The full image name is automatically determined based on the local directory name. If not provided the registry will be taken from func.yaml (Env: $FUNC_REGISTRY)")
 	cmd.Flags().BoolP("push", "u", true, "Attempt to push the function image to registry before deploying (Env: $FUNC_PUSH)")
 	setPathFlag(cmd)
@@ -103,17 +103,15 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 		return
 	}
 
-	//if --image contains '@', validate image digest and disable build if valid, otherwise return an error
+	//if --image contains '@', validate image digest and disable build and push if not set, otherwise return an error
 	imageSplit := strings.Split(config.Image, "@")
 	imageDigestProvided := false
 
 	if len(imageSplit) == 2 {
-		if err := parseImageDigest(imageSplit, config); err != nil {
-			return err
+		if config, err = parseImageDigest(imageSplit, config, cmd); err != nil {
+			return
 		}
 		imageDigestProvided = true
-		config.BuildType = "disabled"
-		config.Image = imageSplit[0]
 	}
 
 	function, err := functionWithOverrides(config.Path, functionOverrides{Namespace: config.Namespace, Image: config.Image})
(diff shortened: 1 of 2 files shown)


cmd/deploy.go
```diff
@@ -63,7 +63,7 @@ that is pushed to an image registry, and finally the function's Knative service
 	// Flags shared with Build specifically related to building:
 	cmd.Flags().StringP("builder", "", "pack", "build strategy to use when creating the underlying image. Currently supported build strategies are 'pack' and 's2i'.")
 	cmd.Flags().StringP("builder-image", "", "", "builder image, either an as a an image name or a mapping name.\nSpecified value is stored in func.yaml (as 'builder' field) for subsequent builds. ($FUNC_BUILDER_IMAGE)")
-	cmd.Flags().StringP("image", "i", "", "Full image name in the form [registry]/[namespace]/[name]:[tag] (optional). This option takes precedence over --registry (Env: $FUNC_IMAGE)")
+	cmd.Flags().StringP("image", "i", "", "Full image name in the form [registry]/[namespace]/[name]:[tag]@[digest] (optional). This option takes precedence over --registry (Env: $FUNC_IMAGE). If digest is given, build and push have to be disabled.")
```
lance, line 66: Nit
```suggestion
	cmd.Flags().StringP("image", "i", "", "Full image name in the form [registry]/[namespace]/[name]:[tag]@[digest] (optional). This option takes precedence over --registry (Env: $FUNC_IMAGE). If digest is given, build and push are disabled.")
```

cmd/deploy.go
```diff
@@ -472,3 +490,25 @@ func validateBuildType(buildType string) error {
...
+		return fmt.Errorf("value '%s' in --image has invalid prefix syntax for digest (should be 'sha256:')", config.Image)
+	}
+
+	if len(imageSplit[1][7:]) != 64 {
+		return fmt.Errorf("sha256 hash in '%s' from --image has the wrong length (%d), should be 64", imageSplit[1], len(imageSplit[1][7:]))
+	}
+
+	// if BuidType was not explicitly set via CLI, set to 'disabled' without an error
+	if config.BuildType != "" && config.BuildType != "disabled" {
+		return fmt.Errorf("build type '%s' is not accepted with --image with digest. Use 'disabled' or none", config.BuildType)
```
lance, line 506: Nit
```suggestion
		return fmt.Errorf("build type '%s' is not accepted when using --image with digest. Use 'disabled' or none", config.BuildType)
```
Also, I worry that `use "disabled" or none` could be misleading and cause folks to use `--build none`, which will also produce an error. Here is another suggestion that could work.
```suggestion
		return fmt.Errorf("The --build flag is not valid when using --image with a digest.")
```

cmd/deploy.go
```diff
@@ -472,3 +490,25 @@ func validateBuildType(buildType string) error {
...
+	}
+
+	if config.Push {
+		return fmt.Errorf("--image was specified with digest, therefore --push is not allowed")
```
lance, line 510:
```suggestion
		return fmt.Errorf("The --deploy flag is not valid when using --image with a digest")
```
To be consistent with the second suggestion above for the `--build` flag.

gauron99: i realized the --push flag is true by default and therefore it has to be specified as `--push=false` in order to work with digests so i thought of:
1. change error message to something more explicit like: "the --push flag has to be false when using --image with a digest"
2. set --push to false explicitly in the code and print out a warning if left as true (by default) or by a user via CLI

(following point 2 i could explicitly set `--build` to disabled and print out a warning if it was not disabled)

would any of these suggestions be good or should i keep your initial suggestion?

lance: My personal preference is number 2, setting `--push` to false in the code and printing a warning. I think `--build` could possibly work the same way. But we must tread carefully. Explicitly setting `--push` and `--build` configuration to essentially noop should probably only happen if the user has not specified anything at all, and instead these values are just the defaults. If the user has explicitly set the `--build` or `--push` flag to something invalid, and ALSO has an `--image` tag with a digest, we can't really know the intent, so perhaps failing fast in this case would be better.

gauron99:
*UPDATE*

when image with digest is provided it is checked if `--build` or `--push` were set via CLI. If not, print a warning and keep going. If they were set to a wrong value, error is returned. I also updated error messages based on previous suggestions

cmd/deploy.go
```diff
@@ -63,7 +63,7 @@ that is pushed to an image registry, and finally the function's Knative service
 	// Flags shared with Build specifically related to building:
 	cmd.Flags().StringP("builder", "", "pack", "build strategy to use when creating the underlying image. Currently supported build strategies are 'pack' and 's2i'.")
 	cmd.Flags().StringP("builder-image", "", "", "builder image, either an as a an image name or a mapping name.\nSpecified value is stored in func.yaml (as 'builder' field) for subsequent builds. ($FUNC_BUILDER_IMAGE)")
-	cmd.Flags().StringP("image", "i", "", "Full image name in the form [registry]/[namespace]/[name]:[tag] (optional). This option takes precedence over --registry (Env: $FUNC_IMAGE)")
+	cmd.Flags().StringP("image", "i", "", "Full image name in the form [registry]/[namespace]/[name]:[tag]@[digest] (optional). This option takes precedence over --registry (Env: $FUNC_IMAGE). If digest is given, build and push are disabled.")
```
zroubalik, line 66: Just a nit, I'd highlight that digest is not mandatory and also put the `(Env: $FUNC_IMAGE)` to the end of the section.
```suggestion
	cmd.Flags().StringP("image", "i", "", "Full image name in the form [registry]/[namespace]/[name]:[tag]@[digest] (optional). This option takes precedence over --registry. Specifying digest in the full image name is not mandatory, but if it is given, 'build' and 'push' phases are disabled. (Env: $FUNC_IMAGE)")
```

Something like this^^

zroubalik (approved):
/lgtm
/hold for others

cmd/deploy.go
```diff
@@ -472,3 +488,40 @@ func validateBuildType(buildType string) error {
...
+	}
+
+	// if the --push flag was set by a user to 'true', return an error
+	if cmd.Flags().Changed("push") && config.Push {
+		return config, fmt.Errorf("the --push flag '%v' is not valid when using --image with digest", config.Push)
+	}
+
+	// if the --push flag was not set by a user, print a warning
+	if !cmd.Flags().Changed("push") {
+		fmt.Printf("Warning: --push flag not set when using --image with digest, setting to 'false'\n")
```
lance, line 519: Similar to above, I think this might be unnecessary. But this has me thinking, what happens if the image with the given digest exists only locally? Is it necessary for the user to manually push the image to the registry using `docker push`? Images built by the `func` CLI are only pushed to the registry when you run `func deploy`, so it seems perhaps we don't want to prohibit pushing the image?

gauron99: I unified the printing message and removed the conditions so it gets printed out every time image with digest is successfully parsed, does this work or were you thinking something else?

zroubalik (approved):
/lgtm

@lance WDYT? I need this and https://github.com/knative-sandbox/kn-plugin-func/pull/1099 as a prerequisite for the on cluster s2i stuff, so I am blocked atm.

cmd/deploy.go
zroubalik: @lance that's a great point with the push. Doesn't push change the image digest though? I'd keep the behavior as it is for now, we can always change this. This is anyway something that users probably won't use directly, but mainly through Tekton

PR #1099, opened 2022-07-08
feat: persist builder value in func.yaml

### LINKED ISSUE #1087: `builder` should be persisted to the func.yaml
Currently there’s only option to specify `builder` (`pack`/`s2i`) via cmd argument, there’s no way how to permanently specify `s2i` in func.yaml, it always defaults to `pack`.

We should persist the configuration to `func.yaml`.

vyasgun:
# Changes
Added a `Builder` field to Function type which will be populated by setting the `--builder` flag during build and deploy and used as a default when the flag is not used.

- Added `Builder` field to Function type.
- Added code to persist the `builder` value in the func.yaml file.
- Added code to use the `builder` value in func.yaml as default when none is specified.

Fixes #1087 

**Release Note**

```release-note
Persists builder in func.yaml
```


cmd/build.go
```diff
@@ -144,21 +144,18 @@ func runBuild(cmd *cobra.Command, _ []string, newClient ClientFactory) (err erro
...
+	if config.Builder == "" {
+		if function.Builder == "" {
+			config.Builder, function.Builder = "pack", "pack"
+		} else {
+			config.Builder = function.Builder
+		}
+	} else {
+		function.Builder = config.Builder
+	}
+
```
lkingland, line 158: The command instance allows us to check if a flag was actually provided, so perhaps this will suffice:

```go
  if f.Builder == "" || cmd.Flags().Changed("builder") {
    f.Builder = config.Builder
    if err = f.Write(); err != nil {
      return
    }
  }
```

cmd/build.go
```diff
@@ -50,7 +50,7 @@ and the image name is stored in the configuration file.
 		PreRunE:    bindEnv("image", "path", "builder", "registry", "confirm", "push", "builder-image", "platform"),
 	}
 
-	cmd.Flags().StringP("builder", "b", "pack", "build strategy to use when creating the underlying image. Currently supported build strategies are 'pack' and 's2i'.")
+	cmd.Flags().StringP("builder", "b", "", "build strategy to use when creating the underlying image. Currently supported build strategies are 'pack' and 's2i'.")
```
lkingland, line 53: It's important to retain the flag default, such that it appears in help text.

lkingland (changes requested):
Thanks for the contribution.  In addition to the two comments below, It would be nice to have a test or two, even if they are not foolproof:

For example `build_test.go` could include:
```go
// TestBuild_BuilderPersistance ensures that the builder chosen is read from
// the Function by default, and is able to be overridden by flags/env vars.
func TestBuild_BuilderPersistance(t *testing.T) {
  root, rm := Mktemp(t)
  defer rm()

  var err error

  // Create a Function without a builder
  f := fn.Function{
    Root:    root,
    Name:    "myFunc",
    Runtime: "go",
  }
  if err := fn.New().Create(f); err != nil {
    t.Fatal(err)
  }

  // Build the Function, relying on the flag deafult to set a value for builder
  cmd := NewBuildCmd(NewClientFactory(func() *fn.Client {
    return fn.New(fn.WithRegistry("registry.example.com/namespace))
  }))
    t.Fatal(err)
  }

  // Assert the Function has persisted a value of builder (has a default)
  if f, err = fn.NewFunction(root); err != nil {
    t.Fatal(err)
  }
  if f.Builder == "" {
    t.Fatal("value of builder not persisted using a flag default")
  }

  // Build the Function, specifying a Builder
  cmd.SetArgs([]string{"--builder=s2i"})
    t.Fatal(err)
  }

  // Assert the Function has persisted the value of builder
  if f, err = fn.NewFunction(root); err != nil {
    t.Fatal(err)
  }
  if f.Builder != "s2i" {
    t.Fatal("value of builder flag not persisted when provided")
  }

  // Build the Function again using a different builder
  cmd.SetArgs([]string{"--builder=pack"})
    t.Fatal(err)
  }

  // Assert the Function has persisted the new value
  if f, err = fn.NewFunction(root); err != nil {
    t.Fatal(err)
  }
  if f.Builder != "pack" {
    t.Fatal("value of builder flag not persisted on subsequent build")
  }
}
```
With the same test wherever else that logic is needed (ie. `deploy_test.go`)  👍🏻

cmd/build.go
vyasgun: Thanks, this looks much better!

cmd/deploy.go
```diff
@@ -154,6 +154,26 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
...
+	var builder fn.Builder
+	if function.Builder == "" || cmd.Flags().Changed("builder") {
+		function.Builder = config.Builder
+	} else {
+		config.Builder = function.Builder
+	}
+	if config.Builder == "pack" {
+		if config.Platform != "" {
+			err = fmt.Errorf("the --platform flag works only with s2i build")
+			fmt.Fprintln(os.Stderr, err)
```
matejvasek, line 167: If we are returning error the print should not be necessary.

cmd/build.go
```diff
@@ -144,24 +144,25 @@ func runBuild(cmd *cobra.Command, _ []string, newClient ClientFactory) (err erro
...
-		config.Registry = ""
-	}
-
-	// Choose a builder based on the value of the --builder flag
-	var builder fn.Builder
 	if config.Builder == "pack" {
 		if config.Platform != "" {
-			fmt.Fprintln(os.Stderr, "the --platform flag works only with s2i build")
+			err = fmt.Errorf("the --platform flag works only with s2i build")
+			fmt.Fprintln(os.Stderr, err)
```
matejvasek, line 164: If we are returning error the print should not be necessary.

cmd/build_test.go
```diff
@@ -150,3 +151,94 @@ created: 2009-11-10 23:00:00`,
...
+	if err := cmd.Execute(); err != nil {
+		t.Fatal(err)
+	}
+
+	var err error
+	// Assert the Function has persisted a value of builder (has a default)
+	if f, err = fn.NewFunction(root); err != nil {
+		t.Fatal(err)
+	}
+	if f.Builder != "pack" {
```
matejvasek, line 184: @vyasgun I am afraid this would fail in midstream where `"s2i"` is default.

cmd/deploy_test.go
```diff
@@ -270,3 +270,94 @@ runtime: go`,
...
+	if err := cmd.Execute(); err != nil {
+		t.Fatal(err)
+	}
+
+	var err error
+	// Assert the Function has persisted a value of builder (has a default)
+	if f, err = fn.NewFunction(root); err != nil {
+		t.Fatal(err)
+	}
+	if f.Builder != "pack" {
```
matejvasek, line 303: @vyasgun I am afraid this would fail in midstream where `"s2i"` is default.

matejvasek:
Minor nit: the two tests looks similar, could they be merged?

cmd/deploy_test.go
matejvasek: maybe just test non-emptiness

matejvasek:
otherwise looks good.

matejvasek:
Also: merging this potentially invalidates some midstream patches I'll take care of it for you.

vyasgun:
@matejvasek thanks for the review. I have changed the tests to not assume a default value and just check for emptiness and also merged both of them.

PR #1103, opened 2022-07-11
fix: use creds from creds store first

matejvasek:
# Changes

- :bug: Use credentials from secure store first. The library for un-encrypted credentials returns empty credentials without any error.
This is issues for `docker.io` which for whatever reason started accepting empty credentials.
 
/kind bug

```release-note
fix: empty credentials issue for docker.io
```


zroubalik (approved):
/lgtm

PR #1109, opened 2022-07-12
fix: image push authorization check

### LINKED ISSUE #1094: Docker Credentials being mixed up by func
Having two docker credentials (personal and enterprise) accounts caused `func` to get confused and not respect the authentication that the docker command is using. 

To reproduce: 
- docker login with a user1
- func create + func deploy to that user1, this means that docker org will be `user1`
- docker logout 
- docker login with user2
- change function docker org to user2
- func deploy -> this will try to deploy the func, but first it will build the image and try to push to `user2` but this fails with: 
```
Error: denied: requested access to the resource is denied
```
- At this point if you do docker push user2/myfunc it will work but func deploy will fail to push to `user2/myfunc`

matejvasek:
# Changes

- :bug: Verify write/push authorization before push, up until now we just checked if login into registry works.

/kind bug

```release-note
fix: verify write/push authorization before push, before we only check if login works
```
resolves #1094

### DIFF
--- a/docker/creds/credentials.go
+++ b/docker/creds/credentials.go
@@ -32,30 +32,28 @@ var ErrUnauthorized = errors.New("bad credentials")
 
 var ErrCredentialsNotFound = errors.New("credentials not found")
 
-// VerifyCredentialsCallback checks if credentials are accepted by the registry.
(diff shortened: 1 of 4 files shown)


zroubalik (approved):
/lgtm

PR #1118, opened 2022-07-15
feat: always store namespace in func.yaml and warn if current ns & func.yaml ns is different

### LINKED ISSUE #1084: A Function's namespace should always be stored in func.yaml
Currently, when a user deploys a function, if they do not specify the namespace using the `--namespace` flag, the default namespace/context is used. But we don't store that information in func.yaml. We only store it if the user explicitly specifies it. 

Instead, the first time a function is deployed, the namespace should be recorded in func.yaml even if we just use the default.

On subsequent deployments, if the user's current context differs from what is found in func.yaml, the user should receive a warning message.

gauron99:
# Changes
*changes to `func deploy`*

- store namespace to func.yaml on first deployment (from current namespace)
- if namespace is not specified via CLI and is different from func.yaml print warning explaining this 
(ex: `func.yaml->namespace = default` &&  `oc->namespace = myns` -> print warning where func is deployed)
- if namespace is specified via CLI, assume user knows what hes doing, do nothing

*update*
- add fake KUBECONFIG to all the tests in deploy_test.go
- replace `context.TODO` with `context.Background` in deploy_test.go
- print warning to stderr

Fixes #1084

### DIFF
--- a/cmd/deploy_test.go
+++ b/cmd/deploy_test.go
@@ -101,6 +101,8 @@ created: 2009-11-10 23:00:00`,
 			errString: "remote git arguments require the --build=git flag",
 		},
 	}
+
+	defer WithEnvVar(t, "KUBECONFIG", fmt.Sprintf("%s/testdata/kubeconfig_deploy_namespace", cwd()))()
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			var captureFn fn.Function
@@ -229,6 +231,8 @@ runtime: go`,
 runtime: go`,
 		},
 	}
+
+	defer WithEnvVar(t, "KUBECONFIG", fmt.Sprintf("%s/testdata/kubeconfig_deploy_namespace", cwd()))()
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			deployer := mock.NewDeployer()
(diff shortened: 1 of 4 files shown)


matejvasek:
@gauron99 the unitests are failing

matejvasek:
you need to add
`defer WithEnvVar(t, "KUBECONFIG", fmt.Sprintf("%s/testdata/kubeconfig_deploy_namespace", cwd()))()`
into `Test_runDeploy`.

cmd/deploy_test.go
```diff
@@ -270,3 +274,88 @@ runtime: go`,
...
+			viper.SetDefault("namespace", tt.namespace)
+			viper.SetDefault("registry", tt.registry)
+			defer viper.Reset()
+
+			// set test case's func.yaml
+			if err := os.WriteFile("func.yaml", []byte(tt.funcFile), os.ModePerm); err != nil {
+				t.Fatal(err)
+			}
+
+			ctx := context.TODO()
```
matejvasek, line 344: this could be `context.Background()`

cmd/deploy_test.go
```diff
@@ -270,3 +274,88 @@ runtime: go`,
...
+		t.Run(tt.name, func(t *testing.T) {
+			deployer := mock.NewDeployer()
+			defer Fromtemp(t)()
+			cmd := NewDeployCmd(NewClientFactory(func() *fn.Client {
+				return fn.New(
+					fn.WithDeployer(deployer))
+			}))
+
+			// set namespace argument if given & reset after
+			cmd.SetArgs([]string{}) // Do not use test command args
```
matejvasek, line 334: Why we cannot use command args?

cmd/deploy_test.go
zroubalik: Yeah, use context.Background() everywhere in the test.

PR #1123, opened 2022-07-18
feat: build for Linux ARM64

matejvasek:
# Changes

- :gift: Add build for Linux/arm64

/kind enhancement

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,8 +48,12 @@ jobs:
           path: func_darwin_arm64
       - uses: actions/upload-artifact@v2
         with:
-          name: Linux Binary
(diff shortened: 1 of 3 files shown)


csantanapr:
Thank you so much for this @matejvasek ❤️

lance (commented):
@matejvasek can you add the binary to the release script here? https://github.com/knative-sandbox/kn-plugin-func/blob/cccb2833d1ad8f77cfb200ef4f43a7e4098af92d/hack/release.sh#L28 

And we should update the brew scripts in https://github.com/knative-sandbox/homebrew-kn-plugins as well.

matejvasek:
@lance PTAL

lance (commented):
/lgtm

PR #1133, opened 2022-07-20
fix: stop build progress loop when build completes

lance:
This change prevents the incremental build messages that loop while
building from continuing during subsequent deploy phases by providing a
cancelable context to the `printBuildActivity()` function parameters.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/974

-->
```release-note
Fixes a bug where build output overwrites user-facing credentials prompts.
```
/kind fix

### DIFF
--- a/client.go
+++ b/client.go
@@ -608,7 +608,8 @@ func ensureRuntimeDir(f Function) error {
 // not contain a populated Image.
 func (c *Client) Build(ctx context.Context, path string) (err error) {
 	c.progressListener.Increment("Building function image")
-
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	// If not logging verbosely, the ongoing progress of the build will not
 	// be streaming to stdout, and the lack of activity has been seen to cause
 	// users to prematurely exit due to the sluggishness of pulling large images


matejvasek:
@lance why adding channel? Shouldn't this be done via context cancellation? There is already `case <-ctx.Done():` in code.

matejvasek:
```diff
diff --git a/client.go b/client.go
index be18f4d7..4ee41972 100644
--- a/client.go
+++ b/client.go
@@ -609,12 +609,14 @@ func ensureRuntimeDir(f Function) error {
 func (c *Client) Build(ctx context.Context, path string) (err error) {
        c.progressListener.Increment("Building function image")
 
+       ctx, cancel := context.WithCancel(ctx)
+       defer cancel()
+
        // If not logging verbosely, the ongoing progress of the build will not
        // be streaming to stdout, and the lack of activity has been seen to cause
        // users to prematurely exit due to the sluggishness of pulling large images
-       t := make(chan int)
        if !c.verbose {
-               c.printBuildActivity(ctx, t) // print friendly messages until context is canceled
+               c.printBuildActivity(ctx) // print friendly messages until context is canceled
        }
 
        f, err := NewFunction(path)
@@ -649,11 +651,10 @@ func (c *Client) Build(ctx context.Context, path string) (err error) {
                message = fmt.Sprintf("Function image built: %v", f.Image)
        }
        c.progressListener.Increment(message)
-       close(t)
        return
 }
 
-func (c *Client) printBuildActivity(ctx context.Context, t chan int) {
+func (c *Client) printBuildActivity(ctx context.Context) {
        m := []string{
                "Still building",
                "Still building",
@@ -675,10 +676,6 @@ func (c *Client) printBuildActivity(ctx context.Context, t chan int) {
                                c.progressListener.Stopping()
                                ticker.Stop()
                                return
-                       case <-t:
-                               c.progressListener.Stopping()
-                               ticker.Stop()
-                               return
                        }
                }
        }()
```

lance:
@matejvasek I was concerned about multi-step client functions such as `New()` which calls `Create()`, `Build()` and `Deploy()` with the same context. Maybe that doesn't matter?

matejvasek:
> @matejvasek I was concerned about multi-step client functions such as `New()` which calls `Create()`, `Build()` and `Deploy()` with the same context. Maybe that doesn't matter?

@lance not sure what you mean, you would create local (child) context, its cancellation shouldn't cancel parent.

PR #1137, opened 2022-07-21
test: remove verbose from test invoke create/deploy commands

jrangelramos:
# Changes

:broom: Just removing `--verbose` from `create` and `deploy` commands on `invoke` e2e tests  just to make e2e logs a bit cleaner. Verbose option is kept on `invoke` command since it is the target command being tested

/kind cleanup


lance (commented):
/lgtm
/approve

PR #1139, opened 2022-07-26
feat: use rendered markdown for CLI help text

lance:
# Changes

- 🎁 uses the Cobra markdown generation and glamour markdown rendering to display rendered markdown for most help commands
- 🎁 modifies existing docs generation to use `go run docs/generator/main.go` which extracts the help in markdown format

Most commands use the default help text renderer. Others like `create` have a bunch of custom stuff going on that results in a little hackiness. Submitting the PR anyway. If merged, we can iterate and improve on it.

/kind enhancement

Fixes https://github.com/knative-sandbox/kn-plugin-func/issues/1124

**Release Note**

```release-note
renders markdown for help text
```


lance:
Some thoughts on this PR for discussion.

- Output of rendered markdown causes some tests to fail due to expectations around text-only output (no control characters/codes).
- Some users may prefer simpler text output (no control characters/codes), so even if we keep this, we should consider support for the [`NO_COLOR` environment variable](https://nocolor.org).
- This implementation requires a bit of hackery to directly embed markdown in the text output, which makes writing and updating help text difficult
- Cobra already supports generated Markdown output for non-terminal text. If we were to consistently use the Cobra variables for things like `cmd.Example` and other named sections, we could generate markdown help documents without affecting the CLI output.

Given these issues, I'm leaning towards not using this implementation, and rather going with generated Cobra output. Thoughts @knative-sandbox/func-tastic-approvers ?

lance:
This PR is just feeling like too much of a hack to keep. Going with https://github.com/knative-sandbox/kn-plugin-func/pull/1169 for now and closing this. We can revisit if needed in the future.

PR #1144, opened 2022-07-29
fix: use space prefix for deploy output text

lance:
- 🐛  change deploy output text so that it is prefixed with a space, lining up with other text output on the left-hand column, and hopefully fixing the problem where the spinner icon overwrites part of the URL for Windows users.

I also added a couple of emoji to make the output a little friendlier.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/1138

### DIFF
--- a/client.go
+++ b/client.go
@@ -699,13 +699,13 @@ func (c *Client) Deploy(ctx context.Context, path string) (err error) {
 	}
 
 	// Deploy a new or Update the previously-deployed function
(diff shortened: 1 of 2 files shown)


lance:
@jrangelramos can you please try a build of my branch to confirm if this resolves the problem for you? I think it only happens on Windows.

zroubalik (approved):
/lgtm
/approve

PR #1154, opened 2022-08-03
doc: fix doc reference on nodejs template

jrangelramos:
# Changes

- :broom: fix doc reference on nodejs template

/kind cleanup


zroubalik (approved):
/lgtm
/approve

thanks, good catch :)

PR #1160, opened 2022-08-10
Added binary builds for platforms ppc64le and s390x

haubenr:
# Changes

This PR adds the binaries for platforms ppc64le and s390x to the list of release artifacts. Although at this point function support on these platforms is limited (thanks to paketo.io only supporting x86_64) this is the first step in properly suppporting functions on ppc64le and s390x (possibly via the s2i feature).

/kind enhancement


lance:
/lgtm
/hold for tests
Thanks @haubenr!

PR #1168, opened 2022-08-15
fix: ensure registry in func.yaml is respected

lance:
# Changes

:bug: If the user has a value specified in func.yaml for the image registry, that value should be used by the build command, if not otherwise specified as a flag or via the environment.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/1159

/kind bug

**Release Note**

```release-note
Fixes function builds when a registry exists in func.yaml but an image has not yet been built
```


lkingland (approved):
/lgtm

PR #1169, opened 2022-08-15
doc: use cobra markdown generator for command docs

lance:
# Changes

This commit removes the docs/reference/commands.txt file and instead uses the Cobra built-in documentation generator to create markdown documents for every command.

- 🎁 use Cobra to generate markdown
- 🗑️ remove commands.txt reference doc and its generator

/kind cleanup
/kind documentation


Makefile
```diff
@@ -98,7 +98,7 @@ clean: clean_templates ## Remove generated artifacts such as binaries and schema
 
 docs: build
 	# Generating command reference doc
-	@./hack/build-docs.sh
+	go run docs/generator/main.go
```
lkingland, line 101: 👍🏻

lkingland (approved):
I really like how this is going so far.  The Markdown generated by Cobra seems to do just fine.   I hope the custom templates can get worked out.

Will this replace #1139 ?

/lgtm so far

lance:
> Will this replace #1139 ?

Yes, this is meant to replace that PR as an alternative. I think it's probably the better path in the long run in spite of the blingy allure of glow.

lance:
Not sure what to do about the verify check that reports this error.
```
Found diffs in: third_party/VENDOR-LICENSE/github.com/opencontainers/selinux/go-selinux/LICENSE
fatal: ambiguous argument 'third_party/VENDOR-LICENSE/github.com/opencontainers/selinux/go-selinux/LICENSE': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Error: Please run ./hack/update-codegen.sh.

knative-sandbox/kn-plugin-func is out of date. Please run hack/update-codegen.sh
Error: Process completed with exit code 1.
```
I have run that command and it generated no changes to the files. I tried deleting the offending file and re-running to no avail.

lance:
/override style / Lint
/override verify / Verify Deps and Codegen

Dependencies should be okay per: https://knative.slack.com/archives/CCSNR4FCH/p1660681106952929
Lint is unavoidable given how the `indent` function works in templates substitution

lance:
/override verify / Verify Deps and Codegen (1.18.x, ubuntu-latest)

lance:
I don't understand why the unit tests are failing. It looks like it could be a timing issue in `TestClient_BuildDetects` with the failure message
```
    client_test.go:1408: client did not detect file timestamp change as indicating build stalenes
```
@lkingland any thoughts here? Nothing in the code has been changed with this PR, but I have not had a passing test run even after several attempts. This seems to happen on Windows only, but I'm not 100% certain of that.

lkingland:
> I don't understand why the unit tests are failing. It looks like it could be a timing issue in `TestClient_BuildDetects` with the failure message
> 
> ```
>     client_test.go:1408: client did not detect file timestamp change as indicating build stalenes
> ```
> 
> @lkingland any thoughts here? Nothing in the code has been changed with this PR, but I have not had a passing test run even after several attempts. This seems to happen on Windows only, but I'm not 100% certain of that.

The tests are working for me right now (see test PR #1175), so I can't reproduce.  It may be a timing issue like you say, and is not rearing its ugly head for me because I am running the CI tasks after prime time.  If your PR still gives you issues on the
next attempt, I would guess that the problem is the following lines beginning client_test.go:1399:
```go
  if !client.Built(root) {
    t.Fatal("freshly built function reported Built==false (1)")
  }

  time.Sleep(1 * time.Second) // Wait for exceptionally slow CI systems
  if err := os.Chtimes(filepath.Join(root, "func.yaml"), time.Now(), time.Now()); err != nil {
    fmt.Println(err)
  }
  time.Sleep(1 * time.Second) // Wait for exceptionally slow CI systems

  if client.Built(root) {
    t.Fatal("client did not detect file timestamp change as indicating build staleness")
  }
```
The first call to `client.Built` calculates the hash based off the current system time.
The next instruction is to change the file modificaiton times of `func.yaml`, which is intended to simulate a simple fileystem modification (essentially emulating `touch func.yaml`)
The second call to `client.Built` after the `touch` recalculates the filesystem hash based off the new times.

Perhaps it is possible that the CI environment is optimized in such a way that the timestamp update using `time.Now` does not result in a registered change to the filesystem (no `flush`?  some other magic?)

This hypothesis could be checked by adding a manual wait, which not only ensures the clock has advanced, but also that the active thread has been released and (presumably) a filesystem flush...

```go
  // Prior to a filesystem edit, it will be Built.
  if !client.Built(root) {
    t.Fatal("freshly built function reported Built==false (1)")
  }

  // Release thread and wait to ensure that the clock advances even in constrained CI environments 
  time.Sleep(100 * time.Millisecond) 

  // Edit the filesystem by touching a file (updating modified timestamp)
  if err := os.Chtimes(filepath.Join(root, "func.yaml"), time.Now(), time.Now()); err != nil {
    fmt.Println(err)
  }
 
  // Release the thread and ensure the clock advances even in constrained CI environments
  time.Sleep(1 * time.Millisecond)

  if client.Built(root) {
    t.Fatal("client did not detect file timestamp change as indicating build staleness")
  }
```

If that does indeed reliably fix the issue, then this test is more brittle than I would like, and we could refactor it slightly
by instead creating (and closing) an empty file before checking that `!client.Built`.

lance:
/override verify / Verify Deps and Codegen (1.18.x, ubuntu-latest)

lance:
/override verify / Verify Deps and Codegen (1.18.x, ubuntu-latest)

docs/reference/func_build.md
```diff
@@ -0,0 +1,68 @@
...
+### Options
+
+```
+  -b, --builder string         build strategy to use when creating the underlying image. Currently supported build strategies are "pack" or "s2i". (default "pack")
+      --builder-image string   builder image, either an as a an image name or a mapping name.
+                               Specified value is stored in func.yaml (as 'builder' field) for subsequent builds. ($FUNC_BUILDER_IMAGE)
+  -c, --confirm                Prompt to confirm all configuration options (Env: $FUNC_CONFIRM)
+  -h, --help                   help for build
+  -i, --image string           Full image name in the form [registry]/[namespace]/[name]:[tag] (optional). This option takes precedence over --registry (Env: $FUNC_IMAGE)
+  -p, --path string            Path to the project directory (Env: $FUNC_PATH) (default "/Users/lball/src/github.com/knative-sandbox/kn-plugin-func")
```
matejvasek, line 52: @lance the default --path seems to be odd.
This happens in multiple files.

matejvasek: This is probably not related to this PR.

matejvasek: that's in cmd/root.go

matejvasek: maybe we should replace `cwd()` with `""` and mention that current dir is used if --path is empty.

matejvasek: `cmd.Flags().StringP("path", "p", "", "Path to the project directory if empty current dir is used (Env: $FUNC_PATH)")`

lance: I didn't even notice that before - good catch. That's happening here: https://github.com/knative-sandbox/kn-plugin-func/blob/3ac5d468bfdbb326375ebc5bc640729b59fec7f2/cmd/root.go#L336-L339

I wonder if we can determine runtime context and instead of `cwd()` when generating documentation, use something like `"$CWD"` or `"<current working directory>"`

matejvasek:
@lance you probably need to put:|
```go
if pathFlag == "" {
    pathFlag = cwd()
}
```

matejvasek:
actually maybe at multiple places :disappointed:

lance:
> @lance you probably need to put:|
> 
> ```go
> if pathFlag == "" {
>     pathFlag = cwd()
> }
> ```

Actually, it's not so simple, and I think going with an env var would be more straightforward. Setting the default to `""` requires all commands to check for empty string and use `cwd()` if necessary. That adds code changes to a lot of files for this PR, and I'm not sure the motivation is really a good one for something that will touch so much.

docs/reference/func_build.md
lkingland: > maybe we should replace `cwd()` with `""` and mention that current dir is used if --path is empty.

I prefer this method if possible as well

lkingland:
If we leave empty string as meaning "no path defined", but allow for the special value `"."` to mean "current working directory", then we only need to convert `.` to `cwd()` wherever we create a command config struct.

That is indeed a little tricky, but I think it could be quick.

lance:
> > ... adds code changes to a lot of files for this PR ..
> 
> Agreed, see PR #1184 which should fix the path flag default and associated generated commands.txt. If that looks good we can merge and rebase _this_ PR, which may fix those failing tests.

Holding this PR until that lands and I can rebase.

lance:
@lkingland @matejvasek #1184 does a nice job of resolving the issues with `--path`. PTAL

lance:
> @lkingland @matejvasek #1184 does a nice job of resolving the issues with `--path`. PTAL

Hmm - seems I spoke too soon

lance:
Should be gtg now.

matejvasek:
@lance  what about https://github.com/knative-sandbox/kn-plugin-func/runs/7999903265?check_suite_focus=true

lance:
> @lance what about https://github.com/knative-sandbox/kn-plugin-func/runs/7999903265?check_suite_focus=true

I'm OK with an override on that wdyt?

matejvasek:
>I'm OK with an override on that wdyt?

What I am afraid of is that the error will pop up in subsequent PR and contributors won't have any idea what's wrong.

matejvasek:
We will have broken `main`.

lance:
> We will have broken `main`.

What do you mean?

matejvasek:
>What do you mean?

I mean that tests on main (default) branch wouldn't be passing, possibly breaking release process.
And any PR derived from it would fail PR test even if the PR is OK.

lance:
> > What do you mean?
> 
> I mean that tests on main (default) branch wouldn't be passing, possibly breaking release process. And any PR derived from it would fail PR test even if the PR is OK.

Hmm - I may be mistaken, but I thought linting only checked changed files. https://github.com/knative/actions/blob/72f25a1bb84367de79bdc41777a96f42f405a32a/.github/workflows/style.yaml#L182-L187

If that's the case, I think that this will only come up again when help text in the `create` command is modified.

matejvasek:
> > > What do you mean?
> > 
> > 
> > I mean that tests on main (default) branch wouldn't be passing, possibly breaking release process. And any PR derived from it would fail PR test even if the PR is OK.
> 
> Hmm - I may be mistaken, but I thought linting only checked changed files. https://github.com/knative/actions/blob/72f25a1bb84367de79bdc41777a96f42f405a32a/.github/workflows/style.yaml#L182-L187
> 
> If that's the case, I think that this will only come up again when help text in the `create` command is modified.

I am not sure but I am afraid that the check is run against whole code-base not just diff.

Anyway, let's give it a try and merge it. You will crate another PR if needs be.

matejvasek:
@lance you need rebase.

lance:
> I am not sure but I am afraid that the check is run against whole code-base not just diff.

The `-filter-mode="added"` flag in the reviewdog task is [documented](https://github.com/reviewdog/reviewdog#added-default) as "Filter results by added/modified lines". So I think we should be good.

PR #1170, opened 2022-08-16
chore: Use JS for script instead of shell

matejvasek:
# Changes

* :broom:  Use JS for script instead of shell for Quarkus update.
* :gift: Now querying GH API whether the PR already exists.


zroubalik (commented):
TBH I am not a big fan of adding dependency to node in our setup, I know this serves a specific usecase, but still.

hack/update-quarkus-platform.js
```diff
@@ -0,0 +1,124 @@
...
+
+const getLatestPlatform = async () => {
+    const resp = await axios.get("https://code.quarkus.io/api/platforms")
+    return resp.data.platforms[0].streams[0].releases[0].quarkusCoreVersion
+}
+
+const prExists = async (pred) => {
+
+    let page = 1
+    const perPage = 10;
```
lance, line 20: Nit: Why such a small number of PRs per page? If you changed this to 50 wouldn't that potentially reduce the potential number of API calls required?

lance (commented):
/lgtm

lance:
> TBH I am not a big fan of adding dependency to node in our setup, I know this serves a specific usecase, but still.

Added a `hold` so we can discuss - don't want to just ignore your concern. Honestly, this was my first thought when I saw the PR. But on further consideration it seems OK to me. This is only part of a github action and does not affect the func developer's local experience or requirements at all. And in fact, when we were using release-please, we had the same Node.js dependency doing similar things (creating/checking pull requests based on some given condition). So, IMO this is fine.

hack/update-quarkus-platform.js
matejvasek: I can decide. Larger page takes longer to get, and I assume the pr will be it top 10 PRs most of the time.

matejvasek:
> TBH I am not a big fan of adding dependency to node in our setup, I know this serves a specific usecase, but still.

@zroubalik @lance 
I don't like additional deps too, but I really dislike `bash` for anyting longer that 10 lines.
Also sometimes even `bash`, `xmllint` or `jq` are not installed (or not installed in latest version, like on mac).

I've chosen JS since it has official GitHub API lib and there is at least one more person who knows JS.

Actually I did first iteration of [this in Clojure](https://github.com/matejvasek/faas/commit/d5337874bf048bb6fe1c942d7b74cfc81c3ce357) (and just latter rewritten it to bash).
It requires only one additional binary installable via curl.
But nobody knows the lang.

zroubalik:
/unhold

okay, I am not gonna block this. Still better than the Closure implementation 😄

PR #1171, opened 2022-08-17
quarkus test template change to remove package type property

jrangelramos:
# Changes

:broom: Removing package.type property from quarkus template used on e2e test to align with default quarkus template.

/kind cleanup


lance (commented):
/lgtm

PR #1180, opened 2022-08-22
builder short name constants etc throughout

lkingland:
- :broom: Full adoption of builder short name constants throughout.

Pulls out builder names, validators and error types into their own package.

Moves validators of builder names into `main`, which is closest to where knowledge of the correct set resides.

/kind chore


function_builder.go
```diff
@@ -1,33 +0,0 @@
-package function
```
lkingland, line 1: most of this file moved to `builders/builder.go`

function_builder.go
```diff
@@ -1,33 +0,0 @@
...
-
-func AllBuilders() []string {
-	return []string{BuilderPack, BuilderS2i}
-}
-
-// ErrInvalidBuilder indicates that the passed builder was invalid.
-type ErrInvalidBuilder error
-
-// ValidateBuilder validatest that the input Build type is valid for deploy command
-func ValidateBuilder(builder string) error {
```
lkingland, line 21: validator is now in `main`, as that is the only place that knows the final bounded set

buildpacks/builder.go
```diff
@@ -15,9 +15,13 @@ import (
 	"github.com/docker/docker/client"
 
 	fn "knative.dev/kn-plugin-func"
+	"knative.dev/kn-plugin-func/builders"
 	"knative.dev/kn-plugin-func/docker"
 )
 
+// DefaultName when no WithName option is provided to NewBuilder
+const DefaultName = "pack"
```
lance, line 23: Shouldn't this be using the constants?
```suggestion
const DefaultName = builders.Pack
```

buildpacks/builder.go
```diff
@@ -52,7 +57,7 @@ type Impl interface {
 
 // NewBuilder instantiates a Buildpack-based Builder
 func NewBuilder(options ...Option) *Builder {
-	b := &Builder{}
+	b := &Builder{name: DefaultName}
```
lance, line 60: Could this just use the constant and eliminate the `var` above?
```suggestion
	b := &Builder{name: builders.Pack}
```

cmd/build.go
```diff
@@ -92,6 +104,13 @@ func ValidNamespaceAndRegistry(path string) survey.Validator {
 	}
 }
 
+// KnownBuilders are a typed string slice of builder short names which this
+// CLI understands.  Includes a customized String() representation intended
+// for use in flags and help text.
+func KnownBuilders() builders.Known {
+	return builders.Known([]string{builders.Pack, builders.S2I})
```
lance, line 111: Shouldn't this use `builders.All()`?
```suggestion
	return builders.Known([]string{builders.Pack, builders.S2I})
```

s2i/builder.go
```diff
@@ -26,13 +26,12 @@ import (
...
+	"knative.dev/kn-plugin-func/builders"
 	"knative.dev/kn-plugin-func/docker"
 )
 
-var (
-	// ErrRuntimeRequired indicates the required value of function runtime was not provided
-	ErrRuntimeRequired = errors.New("runtime is required to build")
-)
+// DefaultName when no WithName option is provided to NewBuilder
+const DefaultName = "s2i"
```
lance, line 34: Same as in buildpacks builder. Use constants?
```suggestion
const DefaultName = builders.S2I
```

s2i/builder.go
```diff
@@ -87,7 +93,7 @@ func WithPlatform(platform string) Option {
 
 // NewBuilder creates a new instance of a Builder with static defaults.
 func NewBuilder(options ...Option) *Builder {
-	b := &Builder{}
+	b := &Builder{name: DefaultName}
```
lance, line 96: As with buildpack builder, the above `var DefaultName` could be removed, I believe.
```suggestion
	b := &Builder{name: builders.S2I}
```

lance (commented):
Looks good overall. Just a couple of questions about constant usage in 2 places.

buildpacks/builder.go
lkingland: Yes, that of course works well, but this way has some benefits for godoc.  The exported "DefaultName" is highlighted quite prominently in godocs, whereas if I directly reference the variable the auto-entry in the docs is skipped.  Also, it's been

buildpacks/builder.go
lkingland: Yep!

s2i/builder.go
lkingland: Yes, I can do that.  The implicit idea was that "this is the constant that will be used if the client is not using the `buildpacks` package to assign short names.  But that is probably just as clear using the direct constant reference.

cmd/build.go
lkingland: Yes, we could, but that isn't _exactly_ reflecting the intention and may lead to confusion later.  The point is a bit nuanced but worth perhaps taking a swing at: I was trying to emphasize that this function is where the exact set of builders are reported _which the CLI supports_ (can create concrete implementations for when used).  The `builders.All` returns the set of _known_ builder short names globally for map references, etc.  These are actually not the same set.

While it's true that in our func CLI, those two sets are currently equivalent, I feel like writing it in this way emphasizes the point that that they are not:  a user of the Client package may have less, or more, that that "known builders".  This is also why the builder validator is not in `builders`.  By skipping straight to `return builders.Known(builders.All())`, I felt that ran the risk of leading a reader to incorrectly presume that the set of known builder _names_ and the set of builders supported by a user (the cli in this case) of the client library are necessarily equivalent.

This may be too pedantic, and does look like an oversight.  Still think I should change after reading the justification?

lkingland:
Thanks @lace for the review.  All comments should now have been processed

lance:
@lkingland needs rebase - otherwise lgtm

lance (commented):
/lgtm

PR #1181, opened 2022-08-22
fix: in cluster dialer security context is non-privileged

matejvasek:
# Changes

- :bug: In-cluster-dialer pod security context is non-privileged now.

### DIFF
--- a/k8s/dialer.go
+++ b/k8s/dialer.go
@@ -149,6 +149,7 @@ func (c *contextDialer) startDialerPod(ctx context.Context) (err error) {
 						AllowPrivilegeEscalation: new(bool),
 						RunAsNonRoot:             &runAsNonRoot,
 						Capabilities:             &coreV1.Capabilities{Drop: []coreV1.Capability{"ALL"}},
+						SeccompProfile:           &coreV1.SeccompProfile{Type: coreV1.SeccompProfileTypeRuntimeDefault},
 					},
 				},
 			},
--- a/k8s/dialer.go
+++ b/k8s/dialer.go
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	socatImage = "quay.io/boson/alpine-socat:1.7.4.3-r"
+	socatImage = "quay.io/boson/alpine-socat:1.7.4.3-r1-non-root"
 )
 
 // NewInClusterDialer creates context dialer that will dial TCP connections via POD running in k8s cluster.
@@ -129,6 +129,7 @@ func (c *contextDialer) startDialerPod(ctx context.Context) (err error) {
 		}
 	}()
 
+	runAsNonRoot := true
 	pod := &coreV1.Pod{
 		ObjectMeta: metaV1.ObjectMeta{
 			Name:        c.podName,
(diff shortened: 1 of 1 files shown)


lance (commented):
Nice - assuming it works with @jrangelramos tests this lgtm

matejvasek:
@jrangelramos please try again

matejvasek:
I find odd I have to set up it this way. Is there any factory function that would present sane defaults for me?

cc @rhuss

lkingland (approved):
/lgtm

k8s/dialer.go
```diff
@@ -102,6 +102,7 @@ func (c *contextDialer) startDialerPod(ctx context.Context) (err error) {
 	if err != nil {
 		return
 	}
+	c.restConf.WarningHandler = restclient.NoWarnings{}
```
rhuss, line 105: why is this needed ?

matejvasek: that's not exactly needed, it should be rather used depending on `--verbose -v` flags
so user is not scared in non verbose mode

matejvasek:
@rhuss is seems that my issue was related to `Sandbox` specific settings not 4.10 vs 4.11.

If I don't set the seccomp profile the there is warning on 4.11,
but if I set the seccomp profile then it is not working in `Sandbox` cluster.

k8s/dialer.go
matejvasek: I just don't want to scare users.

k8s/dialer.go
```diff
@@ -143,6 +145,12 @@ func (c *contextDialer) startDialerPod(ctx context.Context) (err error) {
 					Stdin:     true,
 					StdinOnce: true,
 					Args:      []string{"-u", "-", "OPEN:/dev/null,append"},
+					SecurityContext: &coreV1.SecurityContext{
+						Privileged:               new(bool),
+						AllowPrivilegeEscalation: new(bool),
```
rhuss, line 150: why do you need to set both to `false` ? Isn't this the default value anyway ?

k8s/dialer.go
rhuss: I see, fine for me.

k8s/dialer.go
matejvasek: by default it's null, IIRC it caused troubles

matejvasek:
@rhuss 
In the end I can decide between:
* No warnings on 4.11 but totally broken `Sandbox`.
* Single warning on 4.11 but working `Sandbox`.

k8s/dialer.go
matejvasek: It should be explicitly ptr to false.

matejvasek:
By `Sandbox` I mean OpenShift evaluation instances.

k8s/dialer.go
matejvasek: I would too assumed it should work with zero/null value, but it doesn't.
It also applies to `RunAsNonRoot` despite the fact the user of the image is not `root`.

PR #1184, opened 2022-08-23
chore: use dot as path default rather than absolute path

lkingland:
:broom: use `.` as default path

Flags and generated help text show `"."` as default for `--path`.  The value is converted into the absolute path when a config command struct is generated from the flags.

This is related to #1169

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -335,7 +335,7 @@ func mergeEnvs(envs []fn.Env, envToUpdate *util.OrderedMap, envToRemove []string
 
 // setPathFlag ensures common text/wording when the --path flag is used
 func setPathFlag(cmd *cobra.Command) {
-	cmd.Flags().StringP("path", "p", cwd(), "Path to the project directory (Env: $FUNC_PATH)")
+	cmd.Flags().StringP("path", "p", ".", "Path to the project directory (Env: $FUNC_PATH)")
 }
 
 type Version struct {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -144,16 +144,6 @@ func interactiveTerminal() bool {
 	return err == nil && ((fi.Mode() & os.ModeCharDevice) != 0)
 }
 
-// cwd returns the current working directory or exits 1 printing the error.
-func cwd() (cwd string) {
-	cwd, err := os.Getwd()
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "Unable to determine current working directory: %v", err)
-		os.Exit(1)
-	}
-	return cwd
-}
-
 // bindFunc which conforms to the cobra PreRunE method signature
 type bindFunc func(*cobra.Command, []string) error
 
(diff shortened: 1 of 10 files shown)


matejvasek:
I think there are 4 more places where `getPathFlag()` should be called instead of `viper.GetString("path")`.

matejvasek:
```
cmd/config.go:156:		Name: deriveName(name, viper.GetString("path")),
cmd/delete.go:116:		Name:      deriveName(name, viper.GetString("path")), // args[0] or derived
cmd/info.go:99:		Name:      deriveName(name, viper.GetString("path")),
cmd/run.go:178:		Path:        viper.GetString("path"),
```

cmd/root.go
```diff
@@ -335,7 +325,28 @@ func mergeEnvs(envs []fn.Env, envToUpdate *util.OrderedMap, envToRemove []string
...
+		path = cwd()
+	}
+	return path
+}
+
+// cwd returns the current working directory or exits 1 printing the error.
+func cwd() (cwd string) {
+	cwd, err := os.Getwd()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Unable to determine current working directory: %v", err)
```
matejvasek, line 346: Wouldn't be `panic()` better?

cmd/root.go
```diff
@@ -335,7 +325,28 @@ func mergeEnvs(envs []fn.Env, envToUpdate *util.OrderedMap, envToRemove []string
...
 // setPathFlag ensures common text/wording when the --path flag is used
 func setPathFlag(cmd *cobra.Command) {
-	cmd.Flags().StringP("path", "p", cwd(), "Path to the project directory (Env: $FUNC_PATH)")
+	cmd.Flags().StringP("path", "p", ".", "Path to the project directory (Env: $FUNC_PATH)")
+}
+
+// getPathFlag returns the value of the --path flag.
+// The special value '.' is returned as the abolute path to the current
+// working directory.
+func getPathFlag() string {
```
matejvasek, line 334: Maybe we don't need this function at all.
The `.` path is already considered to be `cwd`.
I see some value in pinning absolute path if `cwd` is changed during execution but `func` is not changing `cwd`.

cmd/root.go
matejvasek: This is not just like my preference.
With `os.Exit()` the process dies instantly and no `defer`s are not invoked.

matejvasek: `os.Exit()` should be only called from `main()`

cmd/root.go
matejvasek: Only tests are changing `cwd`.

lance: > Maybe we don't need this function at all.

I see your point, but do like using `os.Getwd()` to resolve it

lance:
> ```
> cmd/config.go:156:		Name: deriveName(name, viper.GetString("path")),
> cmd/delete.go:116:		Name:      deriveName(name, viper.GetString("path")), // args[0] or derived
> cmd/info.go:99:		Name:      deriveName(name, viper.GetString("path")),
> cmd/run.go:178:		Path:        viper.GetString("path"),
> ```

Good catch @matejvasek

cmd/root.go
lkingland: I am open to that, yes.  This is currently set as nonfatal because I am not convinced being unable to retrieve the current working directory is a problem that is fatal.  As you mentioned the default is to presume CWD.

If this inability to determine current working directory is a fatal error, then we could panic, yes, or change this to be a standard error-returning function which would fal upwards to the root-level catch which exits cleanly with (potentially) a custom numerical code for this particular exit condition.

cmd/root.go
lkingland: Yes, but we may avoid challenges in the future if the core executes without dependency on the parent process being in given working directory.  By converting the `.` to an absolute path in the CLI, we are at least treating it that way

lkingland:
Thanks @lance and @matejvasek .  I have made those changes, running a search for any remaining instances of either `viper.GetString("path")` or `os.Exit()`.  That should be all of them

PR #1185, opened 2022-08-24
fix: image name derivation and flag precidence

### LINKED ISSUE #1125: func build with -i does not use given image name
Using `func build -i docker.io/gauron/app:latest` doesnt create image named `docker.io/gauron/app:latest` but instead 
`image-registry.openshift-image-registry.svc:5000/default/app` (name gets overridden)

* **Reproduce**: 
`func create -l go`
`func build -i docker.io/name/app`

it looks to me like in `Build` function (client.go) the registry is populated with the openshift default registry and used in `DerivedImage` (look at image below)

image
`f.Registry` is empty but `registry` is the default openshift registry right before the if statement (as indicated by the LOOK DEEP print) therefore the if is skipped and image is built without image flag

/kind bug

lkingland:
:broom: clean up image name derivation logic
:bug: image flag now correctly respected
:gift: tests in commands and core codifying `registry` vs `image` treatement
:gift: prompts on deploy should now correctly calculate image default
:broom: clean up namespace derivation logic and add a test

Clean up logic used to derive and update image name based on registry, function name etc.  
Adds requisite tests to both `deploy` and `build`.

/kind cleanup


cmd/deploy.go
```diff
@@ -562,20 +538,39 @@ func parseImageDigest(imageSplit []string, config deployConfig, cmd *cobra.Comma
...
+	if confNamespace != "" {
+		namespace = confNamespace // --namespace takes precidence
+	} else if funcNamespace != "" {
+		namespace = funcNamespace // value from previous deployment (func.yaml) 2nd priority
+	} else {
+		// Try to derive a default from the current k8s context, if any.
+		namespace, err = k8s.GetNamespace("")
+		if err != nil {
+			return "", err
+		}
```
lkingland, line 554: We may want to treat this as nonfatal (perhaps with a warning printed to stderr?)

client_test.go
```diff
@@ -385,10 +384,16 @@ func TestClient_New_ImageNameDerived(t *testing.T) {
...
 
-	// In form: [Default Registry]/[Registry Namespace]/[Service Name]:latest
-	expected := TestRegistry + "/" + f.Name + ":latest"
-	if f.Image != expected {
-		t.Fatalf("expected image '%v' got '%v'", expected, f.Image)
+	// This opaque-box unit test ensures NewImageTag is invoked and applied.
+	// See the Test_NewImageTag clear-box unit test for an in-depth exploration of
+	// how the values of image and registry are treated to create, by default:
+	//   [Default Registry]/[Registry Namespace]/[Service Name]:latest
+	imageTag, err := f.ImageName()
```
lkingland, line 391: This would probably better be `f.Image()` with logic for derivationd inbuilt, but that's a larger proposition because it affects the struct's serialization.

function.go
```diff
@@ -433,6 +366,50 @@ func assertEmptyRoot(path string) (err error) {
...
+// form:    [registry]/[user]/[function]:latest
+// example: quay.io/alice/my.function.name:latest
+//
+// Also nested namespaces should be supported:
+// form:    [registry]/[parent]/[user]/[function]:latest
+// example: quay.io/project/alice/my.function.name:latest
+//
+// Registry values which only contain a single token are presumed to
+// indicate the namespace at the default registry.
+func (f Function) ImageName() (image string, err error) {
```
lkingland, line 383: This would probably better be `f.Image()` with logic for derivation inbuilt, but that's a larger proposition since it affects the struct's serialization.

cmd/build.go
```diff
@@ -278,12 +215,11 @@ func newBuildConfig() buildConfig {
 // Skipped if not in an interactive terminal (non-TTY), or if --confirm false (agree to
 // all prompts) was set (default).
 func (c buildConfig) Prompt() (buildConfig, error) {
-	imageName := deriveImage(c.Image, c.Registry, c.Path)
 	if !interactiveTerminal() || !c.Confirm {
 		return c, nil
 	}
 
-	bc := buildConfig{Verbose: c.Verbose}
+	imageName := deriveImage(c.Image, c.Registry, c.Path)
```
lkingland, line 222: Two things happen here:  The current state of the config is allowed to flow through as the defaults to the prompts, and we move down the line which is calculating the default image name.

cmd/build.go
```diff
@@ -292,18 +228,35 @@ func (c buildConfig) Prompt() (buildConfig, error) {
 				Message: "Project path:",
 				Default: c.Path,
 			},
-			Validate: survey.Required,
```
lkingland, line 295: Best to concentrate validation logic in one place as much as possible, and the client API validates and issues the final error (if any)

cmd/build_test.go
```diff
@@ -223,119 +181,3 @@ created: 2009-11-10 23:00:00`,
 		})
 	}
 }
-
-func testBuilderPersistence(t *testing.T, testRegistry string, cmdBuilder func(ClientFactory) *cobra.Command) {
```
lkingland, line 227: The implementations of the shared tests are moved into deploy_test.go since it is the superset of deploy+build

cmd/build_test.go
```diff
@@ -50,79 +46,41 @@ func TestBuild_ImageFlag(t *testing.T) {
...
-		t.Fatal("Expected error")
-	}
+// TestBuild_ImageAndRegistry
+func TestBuild_ImageAndRegistry(t *testing.T) {
+	testRegistryOrImageRequired(NewBuildCmd, t)
 }
 
-// TestBuild_registryConfigurationInYaml tests that a build will execute successfully
-// when there is no registry provided on the command line, but one exists in func.yaml
-func TestBuild_registryConfigurationInYaml(t *testing.T) {
```
lkingland, line 96: Most of the tests are just moved/shared.  This was more heavily refactored to use the client API rather than a func.yaml as

cmd/deploy.go
```diff
@@ -147,43 +178,21 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
...
-				if err == terminal.InterruptErr {
-					return nil
-				}
-				return
-			}
-		}
-
-		// We have the registry, so let's use it to derive the function image name
-		config.Image = deriveImage(config.Image, config.Registry, config.Path)
-		function.Image = config.Image
```
lkingland, line 178: This prompt block is part of the flow initiated by `--confirm` (`config.Prompt`), with normal execution mode usually expected to result in an immediate non-zero exit code and error message printed to stderr.

zroubalik (approved):
/lgtm great improvement :)

/hold for others

lkingland:
Since parent PR (#1079) already got a few approvals when these changes were originally included therein, I am presuming this derivitave is GTG with the single approval

PR #1188, opened 2022-08-24
Improve `func config envs` [WIP]

matejvasek:
* Added ability to add env non-interactively
* Added ability to list envs as JSON

# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


matejvasek:

lance (commented):
Looks good overall. Thanks! Can you add an example to `ConfigEnvsAddCmd()` help text so that we the `configMap` syntax is shown? Also, I don't see a test for `secret` - is that also possible? Perhaps a test and doc for that as well?

cmd/config_envs.go
```diff
@@ -25,20 +28,20 @@ Prints configured Environment variable for a function project present in
...
 
-			listEnvs(function)
-
-			return
+			return listEnvs(function, cmd.OutOrStdout(), Format(viper.GetString("output")))
 		},
 	}
 
-	configEnvsAddCmd := NewConfigEnvsAddCmd()
+	cmd.Flags().StringP("output", "o", "human", "Output format (human|json) (Env: $FUNC_OUTPUT)")
```
lkingland, line 42: In the other commands I decided to simplify this to just a single flag, `--json`.  I know that means it may be tricky in the future if we add several other output formats, but it sure seems to make the CLI experience nicer now:  For example
```
func templates --json
```
I found that to offer a better experience.  I therefore prefer it that way, personally, but either way we choose, we should probably be consistent.  Thoughts?

lkingland (approved):
Looks good to me as well.  Lance brings up some good points.  My only question was about changing to use the simplified flag `--json` rather than the key/value `output=[x]` approach we started with some time ago.

/lgtm
/hold for Lance's suggestions

matejvasek:
> Also, I don't see a test for secret - is that also possible? Perhaps a test and doc for that as well?

Actually even test for configMap is overkill here, everything is handled as string here (no special cases for configMap or secrets). Magic happens in `deploy` phase.

lance:
> > Also, I don't see a test for secret - is that also possible? Perhaps a test and doc for that as well?
> 
> @lance Actually even test for configMap is overkill here, everything is handled as string here (no special cases for configMap or secrets). Magic happens in `deploy` phase.

Fair enough - but how about an example in `func config envs add` help text?

cmd/config_envs.go
zroubalik: +1 to `-o`

matejvasek:
@lance PTAL

matejvasek:
@lance also I need override `/override style / Lint`

matejvasek:
@lance the override shouldn't be needed now after docgen update.

cmd/config_envs.go
matejvasek: I will use `-o`.

`-ojson` is the same length as `--json`.

PR #1200, opened 2022-08-30
fix: on cluster builds when buildEnvs contains only one build environment

### LINKED ISSUE #1199: On Cluster builds fails when buildEnvs has one environment var
When the project `func.yaml` has one buildEnv set, and build is git (on cluster build), deploy fails with the below error:

```
PipelineRun default/node-fn-git-pack-pipeline-run-jnxk4 parameters have mismatching types with Pipeline default/node-fn-git-pack-pipeline's parameters: parameters have inconsistent types : [buildEnvs]
```

jrangelramos:
# Changes

- :bug: Fix on cluster build, deployment issues when buildEnv has only one env var, example:

/kind bug

Fixes #1199

### DIFF
--- a/pipelines/tekton/resources.go
+++ b/pipelines/tekton/resources.go
@@ -133,7 +133,7 @@ func generatePipelineRun(f fn.Function, labels map[string]string) *pplnv1beta1.P
 		for _, e := range f.BuildEnvs {
 			envs = append(envs, e.KeyValuePair())
 		}
-		buildEnvs = pplnv1beta1.NewArrayOrString(envs[0], envs[1:]...)
+		buildEnvs.ArrayVal = envs
 	}
 
 	params := []pplnv1beta1.Param{


pipelines/tekton/resources.go
```diff
@@ -133,7 +133,11 @@ func generatePipelineRun(f fn.Function, labels map[string]string) *pplnv1beta1.P
 		for _, e := range f.BuildEnvs {
 			envs = append(envs, e.KeyValuePair())
 		}
-		buildEnvs = pplnv1beta1.NewArrayOrString(envs[0], envs[1:]...)
+		if len(envs) == 1 {
+			(*buildEnvs).ArrayVal = []string{envs[0]}
+		} else {
+			buildEnvs = pplnv1beta1.NewArrayOrString(envs[0], envs[1:]...)
+		}
```
matejvasek, line 140: Couldn't be all just `buildEnvs.ArrayVal = envs`?

matejvasek: @jrangelramos  I mean all four line could be just one (no `if` needed).

PR #1201, opened 2022-08-30
chore: update rust dependencies

### LINKED ISSUE #1111: Update rust function dependencies
Dependabot has identified a few issues with the current rust dependencies. These should be updated.

##  regex

Advisory https://groups.google.com/g/rustlang-security-announcements/c/NcNNL1Jq7Yw

Upgrade regex to version 1.5.5 or later. For example:

```
[dependencies]
regex = "1.5.5"
```
```
[dev-dependencies]
regex = "1.5.5"
```

## tokio

Affected versions: < 1.8.4
Patched version: 1.8.4

An issue was discovered in the tokio crate before 1.8.4, and 1.9.x through 1.13.x before 1.13.1, for Rust. In certain circumstances involving a closed oneshot channel, there is a data race and memory corruption.

```
Upgrade tokio to version 1.8.4 or later. For example:

[dependencies]
tokio = "1.8.4"
```
```
[dev-dependencies]
tokio = "1.8.4"
```

## actix-http

Affected versions of this crate did not properly detect invalid requests that could allow HTTP/1 request smuggling (HRS) attacks when running alongside a vulnerable front-end proxy server. This can result in leaked internal and/or user data, including credentials, when the front-end proxy is also vulnerable.

Popular front-end proxies and load balancers already mitigate HRS attacks so it is recommended that they are also kept up to date; check your specific set up. You should upgrade even if the front-end proxy receives exclusively HTTP/2 traffic and connects to the back-end using HTTP/1; several downgrade attacks are known that can also expose HRS vulnerabilities.

Upgrade actix-http to version 2.2.1 or later. For example:

```
[dependencies]
actix-http = "2.2.1"
```
```
[dev-dependencies]
actix-http = "2.2.1"
```

/assign @vyasgun

vyasgun:
# Changes

- Updated rust dependencies

/kind enhancement

Fixes #1111 

**Release Note**

```release-note

```

**Docs**

```docs

```


matejvasek:
@vyasgun this requires also update of the zz_* file.

matejvasek:
@vyasgun looks good, but you should exclude the generated docs, I'll update it in separate PR.

vyasgun:
@matejvasek Thanks for the review. I updated the PR accordingly.

matejvasek:
@vyasgun thanks for your contribution.

PR #1202, opened 2022-08-30
fix: flags more consistent between build/deploy

matejvasek:
# Changes

- :bug: The short flag `-b` now means the same in `build` and `deploy`. 

/kind bug

```release-note
fix: the short `-b` flag now means the same thing for `build` and `deploy`
```


matejvasek:
On related note:
I suspect that the `--build` flag was probably added to disable auto-build for `deploy` and `run` sub-commands.
But it appears that now it also chooses between local and in-cluster build.
Is that correct?

lance:
> On related note: I suspect that the `--build` flag was probably added to disable auto-build for `deploy` and `run` sub-commands. But it appears that now it also chooses between local and in-cluster build. Is that correct?

Yes, that is correct.

/lgtm

PR #1204, opened 2022-08-30
fix: check for `--image` flag in build command

lance:
# Changes

- :bug: check for `--image` flag in build command

If the user provides an `--image` flag on the command line, it should be used. This commit modifies the image resolving code to check whether or not the image name was provided on the command line.

Fixes: https://github.com/knative-sandbox/kn-plugin-func/issues/1125

/kind bug

**Release Note**
```release-note
Fixes a regression where the `--image` flag on `func build` was not being used
```

### DIFF
--- a/cmd/build_test.go
+++ b/cmd/build_test.go
@@ -24,13 +24,7 @@ func TestBuild_ImageFlag(t *testing.T) {
 	root, cleanup := Mktemp(t)
 	defer cleanup()
 
-	// Write a func.yaml config which does not specify an image
-	funcYaml := `name: foo
-runtime: go
-builder: pack
-created: 2022-01-01T00:00:00+00:00
-`
-	if err := ioutil.WriteFile("func.yaml", []byte(funcYaml), 0600); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/build_test.go
```diff
@@ -14,6 +14,48 @@ import (
...
+
+	// Write a func.yaml config which does not specify an image
+	funcYaml := `name: foo
+runtime: go
+builder: pack
+created: 2022-01-01T00:00:00+00:00
+`
+	if err := ioutil.WriteFile("func.yaml", []byte(funcYaml), 0600); err != nil {
+		t.Fatal(err)
+	}
```
lkingland, line 35: Since `func.yaml` is tightly coupled with the core `fn` library, and this `cmd` library depends on that API, I think it may be more usefule to use that api here as well:
```
  if err := fn.New().Create(fn.Function{Runtime: "go", Root: root}); err != nil {
    t.Fatal(err)
  }
```
You are then more guranteed that this unit test remains forward-compatible with what constitutes a valid function for purposes of the test, even if that may entail changes to the initial state of `func.yaml`.

PR #1210, opened 2022-08-31
Adds ability to pull from private repos for git builds

### LINKED ISSUE #849: On-cluster build: Code from private Git repository
`git-clone` supports credentials, auth: https://hub.tekton.dev/tekton/task/git-clone

There should be a user friendly way how to add credentials that will be used by pipeline. Users shouldn't create additinonal resources themselves, all of it should be handled by `func` cli.

gab-satchi:
# Changes

- 🎁 Wires auth workspaces in plugins in order to pull from private repos

/kind enhancement

Fixes #849 

**Release Note**

```release-note
Private repositories can be used with git based on-cluster builds
```

/hold for UX and doc changes

The current UX:
- User creates an appropriate secret with either ssh  or basic-auth credentials as outlined [here](https://hub.tekton.dev/tekton/task/git-clone). Secret must be in the same namespace as the function.
- User adds the secret name under `sshSecret` or `basicAuthSecret` in `func.yaml`
- `func deploy`

There's likely a better UX that doesn't require the user to create the Secrets manually but I'd like more guidance on that. Maybe something under `func config`

### DIFF
--- a/function_git.go
+++ b/function_git.go
@@ -8,9 +8,11 @@ import (
 )
 
 type Git struct {
-	URL        *string `yaml:"url,omitempty"`
-	Revision   *string `yaml:"revision,omitempty"`
-	ContextDir *string `yaml:"contextDir,omitempty"`
+	URL             *string `yaml:"url,omitempty"`
+	Revision        *string `yaml:"revision,omitempty"`
+	ContextDir      *string `yaml:"contextDir,omitempty"`
+	SSHSecret       *string `yaml:"sshSecret,omitempty"`
+	BasicAuthSecret *string `yaml:"basicAuthSecret,omitempty"`
 }
 
 // validateGit validates input Git option from function config
--- a/pipelines/tekton/resources.go
+++ b/pipelines/tekton/resources.go
@@ -13,6 +13,7 @@ import (
 	"knative.dev/kn-plugin-func/builders"
 	"knative.dev/kn-plugin-func/buildpacks"
 	"knative.dev/kn-plugin-func/s2i"
+	"knative.dev/pkg/ptr"
 )
 
 func deletePipelines(ctx context.Context, namespaceOverride string, listOptions metav1.ListOptions) (err error) {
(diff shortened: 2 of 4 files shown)


function_git.go
```diff
@@ -8,9 +8,11 @@ import (
 )
 
 type Git struct {
-	URL        *string `yaml:"url,omitempty"`
-	Revision   *string `yaml:"revision,omitempty"`
-	ContextDir *string `yaml:"contextDir,omitempty"`
+	URL             *string `yaml:"url,omitempty"`
+	Revision        *string `yaml:"revision,omitempty"`
+	ContextDir      *string `yaml:"contextDir,omitempty"`
+	SSHSecret       *string `yaml:"sshSecret,omitempty"`
```
zroubalik, line 14: As mentioned on the issue, it might be actually better if we create secrets for users.

gab-satchi: It will definitely be better and the UX is where I need some help. 

The secret will need to have the private key required for the repo. But to scrape the user's `.ssh` directory and automatically copying keys feels wrong. 

So, instead of `sshSecret` is it sufficient to have `sshPrivateKey` where the user can specify the file path?

salaboy: @zroubalik @gab-satchi for what I understand this secret will contain the tokens for the git repo credentials right? 
If this is only supposed to work with on-cluster builds, right? If so, I would follow the same approach used by tekton where the user is responsible for creating a secret that will be used to connect to a container registry. if this is only relevant for on cluster build I consider this to be a platform configuration more than something that the developer should worry about.

gab-satchi: > for what I understand this secret will contain the tokens for the git repo credentials right?

That's right, with the current implementation, the user will need to either create an `sshSecret` with the private key contents or a `basicAuthSecret` for basic auth (not recommended for security reasons).

> if this is only relevant for on cluster build I consider this to be a platform configuration more than something that the developer should worry about.

One requirement is that the secret needs to be in the same namespace as the function. Would that still be considered a 'platform' configuration?

PR #1213, opened 2022-09-05
fix: on cluster build - populate image name if not set

zroubalik:

The recent changes in function image generation brought regression in on cluster build. If we tried to run a pipeline without image specified, it wasn't able to generate the correct image name.

ie. following `func.yaml` 
```yaml
specVersion: 0.25.0
name: test-function
namespace: default
runtime: node
registry: docker.io/zroubalik
image: ""
imageDigest: ""
build: git
...
```
fails with: 
```
🕐 Creating Pipeline resources
Error: failed to run pipeline: could not parse reference: 
```

This PR improves the error message as well.

# Changes

- :bug: on cluster build - populate image name if not set

/kind bug

### DIFF
--- a/client.go
+++ b/client.go
@@ -728,6 +728,7 @@ func (c *Client) Deploy(ctx context.Context, path string) (err error) {
 }
 
 // RunPipeline runs a Pipeline to Build and deploy the function at path.
+// In a parameter accepts git configuration options that we don't want to persist into func.yaml.
 func (c *Client) RunPipeline(ctx context.Context, path string, git Git) (err error) {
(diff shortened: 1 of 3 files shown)


lkingland (approved):
Looks great!  I definitely missed that the Image population needs to be done separately for remote (pipeline) build/deploys.  I wonder why the E2E didn't catch that 🤔 
/lgtm

jrangelramos:
Hi @lkingland I think it is because the new e2e for oncluster build got merged after the changes of image population. It is pretty recent.

PR #1216, opened 2022-09-05
feat: replace socat with custom dialer

matejvasek:
# Changes

- :gift: The custom dialer is smaller and does everything we need.


jrangelramos:
hi @matejvasek, once #1213 the errors on e2e oncluster tests shall be gone.

cmd/dial/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM registry.redhat.io/rhel8/go-toolset as builder
+
+COPY main.go /opt/app-root/src/
+
+RUN CGO_ENABLED=0 go build -o dial -ldflags "-s -w" main.go
+
+FROM scratch
+
+COPY --from=builder --chown=1001:1001 /opt/app-root/src/dial /dial
```
zroubalik, line 9: Don't you want to use ubi-mini/micro in here?

matejvasek: @zroubalik `scratch` should be even better I guess, there is less attack surface. Also I asked guy from productization and he said it's OK.

matejvasek: Also for what we need I consider even micro obscenely big (not to mention minimal).

I could even achieve 100kB image with pure C instead of Go but `musl C lib` is not parrt of RHEL :disappointed: 

I've chosen Go since productization is easy with it.

matejvasek: without `musl` I cannot create static binary with C

zroubalik: ACK, I thought that `scratch` will be a productization problem. So it is fine and please disregard this :)

zroubalik (commented):
Would be nice if you can also publish an image automatically in the CI in case there are any chnages in `cmd/dial`.
Something like:
https://github.com/knative-sandbox/kn-plugin-func/blob/4041d609dd47adfeedf3e72c4aa32fd50d87d520/.github/workflows/ci.yaml#L69

matejvasek:
After discussion with productiaztion, I might rather move this to separate repo.

zroubalik:
> After discussion with productiaztion, I might rather move this to separate repo.

So, the code will be here and image in another repo? Shouldn't both live at one place?

matejvasek:
I've made the dialer more `socat`-like.

matejvasek:
@lance I pushed the image to ghcr.io but I cannot make it public.
https://github.com/orgs/knative-sandbox/packages/container/package/kn-plugin-func%2Fsocat

matejvasek:
Maybe we could use `ko` to build an image?

matejvasek:
@lance @zroubalik @lkingland PTAL

cmd/socat/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/matejvasek/socat
```
lkingland, line 1: Perhaps should be `github.com/knative-sandbox/kn-plugin-func/cmd/socat`

k8s/dialer.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	socatImage = "quay.io/boson/alpine-socat:1.7.4.3-r1-non-root"
+	socatImage = "quay.io/boson/socat:micro"
```
lkingland, line 26: Might be nice to get a version number in here somewhere.

cmd/socat/main.go
```diff
@@ -0,0 +1,143 @@
...
+	err := cmd.Execute()
+	if err != nil {
+		os.Exit(1)
+	}
+}
+
+func newRootCmd() *cobra.Command {
+	var uniDir bool
+	cmd := cobra.Command{
+		Use:   "socat <address> <address>",
```
lkingland, line 24: Might be worth calling this tcpcat to avoid confusion with the actual socat

cmd/socat/main.go
```diff
@@ -0,0 +1,143 @@
...
+		Short: "Minimalistic socat.",
+		Long: `Minimalistic socat.
+Implements only TCP, OPEN and stdio ("-") addresses with no options.
+Only supported flag is -u.`,
```
lkingland, line 28: Might want to just say what this dialer does, connects two tcp address, either bi or unidirectionally

cmd/socat/main.go
```diff
@@ -0,0 +1,143 @@
...
+	w io.WriteCloser
+}
+
+func (r rwc) Read(p []byte) (n int, err error) {
+	return r.r.Read(p)
+}
+
+func (r rwc) Write(p []byte) (n int, err error) {
+	return r.w.Write(p)
+}
```
lkingland, line 119: These methods which just map reading and writing onto their sub-structs would go away if you use struct embedding for the ReadCloser and WriteCloser on the rwc struct.

Since that removes the named atrributes `r.r` and `r.w`, there would be an ambiguous `.Close`, which should be referenceable using the type instead:  `r.ReadCloser.Close()`

cmd/socat/main.go
```diff
@@ -0,0 +1,143 @@
...
+		var laddr net.TCPAddr
+		raddr, err := net.ResolveTCPAddr(typ, addr)
+		if err != nil {
+			return nil, fmt.Errorf("name does not resolve: %w", err)
+		}
+		return net.DialTCP(typ, &laddr, raddr)
+	case "open":
+		return os.OpenFile(addr, os.O_RDWR, 0644)
+	}
+	return nil, fmt.Errorf("unsupported address: %q", address)
```
lkingland, line 73: Might be nice to stick this in the `default` case of the switch statement and say `unsupported protocol`

cmd/socat/main.go
```diff
@@ -0,0 +1,143 @@
...
+		return err
+	}
+	return r.r.Close()
+}
+
+func (r rwc) CloseWrite() error {
+	return r.w.Close()
+}
+
+func tryCloseWriter(c connection) error {
```
lkingland, line 133: This error return is being ignored, but it's return here suggests it should/could affect execution flow.

Perhaps, since it is not used, it is better to not return an error and write a message to stderr if an error is encountered trying to close the writer

cmd/socat/main.go
```diff
@@ -0,0 +1,143 @@
...
+	var uniDir bool
+	cmd := cobra.Command{
+		Use:   "socat <address> <address>",
+		Short: "Minimalistic socat.",
+		Long: `Minimalistic socat.
+Implements only TCP, OPEN and stdio ("-") addresses with no options.
+Only supported flag is -u.`,
+		Args: cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			right, err := createConnection(args[0])
```
lkingland, line 31: Minor nit, but if this is arg 0, and the first argument to `connect` later; perhaps this should be called `left` and the other `right`.

cmd/socat/main.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (r rwc) Read(p []byte) (n int, err error) {
+	return r.r.Read(p)
+}
+
+func (r rwc) Write(p []byte) (n int, err error) {
+	return r.w.Write(p)
+}
+
+func (r rwc) Close() error {
```
lkingland, line 121: I am not seeing where `rwc.Close()` is used.  Perhaps this could be used in place of `tryCloseWriter`, removing the need for that function, for the type coercion, and the associated interface

lkingland (approved):
/lgtm
/hold for others
Just a few minor questions and suggestions.
Would love to have a test

cmd/socat/main.go
matejvasek: But this deliberately mimics `socat` interface.

cmd/socat/main.go
matejvasek: my bad

cmd/socat/main.go
matejvasek:
```go
type writeCloser interface {
	CloseWrite() error
}
```
Is definitely not interchangeable with io.Closer. The former just closes one direction not both.
The `*net.TCPConn` implements both in expected manner.

I will add close to `Close()` just for completeness although it shouldn't be needed.

cmd/socat/go.mod
matejvasek: I am not sure. Because I want it to work in Docker file as standalone CLI app, it must not be sub-module of the func project. Probably it should be even in separate repo.

k8s/dialer.go
zroubalik: +1

lance:
@matejvasek any updates here?

matejvasek:
no update, we might not need this, I will see.

lance:
@matejvasek any status on this?

lance:
@matejvasek just checking in here - do you think this is still needed?

PR #1219, opened 2022-09-06
fix: downstream e2e fails with unknown default builder image

lkingland:
Downstream changes default builder to S2I, which currently has no
default Go builder image.  This PR pins the use of pack for the invocation test, which has support for all our core runtimes.

/kind bug

### DIFF
--- a/test/_e2e/cmd_invoke_test.go
+++ b/test/_e2e/cmd_invoke_test.go
@@ -59,7 +59,7 @@ func TestInvoke(t *testing.T) {
 
 	run(t, bin, prefix, "create", "--language=go", "--template=cloudevents", cwd)
 	set(t, "handle.go", TestInvokeFunctionImpl)
-	run(t, bin, prefix, "deploy", "--registry", GetRegistry())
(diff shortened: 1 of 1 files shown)


test/_e2e/cmd_invoke_test.go
```diff
@@ -57,7 +57,7 @@ func TestInvoke(t *testing.T) {
 	)
 	defer cleanup()
 
-	run(t, bin, prefix, "create", "--language=go", "--template=cloudevents", cwd)
+	run(t, bin, prefix, "create", "--language=node", "--template=cloudevents", cwd)
 	set(t, "handle.go", TestInvokeFunctionImpl)
 	run(t, bin, prefix, "deploy", "--registry", GetRegistry())
```
jrangelramos, line 62: Hi @lkingland . Just an idea, since here the goal is to test invoke, the builder used to generate the function itself won't matter much (IMO) so probbly we could force it to use `pack` so it won't fail on midstream

```suggestion
	run(t, bin, prefix, "deploy", "--builder", "pack", "--registry", GetRegistry())
```
So you don't need to change the whole test implementation, wdyt?

PR #1232, opened 2022-09-12
chore: dialer improvements

matejvasek:
# Changes

- :gift: Independence of `socat` image entrypoint.
- :gift: The `socat` image is configurable via environment variable.
- :test_tube: Make test assert less strict.

### DIFF
--- a/k8s/dialer.go
+++ b/k8s/dialer.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"os"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


k8s/dialer.go
```diff
@@ -141,10 +147,10 @@ func (c *contextDialer) startDialerPod(ctx context.Context) (err error) {
 			Containers: []coreV1.Container{
 				{
 					Name:      c.podName,
-					Image:     socatImage,
+					Image:     img,
 					Stdin:     true,
 					StdinOnce: true,
-					Args:      []string{"-u", "-", "OPEN:/dev/null,append"},
```
lkingland, line 147: Pretty sure the new custom dialer fails with "flags not supportd" if the address has commas; so this is good this is now changed.

lkingland (approved):
/lgtm
/hold for any other commentary

k8s/dialer.go
matejvasek: it will only issue warning to stderr, not error

matejvasek: also I removed the option in this PR :)

matejvasek: that `,append` is totally unnecessary

zroubalik (approved):
/lgtm

PR #1233, opened 2022-09-12
fix: s2i on cluster build of Quarkus functions

zroubalik:

Java related S2I builder images have S2I build scripts in `/usr/local/s2i`, default location for other languages is `/usr/libexec/s2i`.

# Changes

- :bug: s2i on cluster build of Quarkus functions

/kind bug


matejvasek:
/lgtm
/hold for others

Could it be also done via envvar in func.yaml?

lance:
> Could it be also done via envvar in func.yaml?

I think it could be done as a build env var, yes. Otherwise, /lgtm

zroubalik:
No, env vars are handled separately.

PR #1239, opened 2022-09-14
fix: cleanup of auto-podman-service when receiving SIGINT/SIGTERM

matejvasek:
# Changes

- :bug: The `podman` auto-svc has own control group now.
This prevents `podman` process termination on SIGKILL/SIGTERM sent to `func`.
We must keep `podman` alive for cleanup (e.g. container removal).

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -1,20 +1,14 @@
 package docker
 
 import (
-	"bytes"
-	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io"
 	"net/http"
 	"net/url"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"runtime"
-	"syscall"
-	"time"
 
 	"github.com/docker/docker/client"
 	"golang.org/x/crypto/ssh"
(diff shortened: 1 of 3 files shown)


docker/docker_client_linux.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return
+	}
+
+	podmanSocket := filepath.Join(tmpDir, "podman.sock")
+	dockerHost = fmt.Sprintf("unix://%s", podmanSocket)
+
+	cmd := exec.Command("podman", "system", "service", dockerHost, "--time=0")
+
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true, Pgid: 0}
```
matejvasek, line 30: Only effective line added in this PR.

matejvasek:
Without this `func run` termination with `control+c` will result in errors and un-deleted containers in daemon.

docker/docker_client.go
```diff
@@ -172,79 +166,6 @@ func podmanPresent() bool {
 	return err == nil
 }
 
-// creates a docker client that has its own podman service associated with it
-// the service is shutdown when Close() is called on the client
-func newClientWithPodmanService() (dockerClient client.CommonAPIClient, dockerHost string, err error) {
```
matejvasek, line 177: I just moved this to Linux specific file since I need Linux specific code.

docker/docker_client_linux.go
zroubalik: Thanks, that's helpful :)

zroubalik (approved):
/lgtm

jrangelramos:
thanks @matejvasek !!

PR #1240, opened 2022-09-14
[feat] Adding Releasability GH Workflow

salaboy:
# Changes

- :gift: This PR adds the releasability workflow needed to join the Knative release train.

/kind enhancement

This is just to follow the release procedures

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/.github/workflows/knative-releasability.yaml
+++ b/.github/workflows/knative-releasability.yaml
@@ -1,0 +1,32 @@
+# Copyright 2022 The Knative Authors.
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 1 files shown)


lance (commented):
I **think** that this should be automatically synced if we remove kn-plugin-func from this file https://github.com/knative-sandbox/knobots/blob/20f19dc30534eed1ea630d39484607481242534c/actions-omitted.yaml#L21-L23

dprotaso:
Lance is correct

salaboy:
@dprotaso so should I close this PR?

PR #1246, opened 2022-09-14
fix: updated func-deploy tekton task to aling with the new deploy command build flag options

jrangelramos:
# Changes

- :bug:  updated func-deploy tekton task to alingh with to new deploy flags

/kind bug

`func deploy` command now uses `--build false` instead of `--build disabled`, so changing `func-deploy` tekton task to reflect this change


zroubalik (approved):
/lgtm

matejvasek:
Good catch! @jrangelramos

lkingland:
Thanks for noticing this @jrangelramos

PR #1252, opened 2022-09-15
chore: change Makefile for multiple tags at HEAD

lance:
# Changes

With the move to the full Knative release train, we now get two tags on HEAD for the release commit. In the current case, it was `v0.34.0` for the "internal" version, and `knative-1.7.0`. The Makefile attempts to determine the version for static linking by looking at the tag at HEAD. With the git command returning two tags, `make` broke. This change takes the first of the two tags (will always be `knative-x.x.x`) which I think is less confusing for CLI users than `v0.34.0`.

I've also updated docs with a couple of changes that weren't reflected there but landed in 1.7.

Finally, I've removed versions.txt as that is no longer used, and docs/RELEASING.md as our processes have changed.

- 🐛 Fixed Makefile to handle multiple tags at HEAD
- 🧹 Removed versions.txt
- 🧹 Removed docs/RELEASING.md

/kind chore


lkingland (approved):
/lgtm

Definitely fixes this issue!  However, I would have assumed the correct solution is for `func version` to be the actual version  of `func` (and hopefully in semver format).  Why is it preferable for `func version` to be `knative-x.y.z` rather than the version of `func` itself?
> [knative-x.x.x] ... I think is less confusing for CLI users than `v0.34.0`.
I don't want to hold up the train, but for _this_ CLI user, I find the latter to be less confusing.
/hold for others to help me understand!

lance:
> Why is it preferable for `func version` to be `knative-x.y.z` rather than the version of `func` itself?

Maybe it's not? 🤷 I just assumed that `knative-1.7.0` for example was better because it aligns with the version numbering on the knative.dev page.
image
But I concede that semver could be better. Happy to `s/head/tail` in this PR if folks think that makes more sense.

lkingland:
> > Why is it preferable for `func version` to be `knative-x.y.z` rather than the version of `func` itself?
> 
> Maybe it's not? 🤷 I just assumed that `knative-1.7.0` for example was better because it aligns with the version numbering on the knative.dev page. image But I concede that semver could be better. Happy to `s/head/tail` in this PR if folks think that makes more sense.

While I certainly wish we could "collapse" the two versions into one, knative is a dependency, and the version numbers do seem to diverge.  For example, if we want to follow the semver "standard", then we would increment our minor version each time we add a new feature, correct?  If `func` is version `1.7` in order to align with the knative dependency, the wouldn't that mean we would be unable to increment our version number?

And in a particualry dystopic scenario, should we ever need to release a backwards-incompatible version of `func`... would we have to wait until knative 2.0?

I am no expert on Version-fu, but it seems like we need to keep the versions separate or lose the benefits of versions (something for which Go actually has quite a bit of builtin tooling nowadays).

Perhaps there is some way we could have our cake and eat it too:  `func version --verbose` could report the func version and the latest version of knative for which it is tested, and of course `func status` should eventually have this handy information?

lkingland:
/lgtm
Re-running tests; looks like they timed out during cluster allocation

lkingland:
@lance this seems ready to go, but I will let you initiate the merge through the removal of the 'hold' flag, since we are right in the middle of other build/release related upates, and this could be a confounding factor.

Issue #573, opened 2021-10-11
`func create` command should default to confirm prompt


nainaz:
# Context
During User interviews. When users are creating functions, default choice of language wasn't the expected experience and confused them to think that Node is the only choice and doesn't let the user know that there are other choices available for language/runtime and/or templates. In addition, user had to consult the Help text multiple times and it resulted in disjointed experience.

# Story 

**As Deandra the Developer**
**I want** to understand what options are available to me when creating a new function
**So that** I become familiar with all of the options that func supports

# Acceptance Criteria

- [ ] Guide the user through all the options available while creating func project 
```
GIVEN I typed func create into my terminal 
WHEN I hit enter 
THEN I am guided by an interactive CLI (which currently requires -c flag) 
```

omerbensaadon:
@nainaz I'm going to change this to default to -c the first time the user runs func create. 

I will open another issues which asks the user if they would like to store their option as their default runtime... 

Thoughts on this? @lance @lkingland @zroubalik where could this default be stored?

lance:
@omerbensaadon we have a `func config` command that we have discussed adding things like this to. Currently `func config` affects only the local project, but the tentative plan is to have `func config global` to manage global configuration options such as this (or possibly `func config -g`.

Already, we store installed remote template repositories that have been added with `func repo add` in `~/.config/func/repositories`. Global configuration such as the default runtime to use on create will be stored in a configuration file in `~/.config/func/config.yaml` or similar.

lkingland:
I am pretty sure the core problem here:
> default choice of language wasn't the expected experience
Will be solved by implementing the git-like behavior of prompting a user to provide the `--language` flag, but informing them this can be defaulted for future invocations using `git config X --local`.
This might make changing the default behavior to the `-c` flow unnecessary, but we can take a look again at that time!

lance:
@nainaz there are a couple of overlapping issues related to this. I closed one AND created one of them yesterday. :stuck_out_tongue: 

https://github.com/knative-sandbox/kn-plugin-func/issues/768 was created in favor of https://github.com/knative-sandbox/kn-plugin-func/issues/635, which I closed. And I think we should consider closing this one as well. wdyt?

omerbensaadon:
Glad y'all are finally getting around to some of this stuff!! 

I think this one is actually quite important for onboarding new users. IMO, we want to keep devs "hacking" for as long as possible before they have to reference any documentation. 

[...]

PR #1253, opened 2022-09-16
feat: func-create-command-should-default-to-confirm-prompt

grafvonb:


grafvonb:
Closes #573

grafvonb:
@lance ready for review

cmd/create.go
```diff
@@ -8,6 +8,8 @@ import (
 	"text/tabwriter"
 	"text/template"
 
+	flag "github.com/spf13/pflag"
+
```
zroubalik, line 12: Just a nit, any reason for this empty line?

cmd/create.go
```diff
@@ -247,12 +249,22 @@ func newCreateConfig(cmd *cobra.Command, args []string, newClient ClientFactory)
 		Confirm:          viper.GetBool("confirm"),
 		Verbose:          viper.GetBool("verbose"),
 	}
+
+	// Using Visit func to count the number of flags set at prompt. If 0 then set confirm mode to true.
```
zroubalik, line 253: At the moment the only mandatory flag is the one for language (`-l`) shouldn't we check the presence of this instead?

if user runs `func create -t http my-func` the confirm mode will not be enabled, but the language will not be selected.

cmd/create.go
grafvonb: Nope, copy&paste from `goimports`, I updated it as expected and `goimports` does not complain. Strange.

cmd/create.go
grafvonb: @lance any comments on that? I've just implemented the requested functionality "if `pure func create` then `run using confirm mode`"...

lkingland: This is... a little opaque.
While I admire the cleverness of "if no flags, default to confirm"; I think this will end up causing more confusion than clarity

lkingland (commented):
Thanks for the other contributions @grafvonb!  Unfortunately I am not going to be able to get behind _this_ one.  I understand that the linked issue is quite clear on what it is asking for, but we have a problem that a sizeable portion (I have no data on percentages :) ) of command-line users are not fond of processes that _default_ to hanging awaiting user input... and as such that UX should be avoided if possible.  Command should fail fast, with a non-zero exit code and a helpful message.

This issue will be solved when we deploy an improved `create` experience alongside a global config with an initial setup (both of which I am assigned to in addition to this one, as I seem them as interrelated).

Since flag values are retained through to prompts; in the event that these aforementioned changes do not address the root cause of the linked issue, and we do end up capitulating to default to a confirm prompt... I am fairly sure the PR could be:
```go
cmd.Flags().BoolP("confirm", "c", true, "Prompt to confirm all options interactively (Env: $FUNC_CONFIRM)")
```

grafvonb:
@lkingland Thank you for this detailed explanation of the current design considerations. As I understand, in the current situation, this PR makes no sense. Then I will close it, ok?

lkingland:
> @lkingland Thank you for this detailed explanation of the current design considerations. As I understand, in the current situation, this PR makes no sense. Then I will close it, ok?

Yes; sorry.  I am sure it's about as clear as mud now.

PR #1254, opened 2022-09-19
[feat] refactoring func.yaml struct to match cmd lifecycle

### LINKED ISSUE #817: Refactoring `func.yaml` file to follow a CRD format
Currently, the `func.yaml` file describes a Func project and it follows its own conventions and [schema](https://github.com/knative-sandbox/kn-plugin-func/blob/e918f74b9e177989ac520ae430dd5f232b393deb/schema/func_yaml-schema.json) for defining and validating the available configuration values. 

By refactoring `func.yaml` to a CRD-like resource we can: 
- Reuse existing Kubernetes API machinery to write & validate the configuration
- If custom metadata wants to be added to the Kubernetes resources generated by the `func` CLI we can follow the same principles as other Kubernetes resources do.
- On-ramp users to the Kubernetes world, there are a number of other projects following this approach like for example KinD, Keptn and Jenkins X. Func configuration files feel similar to other Kubernetes resources, which will be created by the `func` CLI
- It enables the idea of having a Kubernetes Controller to deal with these resources in the future, or it makes transformations easier to deal with Kubernetes tools, for example using mutation web hooks.
- By having a section for mutable fields (status), we establish a clear separation between configuration details and data is is updated by the `func` CLI

The following is an example of what this refactoring would look like:

```
apiVersion: func.knative.dev/v1alpha1 // instead of 0.19.0?
kind: Function
metadata: 
  name: level-1
  namespace: "default"
spec:
  runtime: springboot
  registry: "" // This should come from a global configuration? 
  invocation:
    format: http
  build: git
    git:
      url: https://github.com/salaboy/level-1.git
      revision: main
    builder: paketobuildpacks/builder:tiny
    builders:
      default: paketobuildpacks/builder:tiny
    buildpacks: []
    buildEnvs:
    - name: BP_NATIVE_IMAGE
       value: "true"
  template:
    spec: 
      image: salaboy/level-1/level-1:latest
      volumes: []
      env: []
      annotations: {}
      options: {}
      labels: []
      healthEndpoints:
        liveness: /health/liveness
        readiness: /health/readiness
status: // All modifications done by the CLI when updating state should go in status
  imageDigest: ""
  created: 2022-02-01T10:10:13.540854Z
``` 

@lance @evankanderson @zroubalik @matejvasek @xtreme-sameer-vohra (and community in general) feedback is highly appreciated. I've created the issue to see if we can discuss this in our weekly meeting tomorrow. 
Still in Draft, as I think that there are more advantages than the one listed here, wdyt?

salaboy:
# Changes

- :gift: updates func.yaml structure to have `build`, `run` and `deploy` sections
- :gift: add migration strategy to 1.0.x version ? should we call it 1.8? or the internal 0.34.x?
-

/kind api-change

Fixes #817

**Release Note**

```release-note
Changes func.yaml file structure to include nested build, deploy and run sections
```

**Docs**

```docs

```


function_migrations.go
```diff
@@ -203,6 +204,137 @@ func migrateToSpecVersion(f Function, m migration) (Function, error) {
 	return f, nil
 }
 
+// This migration makes sure use the 1.0.0 format. To avoid unmarshalling issues with the old format
+// this migration needs to be executed first. Further migrations will operate on this new struct
+func migrateTo100Structure(f1 Function, m migration) (Function, error) {
+	// Load the Function using pertinent parts of the previous version's schema:
+	f0Filename := filepath.Join(f1.Root, FunctionFile)
```
salaboy, line 211: @lkingland I am also experiencing some issues with this line. As I assumed that migrations are additive, but this line basically reload the file for every migration that we define, hence the latest migration will, in some way, ignore all the previous ones, because it is loading the file from the file system instead of the function that is returned from the previous migration. 

The test that is currently failing (`migrateToBuilderImages`) show cases this issue. Advice is appreciated here.. as I might be missing some details or design decisions.

function_migrations.go
```diff
@@ -203,6 +204,137 @@ func migrateToSpecVersion(f Function, m migration) (Function, error) {
...
+	}
+
+	f1.Name = f0.Name
+	f1.Runtime = f0.Runtime
+	f1.Template = f0.Template
+	f1.Registry = f0.Registry
+	f1.Image = f0.Image
+	f1.ImageDigest = f0.ImageDigest
+	f1.Invocation = f0.Invocation
+
```
lkingland, line 228: We shouldn't need to map every field, only those which have "moved":

1)  Assume `f1` (current version) is populated with everything that mapped over unchanged from the  previous version. (we're not in strict mode, so fields which no longer exist such as `.Git` should be ignored).
2) Load `func.yaml` _again_ (f0), but this time using the `previous version` schema (loads `.Git` etc)
3) map over anything which was moved (`.Git` -> `.Build.Git`).

This is because, at this granular resolution we should not have any unmarshalling errors _between_ migrations.  Those only occur when trying to jump f0 -> fN.  For example, if a field is removed then added back with a new schema (such as `.Build`, which used to be a string and is now a struct), that would cause an unmarshalling error during the initial unmarshalling, but when that fails, migrations are triggered and _between_ migrations there would be no error, because one migration removes the field, the next migration adds it back with the new type.

So I think you might want to rename `f1` back to `f` here like the other migrations; presume it is populated with everytyhing which has not changed (`Name`, `Runtime` etc), and then only use `f0` to grab the stuff that would be lost otherwise (all the fields relocated to the `XSpec` sub-structs).

lkingland (changes requested):
Nice step forward.  Hopefully the longwinded answers below help sort out those test failures.

function_migrations.go
salaboy: yes.. a migrated.Write() will be the right thing to do.. let me try with that and see how that goes.

cmd/repository_test.go
```diff
@@ -11,7 +11,7 @@ import (
 
 // TestRepository_List ensures that the 'list' subcommand shows the client's
 // set of repositories by name for builtin repositories, by explicitly
-// setting the repositories path to a new path which includes no others.
+// setting the repositories' path to a new path which includes no others.
```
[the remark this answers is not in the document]
salaboy: @lkingland, I would love to be the one spotting the issues, but in some weird way Grammarly made it into the IDE and it is highlighting the issues for me :)

function_migrations.go
salaboy: ok adding `migrated.Write()` opens a whole new can of worms :) The problem with that approach is that after running the first migration, the file moves to the latest format, and without a field by field migration on the first one executed, we will loose data.
Also this makes all the test to modify the files, which will require copying the files to a temp location to then do the migrations. 

I will try the second approach which is doing the additive migrations on a memory structure that is written only at the end. I am guessing that was the original approach based on how the test are working today.

salaboy: For now, I've fixed the issue to make sure that the migration is really additive.. this basically means that I am not replacing values that were modified by previous versions. For example in arrays or maps I am just bringing the old values forward, without destroying values that might be in the new structure already.

function.go
```diff
@@ -97,6 +68,44 @@ type Function struct {
 
 	// Health endpoints specified by the language pack
 	HealthEndpoints HealthEndpoints `yaml:"healthEndpoints"`
+}
+
+type Function struct {
```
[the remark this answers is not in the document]
salaboy: done

function.go
```diff
@@ -150,23 +169,34 @@ func NewFunctionWith(defaults Function) Function {
...
 	}
-	bb, err := ioutil.ReadFile(filename)
+	bb, err := os.ReadFile(filename)
 	if err != nil {
 		return
 	}
-	if err = yaml.Unmarshal(bb, &f); err != nil {
-		err = formatUnmarshalError(err) // human-friendly unmarshalling errors
-		return
+	var functionMarshallingError error
```
salaboy, line 183: @lkingland @lance help is appreciated here.. I do not like this error handling, but there was nothing in place to cover this use cases. For now I think this is ok to report back the errors (both marshalling and migration) but it can be improved in the future. If you know about other places where we are dealing with similar errors in a different way please point me in that direction or we can create another issue to improve this code later on.

cmd/client.go
```diff
@@ -25,7 +25,7 @@ type ClientConfig struct {
 	// currently configured in the client's connection should be used.
 	Namespace string
 
-	// Verbose logging.  By default logging output is kept to the bare minimum.
+	// Verbose logging.  By default, logging output is kept to the bare minimum.
```
lance, line 28: Personally, I would prefer to keep small changes like this to commits that also modify the file in some more meaningful way. In this case, adding a comma is the only change to the file. Minor nit - no need to change it, unless you're already making other updates.

docs/reference/func.md
```diff
@@ -4,7 +4,7 @@ Serverless functions
 
 ### Synopsis
 
-Serverless functions v0.0.0-source-2022-09-20T08:13:19&#43;02:00
+Serverless functions v0.0.0-source-2022-09-21T09:06:26&#43;01:00
```
lance, line 7: After rebasing the timestamp should not be included here.

function.go
lance: Off the top of my head, I cannot think of a place in the code where we are handling yaml marshaling differently. I do agree that this could be improved. A new issue would be good for that, imo. For now this works.

lance (approved):
Looks great! Ignore my comment about the comma - I saw in a later file your comment about the IDE doing some good cleanup.
/lgtm

docs/reference/func.md
salaboy: done

function.go
lkingland: Agreed with @lance we can improve this in a separate PR; with this working for now.  I might just upgrade to typed errors (see other comment for details)

function_test.go
```diff
@@ -135,3 +136,34 @@ func Test_Interpolate(t *testing.T) {
...
+	root := "testdata/testFunctionMarshallingError"
+
+	// Load the function to see it fail with a marshalling error
+	_, err := fn.NewFunction(root)
+	if err != nil {
+		if !strings.Contains(err.Error(), "line 2: cannot unmarshal !!seq into string") {
+			t.Fatalf("expected unmarshalling error")
+		}
+
+	}
```
lkingland, line 152: A minor nit here; and there are indeed a few ways to go about this (ranging from simple to complete);  but any time we want to test that a _specific_ error is being returned, it's better to use a typed error; creating an error type for the error to be tested.  This is not only more clear and less brittle than a string test, but also will greatly ease the potential future pain of internationalization/localization of text later:

For example, assuming we have an error type:
```go
type ErrUnmarshalling struct{
  Err error
}
func (e ErrUnmarhsalling) Error() string {
   return fmt.Sprintf("unable to read function: %s", e.Err)
}
```
And a reporting of that error somewhere:
```go
if err = yaml.Unmarshal(bb, &f0); err != nil {
		return f, ErrUnmarshalling{err}
}
```
Then the tests can be:
```go
if err := fn.NewFunction(root); !errors.Is(err, ErrUnmarshalling) {
  t.Fatalf("expected unmarshalling error not received")
}
```

salaboy: I am all in for this.. I will work on that fix after this large PR is merged

salaboy:
@lkingland @lance all comments addressed and all test green

lance (commented):
/lgtm

lkingland (approved):

PR #1255, opened 2022-09-19
feat: `invoke` - Adding option allowing insecure connections over SSL

zroubalik:

# Changes

- :gift: `invoke` - Adding option allowing insecure connections over SSL

/kind enhancement

**Release Note**

```release-note
`invoke` - Adding option allowing insecure connections over SSL
```

### DIFF
--- a/cmd/invoke.go
+++ b/cmd/invoke.go
@@ -26,7 +26,7 @@ NAME
 SYNOPSIS
 	{{.Name}} invoke [-t|--target] [-f|--format]
 	             [--id] [--source] [--type] [--data] [--file] [--content-type]
-	             [-s|--save] [-p|--path] [-c|--confirm] [-v|--verbose]
+	             [-s|--save] [-p|--path] [-i|--insecure] [-c|--confirm] [-v|--verbose]
 
 DESCRIPTION
 	Invokes the function by sending a test request to the currently running
@@ -97,9 +97,12 @@ EXAMPLES
 	o Invoke an arbitrary endpoint (CloudEvent)
 		$ {{.Name}} invoke -f=cloudevent -t="https://my-event-broker.example.com"
 
+	o Allow insecure server connections when using SSL
+		$ {{.Name}} invoke --insecure
+
 `,
 		SuggestFor: []string{"emit", "emti", "send", "emit", "exec", "nivoke", "onvoke", "unvoke", "knvoke", "imvoke", "ihvoke", "ibvoke"},
-		PreRunE:    bindEnv("path", "format", "target", "id", "source", "type", "data", "content-type", "file", "confirm"),
+		PreRunE:    bindEnv("path", "format", "target", "id", "source", "type", "data", "content-type", "file", "insecure", "confirm"),
 	}
 
 	// Flags
@@ -112,6 +115,7 @@ EXAMPLES
 	cmd.Flags().StringP("content-type", "", fn.DefaultInvokeContentType, "Content Type of the data. (Env: $FUNC_CONTENT_TYPE)")
 	cmd.Flags().StringP("data", "", fn.DefaultInvokeData, "Data to send in the request. (Env: $FUNC_DATA)")
 	cmd.Flags().StringP("file", "", "", "Path to a file to use as data. Overrides --data flag and should be sent with a correct --content-type. (Env: $FUNC_FILE)")
+	cmd.Flags().BoolP("insecure", "i", false, "Allow insecure server connections when using SSL. (Env: $FUNC_INSECURE)")
 	cmd.Flags().BoolP("confirm", "c", false, "Prompt to confirm all options interactively. (Env: $FUNC_CONFIRM)")
 
 	cmd.SetHelpFunc(defaultTemplatedHelp)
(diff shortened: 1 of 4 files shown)


cmd/invoke.go
```diff
@@ -375,5 +382,17 @@ func (c invokeConfig) prompt(client *fn.Client) (invokeConfig, error) {
 		return c, err
 	}
 
+	qs = []*survey.Question{
+		{
+			Name: "Insecure",
+			Prompt: &survey.Confirm{
+				Message: "Allow insecure server connections when using SSL",
+				Default: false,
```
lkingland, line 390: The default could be set to the current value of `cfg.Insecure` rather than `false` such that a call to `--confirm` which also includes `--insecure` will set the confirm prompt appropriately.
```shell
func invoke --insecure --confirm
```
... would then prompt but with the default set to `true`

lkingland (approved):
Nice!  I might just consider changing that default value in the prompt to be dynamic
/lgtm
/hold for others

cmd/invoke.go
zroubalik: Right!

zroubalik: Fixed

docs/reference/func_deploy.md
```diff
@@ -95,7 +95,7 @@ func deploy
   -p, --path string             Path to the project directory (Env: $FUNC_PATH) (default ".")
       --platform string         Target platform to build (e.g. linux/amd64).
   -u, --push                    Push the function image to registry before deploying (Env: $FUNC_PUSH) (default true)
-  -r, --registry string         Registry + namespace part of the image to build, ex 'ghcr.io/myuser'.  The full image name is automatically determined based on the local directory name. If not provided the registry will be taken from func.yaml (Env: $FUNC_REGISTRY)
+  -r, --registry string         Registry + namespace part of the image to build, ex 'ghcr.io/myuser'.  The full image name is automatically determined. (Env: $FUNC_REGISTRY) (default "image-registry.openshift-image-registry.svc:5000/demo")
```
lance, line 98: This change seems unrelated to the overall PR - and I don't think that the default in upstream is the openshift image registry.

lance (commented):
Overall great contribution - thanks! I'm just not sure about the change to func_deploy.md

docs/reference/func_deploy.md
zroubalik: I have no idea how does this actually appeared here.

zroubalik: Good catch, thanks. Fixed.

matejvasek:
Would it make sense to have this as top level flag? It could be for instance useful for users with self-signed private OCI registry.

matejvasek:
@jrangelramos would you find useful to have the flag global for testing purposes?

matejvasek:
Thinking of it more deeply: it wouldn't be easy to implement this for `push` operation,
because sometimes push is done directly by `docker` daemon, and there is no easy way to force daemon do it.

jrangelramos:
@matejvasek I think ti would be useful for deploy command for sure, specially if the user options could be forwarded to on-cluster builds deploy step as well

docs/reference/func.md
```diff
@@ -4,7 +4,7 @@ Serverless functions
 
 ### Synopsis
 
-Serverless functions v0.0.0-source-2022-09-12T14:54:32-06:00
+Serverless functions v0.0.0-source-2022-09-19T12:01:46&#43;02:00
```
matejvasek, line 7: This probably doesn't have to be part of the PR?

docs/reference/func_deploy.md
```diff
@@ -12,7 +12,7 @@ SYNOPSIS
 	func deploy [-R|--remote] [-r|--registry] [-i|--image] [-n|--namespace]
 	             [-e|env] [-g|--git-url] [-t|git-branch] [-d|--git-dir]
 	             [-b|--build] [--builder] [--builder-image] [-p|--push]
-	             [-c|--confirm] [-v|--verbose]
+	             [--platform] [-c|--confirm] [-v|--verbose]
```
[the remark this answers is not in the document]
matejvasek: Should it be part of this PR?

matejvasek:
> if the user options could be forwarded to on-cluster builds deploy step as well. For sure that will save some user of some headaches.. :)

I am afraid that would be out of scope. We cannot easily force `docker daemon` or `k8s cluster` to trust connections.

matejvasek:
Looks good, maybe with exception of the two `*.md` file changes.

docs/reference/func_deploy.md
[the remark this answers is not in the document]
zroubalik: The `--platform` was probably missing in the generated doc. This change is just making the docs correct, or not?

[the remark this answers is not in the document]
lkingland: This seems a reasonable cleanup IMHO.  There was a missing flag name in the help text synopsis.  This replicates the fixing of that omission to the generated help text.  Seems fine?

zroubalik:
I have removed the docs fix from this PR and created another one. @matejvasek PTAL

PR #1257, opened 2022-09-20
fix: incorrect deploy namespace warnings

### LINKED ISSUE #1256: Unnecessary Warning on First Deploy
The system now warns the user if a Funciton is being deployed to a different namespace than is active in the users kube context, or if the function is already deployed and being deployed to a different namespace.  This warning should _not_ be generated on the very first deployment, because the function has not yet been deployed:

```shell
Warning: Function is in namespace '', but currently active namespace is 'default'. Continuing with redeployment to ''.
```
Thie above warning is ineffectual.  This warning should not be printed if the Function is currently undeployed.

/kind bug good-first-issue

lkingland:
- :bug: squelch invalid initial warning
- :bug: fix warning message text's last parameter

/kind bug
Fixes #1256

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -690,10 +690,16 @@ func namespace(cfg deployConfig, f fn.Function, stderr io.Writer) (namespace str
 		}
 	}
 
+	// If the Function is not yet deployed, then immediately return the chosen
+	// final namespace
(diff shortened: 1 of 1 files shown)


lance (commented):
/lgtm

Issue #1097, opened 2022-07-06
Streamline `func create` for newcomers


salaboy:
Currently `func create` takes a mandatory argument (language) in the shape of a flag. It also requires to be run inside an empty directory (unless you specify the path) which is not optimal or expected by all users. 

## Mandatory arguments

Currently, the "language" flag is the only mandatory parameter so, 

If you type `func create` on an empty directory you get: 
```
> func create 
Error: Required flag "language" not set.
Available language runtimes are:
  go
  node
  python
  quarkus
  rust
  springboot
  typescript
```

This doesn't help a newcomer to know what the `language` flag is, for some people this will mean `--language` or `-l` but it might not be clear for everyone. Notice that this is a required parameter hence making this a positional argument might make some sense. In such a case we can make something like this: 

```
>func create

Language is missing, please add the language argument: 
Example: func create <language> choosing one from the following list
Available language runtimes are:
  go
  node
  python
  quarkus
  rust
  springboot
  typescript
```

Checking the `func create --help`, it shows an example that is not working: 
```
 o Create a Node.js Function in the directory 'myfunc'.
          $ func create myfunc

```

## Directory creation

Creating or not creating a directory for the function is a controversial topic that, in my opinion, we need to fix. 
If we follow the `git`, `npm`  or even `go mod` approach you are always working in a directory where your project will live. For these cases, the user needs to create a directory. Current `func create` checks if the user is in an empty directory to create a function if the name of the function is not provided as the last argument, in which case it creates a directory. 

For example: 

```
func create -l go func3
Created go Function in /tmp/func3
```
Created a directory and a function inside it. 

For this one I have mixed feelings, as we should decide if we create or not create directories with `func`, supporting both in my opinion causes confusion. 

if we change [path] to be a flag, it will make more sense for cases where a CI tool is running create, so we don't loose the functionality. 

## Open Questions

Do you have any other categories of things to improve for `func create` ? Any other thoughts or ideas? 
@lance @lkingland @zroubalik @matejvasek

lance:
We have talked in the past about entering "confirm" mode when a user types simply `func create` with no flags/parameters. See https://github.com/knative-sandbox/kn-plugin-func/issues/573#issuecomment-1017995470

Personally, I like this a lot.

grafvonb:
Current release v1.7.0 shows after `func create` this:
```
[...]

Issue #901, opened 2022-03-15
Global configuration for functions


lance:
Allow the user to set global configuration options, for example the image registry, with `func config --global`.

nainaz:
**Context**: Enterprises and developers using the same configuration parameters that may be different from the default ones for creating functions currently has to change the values repeatedly. Cluster admin would like to set up the common and standard parameters for the developers

**Story**:

As Dina the developer,

I should be able set some global configuration parameters across my machine/cluster

so that I can have the same configuration override applied to all my functions and not fiddle with the default ones provided every time.




As cluster admin,

I should be able set some global configuration parameters across my machine/cluster

so that I can have the developers user the consistent approved configuration values override applied to all functions created in the org 

**Acceptance Criteria**
GIVEN that I provide the global override of configuration parameter 

WHEN I write and create my functions

THEN globally set configurations should be used rather than the default ones while building and deploying my function.

Issue #636, opened 2021-11-05
`func create`: persist --language choice


omerbensaadon:
- Give users the option to save the runtime as default (y/n) 
- Make sure to let them know where this is saved
- Make sure the template always defaults to -c when not provided as a flag (-t)

lance:
> * Make sure to let them know where this is saved

I don't think that's a great idea. I think the user should just be able to run `func config global`. See: https://github.com/knative-sandbox/kn-plugin-func/issues/573#issuecomment-962003540

lkingland:
I will discuss this further with @lance this week, but the work I am doing on `func invoke` actually requires taking a bite out of the `config` situation:  in particular opening the door to more conventional "config" settings, including git-like behavior of displaying a helpful message informing the user they need to choose a language, but can set a default via `git config ... --global`  (exact syntax TBD)

lkingland:
To expand on this issue a bit, I was envisioning that the first time `func` is run, it will detect there is no user configuration, and the user will be prompted to run through a first-time setup, which is creating the global config (see #901).   One of the questions asked of course being the user's preferred default language (if any).

PR #1260, opened 2022-09-20
feat: readonly global config

lkingland:
- :gift:  adds initial support for a global config in `~/.config/func/config.yaml`
- :gift: default for `func create --language` can now be defined in global config
- :gift: defaulting to confirm prompts can be enabled in the global config
- :gift: modifed defaults are visible in help text.

Example  `~/.config/func/config.yaml`:
```
language: go
confirm: true
```

The config location is mutable using `XDG_CONFIG_HOME`.  A config file is not required for normal operation.  Values are accurately reflected in command help text.

This is the first part of a general-purpose Global Config (#901) which is prerequisite to, and will also partially  address the interrelated issues #573 #636 and #1097. 

/kind enhancement

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,0 +1,121 @@
+package config
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/mitchellh/go-homedir"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	// Filename into which Config is serialized
+	Filename = "config.yaml"
+
+	// DefaultConfigPath is used in the unlikely event that
+	// the user has no home directory (no ~), there is no
+	// XDG_CONFIG_HOME set
+	DefaultConfigPath = ".config/func"
+
+	// DefaultLanguage is intentionaly undefined.
+	DefaultLanguage = ""
+)
+
+type Config struct {
+	// Language Runtime
+	Language string `yaml:"language"`
+
+	// Confirm Prompts
+	Confirm bool `yaml:"confirm"`
+}
+
+// New Config struct with all members set to static defaults.  See NewDefaults
+// for one which further takes into account the optional config file.
+func New() Config {
+	return Config{
+		Language: DefaultLanguage,
+		// ...
+	}
+}
+
+// Creates a new config populated by global defaults as defined by the
+// config file located in .Path() (the global func settings path, which is
+//  usually ~/.config/func)
+func NewDefault() (cfg Config, err error) {
+	cfg = New()       // cfg now populated by static defaults
+	p := ConfigPath() // applies ~/.config/func/config.yaml if it exists
+	if _, err = os.Stat(p); err != nil {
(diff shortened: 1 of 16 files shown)


config/config.go
```diff
@@ -0,0 +1,121 @@
...
+		if os.IsNotExist(err) {
+			err = nil // config file is not required
+		}
+		return
+	}
+	bb, err := os.ReadFile(p)
+	if err != nil {
+		return
+	}
+	err = yaml.Unmarshal(bb, &cfg) // cfg now has applied config.yaml
```
lance, line 59:
```suggestion
    cfg, err = Load(p)
    if os.IsNotExist(err) {
		err = nil // config file is not required
	}
```

lance (commented):
/lgtm with one small suggestion
/hold for others

config/config.go
lkingland: /facepalm
Pretty sure that's the reason I added the `Load` method in the first place.
_codus interruptus_ perhaps!

lkingland: Hey @lance; while in half-sleep late last night I realized why this was the way it was.
The Config object is first instantiated with its values populated with static defaults.
Then the .yaml is loaded into that sams struct, overwriting any static defaults with user-provided global settings.
The `Load` method merely loads exactly the .yaml, ignoring static defaults.
Therefore, updating this method in this manner will ignore the `cfg` object and thus static defaults
(hence the code staticcheck complaining about the unused declaration).
This is the general path: static defaults first, user provided config values overwrite those, environment variables overwrite those, and finally flags.  👍🏻  

I refactored a bit to make this more clear so as to hopefully avoid the same confusion in the future.

lkingland:
> This branch has conflicts that must be resolved
> Use the [web editor](https://github.com/knative-sandbox/kn-plugin-func/pull/1260/conflicts) or the to resolve conflicts.
> Conflicting files
> docker/creds/credentials_test.go


Screen Shot 2022-09-22 at 11 23 01 PM

config/config.go
```diff
@@ -0,0 +1,121 @@
...
+	// DefaultLanguage is intentionaly undefined.
+	DefaultLanguage = ""
+)
+
+type Config struct {
+	// Language Runtime
+	Language string `yaml:"language"`
+
+	// Confirm Prompts
+	Confirm bool `yaml:"confirm"`
```
salaboy, line 31: @lkingland so the idea here is to add for example a global default registry for all functions? I mean in another PR

lkingland: Yes, along with any other settings who would benefit from a shared configurable default.

lance (commented):
/lgtm

config/config.go
zroubalik: 💯

zroubalik (approved):
/lgtm

great job @lkingland !

lance:
Thanks for allowing commits from maintainers @lkingland - I have resolved the merge conflicts. Assuming tests pass, looks good to me.

/lgtm

Issue #1259, opened 2022-09-20
`func create` should validate function name


matejvasek:
As of now it's possible to create a function as `func create 42`.
However such a name causes troubles because it's not valid part of domain.
Calling `func create 42` should result int error.

lance:
Hmm - I agree that `42` is probably not ideal for a function or service name. But I'm trying to find any text that documents this as an error in Kube itself. All formatting restrictions on names that I can find note an alphanumeric requirement, but don't prohibit fully numeric names. See: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names. In that reference, it talks about DNS-1123 naming restrictions which is what we are validating for here. https://github.com/knative-sandbox/kn-plugin-func/blob/abc3097d7563051913573f82cd084add9af8b421/utils/names.go#L29-L41

matejvasek:
@lance the regex there seems to be wrong:
```
Error: knative deployer failed to deploy the Knative Service: admission webhook "[validation.webhook.serving.knative.dev](http://validation.webhook.serving.knative.dev/)" denied the request: validation failed: not a DNS 1035 label: [a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')]: metadata.name
```
note that the name must start with [a-z].

lance:
> @lance the regex there seems to be wrong:
> 
> ```
> Error: knative deployer failed to deploy the Knative Service: admission webhook "[validation.webhook.serving.knative.dev](http://validation.webhook.serving.knative.dev/)" denied the request: validation failed: not a DNS 1035 label: [a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')]: metadata.name
> ```
> 
> note that the name must start with [a-z].

Good catch - I missed that when scanning the name requirements. Start AND end with [a-z]

matejvasek:
@lance maybe the validation changed over time, hard to believe we would got it wrong.

matejvasek:
@dsimansk @rhuss is it possible that validation changes?

lkingland:
[...]

PR #1263, opened 2022-09-20
fix: validation of function name

matejvasek:
# Changes

- :bug: Fixed validation: function name must be DNS 1123 Label.

/kind bug

```release-note
fix: validation of function name -- it must be DNS 1123 Label.
```

resolves: #1259

### DIFF
--- a/utils/names.go
+++ b/utils/names.go
@@ -23,18 +23,18 @@ type ErrInvalidSecretKey error
 // ErrInvalidLabel indicates the name did not pass label key validation, or the value did not pass label value validation.
 type ErrInvalidLabel error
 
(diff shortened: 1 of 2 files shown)


utils/names.go
```diff
@@ -28,13 +28,13 @@ type ErrInvalidLabel error
 // (e.g. 'my-name',  or '123-abc', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?')
```
rhuss, line 28: shouldn't this comment be also updated ? (it would match "123", too)

lance (commented):
/lgtm

PR #1265, opened 2022-09-20
chore: remove timestamp from generated docs

lance:
# Changes

- 🧹 Remove timestamp from generated documentation

This is causing trouble with new PRs that potentially change CLI flags or other user-facing UX. Another option besides completely removing the version information is to modify Version.StringVerbose() so that it uses the value of the most recent semver tag (or Knative tag).

```
❯ git describe --match "v*"
v0.34.0-6-g6d26e64c

❯ git describe --match "knative*"
knative-v1.7.0-6-g6d26e64c
```

This is nice, but could be misleading if the UX has changed since that release.

/kind chore


matejvasek:
@lance linter detected some dead-code.

PR #1266, opened 2022-09-20
chore: linkage assignable socatImage

matejvasek:
A `var` can be set at build (link) time as opposed to a `const`.

### DIFF
--- a/k8s/dialer.go
+++ b/k8s/dialer.go
@@ -23,9 +23,7 @@ import (
 	"k8s.io/client-go/tools/remotecommand"
 )
 
-const (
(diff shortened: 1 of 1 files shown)


matejvasek:
@lance PTAL, this might be useful for next release.

lkingland (approved):
This would be an easy thing for a refactor or a code review to move back to a const (it looks like it should be a const).  I am open on suggestions on how to convey that it is expected to be mutable without resorting to a comment.  In some other languages all caps are used for global constants.. that's definitely not a thing in Go.  ¯\_(ツ)_/¯

PR #1273, opened 2022-09-22
fix: nil deref when custom builder image

matejvasek:
# Changes
- :bug: Fix nil deref when custom builder image.

/kind bug

```release-note
fix: nil deref when custom builder image
```

### DIFF
--- a/function.go
+++ b/function.go
@@ -153,6 +153,8 @@ func NewFunctionWith(defaults Function) Function {
 // selectively consider the minimal validation necesssary to ehable migration.
 func NewFunction(path string) (f Function, err error) {
 	f.Root = path // path is not persisted, as this is the purvew of the FS itself
+	f.BuilderImages = make(map[string]string)
+	f.Annotations = make(map[string]string)
 	var filename = filepath.Join(path, FunctionFile)
 	if _, err = os.Stat(filename); err != nil {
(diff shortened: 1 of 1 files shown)


zroubalik (approved):
/lgtm
/approve

lance (commented):
/lgtm

cmd/build.go
```diff
@@ -123,6 +123,9 @@ func runBuild(cmd *cobra.Command, _ []string, newClient ClientFactory) (err erro
 		f.Builder = config.Builder
 	}
 	if config.BuilderImage != "" {
+		if f.BuilderImages == nil {
+			f.BuilderImages = make(map[string]string, 1)
+		}
```
lkingland, line 128: IMHO this should be in the constructor of `config`.

matejvasek: But config is preparing config struct, not modifying func struct.

Also why would I modify it when I don't have to?
It can be nil (and therefore omitted from `func.yaml`) until we need assign to it.

matejvasek: or you mean `NewFunction()` ?

matejvasek:
@lkingland PTAL

matejvasek:
oops it has been merged before update, nevermind

PR #1274, opened 2022-09-22
feat: persist user preference for the `--remote` flag

lance:
# Changes

- 🎁 This commit modifies the deploy command so that it persists the value of the `--remote` flag upon deployment.
- 🗒️ Depends on https://github.com/knative-sandbox/kn-plugin-func/pull/1260

/kind enhancement

**Release Note**

```release-note
Persists user preference for remote deployment
```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -20,6 +20,7 @@ import (
 	fn "knative.dev/kn-plugin-func"
 	"knative.dev/kn-plugin-func/builders"
 	"knative.dev/kn-plugin-func/buildpacks"
+	"knative.dev/kn-plugin-func/config"
 	"knative.dev/kn-plugin-func/docker"
 	"knative.dev/kn-plugin-func/docker/creds"
 	"knative.dev/kn-plugin-func/k8s"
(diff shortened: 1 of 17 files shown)


cmd/deploy.go
```diff
@@ -267,8 +275,15 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 		return ErrRegistryRequired
 	}
 
+	// Load global config to persist the --remote flag
+	gc, err := fc.NewDefault()
+	if err != nil {
+		fmt.Fprintf(cmd.OutOrStdout(), "error loading config at '%v'. %v\n", fc.ConfigPath(), err)
+	}
+	gc.Remote = cfg.Remote
+
```
lkingland, line 284: Maybe I am mis-reading, but this stanza _should_ not be necessary, because `cfg.Remote` should have already taken into account the static default (false), global config setting (via the flag default value), environment variables and flags (during the construction of the cfg struct)

The default value of the flag --remote will be set to either the static default (false) or the value defined in ~/.config/func/config.yaml`.  If the user decided to (globally) prefer remote, then `cfg.Remote` should reflect this value.

`cfg.Remote` also reflects if the user chose to override their global setting by explicitly providing a --remote flag (or environment variable).

cmd/deploy_test.go
```diff
@@ -1018,3 +1019,60 @@ func TestDeploy_NamespaceRedeployWarning(t *testing.T) {
...
+func TestDeploy_persistRemoteFlag(t *testing.T) {
+	root, rm := Mktemp(t)
+	defer rm()
+
+	// A function with minimum required values for deployment
+	f := fn.Function{
+		Root:     root,
+		Name:     "myfunc",
+		Runtime:  "go",
+		Registry: "example.com/alice",
```
lkingland, line 1034: In this case, we should actually be able to omit everything except `root` and `runtime` (all those noop implementations don't need a name or registry)

cmd/deploy_test.go
```diff
@@ -1018,3 +1019,60 @@ func TestDeploy_NamespaceRedeployWarning(t *testing.T) {
...
+
+	// Ensure that successfully executing the command results in the --remote
+	// flag being persisted. Load the global config and check the value.
+	cfg, err := config.NewDefault()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !cfg.Remote {
+		t.Fatal("expected remote flag to be persisted as true in global config")
+	}
```
lkingland, line 1060: I really like where this PR is headed:  that we can use the global config to define a default (globally) for the --remote flag.

This is, however, a separate feature from _persisting_ the value of the --remote flag (for an individual Function), because that setting would be written to `.func`.

The PR I am working on now actually injects the current Function context into the final value returned when instantiating that global config object.  That's going to be a little bit tricky because this is the first function-scoped value which is not actually part of the core (it pertains only to the `cmd` package).

I might consider using THIS PR to deliver the global config remote value setting, and have this test confirm it is read (perhaps citing a new issue that the global config should support configuring the global default "remote" setting).

lkingland (commented):
I know it's just a draft, but wanted to comment anyway :)

lance:
Closing in favor of https://github.com/knative/func/pull/1402

PR #1286, opened 2022-09-27
feat: ksvc log on deploy

matejvasek:
# Changes

- :gift: Display ksvc ouput on deploy for better debugging.

/kind enhancement

### DIFF
--- a/knative/logs.go
+++ b/knative/logs.go
@@ -1,0 +1,119 @@
+package knative
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"sync"
+	"time"
+
+	"golang.org/x/sync/errgroup"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"knative.dev/kn-plugin-func/k8s"
+)
+
+// GetKServiceLogs will get logs of Knative service.
+//
+// It will do so by gathering logs of user-container of all affiliated pods.
+// In addition, filtering on image can be done so only logs for given image are logged.
+// The image must be the digest format since pods of Knative service use it.
+//
+// This function runs as long as the passed context is active (i.e. it is required cancel the context to stop log gathering).
+func GetKServiceLogs(ctx context.Context, namespace, kServiceName, image string, since *time.Time, out io.Writer) error {
+	client, namespace, err := k8s.NewClientAndResolvedNamespace(namespace)
+	if err != nil {
+		return fmt.Errorf("cannot create k8s client: %w", err)
+	}
+
+	pods := client.CoreV1().Pods(namespace)
+
+	podListOpts := metav1.ListOptions{
+		Watch:         true,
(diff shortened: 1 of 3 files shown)


matejvasek:
If you are asking why only on first deployment:
First deployment is already special case -- only on first deployment we wait for service, on subsequent deployment we only update `ksvc` (almost instantaneous) and return.

matejvasek:
This is useful for instance if user deploys image that is build for different architecture than cluster's.
A message about bad exec would be displayed making it easier to realise what's wrong.

matejvasek:
@lance @lkingland PTAL

knative/logs.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// GetKServiceLogs will get logs of Knative service.
+//
+// It will do so by gathering logs of user-container of all affiliated pods.
+// In addition, filtering on image can be done so only logs for given image are logged.
+// The image must be the digest format since pods of Knative service use it.
+//
+// This function runs as long as the passed context is active (i.e. it is required cancel the context to stop log gathering).
+func GetKServiceLogs(ctx context.Context, namespace, kServiceName, image string, since *time.Time, out io.Writer) error {
```
matejvasek, line 24: @rhuss @dsimansk is there any off the shelf solution that works?
I tried `stern` and `kn-plugin-service-log` neither of them is good enough.
They are not displaying error `exec /bin/sh: exec format error` when deploying image with bad arch.
I don't know why.

zroubalik: https://github.com/knative-sandbox/kn-plugin-func/blob/e041bc84ebe4accc7e099cc45107269164e94066/k8s/logs.go#L13

matejvasek: @zroubalik I've seen that one but I wanted to gather logs in streaming/continuous way not ex post. Also I am gathering logs from multiple pods.

matejvasek: >I wanted to gather logs in streaming/continuous way not ex post.

btw I think that would be also useful for pipeline build with `-v` flag.

matejvasek: >I wanted to gather logs in streaming/continuous way not ex post.

I want to print error ASAP, not after 2 minutest of futile CrashLoopBackOff-Error loop.

zroubalik: > > I wanted to gather logs in streaming/continuous way not ex post.
> 
> btw I think that would be also useful for pipeline build with `-v` flag.

That would be nice.

I am not sure if this is the canonical way of reading logs, but `tkn` (Tekton cli) has this capability and it's quite nice, might be worth checking that out: 
https://github.com/tektoncd/cli/blob/4f6674193b2b6f41c32e14b1e0978d7e1b66c39a/pkg/pods/container.go#L88

matejvasek:
@lance @zroubalik @lkingland PTAL

matejvasek:
@lance @lkingland @dsimansk PTAL

lkingland (approved):
/lgtm
So nice to have some logs

/hold for others

lance (commented):
/lgtm

PR #1300, opened 2022-10-05
feat: interactive prompt for on-cluster deploy

### LINKED ISSUE #852: On-cluster build: Implement interactive prompt
Currently, on-cluster builds are accomplished through the `func deploy` command. While there is a `--confirm` prompt for local builds it does not account for the possibility that the user would like to build on the cluster, and therefore does not prompt for that information. See:

image

The `--confirm` mode should start by asking the user if they would like to build locally or on the cluster and proceed from there.

vyasgun:
# Changes

- 🎁 Prompt user to confirm if they would like to build locally or on-cluster when `--confirm` flag is used

/kind enhancement

Fixes #852 

**Release Note**

```release-note
Prompts user for remote deployment
```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -567,6 +567,20 @@ func (c deployConfig) Prompt() (deployConfig, error) {
 
 	var qs = []*survey.Question{
 		{
+			Name: "remote",
+			Prompt: &survey.Input{
+				Message: "Build on remote cluster",
+				Default: strconv.FormatBool(c.Remote),
+			},
+		},
+		{
+			Name: "GitURL",
+			Prompt: &survey.Input{
+				Message: "Git URL",
+				Default: c.GitURL,
+			},
+		},
+		{
 			Name: "namespace",
 			Prompt: &survey.Input{
 				Message: "Destination namespace:",


cmd/deploy.go
```diff
@@ -566,6 +566,20 @@ func (c deployConfig) Prompt() (deployConfig, error) {
 	}
 
 	var qs = []*survey.Question{
+		{
+			Name: "remote",
+			Prompt: &survey.Input{
+				Message: "Build on remote cluster",
+				Default: strconv.FormatBool(c.Remote),
```
lkingland, line 573: I think this will require that the user literally type "yes", "true" etc.
Perhaps take a look at the documentation for the library we are using for this "https://github.com/AlecAivazis/survey", or other commands which prompt for boolean values.  A good choice for boolean values is  `survey.Confirm` for 
```
Name: "remote",
Prompt: &survey.Confirm{
    Message: "Trigger a remote (on-cluster) build?",
    Default: c.Remote,
}
```

lkingland (commented):
Thanks for the PR!

With a small change to use `survey.Confirm` for the boolean (see below), this should be good to go.

lance (commented):
/lgtm
/approve

Issue #1297, opened 2022-10-04
The `--registry` flag is ignored when `image: <imagename>` exists in func.yaml


lance:
```
❯ func build --registry=docker.io/lanceball
   🙌 Function image built: quay.io/lanceball/hello:latest
```

Note that it built a `quay.io` image instead. We should probably detect when there is a conflict here and warn the user.

/kind bug

grafvonb:
@lance What is the expected behavior in such a case? Personally, I would expect `docker.io/lanceball/hello:latest` to be built. However, `func.yaml` should not be updated, right?

lance:
> @lance What is the expected behavior in such a case? Personally, I would expect `docker.io/lanceball/hello:latest` to be built. However, `func.yaml` should not be updated, right?

Good question. I think you are probably right, but we may want to print some text to the console with a warning saying that the image registry is different than what is found in `func.yaml`, and if the user wants to make this change permanent, the file should be updated.

Or conversely, we could update the file, but print a warning that it was updated with a new registry name.

There is a similar warning when specifying a conflicting namespace in the CLI.

If we wanted to be super sexy about it, we could prompt to ask if the user wants the setting to be permanent (and therefore update the file).

lance:
@grafvonb that sounds great - thanks!

grafvonb:
@lance The "super sexy" behavior cannot be implemented without a global redesign of the way our `func` handles function configuration. In `build.go`, @lkingland implemented this after reading and setting all the important configuration properties:
```
	// This preemptive write call will be unnecessary when the API is updated.
	// To use function instances instead of file paths. For now it must write
	// even if the command fails later.  Not ideal.
	if err = f.Write(); err != nil {
		return
	}
```
To avoid the ambiguity described above and keep the solution simple, I suggest the following:
1. check if `registry` != registry part if `image`.
2. if `true` then warn and update the registry part of `image`.

This way other steps/commands will work as if nothing happened.

What is your opinion?

lkingland:
> cannot be implemented without a global redesign of the way our `func` handles function configuration

Funny you should mention it in just that way.  An update to enforce consitent configuration precidence is in the works as part of offering a global configuration file.  In short, the value in func.yaml will be used unless a new value is expressly provided via flag(which has the highest precedence).  For now your solution seems like a good solution in the present circumstance.

We are also printing warning messages when the value of `--namespace` could potentially cause confusing situations for the user.

[...]

PR #1310, opened 2022-10-09
fix: add handling in case of differences between registry value and image tag (#1297)

grafvonb:
Solution:
1. if there is a mismatch in `func.yaml` between `registry` and `image` the latter will be updated with `registry`
2. if there is a mismatch between `func.yaml` registry and image comparing to flag value `--registry` both will be updated with the flag value
3. if `--image` was provided, use it and do NOT update even if there is a mismatch with the `registry`.


grafvonb:
Ready for review.

/assign @lkingland

lance (commented):
Looks great @grafvonb thank you!

You have added a test for the case where registry differs in func.yaml from the value in the image name. Any chance you can add test cases for the remaining scenarios?

- [x] if there is a mismatch in func.yaml between registry and image the latter will be updated with registry
- [x] if there is a mismatch between func.yaml registry and image comparing to flag value --registry both will be updated with the flag value
- [x] if --image was provided, use it and do NOT update even if there is a mismatch with the registry.

grafvonb:
> You have added a test for the case where registry differs in func.yaml from the value in the image name. Any chance you can add test cases for the remaining scenarios?

@lance sure 👍🏻

grafvonb:
@lance done, ready for review

lance (commented):
/lgtm

Thanks for the contributions @grafvonb!

PR #1330, opened 2022-10-13
test: added test for get pod's logs

matejvasek:
# Changes

- :test_tube: Test get pod's logs.

/kind test

### DIFF
--- a/k8s/logs_test.go
+++ b/k8s/logs_test.go
@@ -1,0 +1,91 @@
+//go:build integration
+// +build integration
+
+package k8s_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	"knative.dev/func/k8s"
+)
+
+func TestGetPodLogs(t *testing.T) {
+	var err error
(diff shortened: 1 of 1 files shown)


matejvasek:
@zroubalik @dolfolife

k8s/logs_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	pod, err = cliSet.CoreV1().Pods(testingNS).Create(ctx, pod, metav1.CreateOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Log("created pod: " + testingPodName)
+
+out:
+	for i := 0; i < 600; i++ {
```
lance, line 71: Is there not an kube API call to wait until the pod is ready?

matejvasek: It is quite low level. I am aware only of watch interface which is not much better than this with respect to line of code.

matejvasek: https://github.com/knative/func/blob/84741e24794e039e17dd2d8292bf37502f376c88/k8s/persistent_volumes.go#L168-L193
Is it any better?

lance: > Is it any better?

I would not be surprised if it's doing something very similar.

matejvasek: >I would not be surprised if it's doing something very similar.

Nah I think it is smarter than that, like doing some long polling on server or something.

Issue #1278, opened 2022-09-24
Improve coverage of knative package (currently close to 6%)


salaboy:
See: https://app.codecov.io/gh/knative-sandbox/kn-plugin-func/tree/main/knative

salaboy:
@lance @zroubalik @lkingland can we merge this one now?

gabo1208:
Now this should be the codecov chore link https://app.codecov.io/gh/knative/func

lance:
@salaboy we are now at 64% test coverage on the knative package. This puts us well into the "Usable" state as defined here: https://github.com/knative/community/pull/1157

I think we want to continue improving this to get to "Stable" (70%) but for the purposes of this issue, I think we can close it now and if we have a specific target (like 70%) we can create a new issue for that. It might be good, actually to have an Epic similar to the [on-cluster build epic](https://github.com/knative/func/issues/620), with subtasks to track specific goals for specific packages.

Feel free to reopen this if you disagree and would like to discuss.

PR #1338, opened 2022-10-14
test: initial integration test for knative related functionality

matejvasek:
# Changes

- :test_tube: Initial integration tests for `knative` related functionality.

Related: https://github.com/knative/func/issues/1278

### DIFF
--- a/knative/integration_test.go
+++ b/knative/integration_test.go
@@ -45,6 +45,11 @@ func TestIntegration(t *testing.T) {
 	t.Cleanup(func() { _ = cliSet.CoreV1().Namespaces().Delete(ctx, namespace, metav1.DeleteOptions{}) })
 	t.Log("created namespace: ", namespace)
 
+	minScale := int64(3)
+	maxScale := int64(100)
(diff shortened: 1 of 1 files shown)


matejvasek:
@dolfolife PTAL

knative/integration_test.go
```diff
@@ -0,0 +1,157 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	fn "knative.dev/func"
+	"knative.dev/func/k8s"
+	"knative.dev/func/knative"
+)
+
+func TestIntegration(t *testing.T) {
```
lance, line 23: This looks great! Looks like this change is great for our coverage. But would it not be better in the long run to have individual tests for deploy, describe and list? I know that means deploying three separate times, so maybe it's overkill.

matejvasek: I want to test that all the operation works together.

Issue #1321, opened 2022-10-12
docs rebuild encodes local environment


lkingland:
Problem:
   When docs are regernerated, a user's current Function config path is used, leading to the docs potentially containing user settings.  For example, if repositories `functastic` and `knative` are installed in ~/.func/repositories, the resultant `docs/reference/func_create.md` would have the following in its diff after docs regeneration on `make`:
```diff
  --------     --------
	  go           cloudevents
	  go           http
+	  go           functastic/improve
+	  go           functastic/uppercase
+	  go           knative/improve
+	  go           knative/uppercase
...
```
Short-term Solution:
The same solution used in tests to avoid this should be workable here:  set `XDG_CONFIG_HOME` to an empty temp directory before docs generation, then set it back in a defer.

Longer-term Solution:
A possible longer term solution is to implement the (see [FUNC_HOME environment variable #1320](https://github.com/knative/func/issues/1320)), which will require a small refactor of Client.

mcruzdev:
Hi @lance, can you assign it for me?

lance:
/assign @mcruzdev 
Thanks for the offer!

mcruzdev:
Hello @lkingland, how can I simulate it?

mcruzdev:
Hi @lkingland, how can I simulate it?

Can you help me with step by step?

lkingland:
@lance I can't seem to reproduce the problem on my system either. I remember it last happened with one of your rebuilds. Can you also confirm this is no longer happening? Perhaps we can close this issue as hopefully no longer being relevant

lance:
@mcruzdev @lkingland this is odd - I can't reproduce this now either. Looking at the [generator history](https://github.com/knative/func/commits/main/docs/generator) there is nothing to indicate that this was somehow fixed. 

```
❯ func templates
LANGUAGE     TEMPLATE
go           cloudevents
go           http
go           knative/improve
go           knative/uppercase
node         cloudevents
node         http
nodejs       knative/metacontroller
python       cloudevents
python       http
python       knative/numpy
quarkus      cloudevents
quarkus      http
rust         cloudevents
rust         http
springboot   cloudevents
springboot   http
springboot   knative/uppercase
typescript   cloudevents
typescript   http

func on  main [$!] via 🐹 v1.19.1 on ☁️  lanceball@gmail.com
❯ make docs
# Generating command reference doc
go run docs/generator/main.go

[...]

PR #1344, opened 2022-10-17
chore: global config file paths

lkingland:
- :bug: config paths on disk only created when necessary
- :gift: Adds FUNC_CONFIG_PATH environment variable

/kind enhancement

Related: #901

### DIFF
--- a/client.go
+++ b/client.go
@@ -11,10 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
-	"runtime/debug"
 	"time"
-
-	"knative.dev/func/config"
 )
 
 const (
@@ -202,7 +199,6 @@ func New(options ...Option) *Client {
 		dnsProvider:       &noopDNSProvider{output: os.Stdout},
 		progressListener:  &NoopProgressListener{},
 		pipelinesProvider: &noopPipelinesProvider{},
-		repositoriesPath:  filepath.Join(config.Path(), "repositories"),
 		transport:         http.DefaultTransport,
 	}
 	for _, o := range options {
@@ -214,20 +210,13 @@ func New(options ...Option) *Client {
 	c.templates = newTemplates(c)
 	c.instances = newInstances(c)
 
-	// Trigger the creation of the config and repository paths
-	_ = config.Path()
-	_ = c.RepositoriesPath() // Repositories is Client-specific
-
 	return c
 }
 
 // RepositoriesPath accesses the currently effective repositories path,
-// which defaults to [ConfigPath]/repositories but can be set explicitly using
-// the WithRepositoriesPath option when creating the client..
-// The path will be created if it does not already exist.
+// which can be set using the WithRepositoriesPath option.
 func (c *Client) RepositoriesPath() (path string) {
 	path = c.repositories.Path()
-	mkdir(path) // make sure it exists
 	return
 }
 
@@ -1046,14 +1035,3 @@ func (p *NoopProgressListener) Increment(m string) {}
 func (p *NoopProgressListener) Complete(m string)  {}
 func (p *NoopProgressListener) Stopping()          {}
 func (p *NoopProgressListener) Done()              {}
-
-// mkdir attempts to mkdir, writing any errors to stderr.
-func mkdir(path string) {
-	// Since it is expected that the code elsewhere never assume directories
-	// exist (doing so is a racing condition), it is valid to simply
-	// handle errors at this level.
-	if err := os.MkdirAll(path, 0700); err != nil {
-		fmt.Fprintf(os.Stderr, "Error creating '%v': %v", path, err)
-		debug.PrintStack()
-	}
-}
--- a/client_test.go
+++ b/client_test.go
@@ -51,35 +51,6 @@ func TestClient_New(t *testing.T) {
 	}
 }
 
-// TestClient_InstantiationCreatesRepositoriesPath ensures that instantiating the
-// client has the side-effect of ensuring that the repositories path exists
-// on-disk, and also confirms that the XDG_CONFIG_HOME environment variable is
-// respected when calculating this home path.
-func TestClient_InstantiationCreatesRepositoriesPath(t *testing.T) {
-	root := "testdata/example.com/testNewCreatesRepositoriesPath"
-	defer Using(t, root)()
-
-	rootAbs, err := filepath.Abs(root) // absolute path to root
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	// Instruct the system to use the above test root directory as the home dir.
-	t.Setenv("XDG_CONFIG_HOME", rootAbs)
-
-	// The expected full path to the repositories should be:
-	expected := filepath.Join(rootAbs, "func", "repositories")
-
-	// Create a new client, which should perform initialization checks such as
-	// ensuring the full path to repositories exists.
-	_ = fn.New(fn.WithRegistry(TestRegistry))
-
-	// Confirm that the repositories path exists.
-	if _, err := os.Stat(expected); err != nil {
-		t.Fatal(err)
-	}
-}
-
 // TestClient_New_RuntimeRequired ensures that the the runtime is an expected value.
 func TestClient_New_RuntimeRequired(t *testing.T) {
 	// Create a root for the new function
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -7,6 +7,7 @@ import (
 
 	fn "knative.dev/func"
 	"knative.dev/func/buildpacks"
+	"knative.dev/func/config"
 	"knative.dev/func/docker"
 	"knative.dev/func/docker/creds"
 	fnhttp "knative.dev/func/http"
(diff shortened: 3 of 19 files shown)


client_test.go
```diff
@@ -51,35 +51,6 @@ func TestClient_New(t *testing.T) {
 	}
 }
 
-// TestClient_InstantiationCreatesRepositoriesPath ensures that instantiating the
-// client has the side-effect of ensuring that the repositories path exists
-// on-disk, and also confirms that the XDG_CONFIG_HOME environment variable is
-// respected when calculating this home path.
-func TestClient_InstantiationCreatesRepositoriesPath(t *testing.T) {
```
lkingland, line 58: This test was reloacted to both config (testing that CreatePaths works as expected) and the cli (ensuring add repositories works requires the path exist)

cmd/client.go
```diff
@@ -65,15 +66,16 @@ func NewClientFactory(n func() *fn.Client) ClientFactory {
 //	defer done()
 func NewClient(cfg ClientConfig, options ...fn.Option) (*fn.Client, func()) {
 	var (
-		p  = progress.New(cfg.Verbose)            // updates the CLI
-		t  = newTransport(cfg.InsecureSkipVerify) // may provide a custom impl which proxies
-		c  = newCredentialsProvider(t)            // for accessing registries
+		p  = progress.New(cfg.Verbose)                // updates the CLI
+		t  = newTransport(cfg.InsecureSkipVerify)     // may provide a custom impl which proxies
+		c  = newCredentialsProvider(config.Path(), t) // for accessing registries
```
lkingland, line 71: Passing in the path both removes the implicit read of an environment variable from non-main, and avoids a circular dependency with config.

cmd/create.go
```diff
@@ -239,7 +239,9 @@ func newCreateConfig(cmd *cobra.Command, args []string, newClient ClientFactory)
 	// it is still available as an environment variable.
 	repositoriesPath = os.Getenv("FUNC_REPOSITORIES_PATH")
 	if repositoriesPath == "" { // if no env var provided
-		repositoriesPath = fn.New().RepositoriesPath() // use ~/.config/func/repositories
+		client, done := newClient(ClientConfig{})
```
lkingland, line 242: Can't use the default client here, because we want to (indirectly) use the path provided to Client via `config.RepositoriesPath`

config/config.go
```diff
@@ -88,7 +107,10 @@ func (c Config) Save(path string) (err error) {
 // 3.  The value of $XDG_CONFIG_PATH/func if the environment variable exists.
 // The path is created if it does not already exist.
 func Path() (path string) {
-	path = DefaultConfigPath
+	// default path is a relative path used in the unlikely event that
+	// the user has no home directory (no ~), there is no
+	// XDG_CONFIG_HOME set
+	path = filepath.Join(".config", "func")
```
lkingland, line 113: This is a bit of an arbitrary choice.  We might want to consider having the deafult path be simply "." in this admitedly quite edge edge case

client.go
```diff
@@ -202,7 +199,6 @@ func New(options ...Option) *Client {
 		dnsProvider:       &noopDNSProvider{output: os.Stdout},
 		progressListener:  &NoopProgressListener{},
 		pipelinesProvider: &noopPipelinesProvider{},
-		repositoriesPath:  filepath.Join(config.Path(), "repositories"),
```
lkingland, line 205: Repository path is now entirely optional from the standpoint of the core library.  When not provided, it simply presumes all repository-related operations are with builtin repositories only, with the exception of things like "add", which will of course error with a message.  This allevaites the problem where, in some circumstances, the client would start creating paths on disk when not necessary.

docker/creds/credentials_test.go
```diff
@@ -576,7 +577,7 @@ func withPopulatedFuncAuthConfig(t *testing.T) {
 
 	var err error
 
-	authConfig := filepath.Join(config.Path(), "auth.json")
+	authConfig := filepath.Join(testConfigPath(t), "auth.json")
```
lkingland, line 580: This change to docker/creds necessary to resolve a circular dependency

docker/creds/credentials_test.go
```diff
@@ -643,20 +644,33 @@ func correctVerifyCbk(ctx context.Context, image string, credentials Credentials
 	return creds.ErrUnauthorized
 }
 
-func withCleanHome(t *testing.T) {
-	t.Helper()
-	homeName := "HOME"
+func testHomeEnvName(t *testing.T) string {
```
matejvasek, line 647: Why does this take `*testing.T`?

config/config.go
```diff
@@ -88,7 +107,10 @@ func (c Config) Save(path string) (err error) {
...
 
 	// ~/.config/func is the default if ~ can be expanded
 	if home, err := homedir.Expand("~"); err == nil {
```
matejvasek, line 116: I wonder, why do we need this sophisticated method? Wouldn't `os.UserHomeDir()` suffice?

docker/creds/credentials_test.go
```diff
@@ -643,20 +644,33 @@ func correctVerifyCbk(ctx context.Context, image string, credentials Credentials
...
 	tmpHome := t.TempDir()
-	t.Setenv(homeName, tmpHome)
+	t.Setenv(testHomeEnvName(t), tmpHome)
 
 	if runtime.GOOS == "linux" {
 		t.Setenv("XDG_CONFIG_HOME", filepath.Join(tmpHome, ".config"))
 	}
 }
 
+func testConfigPath(t *testing.T) string {
```
matejvasek, line 664: I believe this should return the same value as `config.Path()` most of the times, doesn't it?

config/config.go
matejvasek: Was there problem in some containerised contexts?

config/config.go
```diff
@@ -88,7 +107,10 @@ func (c Config) Save(path string) (err error) {
 // 3.  The value of $XDG_CONFIG_PATH/func if the environment variable exists.
 // The path is created if it does not already exist.
 func Path() (path string) {
```
matejvasek, line 109: Wouldn't be better name `ConfigDir()`, from the name it's not clear if it returns path to global config, or dir containing it.

docker/creds/credentials_test.go
matejvasek: Hmm actually not, because `homedir` pkg is caching. Why do we use that package and not just `os.UserHomeDir()`.

config/config_test.go
```diff
@@ -92,3 +100,109 @@ func TestNewDefault(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	if err := cfg.Write(config.ConfigPath()); err != nil {
+		t.Fatal(err)
+	}
+}
+
+// TestDefaultNamespace ensures that, when thre is a problem determining the
+// active namespace, the static DefaultNamespace ("default") is used.
+func TestDefaultNamespace(t *testing.T) {
```
matejvasek, line 200: Might be nice to also have a test with mocked `KUBECONFIG` containing `ns` set.

docker/creds/credentials_test.go
matejvasek: Or we could disable caching in tests, `homedir.DisableCache = true`.

matejvasek:
@lkingland 
I tried to clean up my configs (both `func`'s and `docker`'s) and build&push imge:
```
Error: failed to get credentials: faild to set the helper to the config: open /home/mvasek/.config/func/auth.json: no such file or directory
```
The config dir `~/.config/func/` is probably not created automatically.

matejvasek:
@lkingland 
```
 func templates --json
```
Now prints only embedded templates not the ones I have in `~/.config/func/repositories`.
Is that really what we want?

config/config_test.go
```diff
@@ -92,3 +100,109 @@ func TestNewDefault(t *testing.T) {
...
+
+	if _, err := os.Stat(filepath.Join(config.Path(), "repositories")); err != nil {
+		if os.IsNotExist(err) {
+			t.Fatalf("config path '%v' not created", config.Path())
+		}
+		t.Fatal(err)
+	}
+
+	// Trying to create when repositories path is invalid should error
+	_ = os.WriteFile("./invalidRepositoriesPath.txt", []byte{}, os.ModePerm)
```
matejvasek, line 130: Won't this create garbage in working dir?

config/config.go
lkingland: Making the name imply if it is a filename or a directory sounds good to me. `config.ConfigDir()` is a bit redundant.  How about `config.Dir()` and `config.Filename()`?

config/config.go
lkingland: os.UserHomDir sounds like a better choice to me as well.  Changed.

config/config_test.go
lkingland: Yes, which is the intention:  to create a garbage file which trggers an error when attempting to create the path.
It might be worth noting that this all takes place within a temp directory specifically to give the test a sandbox to play in 👍🏻

config/config_test.go
lkingland: Yes, apparently I agreed with you already.  This test is much [more fully fleshed out](https://github.com/knative/func/pull/1353/files#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82R201) in #1353.  Let me know if that's not what you were thinking.

docker/creds/credentials_test.go
lkingland: For the `t.Helper(t)` call which was missing... and now is not.

lkingland:
```
templates --json
```
now respects the global path for repositories.

lkingland:
> The config dir ~/.config/func/ is probably not created automatically.

Yes, I had to patch the func client in a few places as well, and I must have missed one here in the Docke Credentials check code.  A nonexistent config directory/file should be nonfatal.

Can you help me understand why, when using an entirely empty home (and config) directory, the docker/creds unit tests succeed, as do the E2Es?  Is this perhaps a problem only when we are both performing an integration test AND the `auth.json` path does not exist?

matejvasek:
@lkingland I tried to populate three values it the config:
```
cat ~/.config/func/config.yaml                                                                                                                                                  
registry: quay.io/mvasek
builder: s2i
verbose: true
```
None of them seems to really apply, it that something to be done?

matejvasek:
> @lkingland I tried to populate three values it the config:
> 
> ```
> cat ~/.config/func/config.yaml                                                                                                                                                  
> registry: quay.io/mvasek
> builder: s2i
> verbose: true
> ```
> 
> None of them seems to really apply, it that something to be done?

On related note:
You will probably replace calls to `GetDefaultRegistry()` with access to the config (`GetDefaultRegistry()` will be removed).
I will have to rework how we are defaulting to OpenShift internal registry, but I am thinking about it for some times and we need it anyway.

config/config.go
matejvasek: It think that's good.

lkingland:
Thanks Matej, those are both is as expected.  

This PR fixes paths associated with moving logic into the config package, but stops short of actually incorporating new values into the CLI.  THat includes all the values you mentioned and the method GetDefaultRegistry.  This PR is focused on getting file paths right, fixing circular dependencies etc.

This was an attempt to keep the size of the Global Config PRs down.  I shoud probably have split this one down _even more_ and not added those members to the struct, but I wated the reader to get a hint of how this fits in th larger picture.

Thanks for the review 👍🏻

matejvasek (approved):

lkingland:
I believe this also fixes #1321

PR #1352, opened 2022-10-18
feat: namespace globally configurable

lkingland:
:gift: default namespace globally configurable
:gift:  namespace flag help more contetually relavant
:broom:  removes namespace from commands where it is not applicable

Moves namespace derivation logic to global config and prepares commands for integration by defining namespace directly on each.

Related: #901 

/kind enhancement

```release-notes
* default namespace is now globally configurable
* namespace flag help is more contextually relevant
* removes the namespace flag from commands where it is not applicable
```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -729,9 +729,13 @@ func namespace(cfg deployConfig, f fn.Function, stderr io.Writer) (namespace str
 	} else if f.Deploy.Namespace != "" {
 		namespace = f.Deploy.Namespace // value from previous deployment (func.yaml) 2nd priority
 	} else {
-		// Try to derive a default from the current k8s context, if any.
-		if namespace, err = k8s.GetNamespace(""); err != nil {
-			fmt.Fprintln(stderr, "Warning: no namespace provided, and none currently active. Continuing to attempt deployment")
+		// If global config setting exists, use that, followed by the active
+		// k8s namesapce if not set.
+		gc, err := config.NewDefault()
+		if err != nil {
+			fmt.Fprintf(stderr, "Warning: error reading global config.%v", err)
+		} else {
+			namespace = gc.Namespace
 		}
 	}
 
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -733,7 +733,7 @@ func namespace(cfg deployConfig, f fn.Function, stderr io.Writer) (namespace str
 		// k8s namesapce if not set.
 		gc, err := config.NewDefault()
 		if err != nil {
-			fmt.Fprintf(stderr, "Warning: error reading global config.%v", err)
+			fmt.Fprintf(stderr, "warning: error reading global config.%v", err)
 		} else {
 			namespace = gc.Namespace
 		}
--- a/cmd/info.go
+++ b/cmd/info.go
@@ -12,6 +12,7 @@ import (
 	"gopkg.in/yaml.v2"
 
 	fn "knative.dev/func"
+	"knative.dev/func/config"
 )
 
 func NewInfoCmd(newClient ClientFactory) *cobra.Command {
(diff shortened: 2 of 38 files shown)


config/config.go
```diff
@@ -7,77 +7,96 @@ import (
...
+	switch {
+	case openshift.IsOpenShift():
+		return openshift.GetDefaultRegistry()
+	default:
+		return ""
+	}
+}
+
+// DefaultNamespace for remote operations is the currently active
+// context namespace (if available) or the fallbacl "default".
```
lance, line 39: Nit: (surprised reviewdog did not catch this)
```suggestion
// context namespace (if available) or the fallback "default".
```

config/config.go
```diff
@@ -100,21 +122,43 @@ func Path() (path string) {
...
+	path := filepath.Join(Path(), Filename)
+	if e := os.Getenv("FUNC_CONFIG_FILE"); e != "" {
+		path = e
+	}
+	return path
+}
 
-	// usually ~/.config/func/config.yaml
-	return filepath.Join(Path(), Filename)
+// RepositoriesPath returns the full at which to look for repositories.
```
lance, line 138: Nit:
```suggestion
// RepositoriesPath returns the full path at which to look for repositories.
```

config/config_test.go
```diff
@@ -20,15 +21,22 @@ func TestNewDefaults(t *testing.T) {
...
 	}
+
+	// and ensure error
+	cfg, err = config.Load("invalid/path")
+	if err == nil {
+		t.Fatal("did not receive expected error loading nonexistent config path")
+	}
 }
 
 // TestSave ensures that saving an update config persists.
```
lance, line 42: Nit: 
```suggestion
// TestWrite ensures that saving an update config persists.
```

lance (commented):
Looks good @lkingland! Thank you. I only have a couple of very small nits.

lance:
@lkingland is this meant to be a draft PR?

lkingland:
> @lkingland is this meant to be a draft PR?

Yes, it is dependent on 1344, and will need to be rebased when that is merged

cmd/deploy.go
```diff
@@ -729,9 +729,13 @@ func namespace(cfg deployConfig, f fn.Function, stderr io.Writer) (namespace str
...
 		namespace = f.Deploy.Namespace // value from previous deployment (func.yaml) 2nd priority
 	} else {
-		// Try to derive a default from the current k8s context, if any.
-		if namespace, err = k8s.GetNamespace(""); err != nil {
-			fmt.Fprintln(stderr, "Warning: no namespace provided, and none currently active. Continuing to attempt deployment")
+		// If global config setting exists, use that, followed by the active
+		// k8s namesapce if not set.
+		gc, err := config.NewDefault()
+		if err != nil {
+			fmt.Fprintf(stderr, "Warning: error reading global config.%v", err)
```
lance, line 736: Teeny tiny nit - could be handled in a follow up PR. Adds a space between the warning and the error, and lowercases the first letter (iirc, that's what we are doing across the codebase).
```suggestion
			fmt.Fprintf(stderr, "warning: error reading global config. %v", err)
```

cmd/info.go
```diff
@@ -35,7 +36,15 @@ the current directory or from the directory specified with --path.
...
 
+	// Config
+	cfg, err := config.NewDefault()
+	if err != nil {
+		fmt.Fprintf(cmd.OutOrStdout(), "error loading config at '%v'. %v\n", config.File(), err)
+	}
+
+	// Flags
 	cmd.Flags().StringP("output", "o", "human", "Output format (human|plain|json|xml|yaml|url) (Env: $FUNC_OUTPUT)")
+	cmd.Flags().StringP("namespace", "n", cfg.Namespace, "The namespace in which to look for the named function. (Env: $FUNC_NAMESPACE)")
```
lance, line 47: Why is there not similar logic as in `cmd/deploy.go` - here and in the other commands (`delete` and `list`)?

cmd/root_test.go
```diff
@@ -19,22 +19,29 @@ import (
...
-		name          string
-		args          []string
-		skipNamespace bool
+		name     string
+		args     []string
+		expected bool
 	}{
 		{
-			name: "provided as root flags",
-			args: []string{"--verbose", "--namespace=namespace", "list"},
```
lance, line 28: Why remove the `--namespace` flag here and below? It seems like this might be removing behavior we want to test, unless I'm reading this incorrectly (always a possibility!).

lance (commented):
Overall looks good. I have a small nit and a few questions.

cmd/root_test.go
lkingland: Good question.  You can see from the handfull of other open PRs I am adding more deliberate usage and tests to the commands which utilize namespace as well.

I am systematically moving away from globally shared flags for a couple of reasons:  first, the help text is much better off being custom tailored to the command for which the flag applies.  Second, it is a bit complicated to test the flag is being used properly, because tests have to first instantiate a root command object.  Finally, since the namespace really isn't applicable to every command, it seems sloppy to have it included in the help text.

I can see two flags beng truly globally shared:  "verbose" and "output".  The others seem like more confusion than it's worth.

cmd/info.go
lkingland: This line was incorrectly included in this PR, thanks for noticing.  I separated out the namespace usage (and tests) each into their own separate PR for easier review.

lkingland:
Rebased to resolve merge conflict

PR #1354, opened 2022-10-18
fix: remove ldflags from docs generation make target

lkingland:
Removes ldflags from docs generation make command as they are not used.

Their existence causes a conflict with the related PR which moves --path flag processing into init, causing a (nonfatal but anoying) error that ldflags were used but not defined.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ clean: clean_templates ## Remove generated artifacts such as binaries and schema
 
 docs:
 	# Generating command reference doc
-	go run docs/generator/main.go -ldflags $(LDFLAGS)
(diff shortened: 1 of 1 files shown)


lance:
@lkingland running failed e2e tests again, but otherwise looks good. Curious if you meant to create this as a draft PR.

/lgtm

lance:
> @lkingland running failed e2e tests again, but otherwise looks good. Curious if you meant to create this as a draft PR.
> 
> /lgtm

I'm assuming that you didn't mean to make this WIP. I'm going to unhold it and if it really was a WIP, we can revert it.

/unhold

PR #1357, opened 2022-10-18
test: extended integration test for deployment

matejvasek:
# Changes

- :test_tube: Extended knative deployment integration test: added more tests for envvars and volumes.
- :bug: Fixed race-condition.


lance:
/lgtm

This is great - thanks @matejvasek !

PR #1373, opened 2022-10-21
fix: function constructor errors on valid path

### LINKED ISSUE #1371: function constructor errors on invalid function
The `NewFunction(path)` constructor should only error when the path is invalid, but is currently erroring when the path defined does not contain a function.  This appears to be a regression introduced in a recent pr, so the fix should include a test case such that it doesn't crop back in again.

In particular:
```
f, err := fn.NewFunction(invalidPath) // should error if path does not exist or is a file
```
```
f, err := fn.NewFunction(emptyDir) // should not error, but f.Initialized() == false 
```

lkingland:
:bug: NewFunction errors on valid path

The NewFunction constructor should only error on invalid paths; not when the given path has not yet been initialized.
Fixes #1371 

```release-notes
fixes an issue where the NewFunction constructor could error if the path has not yet been initialized as a function
```

### DIFF
--- a/function.go
+++ b/function.go
@@ -173,10 +173,27 @@ func NewFunction(path string) (f Function, err error) {
 	f.Root = path // path is not persisted, as this is the purview of the FS itself
 	f.Build.BuilderImages = make(map[string]string)
 	f.Deploy.Annotations = make(map[string]string)
+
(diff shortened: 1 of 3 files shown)


lance (commented):
/lgtm

PR #1381, opened 2022-10-24
feat: describe command namespace

lkingland:
- :gift: describe function in given namespace
- :bug: describe function by name _or_ at path
- :broom: fix inconsistency with `client.Describer`
- :broom: add tests

Added ability to specify a namespace in the same manner as `list` and `delete`.
In order to be more consistent with `kubectl describe` and the client's `Describer` interface, the command was renamed `describe` with aliases for both `desc` and `info`.

/kind techdebt

### DIFF
--- a/cmd/describe_test.go
+++ b/cmd/describe_test.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"path/filepath"
 	"testing"
 
 	fn "knative.dev/func"
@@ -94,7 +95,8 @@ func TestDescribe_Namespace(t *testing.T) {
 
 	client := fn.New(fn.WithDescriber(mock.NewDescriber()))
 
-	// Ensre that the default is "default" when no context can be identified
+	// Ensure that the default is "default" when no context can be identified
+	t.Setenv("KUBECONFIG", filepath.Join(cwd(), "nonexistent"))
 	cmd := NewDescribeCmd(func(cc ClientConfig, _ ...fn.Option) (*fn.Client, func()) {
 		if cc.Namespace != "default" {
 			t.Fatalf("expected 'default', got '%v'", cc.Namespace)
--- a/cmd/describe_test.go
+++ b/cmd/describe_test.go
@@ -25,7 +25,6 @@ func TestDescribe_ByName(t *testing.T) {
 	cmd := NewDescribeCmd(NewClientFactory(func() *fn.Client {
 		return fn.New(fn.WithDescriber(describer))
 	}))
-
 	cmd.SetArgs([]string{testname})
 	if err := cmd.Execute(); err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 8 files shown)


cmd/describe_test.go
```diff
@@ -0,0 +1,142 @@
...
+	describer.DescribeFn = func(n string) (i fn.Instance, err error) {
+		if n != "testname" {
+			t.Fatalf("expected describer to receive name 'testname', got '%v'", n)
+		}
+		return
+	}
+	cmd := NewDescribeCmd(NewClientFactory(func() *fn.Client {
+		return fn.New(fn.WithDescriber(describer))
+	}))
+	if err := cmd.Execute(); err != nil {
```
matejvasek, line 65:
```suggestion
	}))
	cmd.SetArgs([]string{})
	if err := cmd.Execute(); err != nil {
```

cmd/describe_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+}
+
+// TestDescribe_Namespace ensures that the namespace provided to the client
+// for use when describing a function is set
+// 1. The flag /env variable if provided
+// 2. The namespace of the function at path if provided
+// 3. The user's current active namespace
+func TestDescribe_Namespace(t *testing.T) {
+	root := fromTempDirectory(t)
```
matejvasek, line 93:
```suggestion
func TestDescribe_Namespace(t *testing.T) {
	t.Setenv("KUBECONFIG", "/tmp/non-existent")
	root := fromTempDirectory(t)
```

cmd/describe_test.go
```diff
@@ -0,0 +1,142 @@
...
+	if err := client.Create(f); err != nil {
+		t.Fatal(err)
+	}
+	cmd = NewDescribeCmd(func(cc ClientConfig, _ ...fn.Option) (*fn.Client, func()) {
+		if cc.Namespace != "deployed" {
+			t.Fatalf("expected 'deployed', got '%v'", cc.Namespace)
+		}
+		return client, func() {}
+	})
+	if err := cmd.Execute(); err != nil {
```
matejvasek, line 126:
```suggestion
	})
	cmd.SetArgs([]string{})
	if err := cmd.Execute(); err != nil {
```

matejvasek:
There is a lint issue.

cmd/describe_test.go
lkingland: Ignoring current user's KUBECONFIG and clearing args from IDEs are now being addressed in separate PRs

cmd/describe_test.go
lkingland: Ignoring current user's KUBECONFIG and clearing args from IDEs are now being addressed in separate PRs

lkingland:
Thanks for the code review @matejvasek ; I think this one is ready for another look.

matejvasek:
@lkingland 
```
expected 'default', got 'func'
```
Assert need adjustment and/or you need also rebase.

matejvasek:
@lance PTAL

PR #1383, opened 2022-10-25
fix: do not upload .git/ for on cluster build

matejvasek:
# Changes

- :gift: Do not upload `.git/` directory when doing on cluster build.

/kind enhancement


matejvasek:
@jrangelramos PTAL

lance (approved):
/lgtm

PR #1385, opened 2022-10-26
[WIP] rsync on volume

matejvasek:
# Changes

- :gift: Added function that allows to run rsync daemon on volume in cluster and expose it locally using `net.Listener`.
  This can be useful for more efficient source push for on-cluster-build.

/kind enhancement


matejvasek:
Interesting stuff is in https://github.com/knative/func/pull/1385/commits/3ce37542b4ec6a0fe9b90a96f1390fedb9185f6e

matejvasek:
Still very much WIP.

matejvasek:
@zroubalik you may find this interesting.

zroubalik:
@matejvasek this is cool. My initial idea was to run this from within a Pipeline task, not a temporary pod, so the whole workflow is in sync with the rest of the pipeline. WDYT?

matejvasek:
@zroubalik 
I prefer temporary pod so it is under my control.
Creating it from pipeline would be more complicated.

k8s/persistent_volumes.go
```diff
@@ -224,3 +238,81 @@ func (t *tsBuff) Write(p []byte) (n int, err error) {
...
+	handleRsyncConnection := func(conn net.Conn) error {
+		defer conn.Close()
+		errOut := &tsBuff{}
+		err := exec(client.CoreV1().RESTClient(), restConf,
+			podName, namespace,
+			[]string{"socat", "-", "EXEC:rsync --daemon"},
+			conn, conn, errOut,
+		)
+		if err != nil {
+			return fmt.Errorf("errow while exec (err out: %q): %w", errOut.String(), err)
```
lkingland, line 295:
```suggestion
			return fmt.Errorf("error while exec (err out: %q): %w", errOut.String(), err)
```

k8s/persistent_volumes.go
```diff
@@ -224,3 +238,81 @@ func (t *tsBuff) Write(p []byte) (n int, err error) {
...
+	}
+
+	var eg errgroup.Group
+	for {
+		conn, err := listener.Accept()
+		if err != nil {
+			if errors.Is(err, net.ErrClosed) {
+				break
+			}
+			return fmt.Errorf("cannot accpet connection: %w", err)
```
lkingland, line 307:
```suggestion
			return fmt.Errorf("cannot accept connection: %w", err)
```

k8s/persistent_volumes_test.go
```diff
@@ -143,3 +145,140 @@ func TestUploadToVolume(t *testing.T) {
...
+
+	go func() {
+		err = k8s.ServeRsyncOnVolume(ctx, l, testingPVCName, testingNS)
+		if err != nil {
+			t.Error(err)
+		}
+	}()
+
+	cmd := exec.Command("rsync",
+		filepath.Join("testData", "content", "hello.txt"),
```
lkingland, line 208:
```suggestion
		filepath.Join("testdata", "content", "hello.txt"),
```
Since go tool treats the exact name `testdata` as special, this may be best renamed.  From `go help packages`:
> Directory and file names that begin with "." or "_" are ignored
> by the go tool, as are directories named "testdata".
I am pretty sure it's case-sensitive?

lkingland (approved):
Looks quite useful.  Zbynek makes a good point, but I do like how straightforward this is.
/lgtm (though maybe I should wait until not a draft to approve?)
/hold for two spelling and one directory rename

k8s/persistent_volumes_test.go
matejvasek: >I am pretty sure it's case-sensitive

@lkingland  I am not going to argue, but how that would work on non-case sensitive OS?

matejvasek: like everything is lower by default?

matejvasek: from Go source it looks like it is case-sensitive

PR #1387, opened 2022-10-28
fix: KUBECONFIG unit test interference

lkingland:
- Clear environmental KUBECONFIG when running CLI unit tests


cmd/root_test.go
```diff
@@ -295,13 +295,17 @@ func piped(t *testing.T) func() string {
...
+// an environment clean of developer's settings for use during CLI testing.
+// - sets KUBECONFIG to an empty temp directory
+// - sets XDG_CONFIG_HOME to an empty temp directory
+// - Creates and changes context to an empty temp directory
+// On test Cleanup:
+// - changes context back to the original directory
+// - resets viper (the reason this is "cli-specific")
 func fromTempDirectory(t *testing.T) string {
 	t.Helper()
+	t.Setenv("KUBECONFIG", t.TempDir())
```
matejvasek, line 308: This should be a file not a directory, and it would be nice if the file existed or was minimally populated.

matejvasek: Also it should be well populated for integration tests.

matejvasek: Otherwise it cannot connect to cluster.

matejvasek: So I wouldn't put it here, only at places where it's useful.

matejvasek: Maybe cluster allocate script could create separate kube config like `~/.kube/func-test-config`,
or we could use dedicated `context` that would be in standard config.

lkingland: At time of the above comments, this was a WIP Draft, so they're not entirely pertinent, but two things perhaps worth noting:
Yes, KUBECONFIG has to be set to something, or else the current user's ~/.kube/config will be used, which can also disrupt tests. 
This change affects unit tests only.
Resolving this thread (we can discuss the latest revision)

PR #1392, opened 2022-10-31
feat: registry globally configurable

lkingland:
:gift: registry globally configurable

Deferred evaluation of cluster type is preserved by making registry a helper in the `cmd` package and a calculated field of the global config struct.

/kind enhancement

```release-notes
Image registry is now globally configurable
```


matejvasek:
@lance PTAL

PR #1396, opened 2022-11-01
test: fix tests when running in cluster

matejvasek:
# Changes

- :test_tube: Resetting `KUBECONFIG` environment variable is not sufficient if test runs in cluster.
  We also must unset `KUBERNETES_SERVICE_HOST` environment variable, so in-cluster-config won't kick in.


lkingland (approved):

lance (commented):
/lgtm

PR #1397, opened 2022-11-01
chore(OWNERS): update to new org

lance:
When we moved to knative org, the groups changed. This addresses that.

/kind chore


psschwei (commented):
/approve

PR #1401, opened 2022-11-01
fix: uploading of project containing symlink for on-cluster-build

matejvasek:
# Changes

- :bug: Fix uploading of project containing symlink for on-cluster-build.

/kind bug

```release-note
fix: uploading of project containing symlink for on-cluster-build
```


matejvasek:
@lance PTAL

lance (commented):
/lgtm

PR #1402, opened 2022-11-02
feat: Persist remote value in func.yaml

### LINKED ISSUE #1176: persisting the remote flag
### Summary

We would like to persist the value of `--remote`, but it clearly does not belong where other persisted values do: on the Function struct.  The remote setting should indeed be persisted, but not as part of function state which will affect any environment which checks out function source code (ex: CI/CD environments).  Instead, this flag's persistence should be scoped locally:  to a single copy of a function (checked out codebase).

### Proposed Solution
Store the value of `--remote` in the persistent, but non-source-controlled `.func` directory.  Subsequent invocations of `deploy` will retain the vale of `--remote`, but not interfere with other developers' settings.

### Further Discussion

In PR #1079 and its associated issue #1075 we propose creating a dedicated flag (`--remote`) to indicate if a build/deploy should be handled locally (the default) or via the invocation of a remote Tekton pipeline.

When re-running the `deploy` command, we seem to agree that the values of flags should persist to the function being operated upon, such that subsequent command repetitions do not require re-setting the flags and mutations of the function are propagated through source control.

In most cases, this is a simple serialization of the function struct.  Most of these flags mutate a simple attribute of the Function (adding environment variables, or setting the URL of an associated git repository).  The `--remote` flag, however, is not altering the function's state, but how the `func` _command_ instantiates and invokes implementations of specific interfaces.  It triggers "running in remote mode", but has no effect on the Function itself.  Therefore simply adding a `.Remote` boolean flag to Function to be serialized to disk and stored in source control is not appropriate.

The fact this value has no associated functionality in the core is a further indication that the remote vs local setting is better left unknown to the core and instead used entirely by clients of the core library such as the "cmd" package (func CLI).

This may be illustrated with an example:

Use Case:
-  Alice deploys myFunc (all defaults)
-  Bob checks out myFunc from source control, makes a bug fix and deploys
    using --remote. The function is deployed by triggering a remote Tekton
     task rather than built locally.  This flips the "remote" flag in
    question.
- Bob checks in the new state of the function, which includes this
    remote flag.
- Aice updates their function source code,  makes a small change and runs deploy.
- Bob is now likely confused why the function is not building but instead has triggered some remote task.
- Running deploy again with `--remote=false` results in a dirty git tree.

This illustrates the problem in a way that hopefully shows there are two questions to be answered.  1)  Do we want the func CLI to retain the value of --remote such that subsequent invocations remember this setting?  2)  Do we want one user's choice to change the behavior of the environment of another developer?

By answering yes to the first question but no to the second, we may have a viable middle way.  Yes, we want the choice to  be retained for subsequent command invocations, but perhaps we do _not_ want this setting to be part of the Function's permanent state and to affect other users.  The resultant Function artifact is identical if the build/deploy is run locally or triggered in remote.

Making this distinction also alleviates the aforementioned code-smell of adding a `.Remote` member to the Function struct.  The remote flag would not be part of a function's sate, but would rather be part of the forthcoming Config System (see draft PR).  The value would be persisted to the non-source-controlled metadata directory ".func", which already includes other stateful, but localhost-scoped settings for a function.

vyasgun:
# Changes

- 🎁 Persist the value of `--remote` flag in func.yaml and use the persisted value when the flag hasn't changed

/kind enhancement

Partly addresses #1176 

**Release Note**

```release-note
Persist the value of "remote" field in func.yaml and use it as default
```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -217,6 +217,7 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 		// Sets default AND accepts andy user-provided overrides
 		f.Deploy.Namespace = cfg.Namespace
 	}
+
 	if cmd.Flags().Changed("remote") {
 		f.Deploy.Remote = cfg.Remote
 	} else {
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -217,6 +217,11 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 		// Sets default AND accepts andy user-provided overrides
 		f.Deploy.Namespace = cfg.Namespace
 	}
+	if cmd.Flags().Changed("remote") {
+		f.Deploy.Remote = cfg.Remote
+	} else {
+		cfg.Remote = f.Deploy.Remote
+	}
 	if cfg.Image != "" {
 		f.Image = cfg.Image
 	}
--- a/cmd/deploy_test.go
+++ b/cmd/deploy_test.go
@@ -923,19 +923,14 @@ func TestDeploy_NamespaceRedeployWarning(t *testing.T) {
 // TestDeploy_RemotePersists ensures that the remote field is read from
 // the function by default, and is able to be overridden by flags/env vars.
 func TestDeploy_RemotePersists(t *testing.T) {
-	testRemotePersists(NewDeployCmd, t)
-}
-
-func testRemotePersists(cmdFn commandConstructor, t *testing.T) {
 	t.Helper()
 	root := fromTempDirectory(t)
+	cmdFn := NewDeployCmd
 
 	if err := fn.New().Create(fn.Function{Runtime: "node", Root: root}); err != nil {
 		t.Fatal(err)
 	}
-	cmd := cmdFn(NewClientFactory(func() *fn.Client {
-		return fn.New(fn.WithRegistry(TestRegistry))
-	}))
+	cmd := cmdFn(NewTestClient(fn.WithRegistry(TestRegistry)))
 	cmd.SetArgs([]string{})
 	if err := cmd.Execute(); err != nil {
 		t.Fatal(err)
@@ -945,7 +940,7 @@ func testRemotePersists(cmdFn commandConstructor, t *testing.T) {
 	var f fn.Function
 
 	// Deploy the function, specifying remote deployment(on-cluster)
-	viper.Reset()
+	cmd = cmdFn(NewTestClient(fn.WithRegistry(TestRegistry)))
 	cmd.SetArgs([]string{"--remote"})
 	if err := cmd.Execute(); err != nil {
 		t.Fatal(err)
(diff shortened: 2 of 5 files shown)


cmd/deploy_test.go
```diff
@@ -948,3 +948,74 @@ func TestDeploy_NamespaceRedeployWarning(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	if !f.Deploy.Remote {
+		t.Fatalf("value of remote flag not persisted when provided. Expected '%v' got '%v'", true, f.Deploy.Remote)
+	}
+
+	// Deploy the function again without specifying remote
+	viper.Reset()
+	cmd.SetArgs([]string{})
+	cmd.Flag("remote").Changed = false
```
matejvasek, line 993: Setting this explicitly seems odd. Is it really necessary. Could it be replaced by re-instantiating the `cmd`?

cmd/deploy.go
```diff
@@ -220,6 +220,12 @@ func runDeploy(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 		// Sets default AND accepts andy user-provided overrides
 		f.Deploy.Namespace = config.Namespace
 	}
+
+	if !f.Deploy.Remote || cmd.Flags().Changed("remote") {
```
matejvasek, line 224: Why do we need `!f.Deploy.Remote ||`? Test seems to pass even without it.

cmd/deploy_test.go
```diff
@@ -948,3 +948,74 @@ func TestDeploy_NamespaceRedeployWarning(t *testing.T) {
...
+	// (was not reset back to a flag default)
+	if f, err = fn.NewFunction(root); err != nil {
+		t.Fatal(err)
+	}
+	if !f.Deploy.Remote {
+		t.Fatalf("value of remote flag not persisted when provided. Expected '%v' got '%v'", true, f.Deploy.Remote)
+	}
+
+	// Deploy the function again using a different value for remote
+	viper.Reset()
```
lkingland, line 1008: Same comment as above, just recreate the command.  It might be worth noting that when you rebase, the latest cmd package allows you to:
```go
cmd = NewDeployCmd(NewTestClient(fn.WithRegistry(TestRegistry))
```

lkingland (commented):
Just a few small comments about cleaning up the test, but otherwise looks good.

This PR does not close the referenced issue #1176.   The purpose of that issue is to explain why persisting the --remote flag in the same way we persist the other values is not ideal.  The first sentence:
> We would like to persist the value of --remote, but it clearly does not belong where other persisted values do: on the Function struct.
> [goes on to explain why; in excruciating detail]

This PR persists the value of the --remote flag by adding it to the Function struct.


tl;dr  I do not want _my_ choice to build remotely to affect another developer working on the same function, but I _do_ want my choice to persist between deploy iterations on my own box.

cmd/deploy_test.go
```diff
@@ -948,3 +948,74 @@ func TestDeploy_NamespaceRedeployWarning(t *testing.T) {
...
+	cmd.SetArgs([]string{"--remote"})
+	if err := cmd.Execute(); err != nil {
+		t.Fatal(err)
+	}
+	// Assert the function has persisted the value of remote
+	if f, err = fn.NewFunction(root); err != nil {
+		t.Fatal(err)
+	}
+	if !f.Deploy.Remote {
+		t.Fatalf("value of remote flag not persisted when provided. Expected '%v' got '%v'", true, f.Deploy.Remote)
```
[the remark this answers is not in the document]
vyasgun: I thought it would be helpful to compare the provided value with the persisted value but yes, it's probably not required.

lance:

lkingland:
Thanks @vyasgun ; looks good

PR #1406, opened 2022-11-03
feat: caching for on cluster s2i build

matejvasek:
# Changes

- :gift: Cache maven repository for Java builds.

/kind enhancement, performance

```release-note
feat: caching for on-cluster s2i build
```


matejvasek:
I had to revert buildah caching since it lead up eating to much storage.

PR #1410, opened 2022-11-04
fix: error instantiating empty function

lkingland:
:bug: validate function on write
:broom: adds typed errors for required name, registry and runtime
:broom: defers member requirements until needed

In the spirit of "Be permissive on what is accepted, and strict on what is emitted", function validation now happens on write rather than load, defers as much as possible until needed, and moves closer towards validating syntax, not state.

This fixes a bug where a function missing members was unable to be loaded (and thus fixed), forcing a manual edit of func.yaml

/kind bug

```release-notes
Fixes a scenario where the CLI is unable to load invalid func.yaml files and repair them in memory before rewriting
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -59,6 +59,12 @@ type Client struct {
 // ErrNotBuilt indicates the function has not yet been built.
 var ErrNotBuilt = errors.New("not built")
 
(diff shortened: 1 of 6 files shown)


zroubalik (approved):
/lgtm

PR #1411, opened 2022-11-04
fix: path default is current working directory

lkingland:
:broom: default path is current working directory

Cleans up logic to correctly treat the current working directory as the default effective function path unless a value for path is provided.  Removes all references to "." because the shell is responsible for path expansion.

```release-notes
Ensures current working directory is always the default effective function path
```


zroubalik (approved):
/lgtm

PR #1412, opened 2022-11-05
src: rename config local variable cfg

lkingland:
:broom: rename config local variable

Use `cfg` as the local variable name for `config` in preparation for use of the `config` pacakge throughout

/kind cleanup


zroubalik (approved):
/lgtm
/approve

PR #1413, opened 2022-11-05
fix: build prompts

lkingland:
:broom: build prompts cleanup

Fixes a few small bugs with build prompts by embracing the system:  there is a dedicated method for prompting a user: a command config's `.Prompt()` method.  Its use results in overall reduction in complexity and removal of some small bugs when combining with --confirm.

This primary purpose of this cleanup, however, is to assists in the integration of global config's function context (PR 1416) into the runBuild method by reducing the logic handled therein

/kind cleanup


lance (commented):
/lgtm

PR #1415, opened 2022-11-07
feat: Change Invocation.format to Invoke

### LINKED ISSUE #1280: Invocation struct simplification
## Summary

The function's `Invocation` struct is unnecessary, and `http` should be the default.

## Exposition

The additional fields originally intended to fill out that struct were ultimately unnecessary, and therefore the struct itself is no longer necessary.  Additionally, the value of 'http' should be considered the static default, and as such the value need not be even written to func.yaml when a new function is created unless the user has specified a non-default value (such as `func create -l go -t cloudevents`

## Implementation Details

The `Function.Invocation` member would replaced with a string member `Function.Invoke` which would take the place of what is currently `Invocation.Format`.  It would also be set to `omitempty` in its yaml annotation.

The `Client.Invoke()` method would need to be updated to presume invoking via HTTP as the default case.

Also worth noting is that the manifest.yaml files in templates will need to be updated to 1) remove explicit "http" invocation from the http templates (in some cases allowing the manifst.yaml to be removed entirely.

There would need to be a migration added which moves from using the Invocation struct to the new simple Invoke member.

This will have the added effect of the `func.yaml` not including `invoke` unless the user specifically chose a template which defines a non-default invocation method.

## Related

This assists in the further minimization of the default func.yaml per #1279

gauron99:
# Changes

- Change Invocation structure from f.Invocation.format -> f.Invoke
- update template files & tests
- create migration 0.35.0 from previous versions & add tests

/kind enhancement
/fixes #1280


client.go
```diff
@@ -879,7 +879,10 @@ func (c *Client) Invoke(ctx context.Context, root string, target string, m Invok
 	if err != nil {
 		return
 	}
-
+	// if invoke is empty, use default 'http'
+	if f.Invoke == "" {
+		f.Invoke = DefaultInvocationFormat
+	}
```
lkingland, line 885: I would go a step further and remove this condition from here because the `invoke` method already takes care of this logic.  It might just need to have the defaulting moved up to the very top of the function (before the verbose print statement).  The `invoke` method can expect to default to invoking via HTTP if not given an invocation method explicitly on the function or the message.

function_migrations.go
```diff
@@ -81,6 +81,7 @@ var migrations = []migration{
 	{"0.23.0", migrateToBuilderImages},
 	{"0.25.0", migrateToSpecVersion},
 	{"0.34.0", migrateToSpecsStructure},
+	{"0.35.0", migrateToInvokeStructure},
```
lkingland, line 84:
```suggestion
	{"0.35.0", migrateFromInvokeStructure},
```
Maybe?

function_migrations.go
```diff
@@ -340,3 +362,14 @@ type migrateToSpecVersion_previousFunction struct {
...
+
+// (Defined only for previous versions migration)
+// Invocation defines hints on how to accomplish a function invocation.
+type Invocation struct {
+	Format string `yaml:"format,omitempty"`
+}
+
+// Functions prior to 0.35.0 will have Invocation.Format instead of Invoke
+type migrateToInvokeStructure_previousFunction struct {
+	Invocation Invocation `yaml:"invocation,omitempty"`
```
lkingland, line 374: We could rename this struct something like migrateFromInvokeStructure_invocation. If the `yaml:"invocation"` annotation on the struct field stays the same, it should have no effect on the migration.
That way, this struct is no longer part of the package's exported Godocs, and will not get in the way if we ever end up adding the structure back.

template.go
```diff
@@ -52,8 +52,8 @@ func (t template) Write(ctx context.Context, f *Function) error {
 	if f.Deploy.HealthEndpoints.Readiness == "" {
 		f.Deploy.HealthEndpoints.Readiness = t.config.HealthEndpoints.Readiness
 	}
-	if f.Invocation.Format == "" {
-		f.Invocation.Format = t.config.Invocation.Format
+	if f.Invoke == "" && t.config.Invoke != DefaultInvokeFormat {
```
lkingland, line 55:
```suggestion
	if f.Invoke == "" && t.config.Invoke != "http" {
```
I would change this to the actual value, since that is "locked in time" as being the value pre-migration-35, whereas it is possible (though very unlikely) that we could change DefaultInvokeFormat in the future

lkingland (commented):
Looks good!  I just made a few naming suggestions etc.

function_migrations.go
```diff
@@ -282,6 +283,32 @@ func migrateToSpecsStructure(f1 Function, m migration) (Function, error) {
...
+	bb, err := os.ReadFile(f0Filename)
+	if err != nil {
+		return f1, errors.New("migration 'migrateFromInvokeStructure' error: " + err.Error())
+	}
+	f0 := migrateFromInvokeStructure_invocation{}
+	if err = yaml.Unmarshal(bb, &f0); err != nil {
+		return f1, errors.New("migration 'migrateFromInvokeStructure' error: " + err.Error())
+	}
+
+	if f0.Invocation.Format != "" && f0.Invocation.Format != DefaultInvocationFormat {
```
gauron99, line 303: I was wondering - Should i change this Default format to "locked in time" "http" string as well?

matejvasek:
@gauron99 you will need to regenerate `zz_filesystem_generated.go` because I modified it in another PR.

function_migrations.go
```diff
@@ -340,3 +367,14 @@ type migrateToSpecVersion_previousFunction struct {
 type migrateToBuilderImages_previousFunction struct {
 	Builder string `yaml:"builder"`
 }
+
+// (Defined only for previous versions migration)
+// Invocation defines hints on how to accomplish a function invocation.
+type Invocation struct {
```
lkingland, line 373:
```suggestion
type migrateFromInvokeStructure_invocation struct {
```

function_migrations.go
```diff
@@ -340,3 +367,14 @@ type migrateToSpecVersion_previousFunction struct {
...
+	Format string `yaml:"format,omitempty"`
+}
+
+// Functions prior to 0.35.0 will have Invocation.Format instead of Invoke
+type migrateFromInvokeStructure_invocation struct {
```
lkingland, line 378:
```suggestion
type migrateFromInvokeStructure struct {
```

function_migrations.go
```diff
@@ -282,6 +283,32 @@ func migrateToSpecsStructure(f1 Function, m migration) (Function, error) {
...
+// func.yaml as the default value is implicitly expected. Otherwise if Invoke
+// is non-default value, it will be written in func.yaml.
+func migrateFromInvokeStructure(f1 Function, m migration) (Function, error) {
+	// Load the Function using pertinent parts of the previous version's schema:
+	f0Filename := filepath.Join(f1.Root, FunctionFile)
+	bb, err := os.ReadFile(f0Filename)
+	if err != nil {
+		return f1, errors.New("migration 'migrateFromInvokeStructure' error: " + err.Error())
+	}
+	f0 := migrateFromInvokeStructure_invocation{}
```
lkingland, line 298:
```suggestion
	f0 := migrateFromInvokeStructure{}
```

function_migrations.go
```diff
@@ -340,3 +367,14 @@ type migrateToSpecVersion_previousFunction struct {
...
+
+// (Defined only for previous versions migration)
+// Invocation defines hints on how to accomplish a function invocation.
+type Invocation struct {
+	Format string `yaml:"format,omitempty"`
+}
+
+// Functions prior to 0.35.0 will have Invocation.Format instead of Invoke
+type migrateFromInvokeStructure_invocation struct {
+	Invocation Invocation `yaml:"invocation,omitempty"`
```
lkingland, line 379:
```suggestion
	Invocation migrateFromInvokeStructure_invocation `yaml:"invocation,omitempty"`
```

function_migrations_unit_test.go
```diff
@@ -170,3 +170,30 @@ func TestMigrateToSpecs(t *testing.T) {
 	}
 
 }
+
+// TestMigrateToInvokeStructure tests that migration from f.Invocation.Format to
+// f.Invoke works
+func TestMigrateToInvokeStructure(t *testing.T) {
```
lkingland, line 176:
```suggestion
// TestMigrateFromInvokeStructure tests that migration from f.Invocation.Format to
// f.Invoke works
func TestMigrateFromInvokeStructure(t *testing.T) {
```

lkingland (commented):
Looks good.  I clarified the suggestion to rename the `Invoke` struct via suggestions herein:

gauron99:
I see now what you meant. I changed the names back and added `_previousFunction` for the `struct` because otherwise the name was with conflict with the`func` itself

lance:
@gauron99 looks like you need to regenerate zz_filesystem_generated.go - sorry about that!

zroubalik (approved):
/lgtm

zroubalik:

PR #1418, opened 2022-11-08
fix: interrupts do not interrupt

lkingland:
Interrupt errors are actual errors, and should be treated as such.

For example, when composing a complex command, such as:
```
func invoke && func deploy
```
If `func invoke` is interrupted, the return code must be non-zero, or `func deploy` will be incorrectly executed.

/kind bug


cmd/func/main.go
```diff
@@ -38,7 +39,9 @@ func main() {
 		}}
 
 	if err := cmd.NewRootCmd(cfg).ExecuteContext(ctx); err != nil {
-		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		if !errors.Is(err, terminal.InterruptErr) {
```
matejvasek, line 42: If opposite is true (err is interrupt) we should return 130.

cmd/func/main.go
```diff
@@ -38,7 +39,9 @@ func main() {
...
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		}
 		if ctx.Err() != nil {
 			os.Exit(130)
```
matejvasek, line 46:
```suggestion
		}
		if ctx.Err() != nil || errors.Is(err, terminal.InterruptErr) {
			os.Exit(130)
```

matejvasek:
Looks good apart from the minor suggestion.

cmd/func/main.go
matejvasek: see the suggestion

cmd/func/main.go
lkingland: Maybe I am mis-reading this... but the `|| errors.Is(err, terminal.InterruptErr)` can never be true if err is nil (which is caught by the first part of the conditional).

cmd/func/main.go
lkingland: +1 thanks, ^C should be 130 not 1

PR #1424, opened 2022-11-13
fix: ignore global config on docs gen

lkingland:
:bug: docs generator is affected by environment's global config

When adding the global config system, the docs generator needed to also be updated such that it does not encode the current environment's global settings in the generated docs (flag defaults).

### DIFF
--- a/docs/generator/main.go
+++ b/docs/generator/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"html/template"
 	"os"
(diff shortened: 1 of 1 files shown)


lkingland:
Is there a better way to do this?

lance:
> Is there a better way to do this?

Seems reasonable to me.

lance (commented):
/lgtm

lance:

PR #1425, opened 2022-11-14
src: update build command help text

lkingland:
- :broom:  Update build command help text a bit

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -19,33 +19,49 @@ import (
 
 func NewBuildCmd(newClient ClientFactory) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "build",
-		Short: "Build a function project as a container image",
-		Long: `Build a function project as a container image
+		Use: "build",
+		Long: `
+NAME
+	{{.Name}} build - Build a Function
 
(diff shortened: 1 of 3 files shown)


lance (commented):
Shouldn't this also include an update to the docs?

lance:

lkingland:
> Shouldn't this also include an update to the docs?

Oops! Yes of course.  It should be there now.

docs/reference/func_deploy.md
```diff
@@ -101,7 +101,7 @@ func deploy
 
 ```
       --build string[="true"]   Build the function. [auto|true|false]. [Env: $FUNC_BUILD] (default "auto")
-  -b, --builder string          builder to use when creating the underlying image. Currently supported builders are "pack" and "s2i". (default "pack")
+  -b, --builder string          builder to use when creating the underlying image. Currently supported builders are "pack" and "s2i". (default "s2i")
```
lance, line 104: This seems to have picked up a local change unnecessarily

lkingland: Ah yes, #1424 had not yet been merged!  Rebased...

Issue #1432, opened 2022-11-16
On-cluster build: uploading of the sources to PVC causes "permission denied" error on a mounted volume on AWS EKS


grafvonb:
If you attempt to run the on-cluster build `func deploy -v --remote` using AWS EKS with a defined storage class for the persistent volume based on the dynamic aws-ebs provisioner (so that a PV is created and bound to PVC when the uploader pod is started), a "permission denied" error message is issued:
```
❯ f deploy -v --remote
Creating Pipeline resources
Error: failed to run pipeline: cannot upload sources to the PVC: the command failed: exitcode=1, out="tar: can't create directory 'source': Permission denied\ntar: can't open 'source/.gitignore': No such file or directory"
```

grafvonb:
PR https://github.com/knative/func/pull/1433

PR #1433, opened 2022-11-16
fix: add default PodSecurityContext to the uploader (#1432)

grafvonb:
Adds `PodSecurityContext` with defined non-root `uid` and `gid` and `fsGroup`, so that the ownership of a volume is changed and the non-cluster user can operate on the volume without "permission denied" error (see #1432 for more).

### DIFF
--- a/k8s/security_context.go
+++ b/k8s/security_context.go
@@ -8,6 +8,17 @@ import (
 
 var oneTwentyFour = semver.MustParse("1.24")
 
+func defaultPodSecurityContext() *corev1.PodSecurityContext {
+	// change ownership of the mounted volume to the first non-root user uid=1000
+	runAsUser := int64(1000)
+	runAsGroup := int64(1000)
+	return &corev1.PodSecurityContext{
+		RunAsUser:  &runAsUser,
+		RunAsGroup: &runAsGroup,
+		FSGroup:    &runAsGroup,
+	}
+}
+
 func defaultSecurityContext(client *kubernetes.Clientset) *corev1.SecurityContext {
 	runAsNonRoot := true
 	sc := &corev1.SecurityContext{
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.6
-	github.com/Masterminds/semver v1.5.0 // indirect
+	github.com/Masterminds/semver v1.5.0
 	github.com/Microsoft/go-winio v0.5.2
 	github.com/Netflix/go-expect v0.0.0-20220104043353-73e0943537d2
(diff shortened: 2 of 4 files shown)


grafvonb:
On cluster tests fail with strange errors, which have nothing to do with the change like:
```
    cmd.go:108: Error: failed to run pipeline: function pipeline run has failed with message: 
        
        STEP 1/9: FROM registry.access.redhat.com/ubi8/openjdk-17
        Trying to pull registry.access.redhat.com/ubi8/openjdk-17:latest...
        error creating build container: copying system image from manifest list: Source image rejected: parsing signature https://access.redhat.com/webassets/docker/content/sigstore/ubi8/openjdk-17@sha256=ce855b7824812b85effc211671df4a241ff105f58e7d9236a519996e6e73f4a4/signature-3: unrecognized signature format, starting with binary 0x3c
```

grafvonb:
It's ok now! Let's merge it 😄 

> On cluster tests fail with strange errors, which have nothing to do with the change like:

k8s/security_context.go
```diff
@@ -8,6 +8,17 @@ import (
 
 var oneTwentyFour = semver.MustParse("1.24")
 
+func defaultPodSecurityContext() *corev1.PodSecurityContext {
+	// change ownership of the mounted volume to the first non-root user uid=1000
+	runAsUser := int64(1000)
```
matejvasek, line 13: This should be set only for non-OpenShift.

matejvasek:
@grafvonb I think this needs rebase.

grafvonb:
> @grafvonb I think this needs rebase.

Already done with your changes.

k8s/security_context.go
grafvonb: In your opinion, the best way to achieve this?

matejvasek:
@grafvonb https://github.com/knative/func/actions/runs/3693043236/jobs/6256136941
you need to run `hack/update-codegen.sh`

grafvonb:
> @grafvonb https://github.com/knative/func/actions/runs/3693043236/jobs/6256136941 you need to run `hack/update-codegen.sh`

Yes, I saw that, but it creates a lot of changes, and even a new folder. Should I commit everything as it is?

matejvasek:
> > @grafvonb https://github.com/knative/func/actions/runs/3693043236/jobs/6256136941 you need to run `hack/update-codegen.sh`
> 
> Yes, I saw that, but it creates a lot of changes, and even a new folder. Should I commit everything as it is?

I think you should. CI verify whether it's all right.

matejvasek:
@grafvonb do you use macOS by any chance?
./hack/update-codegen.sh works differently there for some reason causing troubles.

grafvonb:
> @grafvonb do you use macOS by any chance? ./hack/update-codegen.sh works differently there for some reason causing troubles.

Yes, I do. And I have problems with the results of "verify".

Issue #1379, opened 2022-10-22
Provide a way to specify a lifecycle or remove the hardcoded value during build


andrew-su:
#Problem
In an attempt to use `ghcr.io/vmware-tanzu/function-buildpacks-for-knative/functions-builder:0.2.0` to build a function (generated from some [templates](https://github.com/andrew-su/function-buildpacks-for-knative)) using the `func` CLI. I ran into the following from `func build -v`:

```
[truncated]
pulling image quay.io/boson/lifecycle:0.13.2
0.13.2: Pulling from boson/lifecycle
Digest: sha256:b4493261d2be3e0aad701851f7a74d8ccc4dde18901d48469b77605da6fbfd65
Status: Image is up to date for quay.io/boson/lifecycle:0.13.2
Using build cache volume pack-cache-asfgd_x_latest-6f529a43b780.build
===> ANALYZING
Running the analyzer on OS linux with:
Container Settings:
  Args: /cnb/lifecycle/analyzer -gid 0 -log-level debug -daemon -stack /layers/stack.toml -run-image paketobuildpacks/run-jammy-base:0.1.8 -launch-cache /launch-cache localhost:5001/asfgd/x:latest
  System Envs: CNB_USER_ID=1001 CNB_GROUP_ID=1000 CNB_PLATFORM_API=0.9
  Image: quay.io/boson/lifecycle:0.13.2
  User: root
  Labels: map[author:pack]
Host Settings:
  Binds: /var/run/docker.sock:/var/run/docker.sock pack-cache-asfgd_x_latest-6f529a43b780.launch:/launch-cache pack-layers-wyanjsaagv:/layers pack-app-fkpaonxjjd:/workspace
  Network Mode: 
[analyzer] ERROR: failed to : set platform API: platform API version '0.9' is incompatible with the lifecycle
Error: executing lifecycle. This may be the result of using an untrusted builder: failed with status code: 11
```

I noticed that `quay.io/boson/lifecycle:0.13.2` lifecycle is hardcoded into the [pack options](https://github.com/knative/func/blob/main/buildpacks/builder.go#L110). However, the builder I was trying to use already specifies what lifecycle it needs. This may have caused a conflict between them.

Proposal:
1. Remove the hardcoded line and let pack determine the lifecycle from the builder.
2. Provide a way to specify a different lifecycle.

andrew-su:
I removed the lifecycle and rebuilt the CLI locally and it was able to successfully build my function.

lance:
[...]

PR #1438, opened 2022-11-17
Add Google Cloud Function Buildpacks to trusted builder prefixes

wizpresso-steve-cy-fan:
/kind enhancement
# Changes

See title
Fixes: https://github.com/knative/func/issues/1379#issuecomment-1318269062

```release-notes
Adds Google Cloud Function Buildpacks to trusted builder images
```


buildpacks/builder.go
```diff
@@ -39,6 +39,7 @@ var (
 		"gcr.io/paketo-buildpacks",
 		"docker.io/paketobuildpacks",
 		"ghcr.io/vmware-tanzu/function-buildpacks-for-knative",
+		"gcr.io/buildpacks/builder"
```
lance, line 42: This should not be the full image name. Just use the image registry and prefix. With [this change](https://github.com/knative/func/pull/1442), the prefix should also include a slash at the end.
```suggestion
		"gcr.io/buildpacks/"
```

lance:
@wizpresso-steve-cy-fan it seems the build is failing because you need to add a trailing `,` to the line.

buildpacks/builder.go
```diff
@@ -40,6 +40,7 @@ var (
 		"gcr.io/paketo-buildpacks/",
 		"docker.io/paketobuildpacks/",
 		"ghcr.io/vmware-tanzu/function-buildpacks-for-knative/",
+		"gcr.io/buildpacks/"
```
lance, line 43:
```suggestion
		"gcr.io/buildpacks/",
```

Issue #984, opened 2022-04-20
On-cluster build: Enable custom buildpacks


zroubalik:
Currently there's no way how we can build a funcion on cluster with a custom buildpack.
For example Go runtime specifies these additional builpacks:
```yaml
buildpacks:
- paketo-buildpacks/go-dist
- ghcr.io/boson-project/go-function-buildpack:tip
```

`pack` library used for local build handles this properly and pulls those additional builpacks.

Though `/cnb/lifecycle/*` binaries used by `buildpack` Tekton Task has only access to builpacks that are copyied into the buillder image:
```
 /cnb/lifecycle/creator --help
Usage of lifecycle:
  -buildpacks string
    	path to buildpacks directory (default "/cnb/buildpacks")
...
```

We probably need to pull and copy these builpacks to this directory before we initiate the build.

matejvasek:
related: https://github.com/buildpacks/tekton-integration/issues/30

grafvonb:
Status quo:

- https://github.com/knative/func/blob/main/pipelines/resources/tekton/task/func-buildpacks/0.1/func-buildpacks.yaml
- https://github.com/tektoncd/catalog/blob/main/task/buildpacks/0.5/buildpacks.yaml

grafvonb:
Our challenge here, in my opinion, is that from a UX perspective, switching between local and on-cluster builds by simply adding the `--remote` flag should be seamless for the user, i.e. provide the same functionality.

However, we use the `pack` library locally, while `buildpacks` Tekton Task uses the lifecycle directly, as mentioned by @zroubalik above (https://github.com/tektoncd/catalog/blob/main/task/buildpacks/0.5/buildpacks.yaml#L136), which does not provide "the power" of `pack`.

If we look at the ways buildpacks can be specified for `pack` (https://buildpacks.io/docs/app-developer-guide/specify-buildpacks/) and the implementation behind it (https://github.com/buildpacks/pack/tree/main/pkg/buildpack), it will take some time to recreate even the loading and extracting functionality for all the cases.

So, how do we deal with this issue? Maybe we should start small and initially just support one (it's better than nothing), simple URI format, like URL for the custom buildbacks, and extend our `func-buildpacks.yaml` accordingly?

zroubalik:
[...]

PR #1445, opened 2022-11-23
allowing on cluster build for go runtime

Shashankft9:
# Changes

- :broom: allow on cluster build to happen for go runtime, using boson go builder
- :broom: updating the documention for on cluster build to fix the existing steps and what has to be added further for go runtime

/kind cleanup

Fixes https://github.com/knative/func/issues/809

**Release Note**

```release-note

```

**Docs**

```docs

```


matejvasek:
@Shashankft9 does this depend on https://github.com/knative/func/issues/984 ?

Shashankft9:
thanks @matejvasek for linking that issue, seems like a better solution specially since i got to know yesterday that this boson go builder is not maintained anymore, 
I wonder, if this can fly as a temporary solution though, since not much has changed programmatically?

pipelines/tekton/validate.go
```diff
@@ -30,7 +30,7 @@ func validatePipeline(f fn.Function) error {
 			return ErrRuntimeRequired
 		}
 
-		if f.Runtime == "go" || f.Runtime == "rust" {
+		if f.Runtime == "rust" {
```
matejvasek, line 33: Maybe we could remove this `if` completely since following `if` check if there are additional buildpacks.

docs/building-functions/on_cluster_build.md
```diff
@@ -49,12 +49,16 @@ git remote add origin git@github.com:my-repo/my-function.git
...
-                                                    # in the repository root folder
+build:
+  git:
+    url: https://github.com/my-repo/my-function.git   # required, git repository with the function source code
+    revision: main                                    # optional, git revision to be used (branch, tag, commit)
+    # contextDir: myfunction                          # optional, needed only if the function is not located in the repository root folder
+  # builderImages:                                    # optional, needed only if the runtime is golang 
+  #   pack: ghcr.io/boson-project/go-function-builder:tip
+  buildpacks: []
+  builder: ""
```
matejvasek, line 60: Builder should be `"pack"`.

matejvasek: but empty string probably would work too

matejvasek:
@zroubalik PTAL

matejvasek:
@Shashankft9 Go build issue might be fixed by https://github.com/knative/func/pull/1451 .

pipelines/tekton/validate.go
zroubalik: I think it might be better to keep it here, to explicitly tell user about not supported runtime.

zroubalik (commented):
we should wait with this until go is supported with on cluster builds.

Or am I missing anything?

matejvasek:
> we should wait with this until go is supported with on cluster builds.
> 
> Or am I missing anything?

The thing is we are disabling `Go` build no matter what -= even if user provided custom builder that can do build even now.

pipelines/tekton/validate.go
matejvasek: But the issue with Go and Rust is that they use custom buildpacks.
That problem is detected in following `if`.

What this `if` does is it prevents users from building Go and Rust apps even if user provides custom builder that would work all right.

lkingland:
FYI I am using this as an important reference/ starting point for the solution discussed in https://github.com/knative/func/discussions/1589
Not sure if that means we will include these commits directly ¯\_(ツ)_/¯ but I am actively using it for reference

Shashankft9:
hey, a gentle ping:

I did an update on the go builder here: https://github.com/boson-project/packs with the latest pack full cnb and go-dist BP. And then the same changes in this PR are making my on cluster build work for go functions.

I did the update of the builder because it was failing for my functions which were using latest k8s imports. 

I am aware that the plan for goruntime's oncluster build is much broader, but I am just checking back again if there is  scope to still allow a customer builder like this one, following @matejvasek comment
> The thing is we are disabling Go build no matter what -- even if user provided custom builder that can do build even now.

Shashankft9:
hey, a gentle ping:

I did an update on the go builder here: https://github.com/boson-project/packs with the latest pack full cnb and go-dist BP. And then the same changes in this PR are making my on cluster build work for go functions.

I did the update of the builder because it was failing for my functions which were using latest k8s imports. 

I am aware that the plan for goruntime's oncluster build is much broader, but I am just checking back again if there is  scope to still allow a customer builder like this one, following @matejvasek comment
> The thing is we are disabling Go build no matter what -- even if user provided custom builder that can do build even now.

pipelines/tekton/validate.go
matejvasek: @zroubalik @Shashankft9 what about just printing warning here "Go/Rust default builder do not support in cluster build" instead of returning an error? Could we accept that?

docs/building-functions/on_cluster_build.md
lkingland: If these two items are intended as hints, commenting them would probably be best.

The idea here is that the code has static defaults.  Anything placed in the config file is specifically requesting that exact value, so while the first value, `[]buildpacks` is innocuous, the second, `builder=""` is, if literally translated, requesting the system to build without a builder.  In this case we do set the default if the value comes in empty like this, because building without a builder is nonsensical, but I presume this exists to give a hint to the user how to choose a builder and buildpacks, so a commented entry is the right tool for the job.

pipelines/tekton/validate.go
lkingland: > @zroubalik @Shashankft9 what about just printing warning here "Go/Rust default builder do not support in cluster build" instead of returning an error? Could we accept that?

@Shashankft9 Can you add a simple warning here that Go and Rust default builders do not support in-cluster builds?

lkingland (commented):
This looks mostly good to me, but we could probably use a warning, as suggested, and of course a rebase to resolve the conflicts.

Shashankft9:
just signed the CLA again.

@lkingland and others - hey, apologies for the delay on this, got caught up in other work. This is just a quick fix, let me know if it looks alright, I'll add some more tests

matejvasek:
PTAL @lance

matejvasek:
@Shashankft9 I pushed on fixup.

lance:
I think this is the symlink issue on windows that we have been dealing with lately. @matejvasek please lmk if I am mistaken.

```
test build error rename C:\Users\RUNNER~1\AppData\Local\Temp\TestBuilder_Concurrency2845656642\001\.func\builds\by-hash\70edc4a387fa606bf967e52bea2cc6ce6762ce9eeda58e617cfbf99a3514fd73\config.json C:\Users\RUNNER~1\AppData\Local\Temp\TestBuilder_Concurrency2845656642\001\.func\builds\by-hash\70edc4a387fa606bf967e52bea2cc6ce6762ce9eeda58e617cfbf99a3514fd73\oci\blobs\sha256\8a7469839f911077ecb8405595926905343df847dfed79d760a130c1953d91ee: The system cannot find the path specified.	knative.dev/func/pkg/oci	coverage: 71.6% of statements
```

/override "Unit Test (1.20.2, 17, windows-latest)"

Shashankft9:
@matejvasek would be amazing if you can take a look at this too: https://github.com/boson-project/packs/pull/10

matejvasek:
@Shashankft9 does this work now? I am afraid https://github.com/knative/func/pull/1774 is breaking it.

matejvasek:
> Shashankft9 does this work now? I am afraid #1774 is breaking it.

Well if is probably not breaking on-cluster-build, but it breaks local build.

PR #1451, opened 2022-11-30
[WIP] Add runtime

lance:
# Changes

This change pulls the golang function invocation framework from boson-project/packs and adds that code as a package to `func` itself. This does require that a minimal `main()` be included in the function templates for Go, which is a downside. But is it really so bad? This change eliminates the need to maintain a special Go buildpack for functions, and enables the base paketo Golang buildpack to work out of the box. It also immediately enables on-cluster builds for Go functions, as they no longer require a special buildpack to be added. This also makes s2i builds in Go possible as well.

- :gift: Adds the Go function runtime from https://github.com/boson-project/packs/tree/main/buildpacks/go
- :gift: Updates templates to use the func package

## TODO
- [] Remove check preventing on-cluster builds for golang
- [] Update go.mod `replace` directives if/when this PR lands

/kind enhancement

Related: https://github.com/knative/func/issues/984


lance:
Consider backporting to `release-1.8`.

lance:
Note that currently, the s2i golang image does not support non-standard build targets. See: https://github.com/sclorg/golang-container/issues/48

lance:
This allows, for example, a user to use the Google builder with their Go functions. A working example...

```
....
....
Adding cache layer 'google.go.gomod:gopath'
Layer 'google.go.gomod:gopath' SHA: sha256:3b524f2223d745a5a2027195fa62bfbfde642332a037b7268b751a728ddaa69b
Reading buildpack directory: /layers/google.go.build
Reading buildpack directory item: bin
Reading buildpack directory item: bin.toml
Reading buildpack directory item: gocache
Reading buildpack directory item: gocache.toml
Reading buildpack directory item: launch.toml
Reading buildpack directory: /layers/google.utils.label
🙌 Function image built: quay.io/lanceball/go-s2i:latest

~/demo/go-s2i via 🐹 v1.19.1 on ☁️  lanceball@gmail.com took 48s
❯ func run
Function already built.  Use --build to force a rebuild.
Function started on port 8080
```

lance:
Apparently s2i builds will not work out of the box. There needs to be some improvement in the current s2i builder image: https://github.com/sclorg/golang-container/issues/48

matejvasek:
>Note that currently, the s2i golang image does not support non-standard build targets

What would we need to do to make it work.
Wouldn't it suffice to put main pkg into root and move `Handle()` to another dir/pkg?

lance:
> > Note that currently, the s2i golang image does not support non-standard build targets
> 
> What would we need to do to make it work. Wouldn't it suffice to put main pkg into root and move `Handle()` to another dir/pkg?

Yes that would fix it, but I really do like having `main()` out of the way like this. Greater sophistication on the part of the builder's `build` script would be nice. For example, the paketo builder accepts a `BP_GO_TARGETS` environment variable, and the Google builder image uses `go list` to figure out where `main()` lives. (See: https://github.com/GoogleCloudPlatform/buildpacks/blob/cb94a4f0db650902151419fc3a167cad7831b6b6/cmd/go/build/main.go#L132-L135).

lance:
There has also been some talk of mapping `knative-sandbox/func-go` to `http://knative.dev/runtime` with an HTTP entry for `go get` and moving this invocation code to a new repo. That might make the chicken/egg problem in `go.mod` easier to deal with.

lance:
I am proposing that we create a new repository https://github.com/knative-sandbox/func-go mapping to `knative.dev/runtime`. We can iterate on the runtime characteristics there, and this PR will then become a set of much smaller changes primarily to the golang templates. Let's see how that plays out this week.

lkingland:
I like that this pulls the Go function invocation code out of the buildpack, and how it is now nicely placed right here as a library.  Also nice to see the simplification of the template manifest.  What I think we need to discuss is at which stage in a function's lifecycle the function's source code is "wrapped" in the process boundary and placed inside a container.  In particular, this is part of the build step.  So instead of using the template system to do this during function _creation_, let's do something similar but on build of the container, and as a transient artifact in `.func/build[N]`

go.mod
```diff
@@ -53,7 +53,10 @@ require (
 	knative.dev/serving v0.35.1-0.20221114131921-874ccebb8063
 )
 
-require github.com/sabhiram/go-gitignore v0.0.0-20210923224102-525f6e181f06
+require (
+	github.com/gorilla/mux v1.8.0
```
zroubalik, line 57: https://github.com/gorilla#gorilla-toolkit

not sure if we should depend on this package

templates/go/cloudevents/faas/main.go
```diff
@@ -0,0 +1,11 @@
+package main
+
+import (
+	function "function"
+
+	"knative.dev/func/runtime"
+)
+
+func main() {
```
zroubalik, line 9: We should add some comment here pointing users to handle.go. So they don't try to write the function code here.

templates/go/http/faas/main.go
```diff
@@ -0,0 +1,11 @@
+package main
+
+import (
+	function "function"
+
+	"knative.dev/func/runtime"
+)
+
+func main() {
```
zroubalik, line 9: We should add some comment here pointing users to handle.go. So they don't try to write the function code here.

templates/go/manifest.yaml
```diff
@@ -1,5 +1,3 @@
-# Optional list of additional Buildpacks to be applied to the language pack's
-# builder image when the Function is built using a Buildpack builder.
-buildpacks:
-  - paketo-buildpacks/go-dist
-  - ghcr.io/boson-project/go-function-buildpack:tip
+buildEnvs:
+  - name: BP_GO_TARGETS
```
zroubalik, line 2: out of curiosity, why is this env needed?

zroubalik (commented):
A great improvement!

go.mod
lance: Yes I agree. For the time being, however, I think we can keep it as is until we come up with a suitable alternative.

lance: @zroubalik also, note that I've moved this code over to https://github.com/knative-sandbox/func-go which will change this PR to just be modifications to the templates.

templates/go/manifest.yaml
zroubalik: makes sense. And agree that the `go list` approach is nicer :)

lkingland:
Since this is a bit bigger of a topic, [here is a discussion](https://github.com/knative/func/discussions/1589) we could use to explore

lance:
> Since this is a bit bigger of a topic, [here is a discussion](https://github.com/knative/func/discussions/1589) we could use to explore

Very much appreciate your effort to outline this in a discussion.

lance:
Closing as obsolete

PR #1457, opened 2022-12-06
Fix bug where user's local .gitignore file would be rewritten everytime `func run` was ran, and also a small but nice enhancement by reducing binary size by 25%

### LINKED ISSUE #1420: The func run command overwrites .gitignore
## Steps to reproduce
- Create a function using `func create` (e.g. `func create -l python myfunc`)
- Edit the `.gitignore` that is generated by `func create`
- Run `func run`

## Expected behavior
My changes to `.gitignore` are persisted

## Actual behavior
My changes to `.gitignore` are overwritten:
```
# Functions use the .func directory for local runtime data which should
# generally not be tracked in source control:
/.func
```

amaanq:
# Changes

- When running subsequent `func run` commands on a function, the local .gitignore file is overwritten, so a user might lose other files they want to have ignored by git as well
- Fixes by checking if the file exists, and if it does then check each line to see if variations of ".func" exist, but also ensures its the entire directory and not matching against a file in the directory (e.g. .func/file1)

/kind bug

Fixes #1420 

```release-note
If you were previously having an issue where your local .gitignore file was being overwritten every time you ran `func run`, this is now fixed!
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ HASH    := $(shell git rev-parse --short HEAD 2>/dev/null)
 VTAG    := $(shell git tag --points-at HEAD | head -1)
 VTAG    := $(shell [ -z $(VTAG) ] && echo $(ETAG) || echo $(VTAG))
 VERS    ?= $(shell [ -z $(VTAG) ] && echo 'tip' || echo $(VTAG) )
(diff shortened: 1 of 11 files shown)


matejvasek:
@amaanq thanks for the contribution!

You will need to regenerate `zz_filesystem_generated.go` by running `make zz_filesystem_generated.go`.

amaanq:
I put this on the main branch, will put this on its own branch instead, sorry!

PR #1474, opened 2022-12-16
Cleanup e2e on cluster tests

jrangelramos:
# Changes

Some refinements on e2e on cluster tests:
- :broom: Making function name unique per run (to avoid name conflict on subsequent tests)
- :broom: resolving default test registry name according to the target cluster

/kind cleanup


lance (approved):
Looks like a nice bit of cleanup - thanks!
/lgtm

Issue #851, opened 2022-02-22
On-cluster build: Configurable PVC size used by a Pipeline


lance:

ocpvkb:
👍

tvand:
Would it be feasible to generate a YAML file contains all the pipeline specifications initially and then use that, if it exists? That way, users can fix any issues they may have directly in the file. WDYT?

lance:
@zroubalik I think there are a couple of issues already open around this idea in a general sense. There is https://github.com/knative-sandbox/kn-plugin-func/issues/1015 which is a generic request for yaml output for anything that `func` produces, and there is this https://github.com/knative-sandbox/kn-plugin-func/pull/898 which is about refactoring func.yaml and using it to generate resource yaml. I think that 1015 is actually a good place to track this all as a whole, and I will follow up there.

zroubalik:
@lance you are right! Thanks for poiting this out

grafvonb:
@lance Is it still an open issue that should be implemented? 

Currently the PVC has a fixed size of 5Mi defined in `pipelines/tekton/defaults.go` as follows:
```
const (
	// DefaultPersistentVolumeClaimSize represents default size of PVC created for a Pipeline,
	// specified in bytes (eg. 5Mi = 5MiB = 5 * 1024 * 1024)
	DefaultPersistentVolumeClaimSize int64 = 5 * 1024 * 1024
)
```

lance:
@grafvonb yes, this is still something we'd like to implement, although I haven't really given a lot of thought to how this should surface for the user. It could be a flag on the `func deploy --remote` command, possibly. E.g.

```
func deploy --remote --git-url <url> --pvc-size 10MiB
```
But I am open to consideration for other options.

grafvonb:
@lance Is there a sketch for handling configuration options? I think @lkingland mentioned it somewhere.

I know we use `cobra`/`viper`, but we also have `func.yaml`, so maybe it would make sense to extend `viper` (https://github.com/spf13/viper/blob/master/README.md#reading-config-from-ioreader) and add a new source like `func.yaml`? Then we get a full support for all kind of configuration sources that the user could use if needed...

lkingland:
[...]

PR #1475, opened 2022-12-16
feat: change default pvc size to 256Mi (#851)

grafvonb:
Quick fix according to reported problem (pvc size too small) and small refactoring for #851; second part (new flag `--pvc-size`) follows as an independent PR.


lance (approved):
/lgtm

Issue #1408, opened 2022-11-04
ASP.NET Core language pack support


stevefan1999-personal:
Ditto to title

lance:
Hi @stevefan1999-personal thanks for opening this issue! Unfortunately, nobody that I am aware of on the team has .NET experience. We would, of course, be open to a pull request if you have any interest in working on this.

clibequilibrium:
@stevefan1999-personal  feel free to check this https://github.com/knative/func/pull/1484

PR #1484, opened 2023-01-07
feat: adds CSharp runtime

clibequilibrium:
Hello knative community,

I had a big need for C# runtime and was able to create one directly modifying func. Unfortunately I wasn't able to add C# runtime as per language pack tutorial because func would complain about default bulder image not being found - which is likely a bug. 

``is not a known builder. The available builder is "pack"``

Instead I just went and added new C# runtime that currently only supports http but you are welcome to add cloudevents.

[Linked to this issue](https://github.com/knative/func/issues/1408) 

## Release notes

```release-notes
Adds C# runtime to built-in templates
```

Thanks

### DIFF
--- a/cmd/templates_test.go
+++ b/cmd/templates_test.go
@@ -33,7 +33,9 @@ rust         http
 springboot   cloudevents
 springboot   http
 typescript   cloudevents
-typescript   http`
+typescript   http
+csharp		 cloudevents
+csharp		 http`
 	output := buf()
 	if output != expected {
 		t.Fatalf("expected:\n'%v'\n\ngot:\n'%v'\n", expected, output)
@@ -78,6 +80,10 @@ func TestTemplates_JSON(t *testing.T) {
     "http"
   ],
   "typescript": [
+    "cloudevents",
+    "http"
+  ],
+  "csharp": [
     "cloudevents",
     "http"
   ]
--- a/templates/csharp/manifest.yaml
+++ b/templates/csharp/manifest.yaml
@@ -1,0 +1,5 @@
+# Optional list of additional Buildpacks to be applied to the language pack's
+# builder image when the Function is built using a Buildpack builder.
+buildpacks:
+  - paketo-buildpacks/dotnet-core
+  - gcr.io/paketo-buildpacks/dotnet-core
--- a/buildpacks/builder.go
+++ b/buildpacks/builder.go
@@ -31,6 +31,7 @@ var (
 		"quarkus":    "gcr.io/paketo-buildpacks/builder:base",
 		"rust":       "gcr.io/paketo-buildpacks/builder:base",
 		"springboot": "gcr.io/paketo-buildpacks/builder:base",
+		"csharp": 	  "gcr.io/paketo-buildpacks/builder:base",
 	}
 
 	// Ensure that all entries in this list are terminated with a trailing "/"
--- a/client_test.go
+++ b/client_test.go
@@ -1235,6 +1235,7 @@ func TestClient_Runtimes(t *testing.T) {
 		"springboot",
 		"test",
 		"typescript",
+		"csharp",
 	}
 
 	if !reflect.DeepEqual(runtimes, expected) {
--- a/cmd/languages_test.go
+++ b/cmd/languages_test.go
@@ -23,7 +23,8 @@ python
 quarkus
 rust
 springboot
-typescript`
+typescript
+csharp`
 	output := buf()
 	if output != expected {
 		t.Fatalf("expected:\n'%v'\ngot:\n'%v'\n", expected, output)
@@ -49,7 +50,8 @@ func TestLanguages_JSON(t *testing.T) {
   "quarkus",
   "rust",
   "springboot",
-  "typescript"
+  "typescript",
+  "csharp"
 ]`
 	output := buf()
 	if output != expected {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -400,7 +400,7 @@ func (v Version) StringVerbose() string {
 //
 // Usage Example:
 //
-//	languages := []string{ "go", "node", "rust" },
+//	languages := []string{ "go", "node", "rust", "csharp" },
 //	survey.Select{
 //	  Options: options,
 //	  Default: surveySelectDefaut(cfg.Language, languages),
--- a/function.go
+++ b/function.go
@@ -30,7 +30,7 @@ type Function struct {
 	// Name of the function.
 	Name string `yaml:"name" jsonschema:"pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"`
 
-	// Runtime is the language plus context.  nodejs|go|quarkus|rust etc.
+	// Runtime is the language plus context.  nodejs|go|quarkus|rust|csharp etc.
 	Runtime string `yaml:"runtime"`
 
 	// Template for the function.
--- a/pipelines/tekton/validate_test.go
+++ b/pipelines/tekton/validate_test.go
@@ -104,6 +104,16 @@ func Test_validatePipeline(t *testing.T) {
 			function: fn.Function{Build: fn.BuildSpec{Builder: builders.S2I}, Runtime: "rust"},
 			wantErr:  true,
 		},
+		{
+			name:     "Unsupported runtime - CSharp - pack builder - without additional Buildpacks",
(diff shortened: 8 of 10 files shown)


lance:
@clibequilibrium thanks for contributing this! I'm curious about your experience trying to write a language pack instead of adding C# directly to `func`.

> Unfortunately I wasn't able to add C# runtime as per language pack tutorial because func would complain about default builder image not being found - which is likely a bug.

In your language pack, did you specify the builder and buildpacks in `manifest.yaml` as described here? https://github.com/knative/func/blob/main/docs/language-packs/language-pack-contract.md#builders. If not, can you please give that a try? If you did, then this is definitely a bug and if possible it would be great if you can create an issue illustrating how to replicate the problem. Thanks!

Also, for this or any other PR of yours to be accepted, you'll need to sign the EasyCLA. Thanks again for the contribution.

cmd/templates_test.go
```diff
@@ -33,7 +33,9 @@ rust         http
 springboot   cloudevents
 springboot   http
 typescript   cloudevents
-typescript   http`
+typescript   http
+csharp		 cloudevents
```
matejvasek, line 37: Cloudevents are mentioned couple of time but it seems this adds only http template.

matejvasek:
You need to run `make zz_filesystem_generated.go` and commit result.

cmd/templates_test.go
clibequilibrium: @matejvasek  

> Instead I just went and added new C# runtime that currently only supports http but you are welcome to add cloudevents.

Yes I didn't dive much into cloudevents but they can be taken from [here](https://github.com/knative/docs/tree/main/code-samples/serving/cloudevents/cloudevents-dotnet)

clibequilibrium:
> @clibequilibrium thanks for contributing this! I'm curious about your experience trying to write a language pack instead of adding C# directly to `func`.
> 
> > Unfortunately I wasn't able to add C# runtime as per language pack tutorial because func would complain about default builder image not being found - which is likely a bug.
> 
> Adding new languages and capabilities via the Language Pack contract is a primary aim for us and it should work as expected. It is our preferred method of extending `func` capabilities.
> 
> In your language pack, did you specify the builder and buildpacks in `manifest.yaml` as described here? https://github.com/knative/func/blob/main/docs/language-packs/language-pack-contract.md#builders. If not, can you please give that a try? If you did, then this is definitely a bug and if possible it would be great if you can create an issue illustrating how to replicate the problem. Thanks!
> 
> Also, for this or any other PR of yours to be accepted, you'll need to sign the EasyCLA. Thanks again for the contribution.

Yes I have specified default builder and buildpacks for my language pack in the manifest. I can try to assemble a small repro so you can follow the steps on your and. 

As for EasyCLA I state that I was not able to sign it with my GitHub private email, since I wouldn't want to disclose my public email for this particular commit, and that therefore I grant the permission to copy/merge my PR if accepted or use my code as Free License

matejvasek:
@clibequilibrium have you committed the manifest.yaml in local clone when developing, or have you just added it to FS?

clibequilibrium:
> @clibequilibrium have you committed the manifest.yaml in local clone when developing, or have you just added it to FS?

It was committed to a remote repository as per language pack docs.

matejvasek:
> > @clibequilibrium have you committed the manifest.yaml in local clone when developing, or have you just added it to FS?
> 
> It was committed to a remote repository as per language pack docs.

That's odd, in such a case mainfest.yaml should have been applied.

matejvasek:
> > > @clibequilibrium have you committed the manifest.yaml in local clone when developing, or have you just added it to FS?
> > 
> > 
> > It was committed to a remote repository as per language pack docs.
> 
> That's odd, in such a case mainfest.yaml should have been applied.

I was able to reproduce the manifest issue only when I added manifest to local clone of func-tastic without committing it. IMO that's still a bug (because local modification should be taken into account), but a different one to which you observed.

clibequilibrium:
@matejvasek 

On my end I had a fresh repo with C# project setup, both root manifest and in folder manifest.

matejvasek:
@clibequilibrium could you please provide reproducer?

matejvasek:
@clibequilibrium I cannot reproduce the issue with additional buildpack non-propagation.

lance:
@clibequilibrium I was able to create a language pack using the files you provided in this PR by copying them into a fresh git repo under a top level `csharp` directory. For example

```
mkdir my-language-pack
cd my-language-pack/
cp -R ../func/templates/csharp .
git init .
Initialized empty Git repository in /Volumes/Source/github.com/knative/my-language-pack/.git/
git add .
git commit --signoff -am "Initial commit"
[main (root-commit) 0a8d5bd] Initial commit
 2 files changed, 20 insertions(+)
 create mode 100644 csharp/http/Program.cs
 create mode 100644 csharp/manifest.yaml
func repo add cs .
func create -l csharp -t cs/http ../my-csharp-func
Created csharp function in /Users/lball/Source/github.com/knative/my-csharp-func/
```

clibequilibrium:
> @clibequilibrium I was able to create a language pack using the files you provided in this PR by copying them into a fresh git repo under a top level `csharp` directory. For example
> 
> 
> 
> ```
> 
> mkdir my-language-pack
> 
> cd my-language-pack/
> 
> cp -R ../func/templates/csharp .
> 
> git init .
> 
> Initialized empty Git repository in /Volumes/Source/github.com/knative/my-language-pack/.git/
> 
> git add .
> 
> git commit --signoff -am "Initial commit"
> 
> [main (root-commit) 0a8d5bd] Initial commit
> 
>  2 files changed, 20 insertions(+)
> 
>  create mode 100644 csharp/http/Program.cs
> 
>  create mode 100644 csharp/manifest.yaml
> 
> func repo add cs .
> 
> func create -l csharp -t cs/http ../my-csharp-func
> 
> Created csharp function in /Users/lball/Source/github.com/knative/my-csharp-func/
> 
> ```

I had no issues creating it, but were you able to run or deploy it?

lance:
> I had no issues creating it, but were you able to run or deploy it?

Ahh - now I see the problem. You need to add the following to your `manifest.yaml`

```
builderImages:
    pack: gcr.io/paketo-buildpacks/builder:base
```

templates/csharp/manifest.yaml
```diff
@@ -0,0 +1,5 @@
+# Optional list of additional Buildpacks to be applied to the language pack's
+# builder image when the Function is built using a Buildpack builder.
```
lance, line 2: Please add the following to this file in your language pack and see if that will work for you.

```
builderImages:
    pack: gcr.io/paketo-buildpacks/builder:base
```

lance: @clibequilibrium were you able to resolve your issue by using the above suggestion?

clibequilibrium: > @clibequilibrium were you able to resolve your issue by using the above suggestion?

I will get back to you today. Thank you for follow up

clibequilibrium: @lance  is it an oversight from docs? Because all of this seems Optional, shouldn't func still use default builderImage as base one if it's not provided in manifest.yaml?

clibequilibrium: `===> DETECTING
ERROR: No buildpack groups passed detection.
ERROR: Please check that you are running against the correct path.
ERROR: failed to detect: no buildpacks participating
-
Error: failed to build the function: executing lifecycle: failed with status code: 20`

https://github.com/clibequilibrium/func-templates

lance: I can successfully create a new project using your repo but it won't build, giving the same error you show above. This error is coming from the paketo builder, saying that it can't find a suitable project to build. I'm not familiar with what kinds of files the paketo builder looks for in order to build a CSharp project, but this is typically the kind of error you would see if it can't find what it's looking for. A good way to test this is to try and build the project using the `pack` CLI and not use `func` at all. Doing that, I see this same error.

https://app.warp.dev/block/aWOnPASSvuuQ31nhNB7KTR

You'll need to add whatever file(s) necessary to indicate to the Paketo builder that this is a csharp project. Also, as an aside - I think you don't need both of these lines - only one, and perhaps neither. I believe this buildpack is included in the Paketo builder, so in that case it should not be necessary to add it in manifest.yaml - but I'm not sure.

https://github.com/clibequilibrium/func-templates/blob/eaa195b57e5cbeaa7447a07b32c2692738fbf35d/manifest.yaml#L13-L15

Issue #215, opened 2020-11-06
Group main help message to put important commands to the top


rhuss:
I.e. the `version`, `help` and `completion` commands should be put into a separate section below the main commands. This can be achieved with the same technique as found for `kn` that uses advanced templating techniques from Cobra.

grafvonb:
Hi @rhuss! Are you still working on this issue? If not, I would try to solve it, based on your suggestion to use your code and ideas from `kn`. I have some years of knowhow in go, Cobra, Kubernetes&Co. I have already briefly analysed your solution for `kn` and the question would be if we need exactly the same for `func` or maybe something simplified.

zroubalik:
@grafvonb would be great if you can take a look on this.

grafvonb:
Great, I'll take over.

zroubalik:
@grafvonb any update on this please?

lance:
Would love to see this in 1.0

grafvonb:
@zroubalik I will submit my status this coming weekend. Is that ok for you?

zroubalik:
@grafvonb sure, thanks a lot!

grafvonb:
@zroubalik as discussed, I have checked the context, made a fork and run the core tests successfully on my Mac locally. So I am ready to start developing this change. Two questions before I start:

1. My plan is to reuse the parts of the code from the `fn` project, as suggested by @rhuss. I just don't want to reinvent the wheel. Is that ok with you from a technical and legal perspective?
2. @lance What is the schedule for the 1.0? I would like to make a schedule as I am doing this in my spare time.

Thanks a lot!

zroubalik:
@grafvonb I suppose you meant `kn` project (https://github.com/knative/client/)? Then it is completely fine to use it and it is actually preferred :) 

Thanks for contributing this!

grafvonb:
Yes, exactly `kn`, as you can see, I think about functions only ;).

zroubalik:
> Yes, exactly `kn`, as you can see, I think about functions only ;).

![approve](https://user-images.githubusercontent.com/726523/182096241-3bf265de-f3ea-493f-919a-007383d839a0.gif)

lance:
@grafvonb you can track our progress towards 1.0 here https://github.com/orgs/knative-sandbox/projects/7/views/13

grafvonb:
Done by https://github.com/knative-sandbox/kn-plugin-func/pull/1152

PR #1485, opened 2023-01-09
chore: refactor commands help to use templates

grafvonb:
As part of #215, the definition of the `root` command has changed and uses templating to generate its command line help. However, all subcommands still use `cmd.SetHelpFunc(defaultTemplatedHelp)` or similar for this, so we currently still have the "old approach" in use. 

It should be changed so that all commands use the templating as the `root` command does, and the `cmd.SetHelpFunc(defaultTemplatedHelp)` is gone. 

The generation of the docs must work as before.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -23,10 +23,10 @@ func NewBuildCmd(newClient ClientFactory) *cobra.Command {
 		Short: "Build a Function",
 		Long: `
 NAME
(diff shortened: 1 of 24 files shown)


grafvonb:
@zroubalik @lance ready for review. I was able refactor and remove lot of redundant code in help context. I think it is now cleaner and easier to use.

lance (commented):
/lgtm
Thanks @grafvonb

PR #1509, opened 2023-01-19
Update springboot templates to Spring Boot 3.0

### LINKED ISSUE #1470: Upgrade Spring Boot to 3.0
With Spring Boot 3.0 out of the door (https://spring.io/blog/2022/11/24/spring-boot-3-0-goes-ga), we should update the templates to support this version. 
But I was wondering if we can have spring-boot2 and spring-boot3 support in the templates? 
This brings a lot of complications, but because it's a new major version, it might be worth the hassle.

trisberg:
- add needed buildEnv properties
- add class to register runtime hints for MessageUtils.MessageStructureWithCaseInsensitiveHeaderKeys
- update README files

Signed-off-by: Thomas Risberg <trisberg@pivotal.io>

# Changes

- :broom: Update springboot templates for Spring Boot 3.0
- :bug: This fixes build issues that will happen after anonymous access to repo.spring.io/release is turned off Jan 26, 2023

/kind enhancement

Fixes #1470

**Release Note**

```release-note

The springboot templates have been updated to use Spring Boot 3.0 and the new Spring 6.0 AOT support. Note: this requires Java 17 when building locally.

```

**Docs**

```docs

```


lkingland:
The only E2E test to fail was the OnCluster `TestRuntime/node_s2i_test`, which failed with:
```
Error: failed to run pipeline: function pipeline run has failed with message: 
...
error creating build container: copying system image from manifest list: writing blob: storing blob to file "/var/tmp/storage2885866326/3": happened during read: unexpected EOF
```
This seems transient to me so I'll re-run the failed tests now.

PR #1519, opened 2023-01-24
fix: allow trailing path flag

lkingland:
# Changes

- :bug: fixes a bug where the --path flag would be ignored if provided after a subcommand

Updates the early-stage `effectivePath` method, which directly evaluates `--path` to use the `pflag` library, which implements a more permissive model for parsing flags (POSIX/GNU-style), including the ability to define a flag _trailing_ subcommands, and is the library used by the command structs internally.

/kind bug

**Release Note**

```release-note
Fixes a bug where --path was sometimes not evaluated.
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"flag"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


lance (commented):
/lgtm

PR #1529, opened 2023-01-30
feat: apply and update client methods

lkingland:
# Changes

- :gift: adds client.Apply
- :gift: adds client.Update
- :broom: renames client.Create to client.Init
- :broom: adds `func init` as an alias for `func create`

Adds two meta-commands, 'Apply' and 'Update', to the client. Renames 'Create' to 'Init' to avoid confusion with 'Update' (and to be more linguistically similar to git commands), and adds it as an alias to the create CLI command.

The impetus for this change is how much easier it makes a forthcoming update to the inbuilt integration tests which create functions with inline implementations.

/kind enhancement

### DIFF
--- a/client.go
+++ b/client.go
@@ -407,8 +407,55 @@ func (c *Client) Runtimes() ([]string, error) {
 // LIFECYCLE METHODS
 // -----------------
 
+// Apply (aka upsert)
+//
+// Invokes all lower-level methods as necessary to create a running function
+// whose source code and metadata match that provided by the passed
+// function instance.
+func (c *Client) Apply(ctx context.Context, f Function) (err error) {
+	if f, err = NewFunction(f.Root); err != nil {
+		return
+	}
+	if !f.Initialized() {
+		return c.New(ctx, f)
+	} else {
+		return c.Update(ctx, f.Root)
+	}
+}
+
+// Update function
+//
+// Updates a function which has already been initialized to run the latest
+// source code.
+//
+// Use Init, Build, Push and Deploy independently for lower level control.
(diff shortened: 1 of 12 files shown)


lance (commented):
This generally looks fine, but are we not concerned about breaking API changes? I'm not opposed to the name change for `Init()` but in theory, this should be aliased in code as well as on the CLI, if possible.

lkingland:
> 

I am not concerned _at this time_, but that will of course be an increasing concern as time and adoption progresses; at which time we can start using the semver standard of increasing the major version on any changes to the API.

> 

So yes I think that is a valid concern, but I consider the client api still a work-in progress, and doubt many are importing it at this time (though I could of course be wrong).  I consider its primary client to be the
CLI, which I am keeping in sync.

lkingland:
Also worth noting, we need to move the exported members of the concrete interface implementations (builder, deployer, etc) into an `internal` package such that they can not become external dependencies of other projects, retaining our ability to iterate on _their_ APIs without the same considerations!

lance:
> I am not concerned at this time, mainly beacuse we are not directing anyone, through guides or documentation, towards importing the client library yet.

Good point.

> at which time we can go 1.0 and start using the semver standard of increasing the major version on any breaking changes to the API, and using aliases or other tricks to avoid doing so too often.

Yes - but the tricky part is synchronizing with all of the other Knative components! Knative explicitly eschews semver. 😉 

> The core library's primary client is the CLI, which I am of course keeping in sync.

➕ 1️⃣

salaboy (approved):
lgtm

PR #1530, opened 2023-01-30
client methods return current function route

lkingland:
# Changes

- :gift: Client API returns actual function route
- :gift: Actual route now printed by CLI on deploy

Client methods Apply, Update and New now look up and return the function's current primary route.  The lower-level method Deploy remains unchanged as it is intended to be a mostly atomic operation which, at that level, would be used with Describe or using the client's instances manager.

/kind enhancement


client.go
```diff
@@ -789,7 +794,16 @@ func (c *Client) Route(path string) (err error) {
...
+	if err = c.dnsProvider.Provide(f); err != nil {
+		return
+	}
+
+	// Return the correct route.
+	instance, err := c.Instances().Remote(ctx, "", path)
+	if err != nil {
+		return
+	}
+	return instance.Route, nil
```
salaboy, line 806: @lkingland ok.. this PR includes the change but not who is using this change right? I can see that the route is not used .. yet..

salaboy (approved):
/lgtm

PR #1539, opened 2023-02-09
fix: func deploy with explicity `--build=false` bypass isBuilt check

### LINKED ISSUE #1535: On Cluster build failing on deploy task with `Error: not built`
On Cluster builds are failing on deploy task with message `Error: not built`.
On the `Deploy` implementation,`func` client now ensures the function is built and source files are not modified (since last built) by comparing a generated hash based on source file modified timestamps (stored on `.func/built` after build).

For on-cluster builds, the build task (buildpack and s2i) does not generate and save the hash on `.func/built` thus causing the following  task "deploy" to fail.

jrangelramos:
# Changes

- :bug: Fix error on build command `Error: not built` on situation where image is built but user explicity specifies `--build=false`

**Use cases it covers**
- On Cluster builds which image is build and deployed on different steps and build stamp is not generated
- Local builds when a user builds a function, then run locally (i.e `mvn quarkus:dev`) or user modifes a non func related file (i.e. README.md)

/kind bug

Fixes #1535

### DIFF
--- a/client.go
+++ b/client.go
@@ -708,8 +708,25 @@ func (c *Client) printBuildActivity(ctx context.Context) {
 	}()
 }
 
+type DeployParams struct {
+	skipBuiltCheck bool
+}
+type DeployOption func(f *DeployParams)
+
+func WithDeploySkipBuildCheck(skipBuiltCheck bool) DeployOption {
+	return func(f *DeployParams) {
+		f.skipBuiltCheck = skipBuiltCheck
+	}
+}
+
 // Deploy the function at path. Errors if the function has not been built.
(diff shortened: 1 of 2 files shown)


matejvasek:
@lance we will need to override tests here.

jrangelramos:
On cluster tests here fails because the [image](https://github.com/knative/func/blob/main/pipelines/resources/tekton/task/func-deploy/0.1/func-deploy.yaml#L27) it uses on deploy does not contains the fix yet. _The chicken and egg problem..._

lance:
/override "On Cluster Test (1.18.x, ubuntu-latest)"

PR #1543, opened 2023-02-12
[WIP] feat: add --pvc-size flag to deploy command (#851)

grafvonb:
Ref: https://github.com/knative/func/issues/851

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"os"
 
+	"knative.dev/func/cmd/prompt"
 	"knative.dev/func/pkg/builders/buildpacks"
 	"knative.dev/func/pkg/config"
 	"knative.dev/func/pkg/docker"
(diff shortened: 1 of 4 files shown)


matejvasek:
@grafvonb out of curiosity have you encounter problems with the default size for some builds?

grafvonb:
> @grafvonb out of curiosity have you encounter problems with the default size for some builds?

Hi, I will test it extensively on AWS for different tech stacks when I am ready, but the version I pushed in this feature branch already works (I want to change the error handling for the parameter and add tests).

However, I have already made interesting observations, for example, the default size of a pvc defined in the `gp2` storage class cannot be overwritten with a lower value (so the default value is actually the minimum value). Thus, our default value of 256Mi does not result in the creation of a pvc with that size on AWS. Instead, it creates a pvc with 1Gi, which is the default value for gp2 in AWS (unless explicitly changed by the administrator). I also tried a deployment with the value `--pvc-size=2Gi` and it was created as expected.

matejvasek:
@grafvonb this needs rebase.

PR #1548, opened 2023-02-13
Use current (as in PR) func image in integration tests

matejvasek:
# Changes

- Use current (as in PR) version of `func` in Tekton task during integration testing.
Without this change the `func` from `main` is used and changes in PR are not tested for on cluster build.


lance (commented):
/lgtm

matejvasek:
@lance @jrangelramos please re-lgtm.

Issue #1547, opened 2023-02-13
Refactor repository layout - make it more hierarchical


zroubalik:
The current repository layout is quite flat and sometimes it might be very confusing to find the correct code block. In the root of the https://github.com/knative/func directory there are packages related to `func` functionality, `.dotted` directories, vendor/thirdparty directories, metadata files test packages, generators and also `func` related files put directly in the root etc. Everything.

**Current state:**
```
.chglog
.codecov.yaml
.git
.gitattributes
.github
.gitignore
.golangci.yaml
.ko.yaml
.unicode-control-characters.config.py
.versionrc
.vscode
AUTHORS
CHANGELOG.md
CODE-OF-CONDUCT.md
LICENSE
Makefile
OWNERS
OWNERS_ALIASES
README.md
builders
buildpacks
client.go
client_int_test.go
client_test.go
cmd
config
config_test.go
docker
docs
filesystem.go
filesystem_test.go
function.go
function_envs.go
function_envs_unit_test.go
function_git.go
function_git_unit_test.go
function_labels.go
function_labels_unit_test.go
function_migrations.go
function_migrations_unit_test.go
function_options.go
function_options_unit_test.go
function_test.go
function_unit_test.go
function_volumes.go
function_volumes_unit_test.go
generate
go.mod
go.sum
hack
http
instances.go
instances_test.go
invoke.go
job.go
k8s
knative
mock
openshift
pipelines
plugin
progress
repositories.go
repositories_test.go
repository.go
repository_test.go
s2i
schema
sortedset.go
ssh
template.go
templates
templates.go
templates_test.go
test
testdata
testing
third_party
utils
vendor
zz_filesystem_generated.go
```

I'd like to refactor the repository and bring more hierarchy into it, I'd put source code related to `func` functionality into `pkg/...` as it is common with other Go based projects, making the root more simple and clear. Inside `pkg` directory I'd also create a new package for `client*.go`, `function*.go`, `repositor*.go` and `template*.go` etc.

**Desired state:**
```
.chglog
.codecov.yaml
.git
.gitattributes
.github
.gitignore
.golangci.yaml
.ko.yaml
.unicode-control-characters.config.py
.versionrc
.vscode
AUTHORS
CHANGELOG.md
CODE-OF-CONDUCT.md
LICENSE
Makefile
OWNERS
OWNERS_ALIASES
README.md
cmd
docs
generate
go.mod
go.sum
hack
pkg
plugin
schema
test
testdata
testing
third_party
[...]

PR #1554, opened 2023-02-14
chore: refactor repository layout

zroubalik:

I'd prefer if we can merge this PR asap, rebasing this will be very painful :))

Moved majority of code into `pkg`. This is the layout now:

```
$ tree -L 2 -d 
.
├── bin
├── cmd
│   ├── func
│   ├── templates
│   └── testdata
├── docs
│   ├── building-functions
│   ├── function-templates
│   ├── generator
│   ├── language-packs
│   └── reference
├── generate
│   └── templates
├── hack
├── pkg
│   ├── builders
│   ├── config
│   ├── docker
│   ├── filesystem
│   ├── function
│   ├── http
│   ├── k8s
│   ├── knative
│   ├── mock
│   ├── openshift
│   ├── pipelines
│   ├── progress
│   ├── ssh
│   ├── testing
│   └── utils
├── plugin
├── schema
│   └── generator
├── templates
│   ├── go
│   ├── node
│   ├── python
│   ├── quarkus
│   ├── rust
│   ├── springboot
│   └── typescript
├── test
│   ├── common
│   ├── e2e
│   └── oncluster
├── third_party
│   └── VENDOR-LICENSE
└── vendor
```

# Changes

- :broom: refactor repository layout

Fixes: #1547


lkingland:
I think the part I like about this the most is that, by default, we are back to there being a sane, Go-friendly default import path for users of the client library:
```
import knative.dev/func/pkg/function
...

function.XYZ
```

This is perhaps a bit pedantic, but since the singula makes it sound like they are expecting to operate on a single function, would it be better for the user if the key were plural?
```
functions.NewFunction?
functions.XYZ?
```
Overall huge improvement.  Yes' let's merge quickly to minimize collateral damage.

/lgtm
/hold for answer if anyone else thinks `functions` plural

lkingland (approved):

PR #1563, opened 2023-02-16
fix: ban podman 4.3 for pack build

matejvasek:
# Changes

- :bug: Ban `podman` 4.3 for pack build. The 4.3 version contains bug causing infinite wait.


matejvasek:
> Is this only an issue for buildpacks and not s2i?

@lance Yes.

lance (commented):
/lgtm

PR #1564, opened 2023-02-17
cli help text updates and flags cleanup

lkingland:
# Changes

- :gift: verbose flag uses global setting throughout
- :broom: confirm flag added using shared visitor throughout
- :broom: path flag added using shared visitor throughout
- :broom: removes --version flag on root as redundant with subcommand
- :broom: splits main help's 'Main Commands' into 'Primary Commands' and 'Development Commands' groups
- :broom: Other general command structure cleanup

/kind cleanup

### Refactor of root help text
  In order to cohere more with kn, the root help text has been reworked to be
  quite similar.  Subcommands were not (yet) updated.  Moving the lone
  global flag, verbose, to subcommands assisted in this homogenization

### Addition of the verbose flag directly to commands
  The complexity added by having a single global (persistent) flag was not
  worth the benefit of being able to elide defining it on each command.
  Furthermore, when used as a plugin (to kn for example) it is easy to be
  confused which set of global commands apply where.  It's clearly simpler
  to just `addVerbose(cmd)` right along with all the other flags.

### Renaming setXFlag functions to addXFlag
  setX usually implies that the object with context to the accessor will be
  mutaed (user.setName("bob"), etc).  Whereas addX resolves this ambiguity
  by implying the passed argument will be mutated.

### Removing the redundant --version flag
  Both to be more coherent with kn, and because it seems to be the emergent
  idiom for Go apps, the builtin --version flag was removed in favor of
  the extant subcommand.

### Ensuring --confirm and --path flags use accessor
  As stated above, there is not enough flag-usage overlap between commands
  to make global ("persistent") flags correct, but there are three which
  are used enough to make a helper function useful: verbose, path and confirm.
  These were added where missing throughout.

### Creation of the Primary Commands and Development Commands groups
  It is important to list the primary CRUD operations as the primary commands,
  so that new users understand the high-level lifecycle of a function.
  This creates that group, moving development/helper commands such as build,
  run and invoke to their own group.

**Release Note**

```release-note
Fixes issue where global settings for --verbose and --confirm were sometimes not considered
Removes the --version flag; please use the 'version' subcommand.
```


zroubalik (approved):
/lgtm

Great job!

/hold for others

cmd/root.go
```diff
@@ -68,19 +56,9 @@ EXAMPLES
...
-	cmd.PersistentFlags().BoolP("verbose", "v", false, "Print verbose logs ($FUNC_VERBOSE)")
-	if err := viper.BindPFlag("verbose", cmd.PersistentFlags().Lookup("verbose")); err != nil {
-		fmt.Fprintf(os.Stderr, "error binding flag: %v\n", err)
-	}
-
 	// Version
-	cmd.Version = cfg.Version.String()
-	cmd.SetVersionTemplate(`{{printf "%s\n" .Version}}`)
+	// cmd.Version = cfg.Version.String()
+	// cmd.SetVersionTemplate(`{{printf "%s\n" .Version}}`)
```
lance, line 61: Are these meant to be commented out?

cmd/run.go
```diff
@@ -43,7 +44,15 @@ to the function's source.  Use --build to override this behavior.
 
 `,
 		SuggestFor: []string{"rnu"},
-		PreRunE:    bindEnv("build", "path", "registry"),
+		PreRunE:    bindEnv("build", "path", "registry", "verbose"),
```
lance, line 47: What about making the help text on lines 20-21 use `container` similarly to the others. E.g. `run the function container locally`?

lance (commented):
Looks great overall - nice clean up. I just have a small suggestion and one question about the `version` values.

cmd/root.go
lkingland: Ah yes, thank you for noticing. 
I was hemming-and-hawing about whether or not to leave it in there.

cmd/run.go
lkingland: Yes, that's a good question:  because from the perspective of a Function developer, they are just running the function.  Additionally, we may want to consider updating the `run` command to support both running on the host or containerized.  So I am trying to be more deliberate when using the word "container" to only be for situations where it is explicitly exposing this concept to the user:  Build.  Of course, that's just my rationale, which indeed might be misguided.

lance: > Additionally, we may want to consider updating the `run` command to support both running on the host or containerized.

I like this idea. I wonder how it would be best manifested. Each runtime has their own way of starting the function locally - and some don't yet explicitly support this. Of course the buildpacks know these things e.g. `npm start` is common for Node.js - so we haven't really considered a common runtime "start" command that does things "natively". My first instinct is to say that it could be indicated in manifest.yaml with sane defaults for already-known/supported runtimes. But I'm aware there exist ways to deal with this question already, e.g. Procfiles. See for example https://github.com/knative/func/blob/main/templates/python/http/Procfile. Something like `func fun --native` would be really nice if we can work out how to best indicate that in a project's metadata - and again with sane defaults for built-in runtimes.

lkingland: Yes, this sounds a lot like what I was thinking of as well @lance.  For example:

Consider how templates work during the initialization phase, now imagine something almost exactly the same but during the build phase; let's call them 'scaffolding'.  Embedded into the binary are the scaffolding for language runtimes we natively support, and they could be included in the language repositories for extensibility.
When a Function is built (either from `build` or `deploy`), the contents of the scaffolding is written into `.func/runs/0/` (or we could use a temp directory, but I think developers might appreciate having it close at hand for inspection).  This scaffolding code consists of everything necessary to run their function _except_ the function code itself.  In the case of Go, this would be `main.go`, which would then import the user's Function code in the relative directory `./function`.  This would be (from the project root) `myFunction/.func/runs/0/function` which itself would be a simple symlink `-> ../../../`.  Included in the scaffolding would be the command, perhaps in the form of a Procfile, which would be able to actually invoke the code.  Upon completion of the run (when the process is canceled), the run directory `.func/runs/0` is removed unless the user specifically requests it be retained using, for example the `-k|--keep` flag.

This allows our Functions to remain dependency-free (and thus gain all the future benefits of a proper FaaS, such as auto-upgrades, etc) without losing the ability to run locally.  It also keeps things quite simple for developers who want to stay at a high level, working mostly in their libraries/code.  For those that want to take lower-level control, or to enable CI/CD to run these "pure" dependency-free functions, this process can be exposed as `func scaffold`:

Should a user decide they would rather not have our system's scaffolding, or take control of it themselves, they could run `func scaffold`, which would write the scaffolding code into the project source-controlled tree (perhaps `./run/...`) and a flag added to func.yaml such as `scaffold=false` indicating the Function developer has taken control of the code into their own hands (and will presumably check in the `/run` directory to source control) and the build/deploy commands will skip this step and instead invoke the Function using the process-wrapping code (main) in `./run`.

For environments like Pipelines, they need only `func scaffold` the Function source code once pulled from Git (no need to commit the result) to convert it into a form which it supports  (a Function wrapped in a process-creating boundary main).

Then, of course a `func run --host` would still scaffold as before, but it could skip creating the container and just straight to invoking the code in `./func/runs/0`.

This would have the added benefit of pulling all of our scaffolding into one repository where we can more easily iterate and test it as well.

lkingland: I think maybe this is a good [topic for a discussion](https://github.com/knative/func/discussions/1589)...

zroubalik (approved):
/lgtm

Great improvement!

Issue #1571, opened 2023-02-20
Annotations removed from `func.yaml` are not removed from deployed function


zroubalik:
1. Deploy function with this func.yaml:
```yaml
  annotations:
    foo: bar
    xxx: zzz
```
2. KSVC with deployed function has both annotations  `foo: bar` and  `xxx: zzz` set
3. Remove `xxx: zzz` from func.yaml and redeploy
```yaml
  annotations:
    foo: bar
```
4. KSVC with deployed function has still both annotations  `foo: bar` and  `xxx: zzz` set

PR #1574, opened 2023-02-20
fix: deploy annotations & labels correctly

zroubalik:
# Changes

- :broom: deploy annotations & labels correctly
- :broom: set timeout for integration test to 30mins

Fixes: https://github.com/knative/func/issues/1571

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ test-typescript: ## Test Typescript templates
 ###################
 
 test-integration: ## Run integration tests using an available cluster.
-	go test -tags integration --coverprofile=coverage.txt ./... -v
+	go test -tags integration -timeout 30m --coverprofile=coverage.txt ./... -v
 
 test-e2e: ## Run end-to-end tests using an available cluster.
 	./test/e2e_lifecycle_tests.sh node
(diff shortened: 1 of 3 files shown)


lkingland:
I've noticed that my commands reliably fail when testing against a local cluster with the same messages seen here: "unable to get logs from container",  but only after a bit of a timeout.  This is appreciably affecting the time it takes to run the integration tests, and may be the reason it is going past the 10 minute limit.
Short fix would be to just increase the timeout of course

zroubalik:
> I've noticed that my commands reliably fail when testing against a local cluster with the same messages seen here: "unable to get logs from container", but only after a bit of a timeout. This is appreciably affecting the time it takes to run the integration tests, and may be the reason it is going past the 10 minute limit. Short fix would be to just increase the timeout of course

@lkingland this make sense. What is the best place to set the timeout?

lkingland:
> @lkingland this makes sense. What is the best place to set the timeout?

I think the go test command in the make target for integration requests.  See PMs :)

zroubalik:
> > @lkingland this makes sense. What is the best place to set the timeout?
> 
> I think the go test command in the make target for integration requests. See PMs :)

Done that :)

PR #1575, opened 2023-02-21
fix: missing map initializations

lkingland:
- :bug: both function constructors now ensure maps are initialized

The helper Function constructor which uses a passed function as defaults, `NewFunctionWith` was missing map initialization.  This could lead to nil pointer exceptions as it is used by `client.Init`.

/kind bug


zroubalik (approved):
/lgtm

PR #1598, opened 2023-03-06
feat: add pvc-size flag to deploy command (#851)

grafvonb:
Ref: #851

Second try, after recreating the changes on the new project structure (old PR: https://github.com/knative/func/pull/1543)

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -166,7 +166,7 @@ EXAMPLES
 	cmd.Flags().Bool("remote", f.Deploy.Remote,
 		"Trigger a remote deployment. Default is to deploy and build from the local system (Env: $FUNC_REMOTE)")
 	cmd.Flags().String("pvc-size", fn.DefaultPersistentVolumeClaimSize,
-		"Configure the PVC size used by a pipeline. This flag can only be set if --remote flag is used.")
+		"Configure the PVC size used by a pipeline during remote build.")
 
 	// Static Flags:
 	// Options which have static defaults only (not globally configurable nor
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -427,7 +427,7 @@ func (c deployConfig) Configure(f fn.Function) (fn.Function, error) {
 	if err != nil {
 		return f, fmt.Errorf("cannot parse the provided PVC size '%s' due to: %w", c.PVCSize, err)
 	}
-	f.Deploy.PVCSize = c.PVCSize
+	f.Build.PVCSize = c.PVCSize
 
 	// ImageDigest
 	// Parsed off f.Image if provided.  Deploying adds the ability to specify a
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -10,14 +10,15 @@ import (
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/ory/viper"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"knative.dev/client/pkg/util"
-
 	"knative.dev/func/pkg/builders"
 	"knative.dev/func/pkg/builders/buildpacks"
 	"knative.dev/func/pkg/builders/s2i"
 	"knative.dev/func/pkg/config"
 	fn "knative.dev/func/pkg/functions"
 	"knative.dev/func/pkg/k8s"
+	"knative.dev/func/pkg/pipelines/tekton"
 )
 
 func NewDeployCmd(newClient ClientFactory) *cobra.Command {
@@ -79,7 +80,7 @@ DESCRIPTION
 EXAMPLES
 
 	o Deploy the function
-	  $ {{.Name}} deploy
+	  $ {{rootCmdUse}} deploy
 
 	o Deploy the function using interactive prompts. This is useful for the first
 	  deployment, since most settings will be remembered for future deployments.
@@ -116,7 +117,7 @@ EXAMPLES
 
 `,
 		SuggestFor: []string{"delpoy", "deplyo"},
-		PreRunE:    bindEnv("confirm", "env", "git-url", "git-branch", "git-dir", "remote", "build", "builder", "builder-image", "image", "registry", "push", "platform", "namespace", "path", "verbose"),
+		PreRunE:    bindEnv("confirm", "env", "git-url", "git-branch", "git-dir", "remote", "build", "builder", "builder-image", "image", "registry", "push", "platform", "namespace", "path", "verbose", "pvc-size"),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runDeploy(cmd, newClient)
 		},
@@ -137,8 +138,8 @@ EXAMPLES
 	// Flags
 	//
 	// Globally-Configurable Flags:
-	//   Options whose value may be defined globally may also exist on the
-	//  contextually relevant function; but sets are flattened via cfg.Apply(f)
+	// Options whose value may be defined globally may also exist on the
+	// contextually relevant function; but sets are flattened via cfg.Apply(f)
 	cmd.Flags().StringP("builder", "b", cfg.Builder,
 		fmt.Sprintf("Builder to use when creating the function's container. Currently supported builders are %s.", KnownBuilders()))
 	cmd.Flags().StringP("registry", "r", cfg.Registry,
@@ -147,10 +148,10 @@ EXAMPLES
 		"Deploy into a specific namespace. Will use function's current namespace by default if already deployed, and the currently active namespace if it can be determined. (Env: $FUNC_NAMESPACE)")
 
 	// Function-Context Flags:
-	// Options whose value is avaolable on the function with context only
+	// Options whose value is available on the function with context only
 	// (persisted but not globally configurable)
 	builderImage := f.Build.BuilderImages[f.Build.Builder]
-	cmd.Flags().StringP("builder-image", "", builderImage,
+	cmd.Flags().String("builder-image", builderImage,
 		"Specify a custom builder image for use by the builder other than its default. ($FUNC_BUILDER_IMAGE)")
 	cmd.Flags().StringP("image", "i", f.Image, "Full image name in the form [registry]/[namespace]/[name]:[tag]@[digest]. This option takes precedence over --registry. Specifying digest is optional, but if it is given, 'build' and 'push' phases are disabled. (Env: $FUNC_IMAGE)")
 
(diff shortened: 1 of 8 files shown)


grafvonb:
Getting currently a little bit strange panic during an old UT (in previous version it was ok):
```
--- FAIL: TestDeploy_BuilderPersists (0.01s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1c38bca]
``` 
Needs investigation.

cmd/deploy.go
```diff
@@ -163,18 +165,18 @@ EXAMPLES
...
 	// persisted with the function)
-	cmd.Flags().StringP("build", "", "auto",
+	cmd.Flags().String("build", "auto",
 		"Build the function. [auto|true|false]. (Env: $FUNC_BUILD)")
 	cmd.Flags().Lookup("build").NoOptDefVal = "true" // register `--build` as equivalient to `--build=true`
 	cmd.Flags().BoolP("push", "u", true,
 		"Push the function image to registry before deploying. (Env: $FUNC_PUSH)")
-	cmd.Flags().StringP("platform", "", "",
+	cmd.Flags().String("platform", "",
 		"Optionally specify a specific platform to build for (e.g. linux/amd64). (Env: $FUNC_PLATFORM)")
```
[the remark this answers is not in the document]
matejvasek: But it fails on
```
=== RUN   TestDeploy_BuilderPersists
Error: cannot parse the provided PVC size '' due to: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
Usage:
```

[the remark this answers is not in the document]
matejvasek:
```go
cmd.Flags().String("pvc-size", "256M", "")
```
could work

cmd/deploy.go
```diff
@@ -163,18 +165,18 @@ EXAMPLES
...
+	// Options which have static defaults only (not globally configurable nor
 	// persisted with the function)
-	cmd.Flags().StringP("build", "", "auto",
+	cmd.Flags().String("build", "auto",
 		"Build the function. [auto|true|false]. (Env: $FUNC_BUILD)")
 	cmd.Flags().Lookup("build").NoOptDefVal = "true" // register `--build` as equivalient to `--build=true`
 	cmd.Flags().BoolP("push", "u", true,
 		"Push the function image to registry before deploying. (Env: $FUNC_PUSH)")
-	cmd.Flags().StringP("platform", "", "",
+	cmd.Flags().String("platform", "",
```
matejvasek, line 179: Isn't this breaking anything?
```
=== RUN   TestDeploy_BuilderPersists
Error: Only S2I builds currently support specifying platform
```

matejvasek: It seems this is all right.

cmd/deploy.go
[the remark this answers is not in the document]
grafvonb: @matejvasek Wow, thank you, I had to be tired yesterday. It was an error when transferring code from the version 1. My new project makes me \o/.

matejvasek:
Something is wrong with build timestamps.

matejvasek:
```
touch func.yaml && /home/mvasek/devel/kn-plugin-func/func build --path /tmp/test-config-labels --registry localhost:50000/user --builder pack && /home/mvasek/devel/kn-plugin-func/func deploy --path /tmp/test-config-labels --registry localhost:50000/user --build=false -v
   🙌 Function image built: localhost:50000/user/test-config-labels:latest
Error: not built
```

matejvasek:
> ```
> touch func.yaml && /home/mvasek/devel/kn-plugin-func/func build --path /tmp/test-config-labels --registry localhost:50000/user --builder pack && /home/mvasek/devel/kn-plugin-func/func deploy --path /tmp/test-config-labels --registry localhost:50000/user --build=false -v
>    🙌 Function image built: localhost:50000/user/test-config-labels:latest
> Error: not built
> ```

It appears that build sets buildStamp at the end of build but it simultaneously creates .gitignore (which affects final stamp). There is apparently some race because sometimes the modified .gitignore is taken into account and sometime is's not.

I don't see how this PR is causing this.

matejvasek:
See https://github.com/knative/func/pull/1608

matejvasek:
@grafvonb  Now I see similar issue is happening with `func.yaml` it is also (similarly to `.gitignore`) modified between build and deploy.

matejvasek:
> @grafvonb Now I see similar issue is happening with `func.yaml` it is also (similarly to `.gitignore`) modified between build and deploy.

In deploy step we are updating `func.yaml` just before running `buildStamp`. Sometimes the write to FS is delayed so it works sometimes it is fast so stamp is not matching.

grafvonb:
> > @grafvonb Now I see similar issue is happening with `func.yaml` it is also (similarly to `.gitignore`) modified between build and deploy.
> 
> In deploy step we are updating `func.yaml` just before running `buildStamp`. Sometimes the write to FS is delayed so it works sometimes it is fast so stamp is not matching.

Can I re-run the pipeline without changing anything or faking with `--amend` of the last commit?

matejvasek:
> > > @grafvonb Now I see similar issue is happening with `func.yaml` it is also (similarly to `.gitignore`) modified between build and deploy.
> > 
> > 
> > In deploy step we are updating `func.yaml` just before running `buildStamp`. Sometimes the write to FS is delayed so it works sometimes it is fast so stamp is not matching.
> 
> Can I re-run the pipeline without changing anything or faking with `--amend` of the last commit?

I suppose you could, or you could add empty commit. But I don't think it will pass. If it was single test then maybe, but there are multiple and do not think that odds are high that all test will pass.

matejvasek:
Could you look into it more closely?

matejvasek:
@grafvonb please try this.
```diff
diff --git a/cmd/deploy.go b/cmd/deploy.go
index 73a3cc0d..8823cccd 100644
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -429,10 +429,13 @@ func (c deployConfig) Configure(f fn.Function) (fn.Function, error) {
 	// ImageDigest
 	// Parsed off f.Image if provided.  Deploying adds the ability to specify a
 	// digest on the associated image (not available on build as nonsensical).
-	f.ImageDigest, err = imageDigest(f.Image)
+	digest, err := imageDigest(f.Image)
 	if err != nil {
 		return f, err
 	}
+	if digest != "" {
+		f.ImageDigest = digest
+	}
 
 	// Envs
 	// Preprocesses any Envs provided (which may include removals) into a final
```

matejvasek:
@grafvonb tests are failing because `func deploy` sets `pvc-size`  before computing build stamp.

matejvasek:
Also why is `pvc-size` under `deploy` section?
It it really `build` variable.

Failing test could be fixed by initialising `pvc-size`  on function creation, so the `deploy` step is not causing any modifications.

grafvonb:
@matejvasek 
> Also why is `pvc-size` under `deploy` section? It it really `build` variable.

Good question and I agree. 

I put `pvc-size` into the `deploy` section because it depends on the `remote`-flag which is, if we follow your comment, imho also a `build` variable even it the help says "Trigger a remote deployment.  Default is to deploy and build from the local system". So at the end they are both listed in the same block in the `func.yaml`:

```
deploy:
  namespace: ""
  remote: true
  pvc-size: 256Mi
  annotations: {}
  options: {}
  labels: []
  healthEndpoints:
    liveness: /health/liveness
    readiness: /health/readiness
```

I can move `pvc-size` to the `build` section but it should be done for the `remote` also which probably means migration of the existing configurations.

Regardless this problem, we can expected a similar situation in the future with a parameters that really belong to the `deploy` section so this should be solved also.

matejvasek:
>I moved the pvc-size to the build section however a e2e error is still there with not built message,

I did not meant it would fix anything, it is just in my opinion more logical pace for it.

matejvasek:
What could fix it is that newly created function will have `pvc-size` pre-set.

matejvasek:
`func create` should create `func.yaml` such that it already contains `pvc-size: 256Mi`.

grafvonb:
> Another option would be to implement semantic hashing for `func.yaml` so only variables affecting image build are hashed to `buildStamp`.

That sounds reasonable. Btw. I just added setting default value for `pvc-size` as suggested.

matejvasek:
@lance @grafvonb Note: we should not forget to update schema in registry on next release.

matejvasek:
The failing e2e passed now I think.

matejvasek:
@lance @zroubalik PTAL

grafvonb:
> The failing e2e passed now I think.

Yup, but I need to add some tests now, due to the new code for default value etc.

grafvonb:
codecov seems to be ok, so it's ready for review and merge

matejvasek:
@grafvonb you will need rebase and rebuild schema because of: https://github.com/knative/func/pull/1633

cmd/deploy.go
```diff
@@ -79,7 +79,7 @@ DESCRIPTION
 EXAMPLES
 
 	o Deploy the function
-	  $ {{.Name}} deploy
+	  $ {{rootCmdUse}} deploy
```
matejvasek, line 82: Why has this changed?

pkg/functions/function.go
```diff
@@ -96,6 +100,10 @@ type BuildSpec struct {
 
 	// Build Env variables to be set
 	BuildEnvs []Env `yaml:"buildEnvs"`
+
+	// PVCSize specifies the size of persistent volume claim used to store function
+	// when using deployment and remote build process (only relevant when Remote is true).
+	PVCSize string `yaml:"pvcSize"`
```
lance, line 106: Can you add `omitempty` here?

matejvasek: Actually not.

matejvasek: Well I am not entirely sure now I think about it. We always default it right @grafvonb , so it is never empty.

grafvonb:
"Analyse" check fails due to:
```
Errors during downloading metadata for repository 'baseos':
    - Downloading successful, but checksum doesn't match. Calculated: 67f0fad4fe457471c91c70a50245e5700a98c4b49d509d35a8b6643eed55e789b3afb5ced888da13b63a3bdfd122e7aecc8b6c485a9dedcd1afca23cf81fa288(sha512)  Expected: 60642857dc1e825a77340c6ffe13eb4c84f0c3c1972dd4dfc7ab8e00c6b1980f63a5d331bd97f5a9808466c657325a1063[18](https://github.com/knative/func/actions/runs/4615303335/jobs/8159049308?pr=1598#step:2:18)40267e80cfb5a87b1adb28d52947(sha512)
```
@matejvasek Who can repair this?

matejvasek (approved):

PR #1602, opened 2023-03-07
bug: Correctly format Node.js code

### LINKED ISSUE #1528: Found 2 spaces of non-formatted JS code
After creating the function with `func create -l node -t http foobar`

https://github.com/obs-nebula/function-five/blob/main/two/index.js#L27-L28

manoelmarques:
# Changes

- :bug: Correctly format Node.js code

/kind bug

Fixes #1528


matejvasek:
@manoelmarques when modifying `./templates/**` you also need re-generate and commit `generate/zz_filesystem_generated.go` file.
```
make generate/zz_filesystem_generated.go
```

matejvasek:
Thanks! @manoelmarques

matejvasek:
https://github.com/knative/func/actions/runs/4357921420/jobs/7617882795
Your OS probably added some file to "./templates/" dir.

matejvasek:
@manoelmarques `.DS_Store` still seems to be there.

matejvasek:
@manoelmarques 
for future I recommend adding:
```sh
#!/bin/sh

if git diff-index --cached --name-only HEAD | grep -i "^templates/.*$" >/dev/null; then
  if ! git diff-index --cached --name-only HEAD | grep -i "^generate/zz_filesystem_generated.go$" >/dev/null; then
    echo "WARNING: You are about to commit changes to the templates directory," \
      "but the generated generate/zz_filesystem_generated.go file is not staged."
    echo "If this is intentional use '--no-verify' flag."
    exit 1
  fi
fi

if git diff-index --cached --name-only HEAD | grep -i "^generate/zz_filesystem_generated.go$" >/dev/null; then
  UNVERSIONED=$(git ls-files --others --exclude-standard --ignored -- templates/ 2>/dev/null)
  if [ -n "$UNVERSIONED" ]; then
    echo "WARNING: You are about to commit generate/zz_filesystem_generated.go," \
      "but the templates directory contains some unversioned files" \
      "that may be unintentionally included in generate/zz_filesystem_generated.go"
    for f in $UNVERSIONED; do
      echo "    $f"
    done
    echo "If this is intentional use '--no-verify' flag."
    exit 1
  fi
fi

exit 0
```
into `.git/hooks/pre-commit`

manoelmarques:
Thanks, I will add the hooks and next time I change templates, verify if .DS_Store exists in the templates folder or use Linux  :-)

PR #1603, opened 2023-03-07
chore: coverage for e2e tests

matejvasek:
# Changes

- :gift: Coverage for the end to end tests. With Go 1.20+ it is possible to instrument final binary to gather coverage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,14 +144,19 @@ test-typescript: ## Test Typescript templates
 test-integration: ## Run integration tests using an available cluster.
 	go test -tags integration -timeout 30m --coverprofile=coverage.txt ./... -v
 
-test-e2e: ## Run end-to-end tests using an available cluster.
+.PHONY: func-instrumented
+
+func-instrumented: ## Func binary that is instrumented for e2e tests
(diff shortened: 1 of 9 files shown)


matejvasek:
PTAL @jrangelramos

Makefile
```diff
@@ -144,14 +144,19 @@ test-typescript: ## Test Typescript templates
 test-integration: ## Run integration tests using an available cluster.
 	go test -tags integration -timeout 30m --coverprofile=coverage.txt ./... -v
 
-test-e2e: ## Run end-to-end tests using an available cluster.
+.PHONY: func-instrumented
+
+func-instrumented: ## Func binary that is instrumented for e2e tests
+	env CGO_ENABLED=1 go build -ldflags $(LDFLAGS) -cover -o func ./cmd/func
```
matejvasek, line 150: I also thought about adding `-race` but that breaks tests.
We have race condition in code but I believe it's benign.

matejvasek:
> In general great contribution!
> 
> knative/client is still on 1.18, shouldn't we be in sync?
> 
> CC @dsimansk

Is 1.18 even supported now?

PR #1608, opened 2023-03-08
fix: build stamp computation

matejvasek:
# Changes

- :bug: Fix build stamp computation.
About the same time that `buildStamp` was computed `.gitignore` was (unnecessary) modified.
This caused timestamp mismatch so function appeared to be not built.
Reason: sometimes (randomly) the `buildStamp` was computed before the `.gitignore` was modified.

/kind bug

```release-note
fix: build stamp computation
```

### DIFF
--- a/pkg/functions/client.go
+++ b/pkg/functions/client.go
@@ -617,12 +617,20 @@ func ensureRuntimeDir(f Function) error {
 		return err
 	}
 
+	_, err := os.Stat(".gitignore")
+	if err == nil {
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


matejvasek:
@lkingland @zroubalik PTAL

matejvasek:
@lkingland why is `ensureRuntimeDir()` even called upon `buildStamp` computation? Shouldn't be the existence ensured on project creation?

lkingland (approved):

PR #1612, opened 2023-03-09
fix: node s2i build when node_modules present

matejvasek:
# Changes

- :bug: Fix NodeJS s2i build when `node_modules` is present on filesystem.
  This is workaround for two bug in another components:
    * The s2i CLI/library is not honoring the `--exclude` flag when used with the `--as-dockerfile` flag.
        https://github.com/openshift/source-to-image/issues/1104
    * The node s2i image is not working if project contains `node_modules` directory with NodeJS modules.
        https://github.com/sclorg/s2i-nodejs-container/issues/333
  
  If only one of the bugs above were fixed this PR wouldn't be necessary.

/kind bug

```release-note
fix: s2i build when node_modules present
```

### DIFF
--- a/pkg/builders/s2i/builder.go
+++ b/pkg/builders/s2i/builder.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/docker/docker/api/types"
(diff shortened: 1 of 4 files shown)


matejvasek:
@zroubalik @lkingland PTAL

lkingland (approved):
/lgtm

PR #1614, opened 2023-03-13
Include --builder option on func run command

### LINKED ISSUE #1609: Include `--builder` option on `func run` command
Here a suggestion for enhancement.

A function build (builder process and image creation) might be executed by one the following commands `build`, `run` or `deploy`.

The `--builder` option is only available for `func build` and `func deploy`. It would be nice (and handy) to also let user select the builder strategy (pack, s2i) on `run` command as well.

manoelmarques:
# Changes

- :gift: Include --builder option on func run command

/kind api-change

Fixes #1609

**Release Note**

```release-note
A new option "--builder" added to "run" command to be used when building.
Default is "pack"
```

**Docs**

```docs

```

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -10,6 +10,9 @@ import (
 	"github.com/spf13/cobra"
 	"knative.dev/client/pkg/util"
 
+	"knative.dev/func/pkg/builders"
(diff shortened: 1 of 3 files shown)


lkingland (approved):
Looks good to me; thanks for adding

/lgtm

PR #1615, opened 2023-03-13
e2e tests restructuring / refactoring

jrangelramos:
# Changes

- :broom: Refactoring and simplification of existing e2e tests
  - Removed duplicates, re-managed packages
  - Refined test files naming pattern to match on cluster tests

-  :broom: S2I Tests
Previous test is removed. Now covered by Lifecycle/Runtime on HTTP and CloudEvents Tests

-  :broom: Invoke Test
Previous test is removed. The `invoke` is exercised by a new Extended Flow scenario.

-  :broom: Lifecycle (Language Runtime) tests
    -  Previous test runs for default builder only using a single language runtime: 
    `create` > `deploy` > `list` > `describe` > update (de-reploy) > delete
    -  New test is executed for all runtimes and builders (**pack**, **s2i**) using only primary commands: 
    `create` > `deploy` > `delete` (other commands has been moved a different test scenario)

- :gift: Added a new test case (Extended Flow) which exercises other secondary commands as `invoke`, `list`, `describe`, `run` as well as local runs:
`create` > `build` > `run` > `invoke`  > `deploy` > `describe` > `list` > `invoke` >  re-deploy (new revision) > `delete`


test/e2e/scenario_basic_flow_test.go
```diff
@@ -0,0 +1,173 @@
...
+	// ---------------------------
+	knFunc.Exec("create", "--language", "node", "--template", "http", funcPath)
+
+	// From here on, all commands will be executed from the func project path
+	knFunc.SourceDir = funcPath
+
+	// ---------------------------
+	// Func Build Test
+	// ---------------------------
+	knFunc.Exec("build", "--registry", common.GetRegistry())
```
lkingland, line 45: A user should generally not need to/want to run 'build'.  This command is only used for the edge-case when a developer just wants the container to be built for some reason.  The `deploy` command should (almost) always do what they want:

Primary Flow:
`create`-> `deploy` -> `curl` -> `delete`
I would perhaps put the explicit 'build' command down near the bottom since it is an optional helper like `run`, `invoke` or `status`

jrangelramos: Hello Luke! Correct! Perhaps I can change the name of the scenario to "extended" or "complete" flow (the name `basic` trully does not reflect the purpose of this test).

The idea is to have one more comprehensive flow that walks thru developer local experience (build,run) as well as deployment on cluster experience (deploy, update, re-deploy, check,...).

It would also cover/replace previous Invoke test (by testing against a function running locally and deployed on cluster) as well as give at least one shot test other (less used) such as `list` and `info` (previously on runtime test).

The basic primary commands (`create` > `deploy` > `curl` > `delete`) as you mentioned I've been thinking in leave in the runtime (lifecycle) test, for both http and cloudevents. https://github.com/knative/func/pull/1615/commits/974a37132778d03e25c1b74899d7f9eb23ee3b96#diff-00b3b4a4cedefae054b320bcd68e937d94eb09caafa1dcde3877cf12fc484542R69 

I'm putting together a doc with the proposal and explaining these changes in comparison with existing tests and I shall share here soon.

jrangelramos:
Closes #1625

test/e2e/scenario_config-envs_test.go
```diff
@@ -94,21 +95,20 @@ func TestConfigEnvs(t *testing.T) {
...
-	project.Builder = builders.Pack
+	funcName := "test-config-envs"
+	funcPath := filepath.Join(t.TempDir(), funcName)
 
-	Create(t, knFunc.TestShell, project)
-	defer func() { _ = project.RemoveProjectFolder() }()
+	knFunc.TestCmd.Exec("create",
+		"--language", "go",
+		"--template", "envs",
+		"--repository", "http://github.com/boson-project/test-templates.git", // TODO Make on config
```
lance, line 109: I don't really love that we are still using this boson-project repo, but I'm not sure what the best alternative would be. I don't really want to pollute knative-sandbox with more repos if we can avoid it.

test/e2e_extended_tests.sh
```diff
@@ -45,7 +45,7 @@ fi
 export E2E_FUNC_BIN_PATH=$(pwd)/func
 
 go clean -testcache
-go test -v -test.v -tags="e2e" ./test/e2e/
+go test -v -test.v -test.timeout=60m -tags="e2e" ./test/e2e/
```
lance, line 48: I guess it does take a long time for these tests to run. One hour is a long timeout though. 🤷

lance (commented):
/lgtm

test/e2e_extended_tests.sh
jrangelramos: Well.. actually it takes around 15 to 20min most... I think I overestimated it (for safe)... maybe I can decrease it to 30m(?)..

test/e2e/scenario_config-envs_test.go
jrangelramos: yea.. I agree!!  I think I have an idea I can try on a different PR.. like using that custom git server (just as oncluster build tests) and pull from there..

PR #1616, opened 2023-03-13
chore: more tests on push to main

matejvasek:


matejvasek:
PTAL @lance

lance (commented):
Great addition @matejvasek - 👍 
/lgtm

PR #1619, opened 2023-03-14
feat: enable and disable features by keyword

### LINKED ISSUE #1566: Dapr annotations always set
The `dapr.io` annotations are now always set, on the `KSVC` metadata:
https://github.com/knative/func/blob/main/pkg/knative/deployer.go#L394-L398

While these have no effect, it does feel a bit verbose that they are always there, given that the majority of KSVC might not use Dapr at all. It would be good if this could be enabled only with a flag or so

lkingland:
# Changes

- :gift: enable and disable ad-hoc features via --enable and --disable flags
- :gift: defaults globally configurable
- :broom: Dapr annotations no longer added by default

```
func deploy --enable=dapr
```

General-purpose allowing the system to be open to enable ad-hoc features of concrete implementations while core stays closed to modification.

/kind enhancement

Closes #1566

**Release Note**

```release-note
Dapr runtime integration can be enabled on a per-function basis with 'func deploy --enable dapr'
Dapr integration can be disabled with 'func deploy --disable dapr'
```

**Docs**

```docs

```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -32,7 +32,7 @@ SYNOPSIS
 	{{rootCmdUse}} deploy [-R|--remote] [-r|--registry] [-i|--image] [-n|--namespace]
 	             [-e|env] [-g|--git-url] [-t|git-branch] [-d|--git-dir]
 	             [-b|--build] [--builder] [--builder-image] [-p|--push]
-	             [--platform] [-c|--confirm] [-v|--verbose]
+	             [--platform] [--enable] [--disable] [-c|--confirm] [-v|--verbose]
 
 DESCRIPTION
 
@@ -76,6 +76,12 @@ DESCRIPTION
 	  of a git repository instead of local source, combine with '--git-url':
 	  '{{rootCmdUse}} deploy --remote --git-url=git.example.com/alice/f.git'
 
+	Enable Integrations
+	  To enable additional integrations supported by the target cluster, use
+	  the --enable flag.  For exampe, if the target cluster has the Dapr runtime
+	  installed, enable the integration on the deployed function using the
+	  flag --enable=dapr.
+
 EXAMPLES
 
 	o Deploy the function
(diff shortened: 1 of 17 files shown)


pkg/functions/function.go
```diff
@@ -72,6 +72,11 @@ type Function struct {
 
 	//DeploySpec define the deployment properties for a function
 	Deploy DeploySpec `yaml:"deploy"`
+
+	// Enable extensible integrations
+	// These values are ad-hoc and opaque from the view of the functions core
+	// used to comunicate keys through to concrete implementations.
+	Enable []string `yaml:"enable,omitempty"`
```
zroubalik, line 79: What about we name this `Extension` ?

cmd/deploy.go
```diff
@@ -380,6 +390,12 @@ type deployConfig struct {
 	// Remote indicates the deployment (and possibly build) process are to
 	// be triggered in a remote environment rather than run locally.
 	Remote bool
+
+	// Enable additional integrations
+	Enable []string
```
zroubalik, line 395: What if we use `extensions` here and to mark it to be disable use a notation with `-` prefix? There is similar approach is used with envs if I am not mistaken.

Enable/Disable sounds too generic to me and might be confusing.

cmd/deploy.go
```diff
@@ -437,12 +465,27 @@ func (c deployConfig) Configure(f fn.Function) (fn.Function, error) {
...
 	// Preprocesses any Envs provided (which may include removals) into a final
 	// set
+	// TODO: remove pointers and use applyVector(f.Run.Envs, c.Env)
 	f.Run.Envs, err = applyEnvs(f.Run.Envs, c.Env)
 	if err != nil {
 		return f, err
 	}
 
-	// .Revision
+	// Enabled Integrations
```
zroubalik, line 474: I'd vote for keeping the `-` notation, it is being used with kn cli as well.

cmd/deploy.go
```diff
@@ -437,12 +465,27 @@ func (c deployConfig) Configure(f fn.Function) (fn.Function, error) {
...
+	// Can disable using the '-' suffix like envs for compatibility, but this is
+	// gently undocumented externally, choosing to isntead prefer the more
```
zroubalik, line 476:
```suggestion
	// gently undocumented externally, choosing to instead prefer the more
```

zroubalik (commented):
Great addition, I have a minor comment regarding the naming of the property.

pkg/functions/function.go
salaboy: @zroubalik @lkingland  Maybe feature extensions?

salaboy:
@lkingland I've added just a comment and I see that @zroubalik has some open comments too but this looks good to me.. unfortunately, this needs rebase already.. sorry for the delay in reviewing this.

cmd/deploy.go
lance: It looks like the `-` notation is still used internally. Personally, I don't mind the syntactic sugar of `enable` / `disable`. It is generic, yes, but that may also be a "good thing".

lance (commented):
/lgtm
but 
/hold
pending resolution of @zroubalik's comments

PR #1630, opened 2023-03-16
feat: Minimal initial func.yaml

### LINKED ISSUE #1279: Minimal initial func.yaml
## Summary

All fields on the `Function` struct should be set to `omitempty`

## Exposition

When a user creates a new Function, their `func.yaml` should include only the values which are either required (language), have no default (created date/time stamp) or deviate from static defaults (`func create -l go -t cl.

With the current version of the functions core library, this would _ideally_ look as such:
```shell
>  mkdir f && cd f && func create -l go
Created go function in [path]

> cat func.yaml
specVersion: 0.25.0
runtime: go
created: 2022-09-26T23:06:48.442316+09:00
```

## Alternative Considered: The fully-populated Config File approach
 
A traditional "config file" style setup would be where the func.yaml is fully populated with all possible values (commented and showcasing static defaults), whereby the user can choose to uncomment a given setting.  However, this usage pattern is directly at odds with the usage pattern we are encouraging: CLI or direct API.  For example, we using our friend `git` as an example, only when configuration is changed using `git config ...` does the `.git/config` file change.  This is the opposite approach to, for example `sshd`, which upon installation provides a comprehensive config file available at `/etc/ssh/sshd_config`.  The reason for the difference is that the former has a CLI which provides for configuration, while the latter does not.  We have a CLI.

This is of course up for debate.  It might be nice to account for a user who prefers  a `func.yaml`-first approach.  That may very well happen in the future, but this minimal approach should probably still be the first step.  We currently have the worst of both worlds, were func.yaml is populated with an intimidating litany of uncommented zero values.

## Caviat: It won't be that simple right now

Even if all fields are set to `omitempty`, the func.yaml will still include several more values.  This is a result of having not been strict in our usage of in-code static defaults early on, and requiring a few values we don't really need, so there are still a few fields which are written to every func.yaml despite not being actually deviations from the default.  The closest we can get _at present_ would be:

```
specVersion: 0.25.0
name: f
runtime: go
healthEndpoints:
  liveness: /health/liveness
  readiness: /health/readiness
buildpacks:
- paketo-buildpacks/go-dist
- ghcr.io/boson-project/go-function-buildpack:tip
created: 2022-09-26T23:06:48.442316+09:00
invocation:
  format: http
```

## Related

See forthcoming issues which request that we rely on the static defaults for buildpacks and health endpoints, only require a name upon deploy, and flatten the `invocation` struct (also relying on `http` as the static default)

manoelmarques:
# Changes

- :gift: Add `omitempty` flag to fields in Function structure in order to create a minimal func.yaml. This change doesn't address `buildpacks` that should be dealt with in a separate PR.

/kind enhancement

Fixes #1279

### DIFF
--- a/pkg/functions/function.go
+++ b/pkg/functions/function.go
@@ -29,10 +29,10 @@ type Function struct {
 	Root string `yaml:"-"`
 
 	// Name of the function.
-	Name string `yaml:"name" jsonschema:"pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"`
+	Name string `yaml:"name,omitempty" jsonschema:"pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"`
 
 	// Runtime is the language plus context.  nodejs|go|quarkus|rust etc.
-	Runtime string `yaml:"runtime"`
+	Runtime string `yaml:"runtime,omitempty"`
 
 	// Template for the function.
 	Template string `yaml:"-"`
@@ -53,7 +53,7 @@ type Function struct {
 	Image string `yaml:"image"`
 
 	// SHA256 hash of the latest image that has been built
-	ImageDigest string `yaml:"imageDigest"`
+	ImageDigest string `yaml:"imageDigest,omitempty"`
 
 	// Created time is the moment that creation was successfully completed
 	// according to the client which is in charge of what constitutes being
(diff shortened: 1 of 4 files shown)


matejvasek:
@manoelmarques are these Lint errors valid?

manoelmarques:
@matejvasek, It is very weird because until yesterday it never gave those errors and they don’t happen locally. It seems to get confused with types within the same package defined in other files.

matejvasek:
@lance please override the lint errors.

matejvasek:
@manoelmarques please try rebasing on `main` to resolve the error.

lkingland (approved):
/lgtm

👍🏻

PR #1640, opened 2023-03-20
Replace path with function struct in client api

### LINKED ISSUE #1283: Client API Function Arguments
Most of the methods on Client accept a _path_ to a Function rather than an _instance_ of a Function:
```go
func (c *Client) Build(ctx context.Context, path string) (err error) {
```
Thes methos should all be changed to accept a Function instance:
```go
func (c *Client) Build(ctx context.Context, f Function) (f Function, err error) {
```

This will allow for the removal of many `f.Write()` calls which were only used to communicate information into the client.  For example, `[cmd/deploy.go:289](https://github.com/knative-sandbox/kn-plugin-func/blob/main/cmd/deploy.go#L289)`:
```go
    if err = f.Write(); err != nil { // TODO: remove when client API uses 'f'
      return
    }
```

This has the added benefit of allowing all the commands to only call .Write once (if at all) when the command successfully completes, which will reduce the liklihood the Function enters an invalid state.  For example, a premature .Write which serializes to disk an invalid value for a Function member would result in the user's working source to be modified (dirty func.yaml) despite the command not completing successfully, requiring the user to manually revert their func.yaml or re-execute the command with valid values.

For example, see `[deploy.go:314](https://github.com/knative-sandbox/kn-plugin-func/blob/main/cmd/deploy.go#L314)` which shows the (ideally) only place where .Write should be called:
```go
  // mutations persisted on success
  return f.Write()
```
Note that this also will require that loading a Function via `NewFunction` be mostly removed from the `func` core library and into `cmd`.  Calls to `f.Initialized()` can be moved the other way.

manoelmarques:
# Changes

- :gift: Replace path with function struct in client api. Some of the internal function in client.go that only acted on the `Function` structure were refactored to function.go. The buildstamp `built` file logic was kept but now is saved by the `Function`

/kind enhancement

Fixes #1283


matejvasek:
@manoelmarques  plese try rebasing on `main` to fix the failing test.

lkingland (approved):
Thanks for the contribution.

I am not so sure about moving so much logic onto the Function struct, which is intended to be a simple data object, but I don't want that to hold up merging, because that can be easily moved around later if necessary.

To summarize quickly:  the current architecture keeps the Function an inert, serializable hunk of _data_ (a noun, in a way) with perhaps some  accessors to compute data akin to compound keys.  The Client is then the source of actions performed with/to a Function.  This separation of nouns and verbs really helps things stay stateless and simple as they get more complex.  I would suggest keeping logic on the client to keep things consistent in general.  I will remember to add this suggestion in our forthcoming architecture doc:  "a Function _is_ a thing.  the Client _does_ things with and to Functions".

Like I said, other than the logic which has been moved from Client to Function, this looks good, and I don't want to hold up a good thing for something easy to move around later.  👍🏻 

Everything else looks good!

manoelmarques:
Hi @lkingland, sorry for that. When I looked at those client functions acting just on the function structure, I thought they would be good candidates to be moved to function. I can move them back into the client in another PR.

manoelmarques:
HI @lkingland on second thought after reviewing the methods I moved to Function, they deal with the hash and saving this hash to disk as well as checking if a Function is "built".  The hash and the Function in my opinion are connected and now that the function is saved much later, the hash has to follow being saved at the same time, by an outer method not responsible for knowing that there was a hash to be saved in the first place. Hence, I added a flag to Function that can be set and when the Function saves itself, it will also save the hash. If there is some failure before, neither function nor hash will be saved to disk and nothing gets out of sync. Moving that logic out of Function can be done but that would mean Function calling the Client to do something on itself that only it knows it need to be done.

For example inside the Function Write:
```
if f.NeedsWriteBuildStamp {
		err = f.writeBuildStamp()
		if err != nil {
			return
		}
		f.NeedsWriteBuildStamp = false
	}
```
Since the Write to disk is done much later, the Function is the only agent now that will write the  Build Stamp based on a flag set by some previous method. At that moment, all that the outer method should do is call "f.Write()". It shouldn't have to worry about saving or not hashes.

lkingland:
> HI @lkingland on second thought after reviewing the methods I moved to Function, they deal with the hash and saving this hash to disk as well as checking if a Function is "built". The hash and the Function in my opinion are connected and now that the function is saved much later, the hash has to follow being saved at the same time, by an outer method not responsible for knowing that there was a hash to be saved in the first place. Hence, I added a flag to Function that can be set and when the Function saves itself, it will also save the hash. If there is some failure before, neither function nor hash will be saved to disk and nothing gets out of sync. Moving that logic out of Function can be done but that would mean Function calling the Client to do something on itself that only it knows it need to be done.
> 
> For example inside the Function Write:
> 
> ```
> if f.NeedsWriteBuildStamp {
>   err = f.writeBuildStamp()
>   if err != nil {
> 	  return
>   }
>   f.NeedsWriteBuildStamp = false
> }
> ```
> 
> Since the Write to disk is done much later, the Function is the only agent now that will write the Build Stamp based on a flag set by some previous method. At that moment, all that the outer method should do is call "f.Write()". It shouldn't have to worry about saving or not hashes which is an internal implementation.

Sounds good.  Let's go with it for a while and see if it still seems clear to our future selves

PR #1645, opened 2023-03-21
fix: pre-compile mock binaries instead of 'go run' in tests

matejvasek:
# Changes

- :gift: Pre-compile mock binaries instead of using `go run`. The `go run` does internally compile which prolongs perceived run-time. This may lead to timeout in some tests.


matejvasek:
PTAL @jrangelramos

matejvasek:
@lance please override the Lint.

matejvasek:
PTAL @jrangelramos

jrangelramos:
thanks @matejvasek 
/lgtm

Issue #856, opened 2022-02-22
On-cluster build: E2E test for build from Git repository (public & private)


lance:
- [x] Deploy on cluster a default function from a public git repository, from a default branch
- [x] Deploy on cluster a default function from a public git repository, from a feature branch
- [x] Deploy on cluster a default function from a public git repository, from a commit hash
- [x] Deploy on cluster a default function from a public git repository, from a tag revision
- [x] Deploy on cluster a default function located in a specific directory from a public git repository, from main branch
- [x] Updated changes to a function get deployed
- [x] Deploy on cluster a default function for each language: node, go, python, quarkus, springboot, typescript using pack
- [x] Deploy on cluster a default function using s2i builder for the languages supported by s2i. 
- [x] Deploy on cluster a fuction from a public GitHub repository
- [ ] Deploy on cluster a fuction from a public GitLab repository
- [ ] Deploy on cluster a default function from a Private git repository using basic http auth
- [ ] Deploy on cluster a default function from a Private git repository using ssh auth
- [x] Triggers on cluster build from cli when func.yaml indicates local
- [x] Triggers on cluster build from cli when func.yaml indicates local from a specific branch
- [x] Triggers on cluster build from cli when func.yaml indicates local from a when a functions is on a specific directory
- [x] Triggers local build from cli when func.yaml indicates build from git
- [x] Ensure all resources created by on cluster build are deleted when the functions is deleted

lance:
@jrangelramos thanks for providing these. I've added them all to the issue description as tasks (which could potentially be turned in to github issues if you want), and added the `Epic` label. As you find other test cases, feel free to just update the main issue description if you can.

zroubalik:
We should add tests covering s2i build strategy for on cluster build.

lance:
@jrangelramos any updates on this?

jrangelramos:
Hi @lance . yes I just updated the description. I shall re-take the remaining tests. Including s2i one this week still.. and may have an implementation next week probably.

lance:
@jrangelramos do you have any update on this one?

jrangelramos:
yes.. that item from GitHub repo is done..

For this one, I'm not seing much benefit on implement this, and thinking about remove it.
[ ] Deploy on cluster a fuction from a public GitLab repository

These remaing 2 depends on features not yet implemented in func.
[ ] Deploy on cluster a default function from a Private git repository using basic http auth
[ ] Deploy on cluster a default function from a Private git repository using ssh auth

[...]

PR #1649, opened 2023-03-24
test: adding e2e github oncluster test scenario

jrangelramos:
# Changes

- :gift: Add e2e github oncluster test scenario

It covers #856 item 
- [x]  Deploy on cluster a fuction from a public GitHub repository

/kind cleanup

### DIFF
--- a/test/oncluster/scenario_github_test.go
+++ b/test/oncluster/scenario_github_test.go
@@ -1,0 +1,90 @@
+//go:build oncluster
+// +build oncluster
+
+package oncluster
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"gotest.tools/v3/assert"
+	common "knative.dev/func/test/common"
+)
+
+/*
+Test scenario covered here:
+ - As a Developer I want my function stored on a public GitHub repo to get deployed on my cluster
+
+Notes:
+ * The function used as input for this scenario is stored in this repository at /test/oncluster/testdata/simplefunc
+
+ * On a CI Pull Request action (env CI="true") the branch used on the on-cluster test is the pull request reference.
+   The equivalent deploy func command would look like this:
+
+   func deploy --remote \
+     --git-url https://github.com/knative/func \
+     --git-dir test/oncluster/testdata/simplefunc \
+     --git-branch refs/pull/1650/head
+
+ * When not on CI,  the branch used will be "main" and the repository is https://github.com/knative/func.git
+
+*/
+
+func resolveGitVars() (gitRepoUrl string, gitRef string) {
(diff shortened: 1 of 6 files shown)


test/oncluster/scenario_github_test.go
```diff
@@ -0,0 +1,90 @@
...
+     --git-branch refs/pull/1650/head
+
+ * When not on CI,  the branch used will be "main" and the repository is https://github.com/knative/func.git
+
+*/
+
+func resolveGitVars() (gitRepoUrl string, gitRef string) {
+	var githubRepository = "knative/func"
+	gitRef = "main"
+	if ci, _ := os.LookupEnv("CI"); ci == "true" {
```
matejvasek, line 38: If you are not interested in `"ok"` value you can use "os.Getenv()".

test/oncluster/scenario_github_test.go
```diff
@@ -0,0 +1,90 @@
...
+     --git-dir test/oncluster/testdata/simplefunc \
+     --git-branch refs/pull/1650/head
+
+ * When not on CI,  the branch used will be "main" and the repository is https://github.com/knative/func.git
+
+*/
+
+func resolveGitVars() (gitRepoUrl string, gitRef string) {
+	var githubRepository = "knative/func"
+	gitRef = "main"
```
matejvasek, line 37: Better to use something else than `main` to prevent accidents.

test/oncluster/scenario_github_test.go
```diff
@@ -0,0 +1,90 @@
...
+	// -- Clone Func from GITHUB and checkout Branch
+	sh := common.NewShellCmd(t, tmpDir)
+	sh.ShouldFailOnError = true
+	sh.ShouldDumpOnSuccess = false
+	sh.Exec("git init " + tmpRepo)
+
+	sh.SourceDir = cloneDir
+	sh.Exec("git remote add origin " + githubRepo)
+	sh.Exec("git fetch --recurse-submodules=yes --depth=1 origin --update-head-ok --force " + githubRef)
+	sh.Exec("git checkout FETCH_HEAD")
```
matejvasek, line 70: Why not just `git clone`?

jrangelramos: I found this approach than clone consider the side the repo is getting... (I bother this git approach from the git-clone taks itself).. it does not clone the entire git tree.. just a shallow copy of the repo. Here we just need access to the `func.yaml` locally. TBH I think `func deploy --remote --git-url ...` should not even require a func locally ...

test/oncluster/scenario_github_test.go
jrangelramos: sure I'll push a fix on that

PR #1655, opened 2023-03-29
Add Python templates for gunicorn and uwsgi + Flask

frenzymadness:
These templates represent the most common and simple ways how to implement FAAS in Python either directly via gunicorn or using Flask and uwsgi.

# Changes

- :gift: Add new Python templates.

/kind enhancement

Because I'm not able to build func itself locally, I'm not able to test `func create` command. But when manually prepared, `func build -v -b=s2i` works fine as well as the resulting container image created from the new templates.


lance:
@frenzymadness can you please sign EasyCLA?

frenzymadness:
> @frenzymadness can you please sign EasyCLA?

Done.

matejvasek:
@frenzymadness  Thanks for the contribution!

If I understand this correctly `uwsgi` and `gunicorn` and pretty much interchangeable right?
From tests with "hello world" function it seems `uwsgi` is a little bit faster, but that probably does't really mean anything.
Which one is in general preferred?

For now I think with should add just the `Flask` template (under `templates/python/flask`, not `templates/python/uwsgi`), doesn't matter which server.

I find the other template with `def main(environ, start_response):` quite low level and a such I wouldn't add it to `func` CLI.



@lance I think we should also adjust `http`/`cloudevents` templates to employ `uwsgi/gunicorn` as replacement for `parliament+waitress`. The `parliament` project would just contain decorators transforming functions to the shape required by `uwsgi/gunicorn`.

Example of simplified decorator:
```python
def func(fn):
    app = Flask(__name__)
    app.add_url_rule("/", None, lambda: fn(Context(request)))
    return app.wsgi_app

@func
def main(context: Context):
    pass
```
in reality it would do somethinkg like [this]( https://github.com/boson-project/parliament/blob/main/parliament/server.py#L20.).

matejvasek:
@frenzymadness when modifying content of `./templates` you also have to run `make generate/zz_filesystem_generated.go` and commit the result.

matejvasek:
The `pack` build works too.

matejvasek:
Good thing about `gunicorn` is that with it we don't need `app.sh` we can just set `APP_MODULE` in `func.yaml` when deploying.

frenzymadness:
> @frenzymadness Thanks for the contribution!
> 
> If I understand this correctly `uwsgi` and `gunicorn` and pretty much interchangeable right? From tests with "hello world" function it seems `uwsgi` is a little bit faster, but that probably does't really mean anything. Which one is in general preferred?

You can find many comparisons of mod_wsgi, uwsgi and gunicorn. It's hard to say which one is the best one but gunicorn is very popular in containers.

> 
> For now I think with should add just the `Flask` template (under `templates/python/flask`, not `templates/python/uwsgi`), doesn't matter which server.
> 
> I find the other template with `def main(environ, start_response):` quite low level and a such I wouldn't add it to `func` CLI.

I agree that it is low level but some users might want it.

> @frenzymadness when modifying content of `./templates` you also have to run `make generate/zz_filesystem_generated.go` and commit the result.

Will do. Should I remove the uwsgi template?

> Good thing about `gunicorn` is that with it we don't need `app.sh` we can just set `APP_MODULE` in `func.yaml` when deploying.

How can I do that in the template? Is there any way how to have a template for the `func.yaml`?

matejvasek:
@frenzymadness I create PR to your branch in your repo.

matejvasek:
>How can I do that in the template? Is there any way how to have a template for the func.yaml?

This is done by adding `manifest.yaml`.
See [this](https://github.com/knative/func/blob/main/docs/language-packs/language-pack-contract.md#language-pack-manifests).
Unfortunately it is not supporting defaulting of runtime envvars (only build time envvars are supported now).
I fixed that in the PR.

matejvasek:
>I agree that it is low level but some users might want it.

OK lets keep it then.

frenzymadness:
I did a rebase which was quite painful because of the generated binary file and squashed the fixup commits to the original one.

matejvasek:
> I did a rebase which was quite painful because of the generated binary file and squashed the fixup commits to the original one.

ah, yes, it is annoying. Most of the time just regenerate it using `make generate/zz_filesystem_generated.go` and `git rebase --continue`.

matejvasek:
> Good thing about `gunicorn` is that with it we don't need `app.sh` we can just set `APP_MODULE` in `func.yaml` when deploying.

Also it tries to setting thread count in accordance with `cgroups` limitations. But it only works for `cgroups` v1 not v2.

lance (commented):
This is great - thanks for the contribution!

> @lance I think we should also adjust http/cloudevents templates to employ uwsgi/gunicorn as replacement for parliament+waitress. The parliament project would just contain decorators transforming functions to the shape required by uwsgi/gunicorn.

Agreed @matejvasek regarding

> https://github.com/boson-project/parliament/issues/25

New issue opened: https://github.com/boson-project/parliament/issues/25

/approve
/lgtm

PR #1661, opened 2023-03-31
Update Rust templates dependencies and UTs for Actix 4.x

andrejusc:
# Changes

- Updated Rust dependencies to latest for Rust templates
- As Actix FW got updated to 4.3.1 - adjusted involved UTs
- Added primitive .gitignore to avoid `target/` from commit

/kind enhancement

Fixes #

**Release Note**

```release-note
Update Rust templates dependencies and UTs for Actix 4.x
```

**Docs**

```docs

```

PR #1658, opened 2023-03-30
Update springboot templates to Spring Boot 3.0.5

andrejusc:
# Changes

- Updated Spring Boot version from 3.0.2 to latest stable 3.0.5
- Updated Spring Cloud version from 2022.0.0 to latest stable 2022.0.2
- Made Maven compiler plugin to use latest version 3.11.0
- Made Maven surefire plugin (used for tests) to use latest version 3.0.0
- Updated Maven wrapper definition to utilize latest 3.9.1

/kind enhancement

Fixes #

**Release Note**

```release-note
Update springboot templates to Spring Boot 3.0.5
```

**Docs**

```docs

```

PR #1663, opened 2023-03-31
[Release 1.9] cherry pick template updates for Rust and SpringBoot

lance:
Cherry pick https://github.com/knative/func/pull/1661 and https://github.com/knative/func/pull/1658 to the `release-1.9` branch.

```release-notes
Update Rust templates dependencies and UTs for Actix 4.x
Update springboot templates to Spring Boot 3.0.5
```


matejvasek:
@andrejusc I am trying the templates but the produced images are not serving HTTP on 8080. Any idea what's wrong?

/hold

matejvasek:
Hmm old templates do the same.

matejvasek:
The HTTP works from within a container, but docker's `-p` forwarding is not working for me.

matejvasek:
`curl: (56) Recv failure: Connection reset by peer`

matejvasek:
I guess it's a bug in buildpacks.

/approve
/lgtm

matejvasek:
The buildpacks are not exposing port on image for some reason.

/unhold

andrejusc:
@lance - I thought there are some Rust image related build/test step inside those checks, but looking through them - only Rust Integration tests are there, but not image tests. I need to study how use case is done for Go case image inside checks (hopefully such is there) and then adopt for Rust, so we'll eliminate such potential breakage in future.
I'm not yet familiar deep enough with buildpacks to comment - will try to analyze more in upcoming days.

andrejusc:
@lance - issue is with code line (which I didn't change) and this binding to listen:
https://github.com/knative/func/blob/main/templates/rust/http/src/main.rs#L28
`.bind(("127.0.0.1", port))?`
If such replace to this:
`.bind(("0.0.0.0", port))?`
it should potentially work.

Could you give it a try? Otherwise - I'll review more on weekend to correct.

lance:
@andrejusc please work with @matejvasek on this next week. I will not be available until April 10.

PR #1664, opened 2023-03-31
fix: Rust template project listen on 0.0.0.0

matejvasek:
# Changes

- :bug: Rust template's Actix listen on 0.0.0.0 

```release-note
fix: rust template project listen on 0.0.0.0
```


matejvasek:
cc @andrejusc

matejvasek:
Without this containerized app cannot work at all.

PR #1666, opened 2023-04-04
add pvc and emptyDir to function_volumes

### LINKED ISSUE #1647: Support PVC and emptyDir in `run.volumes` in func.yaml
## Description

Currently, in func.yaml, you can create volumes from a secret or a configmap using:

```yaml
...
run:
  volumes:
    - secret: "my-secret"
      path: "/secret"
    - configMap: "my-config-map"
      path: "/config"
...
```

It would be nice if other volume types are supported. For example, knative has a [feature where a PVC](https://knative.dev/docs/serving/configuration/feature-flags/#kubernetes-persistentvolumeclaim-pvc) can be used to create a volume, and another where an [empty dir](https://knative.dev/docs/serving/configuration/feature-flags/#kubernetes-emptydir-volume) could be used. This could be achieved by adding more options to the volumes map:

```
...
run:
  volumes:
    - presistentVolumeClaim: "my-pvc"
      path: "/pvc"
    - emptyDir: {}
      path: "/emptydir"
...
```

## Why this is helpful?

Currently, the only way to do this is to deploy the function then edit the resulting service.serving.knative.dev as described in [the features docs](https://knative.dev/docs/serving/configuration/feature-flags/#kubernetes-emptydir-volume). It would be more seamless if the volumes could be specified in the func.yaml directly.

zalsader:
# Changes

- add `persistentVolumeClaim` and `emptyDir` to the allowed function volumes 
- deploy `persistentVolumeClaim` and `emptyDir` volumes
- add `persistentVolumeClaim` and `emptyDir` as options in config 

/kind api-change

Fixes #1647

**Release Note**

```release-note
Allow specifying `persistentVolumeClaim` and `emptyDir` as volumes in functions.
```

**Docs**

```docs
- [knative/docs]: TODO
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -203,6 +203,7 @@ $(BIN_WINDOWS): generate/zz_filesystem_generated.go
 ##@ Schemas
 ######################
 schema-generate: schema/func_yaml-schema.json ## Generate func.yaml schema
+.PHONY: schema/func_yaml-schema.json
 schema/func_yaml-schema.json: pkg/functions/function.go
 	go run schema/generator/main.go
 
--- a/Makefile
+++ b/Makefile
@@ -203,8 +203,7 @@ $(BIN_WINDOWS): generate/zz_filesystem_generated.go
 ##@ Schemas
 ######################
 schema-generate: schema/func_yaml-schema.json ## Generate func.yaml schema
-.PHONY: schema/func_yaml-schema.json
-schema/func_yaml-schema.json: pkg/functions/function.go
+schema/func_yaml-schema.json: pkg/functions/function.go pkg/functions/function_*.go
 	go run schema/generator/main.go
 
 schema-check: ## Check that func.yaml schema is up-to-date
--- a/cmd/config_volumes.go
+++ b/cmd/config_volumes.go
@@ -165,6 +165,11 @@ func runAddVolumesPrompt(ctx context.Context, f fn.Function) (err error) {
 		}
 	}
 
+	// SECTION - display a help message to enable advanced features
+	if selectedOption == optionEmptyDir || selectedOption == optionPersistentVolumeClaim {
+		fmt.Printf("Please make sure to enable the %s extension flag: https://knative.dev/docs/serving/configuration/feature-flags/\n", selectedOption)
+	}
+
 	// SECTION - select the specific resource to be mounted
 	optionsResoures := []string{}
 	switch selectedOption {
@@ -194,12 +199,24 @@ func runAddVolumesPrompt(ctx context.Context, f fn.Function) (err error) {
 		Message: fmt.Sprintf("Please specify the path where the %s should be mounted:", selectedOption),
 	}, &path, survey.WithValidator(func(val interface{}) error {
 		if str, ok := val.(string); !ok || len(str) <= 0 || !strings.HasPrefix(str, "/") {
-			return fmt.Errorf("The input must be non-empty absolute path.")
+			return fmt.Errorf("the input must be non-empty absolute path")
 		}
 		return nil
 	}))
 	if err != nil {
 		return
+	}
+
+	// SECTION - is this read only for pvc
+	readOnly := false
+	if selectedOption == optionPersistentVolumeClaim {
+		err = survey.AskOne(&survey.Confirm{
+			Message: "Is this volume read-only?",
+			Default: false,
+		}, &readOnly)
+		if err != nil {
+			return
+		}
 	}
 
 	// we have all necessary information -> let's store the new Volume
@@ -210,7 +227,10 @@ func runAddVolumesPrompt(ctx context.Context, f fn.Function) (err error) {
 	case optionSecret:
 		newVolume.Secret = &selectedResource
 	case optionPersistentVolumeClaim:
-		newVolume.PresistentVolumeClaim = &fn.PersistentVolumeClaim{ClaimName: &selectedResource}
+		newVolume.PresistentVolumeClaim = &fn.PersistentVolumeClaim{
+			ClaimName: &selectedResource,
+			ReadOnly:  readOnly,
+		}
 	case optionEmptyDir:
 		newVolume.EmptyDir = &fn.EmptyDir{}
 	}
--- a/cmd/config_volumes.go
+++ b/cmd/config_volumes.go
@@ -210,7 +210,7 @@ func runAddVolumesPrompt(ctx context.Context, f fn.Function) (err error) {
 	case optionSecret:
 		newVolume.Secret = &selectedResource
 	case optionPersistentVolumeClaim:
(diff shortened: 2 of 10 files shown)


matejvasek:
@zalsader you need to re-generate schema `make schema-generate` and push it.

zalsader:
> @zalsader you need to re-generate schema `make schema-generate` and push it.

I did that. The command for `make schema-generate` was broken, so I added a fix.

zalsader:
One thing is missing: We need to check that the features are enabled. Do you have suggestions on how to do that?

matejvasek:
> One thing is missing: We need to check that the features are enabled. Do you have suggestions on how to do that?

@zalsader what features do you mean?

zalsader:
> @zalsader what features do you mean?

Sorry I should have been more clear. The PVC and emptyDir are [behind a feature flag](https://knative.dev/docs/serving/configuration/feature-flags/#kubernetes-persistentvolumeclaim-pvc).

lkingland (dismissed):
This looks very good

Just one small change:  please move the validation logic `func validateVolumes` out of `pkg/functions` since in order to execute this (much appreciated) more advanced validation requires  `k8s.io/api/core/v1` and `k8s.io/apimachinery/pkg/api/resource`.  We are keeping the core package `functions` dependency-free, so I think perhaps `pkg/k8s` would be the right place for this validator; taking its dependencies with it.

lkingland:
> > @zalsader what features do you mean?
> 
> Sorry I should have been more clear. The PVC and emptyDir are [behind a feature flag](https://knative.dev/docs/serving/configuration/feature-flags/#kubernetes-persistentvolumeclaim-pvc).

This is a good point, and actually a larger question, because there are other things I would like the functions client library to be able to inspect about the state of the configuration of the cluster, and interact with the user appropriately (or at least generate helpful error messages).

For now, I am fine with this PR being as-is, leaning on documentation, and perhaps issuing a warning to the user in the CLI that if they receive error X they need to enable PVC support.

I am going to put this on our agenda for this weeks meeting, and probably open a discussion.

zalsader:
> For now, I am fine with this PR being as-is, leaning on documentation, and perhaps issuing a warning to the user in the CLI that if they receive error X they need to enable PVC support.
> 
> I am going to put this on our agenda for this weeks meeting, and probably open a discussion.

When is the meeting, and would I be able to join? Where would be an appropriate place to issue this warning or write that documentation?

lkingland:
> > For now, I am fine with this PR being as-is, leaning on documentation, and perhaps issuing a warning to the user in the CLI that if they receive error X they need to enable PVC support.
> > I am going to put this on our agenda for this weeks meeting, and probably open a discussion.
> 
> When is the meeting, and would I be able to join? Where would be an appropriate place to issue this warning or write that documentation?

We have a [Knative Community Calendar](https://calendar.google.com/calendar/u/0/embed?src=knative.team_9q83bg07qs5b9rrslp5jor4l6s@group.calendar.google.com)  which includes a Functions Weekly Meeting on Tuesdays from 14:00-14:30 UTC as well as many other meetings.

Until we have a way to query the connected platform to detect things like whether or not PVCs are enabled, I would issue this warning clearly from within the CLI package when a user adds a PVC volume; perhaps even with a link to the document.  Perhaps someone in the meeting will have a better suggestion as a stop-gap until we have a more permanent solution.

Makefile
```diff
@@ -203,6 +203,7 @@ $(BIN_WINDOWS): generate/zz_filesystem_generated.go
 ##@ Schemas
 ######################
 schema-generate: schema/func_yaml-schema.json ## Generate func.yaml schema
+.PHONY: schema/func_yaml-schema.json
```
lkingland, line 206: I am pretty sure that this is not a .PHONY, because it is actually intended to create the file `schema/func_yaml-schema.json.  Did you mean to say `.PHONY schema-generate`?

lkingland (commented):
Looks good to me @zalsader

I'll `/lgtm` this after giving you a chance to take a look at that .PHONY declaration, and perhaps add a helpful warning to the user about this feature requiring special features being installed in the cluster; if you can find a good spot to emit that from the `cmd` package.
👍🏻

Makefile
zalsader: `make build` depends on `$(CODE)` which includes `schema/func_yaml-schema.json`. Since the file exists, `make` does not attempt to regenerate that file. This caused an issue later in the CI pipeline. Since there was a `.PHONY: generate/zz_filesystem_generated.go`, I thought it would be ok to do the same for the schema file. What do you think we should do instead?

lance: I think @lkingland is correct here. Modifying `pkg/functions/function.go` does indeed cause the schema to be regenerated via this make target:

https://github.com/knative/func/blob/d5b4a8d8e9bfe8a4ffc66215253f75f1feab7d69/Makefile#L206-L207

The `schema-generate` target should have been a no-op in CI unless the `schema/func_yaml-schema.json` changed during the build. This has not historically been a problem, so IMO this change is not needed. Otherwise all looks good!

lkingland: > Since there was a .PHONY: generate/zz_filesystem_generated.go, I thought it would be ok to do the same for the schema file.

I see; thanks for noticing!  I believe that `.PHONY: generate/zz_filesystem_generated.go` should actually be removed as well.  The way it is being generated is not entirely optimal, so it's probably not the best example.

Here's how I understand this rule to work:
```
schema/func_yaml-schema.json: pkg/functions/function.go 
        go run schema/generator/main.go
```
`pkg/functions/function.go` is the dependency. The rule will only execute if the target file either doesn't exist or is older than the dependency file. In this case, if `function.go` is newer than `func_yaml-schema.json` (has modifications), then the rule will be executed, generating the file again via `go run...`.  This is the intended functionality.

I would like to know the problem you saw in CI.  There actually might be something over there we need to fix or clean up!

zalsader: I know what the issue is now; I added a commit. `schema/func_yaml-schema.json` should also depend on `function_*.go`

lance:
Assuming tests pass, this is good to go. Thanks for your contribution @zalsader!

/lgtm

lance:
@zalsader I neglected to notice that @lkingland had requested this change.

> Just one small change: please move the validation logic func validateVolumes out of `pkg/functions` since in order to execute this (much appreciated) more advanced validation requires `k8s.io/api/core/v1` and `k8s.io/apimachinery/pkg/api/resource`. We are keeping the core package functions dependency-free, so I think perhaps `pkg/k8s` would be the right place for this validator; taking its dependencies with it.

Would you mind please handling this as well? Thanks.

zalsader:
> Would you mind please handling this as well? Thanks.

I think it is already done since I removed the extra dependencies.

PR #1669, opened 2023-04-04
doc: refer specific verision of Tekton not latest

matejvasek:


jrangelramos:
Hi @matejvasek I think we could also change here https://github.com/knative/func/blob/main/hack/tekton.sh#L25 to match the suggested version on this PR.

matejvasek:
@zroubalik PTAL

matejvasek:
> Hi @matejvasek I think we could also change here https://github.com/knative/func/blob/main/hack/tekton.sh#L25 to match the suggested version on this PR.

@jrangelramos fixed.

matejvasek:
@zroubalik please override the check.

kvmware:
https://github.com/knative/actions/pull/132 should fix that check fail

PR #1671, opened 2023-04-09
Replace knative.dev/client references to knative.dev/client-pkg

vyasgun:
# Changes

- :broom: Replaced the occurrences of knative.dev/client with knative.dev/client-pkg

/kind cleanup

Part of https://github.com/knative/client/issues/1591

**Release Note**

```release-note

```

**Docs**

```docs

```


lance:
@vyasgun thanks for this - and welcome back! :)

It looks like the verify codegen action failed, which is common if you are using a newer (M1) mac. You can resolve this by running the following command from the root directory in your branch and updating the PR.

```
docker run -it --rm -v$(pwd):/src/:Z golang sh -c 'cd /src && ./hack/update-codegen.sh'
```

vyasgun:
Thanks a lot @lance - happy to be back! The suggestion worked and the action is passing now :)

PR #1684, opened 2023-04-18
test: extend filesystem tests

matejvasek:
# Changes

- :test_tube: Extend filesystem tests.


pkg/filesystem/filesystem_test.go
```diff
@@ -149,6 +87,95 @@ func TestFileSystems(t *testing.T) {
...
+		case fs.ModeSymlink:
+			t, _ := os.Readlink(path)
+			bs = []byte(t)
+		}
+		path, err = filepath.Rel(root, path)
+		if err != nil {
+			return err
+		}
+		files = append(files, FileInfo{
+			Path:       filepath.ToSlash(path),
```
dsimansk, line 168: How does it work with Windows backslash path separator?

matejvasek: The Go's `fs` package uses slashes (like zip or tar does).

matejvasek: This line is here precisely so it works on Windows.

matejvasek:
@lance @jrangelramos  PTAL.

zroubalik (approved):
/lgtm
/approve

PR #1685, opened 2023-04-23
feat: specify create time for image built with buildpacks

cuichenli:
# Changes
Currently the image built with `func build` command all says the image was built 42 years ago. Although it was treated as a feature of buildpacks (in https://github.com/buildpacks/pack/issues/931#issuecomment-719524493), they have fixed this feature recently in https://github.com/buildpacks/imgutil/pull/138 hence I think it is worth to change the behavior here as well. 

/kind enhancement

Note: I am new to the func project and this change may be inappropriate, feel free to reject it. Thanks


matejvasek:
I think we can have this but  I would prefer having this behind some config flag in `~/.config/func/config.yaml`.

cuichenli:
yeah, definitely it would be great to make it configurable. if the owners are ok with this, i will try to implement it.

matejvasek:
@lance @zroubalik wdyt?

lance:
> @lance @zroubalik wdyt?

I agree behind a config flag would be nice

lkingland:
Perhaps a Builder option `WithTimestamp(boolean)` which would be set when the builder is created in `cmd`, based on the value of a flag `--build-timestamp`?

cuichenli:
i have updated accordingly, please let me know if i missed anything. thanks!

matejvasek:
@cuichenli you need to fix one linter error.

cuichenli:
i (think) i have fixed it in the latest commit, can you trigger the build again? thanks

lkingland (approved):
Looks really good. Thanks @cuichenli 

I re-ran the failed tests.  They've been flaky lately

PR #1690, opened 2023-04-25
feat: func domain route

lkingland:
:feat: func --domain to choose routes

Adds the --domain flag which, when provided sets the value of the `func.domain` label on the deployed knative service.  The cluster can then use this to choose which route to create for multi-route clusters. Optional.

Note that this will have no effect unless the cluster is configured to apply routes based on a domain mapping in the `config-domain` config map.  This feature is functional in dev and a future PR will add this configuration both to the cluster created via CI/CD.

Documenting this feature is still an open question, but here us an example of a cluster configuration which by default exposes functions internally only as `function.namespace.cluster.local`, and allows for functions to be selectively exposed using the --domain flag for boson-project.org or boson-project.net:
```
  cluster.local: ""
  boson-project.org: |
    selector:
      func.domain: "boson-project.org"
  boson-project.net |
    selector:
      func.domain: "boson-project.net"
```

/kind enhancement

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


lkingland:
Note, related to #1666, it would be best if we had a way to inspect the state of the current platform upon which the function will be deployed, and issue a warning if this flag will not work as intended, and ideally have a tab-completion UX for choosing the domain for the function's route when more than one are configured.

lance (commented):
/lgtm

This is a nice addition which will also be useful in clusters that have `staging` and/or `test` domains.

PR #1692, opened 2023-04-25
fix: remove preparsing flag error messages

lkingland:
- :bug: removes flag preparsing errors

Errors pre-parsing flags to extract --path were incorrectly logging expected parsing errors when other flags are provided.  Squelched.

/kind bug


zroubalik (approved):
/lgtm
/approve

PR #1693, opened 2023-04-25
feat: func run support all build options

lkingland:
- :gift: `func run` supports build options and host (non-containerized) build option

Refactors the run command to include all options from build, in the same manner as deploy.
Updates help text to match the other command format, and lays the verbal groundwork for the difference between containerized (currently default) and non-containerized runs.
Adds the --container flag (currently defaults to true) which will, when disabled, invoke the localhost-based runner which, in this commit, is not yet active.

/kind enhancement

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -30,7 +30,7 @@ NAME
 
 SYNOPSIS
 	{{rootCmdUse}} deploy [-R|--remote] [-r|--registry] [-i|--image] [-n|--namespace]
-	             [-e|--env] [-g|--git-url] [-t|git-branch] [-d|--git-dir]
+	             [-e|--env] [-g|--git-url] [-t|--git-branch] [-d|--git-dir]
 	             [-b|--build] [--builder] [--builder-image] [-p|--push]
 	             [--platform] [-c|--confirm] [-v|--verbose] [--build-timestamp]
 
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -13,7 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"knative.dev/client-pkg/pkg/util"
 	"knative.dev/func/pkg/builders"
-	"knative.dev/func/pkg/builders/buildpacks"
+	pack "knative.dev/func/pkg/builders/buildpacks"
 	"knative.dev/func/pkg/builders/s2i"
 	"knative.dev/func/pkg/config"
 	fn "knative.dev/func/pkg/functions"
@@ -30,7 +30,7 @@ NAME
 
 SYNOPSIS
 	{{rootCmdUse}} deploy [-R|--remote] [-r|--registry] [-i|--image] [-n|--namespace]
-	             [-e|env] [-g|--git-url] [-t|git-branch] [-d|--git-dir]
+	             [-e|--env] [-g|--git-url] [-t|git-branch] [-d|--git-dir]
 	             [-b|--build] [--builder] [--builder-image] [-p|--push]
 	             [--platform] [-c|--confirm] [-v|--verbose] [--build-timestamp]
 
@@ -218,7 +218,7 @@ func runDeploy(cmd *cobra.Command, newClient ClientFactory) (err error) {
 		return
 	}
 
-	// TODO: this is duplicate logic with runBuild.
+	// TODO: this is duplicate logic with runBuild and runRun.
 	// Refactor both to have this logic part of creating the buildConfig and thus
 	// shared because newDeployConfig uses newBuildConfig for its embedded struct.
 	if f.Registry != "" && !cmd.Flags().Changed("image") && strings.Index(f.Image, "/") > 0 && !strings.HasPrefix(f.Image, f.Registry) {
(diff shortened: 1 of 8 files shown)


cmd/deploy.go
```diff
@@ -30,7 +30,7 @@ NAME
 
 SYNOPSIS
 	{{rootCmdUse}} deploy [-R|--remote] [-r|--registry] [-i|--image] [-n|--namespace]
-	             [-e|env] [-g|--git-url] [-t|git-branch] [-d|--git-dir]
+	             [-e|--env] [-g|--git-url] [-t|git-branch] [-d|--git-dir]
```
lance, line 33:
```suggestion
	             [-e|--env] [-g|--git-url] [-t|--git-branch] [-d|--git-dir]
```
Nice catch - there's one more!

pkg/functions/client.go
```diff
@@ -826,10 +826,7 @@ func (c *Client) Run(ctx context.Context, f Function) (job *Job, err error) {
 	}()
 
 	if !f.Initialized() {
-		// TODO: this needs a test.
-		err = fmt.Errorf("the given path '%v' does not contain an initialized "+
-			"function.  Please create one at this path in order to run", f.Root)
-		return nil, err
+		return nil, fmt.Errorf("Can not run an uninitialized function")
```
lance, line 829:
```suggestion
		return nil, fmt.Errorf("can not run an uninitialized function")
```
Nit: I think convention is to begin error messages with lower-case

cmd/run.go
```diff
@@ -83,146 +125,193 @@ Default builder is pack.
...
+
+	// There is currently no local host runner implemented, so specifying
+	// --container=false should always return an informative error to the user
+	// such that they do not receive the rather cryptic "no runner defined"
+	// error from a Client instance which was instantiated with no runner.
+	// TODO: modify this check when the local host runner is available to
+	// only generate this error when --container==false && the --language is
+	// not yet implemented.
+	if !c.Container {
+		return errors.New("The ability to run functions outside of a container via 'func run' is coming soon.")
```
lance, line 314: Nit: I think convention is to start error messages with lower-case
```suggestion
		return errors.New("the ability to run functions outside of a container via 'func run' is coming soon.")
```

cmd/run.go
```diff
@@ -4,77 +4,119 @@ import (
...
 	cfg, err := config.NewDefault()
 	if err != nil {
 		fmt.Fprintf(cmd.OutOrStdout(), "error loading config at '%v'. %v\n", config.File(), err)
 	}
 
+	// Function Context
+	f, _ := fn.NewFunction(effectivePath())
+	if f.Initialized() {
+		cfg = cfg.Apply(f)
+	}
```
lance, line 86: If the function isn't initialized, it can't be run, right? Is that handled elsewhere?

cmd/run.go
```diff
@@ -83,146 +125,193 @@ Default builder is pack.
...
 	}
-
-	function, err := fn.NewFunction(cfg.Path)
-	if err != nil {
+	if err = cfg.Validate(cmd); err != nil {
 		return
 	}
-	if !function.Initialized() {
-		return fmt.Errorf("the given path '%v' does not contain an initialized function", cfg.Path)
+	if f, err = fn.NewFunction(cfg.Path); err != nil {
```
lance, line 146: I think this is probably where `Initialized()` ends up being checked. I suppose the deleted check earlier in this file was redundant.

cmd/run.go
```diff
@@ -83,146 +125,193 @@ Default builder is pack.
...
 	}
+
+	// NOTE: we do not f.Write() here unlike deploy (and build).
+	// running is ephemeral: a run is not affecting the function itself,
+	// as opposed to deploy commands, which are actually mutating the current
+	// state of the function as it exists on the network.
+	// Another way to think of this is that runs are development-centric tests,
+	// and thus most likely values changed such as environment variables,
+	// builder, etc. would not be expected to persist and affect the next deploy.
+	// Run is ephemeral, deploy is persistent.
```
lance, line 228: I wonder if this should somehow be communicated to the user.

lance (commented):
Looks great! I just have a couple of nits - and it needs a rebase.

cmd/run.go
lkingland: Yes that's handled elsewhere.  I could refactor this a bit to make it more clear, but for right now all of the commands which use "the current function" to override global values only do so if there is actually an initialized functoin at that path (otherwise, as you say, the error of "can't X; no function at Y" is handled later.  This stanza is the pre-execution phase where flag defaults are being populated, so it is a bit more forgiving.    Definitely could refactor this if statement away entirely, however.

cmd/run.go
lkingland: Right, that is the pattern I'm following in the other commands as well.   So if there is an error in the logic than we "get" to fix a bug in all of them.

Generally speaking, I am trying to keep the CLI as dumb as possible while still being useful and helpful to the user.   The implementation of `Run` is the only part of the equation that actually cares if the function has actually been initialized, so the check is done there.

cmd/run.go
lkingland: Yes and at least one time when I was using `run` locally, specifying `--builder`, I was confused why it didn't use the same value the next time around.

So on the one hand I am pretty sure that we do not want invocations of `run` to persist the value of flags which alter `deploy` (and `build`), but I also find that once accustomed to the value of flags persisting between commands, I expect it to be consistently available.

The only solution that comes to mind is that we actually have a separate set of flag values for running (a development/testing, inner-loop sort of paradigm) vs those used for the deployed function. These would be developer-specific (host scoped) and reside in `.func/` in the same way as `--remote`, etc.

This is definitely complex enough to warrant its own issue and PR, so for the purposes of this PR I just want to make sure that invoking `run` with a different builder during development/testing doesn't alter the deployed function configuration!

lance (commented):
/lgtm

PR #1695, opened 2023-04-25
feat: write a build stamp log to .func

lkingland:
- :gift:  writes a log of the files and hashes which contribute to a build stamp
- :bug: skip the function root directory when calculating the fingerprint
- :broom: adds a `.Stamp` method to functions, making stamp updates explicit
- :bug: Building always "Stamps" the function, allowing builds to cache (and fixes a couple bugs in pipeline tests)
- :broom: run tests use client API rather than writing a func.yaml

To aid in the troubleshooting of a fairly sensitive filesystem figerprinting process for build caching, a logfile is written that enumerates the files and their hashes which contributed to the fingerprint.  Also fixes a bug where the root directoy, and its modification timestamp, were considered part of the hash.

Updates the run command's tests to not rely on func.yaml, which is an implementation detail of the client package, instead using the client API to create the function.  Adds an optional function manipulation "setup" method to the tests which allows the test to mutate the function, taking the place logically of providing this exact func.yaml.

/kind enhancement

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,8 +121,11 @@ test-python: ## Test Python templates
 	cd templates/python/http && python3 test_func.py && rm -rf __pycache__
 
 test-quarkus: ## Test Quarkus templates
-	cd templates/quarkus/cloudevents && mvn test && mvn clean
-	cd templates/quarkus/http && mvn test && mvn clean
+	# Quarkus Template Tests are temporarily disabled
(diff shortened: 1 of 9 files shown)


lkingland:
Looks like a conflict snuck its way in there after the last review.  I've now rebased and it should be ready for a re-/lgtm after tests complete 👍🏻

zroubalik (approved):
/lgtm

PR #1696, opened 2023-04-25
fix: run cmd tests depend on func.yaml

lkingland:
- :broom: run tests use client API rather than writing a func.yaml

Updates the run command's tests to not rely on func.yaml, which is an implementation detail of the client package, instead using the client API to create the function.

Adds an optional function manipulation "setup" method to the tests which allows the test to mutate the function, taking the place logically of providing this exact func.yaml.

/kind cleanup

### DIFF
--- a/cmd/run_test.go
+++ b/cmd/run_test.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"context"
 	"fmt"
-	"os"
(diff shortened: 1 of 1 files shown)


lkingland:
Close as incorporated into #1695 as prerequisite

PR #1697, opened 2023-04-25
feat: job scaffolding

lkingland:
- :gift: client Jobs support scaffolding (backwards compatibly)

- Local jobs use a directory rather than filename This sets up for upcoming scaffolding to use as a space for job scaffolding code.
- Places all jobs within a 'runs' directory in ./func rather than 'instances' to further differentiate between an instance and a local run task invoked via .Run
- Updates Jobs' stop handler to have an error return.
- Adds tests which ensure creating and stopping a job are reflected in the client's .Instances().Local() accessor.
- Adds verbose logging support to the Job implementation.
- Adds a cleanup task which will remove orphaned jobs by, in addition to previous logic, checking that there is no longer a process listening on the port indicated by the job.

/kind enhancement

### DIFF
--- a/cmd/invoke_test.go
+++ b/cmd/invoke_test.go
@@ -48,8 +48,8 @@ func TestInvoke(t *testing.T) {
 		}()
 		_, port, _ := net.SplitHostPort(l.Addr().String())
 		errs := make(chan error, 10)
-		stop := func() { _ = s.Close() }
-		return fn.NewJob(f, port, errs, stop)
+		stop := func() error { _ = s.Close(); return nil }
+		return fn.NewJob(f, port, errs, stop, false)
 	}
 
 	// Run the mock http service function interloper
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -121,8 +121,11 @@ test-python: ## Test Python templates
 	cd templates/python/http && python3 test_func.py && rm -rf __pycache__
 
 test-quarkus: ## Test Quarkus templates
-	cd templates/quarkus/cloudevents && mvn test && mvn clean
-	cd templates/quarkus/http && mvn test && mvn clean
+	# Quarkus Template Tests are temporarily disabled
```
matejvasek, line 124: This should be fixed, just rebase.

lance:
Rerunning e2e tests. If those pass...

/lgtm

PR #1698, opened 2023-04-25
src: move template struct definition

lkingland:
- :broom: move template struct definition to template.go 

/kind cleanup


zroubalik (approved):
/lgtm
/approve

PR #1699, opened 2023-04-25
feat: function signature detection

lkingland:
- :gift: function signatures with detector

Adds an enumeration of base accepted method signatures, and implements a detector for go functions.
Introduces the concept of "Instanced" functions.
Mocks up where additional detectors would be added using a stubbed Python detector.

/kind enhancement

### DIFF
--- a/pkg/functions/signatures.go
+++ b/pkg/functions/signatures.go
@@ -1,0 +1,148 @@
+package functions
+
+import (
+	"errors"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+type Signature int
+
+const (
+	UnknownSignature Signature = iota
+	InstancedHTTP
+	InstancedCloudevent
+	StaticHTTP
+	StaticCloudevent
+)
+
+func (s Signature) String() string {
+	return []string{
+		"unknown",
+		"instanced-http",
+		"instanced-cloudevent",
+		"static-http",
+		"static-cloudevent",
+	}[s]
+}
+
+var signatureMap = map[bool]map[string]Signature{
+	true: {
+		"http":       InstancedHTTP,
+		"cloudevent": InstancedCloudevent},
+	false: {
+		"http":       StaticHTTP,
+		"cloudevent": StaticCloudevent},
+}
+
+func signature(instanced bool, invoke string) Signature {
+	if invoke == "" {
+		invoke = "http"
+	}
+	s, ok := signatureMap[instanced][invoke]
+	if !ok {
+		return UnknownSignature
+	}
+	return s
+}
+
+// detectors check for the existence of certain method signatures in the
+// source code at the given root.
+type detector interface {
+	Detect(dir string) (static, instanced bool, err error)
+}
+
+// functionSignature returns the signature implemented by the given function
+func functionSignature(f Function) (s Signature, err error) {
+	d, err := detectorFor(f.Runtime)
+	if err != nil {
+		return UnknownSignature, err
+	}
+	static, instanced, err := d.Detect(f.Root)
+	if err != nil {
+		return
+	}
+	// Function must implement either a static handler or the instanced handler
+	// but not both.
+	if static && instanced {
+		return s, fmt.Errorf("function may not implement both the static and instanced method signatures simultaneously")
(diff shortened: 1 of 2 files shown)


pkg/functions/signatures.go
```diff
@@ -0,0 +1,147 @@
...
+
+// detectorFor runtime returns a signature detector for a given runtime
+func detectorFor(runtime string) (detector, error) {
+	switch runtime {
+	case "go":
+		return &goDetector{}, nil
+	case "python":
+		return &pythonDetector{}, nil
+	case "rust":
+		return nil, errors.New("The Rust signature detector is not yet available.")
```
lance, line 91: Nit
```suggestion
		return nil, errors.New("the Rust signature detector is not yet available")
```

pkg/functions/signatures.go
```diff
@@ -0,0 +1,147 @@
...
+	case "node":
+		return nil, errors.New("The Node.js signature detector is not yet available.")
```
lance, line 93: Nit
```suggestion
		return nil, errors.New("the Node.js signature detector is not yet available")
```

pkg/functions/signatures.go
```diff
@@ -0,0 +1,147 @@
...
+	case "quarkus":
+		return nil, errors.New("The TypeScript signature detector is not yet available.")
```
lance, line 95: Nit
```suggestion
		return nil, errors.New("the TypeScript signature detector is not yet available")
```

pkg/functions/signatures.go
```diff
@@ -0,0 +1,147 @@
...
+	default:
+		return nil, fmt.Errorf("Unable to detect the signature of the unrecognized runtime language %q", runtime)
```
lance, line 97: Nit
```suggestion
		return nil, fmt.Errorf("unable to detect the signature of the unrecognized runtime language %q", runtime)
```

pkg/functions/signatures.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// GO
+
+type goDetector struct{}
+
+func (d goDetector) Detect(dir string) (static, instanced bool, err error) {
+	files, err := os.ReadDir(dir)
+	if err != nil {
+		err = fmt.Errorf("signature detector encountered an error when scanning the function's source code. %w", err)
```
lance, line 108: Nit
```suggestion
		err = fmt.Errorf("signature detector encountered an error when scanning the function's source code %w", err)
```

pkg/functions/signatures.go
```diff
@@ -0,0 +1,147 @@
...
+		return
+	}
+	for _, file := range files {
+		filename := filepath.Join(dir, file.Name())
```
lance, line 112: I wonder if it could be a condition that the filename be `handle.go`

pkg/functions/signatures.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		if d.hasFunctionDeclaration(filename, "Handle") {
+			static = true
+		}
+	}
+	return
+}
+
+func (d goDetector) hasFunctionDeclaration(filename, function string) bool {
+	astFile, err := parser.ParseFile(token.NewFileSet(), filename, nil, parser.SkipObjectResolution)
```
lance, line 127: I wonder if there's a golang parser for all of our runtimes! :)

lance (commented):
Overall looks good to me. Just a few small nits to address for consistency across the codebase. Currently, none of the other runtimes has instanced based functions, but I suppose that shouldn't really matter and may come over time.

pkg/functions/signatures.go
lkingland: Since most languages don't have any file name requirements, I didn't want us to impose anything.
Scanning each file in the directory looking for the expected signature should be trivially fast.

pkg/functions/signatures.go
lkingland: Wouldn't that be helpful!  I wouldn't be suprised if there were a parser for all of them already out there.

Fortunately, we don't _need_ a full parser; we just need to be able to split on tokens and have enough of a state machine to know if it's inside of a block comment or not.  So even if we need to write a small parser for each language ourselves; it should be fairly straightforward.  But let's hope they already exist!  🤞🏻

PR #1702, opened 2023-04-25
feat: oci directory pusher

lkingland:
- :gift:  adds a basic pusher for oci-formatted directories

For use by the initial implementation of Scaffolding, this is a basic container pusher which sends a directory which is formatted according to the OCI standard to a given registry using the default keychain, insecurely, or using explicit creds.

/kind enhancement


test/oncluster/scenario_runtime_test.go
```diff
@@ -15,10 +15,12 @@ import (
...
 
 var runtimeSupportMap = map[string][]string{
-	"node":       {"pack", "s2i"},
-	"go":         {},
-	"python":     {"pack", "s2i"},
-	"quarkus":    {"pack", "s2i"},
+	"node":   {"pack", "s2i"},
+	"go":     {},
+	"python": {"pack", "s2i"},
+	// Temporarily disable Quarkus tests.
```
matejvasek, line 21: This should be fixed.

Makefile
```diff
@@ -121,8 +121,11 @@ test-python: ## Test Python templates
 	cd templates/python/http && python3 test_func.py && rm -rf __pycache__
 
 test-quarkus: ## Test Quarkus templates
-	cd templates/quarkus/cloudevents && mvn test && mvn clean
-	cd templates/quarkus/http && mvn test && mvn clean
+	# Quarkus Template Tests are temporarily disabled
+	# see: https://github.com/knative/func/issues/1711
+	#
+	# cd templates/quarkus/cloudevents && mvn test && mvn clean
+	# cd templates/quarkus/http && mvn test && mvn clean
```
lance, line 128: I believe these should be fixed now

lance (commented):
lol I see @matejvasek and I are making the same comments - I'll stop now.

/lgtm
/hold for requested changes

pkg/oci/pusher_test.go
```diff
@@ -0,0 +1,16 @@
...
+	// Create a function
+	// Build the function
+	// Create an HTTP listener on an OS-chosen port ${port}
+	// Create an endpoint which emulates an OCI compliant push-target
+	// Alter the funciton's image tag to be localhost:${port}
+	// Create an OCI pusher instance and .Push
+	// Confirm with the HTTP listener that the expected request was received
+	//  - Only a shallow test is necessary because the unit tests of
+	//    go-containerregistry/pkg/v1/remote ensure it is correct
+	// Note that the test may need to use the explicit username and password
```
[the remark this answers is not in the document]
lkingland: I agree, and I will either add it as a separate PR or update this PR depending on when the OCI builder (`oci/builder.go`) is available, which is being submitted in a separate PR, and is a bit delayed behind this pusher because it depends on #1728, while this pusher has minimal dependencies on other parts of the scaffolding system.

pkg/oci/pusher.go
```diff
@@ -0,0 +1,139 @@
...
+					progress.OptionSetVisibility(term.IsTerminal(int(os.Stdin.Fd()))),
+					progress.OptionSetDescription("pushing"),
+					progress.OptionShowCount(),
+					progress.OptionShowBytes(true),
+					progress.OptionShowElapsedTimeOnFinish())
+			}
+			_ = bar.Set64(update.Complete)
+			continue
+		case <-p.done:
+			fmt.Println("received done signal, finishing bar")
```
lance, line 125: This seems like debugging output that should be removed
```suggestion
```

pkg/oci/pusher.go
```diff
@@ -0,0 +1,139 @@
...
+			if bar != nil {
+				_ = bar.Finish()
+			}
+			return
+		case <-ctx.Done():
+			fmt.Println("context done, finishing bar")
```
lance, line 131: Same here
```suggestion
```

lance (commented):
/lgtm
/hold to remove debugging output

pkg/oci/pusher.go
lkingland: Thanks for that catch

lance:
@lkingland this is still marked as a work in progress - is it ready to go?

lkingland:
> @lkingland this is still marked as a work in progress - is it ready to go?

Yes, I am trying a slightly new process.  To reduce complexity and thus my cognitive load, I am structuring these Scaffolding-related PRs as being strictly serially processed, presuming a latter PR depends on the one prior in all cases (even when in theory it could be merged concurrently).

Since the prior PR was merged, this one could be taken out of draft and merged as well, but this way its preserving the order.

So far this small capitulation has led to an appreciable improvement in my workflow

PR #1715, opened 2023-05-02
chore: update docs from changes to deploy/run

lance:
The docs were overlooked in the changes to deploy and run in https://github.com/knative/func/pull/1693. This commit just updates the docs so they are consistent with the binary.


zroubalik (approved):
/lgtm

PR #1728, opened 2023-05-09
fix: .gitignore .func

lkingland:
- :bug: .gitignore file path incorrect when creating in a different path that CWD
- :bug: .gitignore /.func was causing the build stamp to update
- :gift: can now force inclusion of .func in source control by commenting the line
- :gift: adds built log journaling via a WithStampJournal option

/kind bug
/kind enhancement

```release-note
Fixes an issue where .gitignore was not correctly updated when running func with a custom --path
Improves build caching
```

### DIFF
--- a/pkg/functions/client.go
+++ b/pkg/functions/client.go
@@ -1,16 +1,22 @@
 package functions
 
 import (
+	"bufio"
+	"bytes"
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"net/http"
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"time"
 
+	"gopkg.in/yaml.v2"
 	"knative.dev/func/pkg/utils"
 )
 
@@ -551,7 +557,7 @@ func (c *Client) Init(cfg Function) (Function, error) {
 	f := NewFunctionWith(cfg)
 
 	// Create a .func diretory which is also added to a .gitignore
-	if err = f.ensureRuntimeDir(); err != nil {
+	if err = ensureRunDataDir(f.Root); err != nil {
 		return f, err
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/functions/function_test.go
```diff
@@ -220,3 +226,176 @@ func TestFunction_MigrationError(t *testing.T) {
...
+	if stamp2 == stamp {
+		t.Fatalf("stamping a changed function did not change stamp.  got '%v' again", stamp2)
+	}
+
+	// Asking to stamp again with a journal should result in there being
+	// a "[timestamp]built.log" file in .func
+	if err = f.Stamp(fn.WithStampJournal()); err != nil {
+		t.Fatal(err)
+	}
+	files, err := ioutil.ReadDir(filepath.Join(root, fn.RunDataDir))
```
lance, line 385: Change this to `os.ReadDir()` to fix the linter error, please

pkg/functions/client_test.go
```diff
@@ -1532,83 +1532,3 @@ func TestClient_CreateMigration(t *testing.T) {
 		t.Fatal("freshly created function should have the latest migration")
 	}
 }
-
-// TestClient_BuiltDetects ensures that the client's Built command detects
-// filesystem changes as indicating the function is no longer Built (aka stale)
-// This includes modifying timestamps, removing or adding files.
-func TestClient_BuiltDetects(t *testing.T) {
```
lkingland, line 1539: This test was just moved in its entirety to the functions test, since the `Built` logic is now a method on Function rather than a package static function.

pkg/functions/function_test.go
lkingland: I should have noticed that

pkg/functions/function.go
```diff
@@ -394,19 +408,46 @@ func (f Function) Write() (err error) {
 //
 // The runtime data directory .func is created in the function root if
 // necessary.
-func (f Function) Stamp() (err error) {
+func (f Function) Stamp(oo ...stampOption) (err error) {
```
lkingland, line 411: Journaling is very much an optional, debugging-focused feature, so I used the option pattern here to retain the default API as the happy-path

pkg/functions/client.go
```diff
@@ -960,6 +966,178 @@ func (c *Client) Push(ctx context.Context, f Function) (Function, error) {
 	return f, nil
 }
 
+// ensureRunDataDir creates a .func directory at the given path, and
+// registers it as ignored in a .gitignore file.
+func ensureRunDataDir(root string) error {
```
lkingland, line 971: These functions support the functionality of the Client methods, so they were moved from function.go.  Keeping Function as closed to a data object as possible, with Client being the actor.

pkg/functions/client_test.go
```diff
@@ -51,6 +52,138 @@ func TestClient_New(t *testing.T) {
 	}
 }
 
+// TestClient_New_RunData ensures that the .func runtime directory is
+// correctly created.
+func TestClient_New_RunDataDir(t *testing.T) {
```
lkingland, line 57: This is the new test

matejvasek:
Running `func run` twice in a row do build also on second run, I think that should not be happening.

matejvasek:
> Running `func run` twice in a row do build also on the second run, I think that should not be happening.

Although  this might not be caused by this PR.

lkingland:
> > Running `func run` twice in a row do build also on the second run, I think that should not be happening.
> 
> Although this might not be caused by this PR. EDIT: it is not caused by this PR, it was probably broken some time ago. Deploy does the same unnecessary rebuild, I would swear this worked at some point.

Thanks for taking a look. Yes caching is something that is still not happening with complete reliability, but this PR gets it working in a few places and makes it more auditable so we can get the others.  Also explicitly calling `.Stamp` when the logic dictates that any changes up to that point need not cause a rebuild should help avoid regression 🤞🏻

PR #1730, opened 2023-05-10
feat: host oci builder

lkingland:
- :gift: adds the "host" builder

Adds an initial builder implementation which uses the host directly for building and containerization for Go functions.

Currently not in an active code path (builder and test only) awaiting implementation of scaffolding and a compatible pusher.

/kind enhancement

### DIFF
--- a/docs/reference/func_build.md
+++ b/docs/reference/func_build.md
@@ -57,7 +57,7 @@ func build
 
 ```
       --build-timestamp        Use the actual time as the created time for the docker image. This is only useful for buildpacks builder.
-  -b, --builder string         Builder to use when creating the function's container. Currently supported builders are "pack" and "s2i". ($FUNC_BUILDER) (default "pack")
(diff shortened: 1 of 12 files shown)


docs/reference/func_build.md
```diff
@@ -57,7 +57,7 @@ func build
 
 ```
       --build-timestamp        Use the actual time as the created time for the docker image. This is only useful for buildpacks builder.
-  -b, --builder string         Builder to use when creating the function's container. Currently supported builders are "pack" and "s2i". ($FUNC_BUILDER) (default "pack")
+  -b, --builder string         Builder to use when creating the function's container. Currently supported builders are "pack", "s2i" and "host". ($FUNC_BUILDER) (default "pack")
```
zroubalik, line 60: Should we list here if it is not in the active code path?

lkingland: True; probably not yet.  That change must have snuck in from a later request.

PR #1732, opened 2023-05-10
feat: go scaffolding

lkingland:
- :gift: go scaffolding, instanced and static

Go functions can now be "scaffolded",  adding the ability to use instanced method signatures.  Backwards compatible.

Note that this code-path will be unreachable from the CLI until the forthcoming PR which adds the `--builder=host` option is merged.

/kind enhancement


templates/go/scaffolding/README.md
```diff
@@ -0,0 +1,10 @@
+# Go Scaffolding
+
+Organized into directories conforming to the naming convention
+[signature]-[invocation] where signature is the method signature group expected
```
lkingland, line 4: Also worth noting is that the scaffolding directories which are hidden (beginning with a dot ".") are not avialable when creating a new function, but are instead used for testing and for illustrative purposes.

templates/go/scaffolding/instanced-cloudevent/main.go
```diff
@@ -0,0 +1,17 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	ce "github.com/lkingland/func-runtimes/go/cloudevents"
```
lkingland, line 7: I would expect this repository to be transferred to the Boson project once this string of PRs comprising the Scaffolding POC are a bit more mature.

pkg/functions/client.go
```diff
@@ -618,14 +618,61 @@ func (c *Client) Build(ctx context.Context, f Function) (Function, error) {
...
+		message = fmt.Sprintf("Function built: %v", f.Image)
 	}
 	c.progressListener.Increment(message)
 	return f, err
 }
 
+// Scaffold writes a functions's scaffolding to a given path.
+// It also updates the included symlink to function source 'f' to point to
+// the current function's source.
+func (c *Client) Scaffold(ctx context.Context, f Function, dest string) (err error) {
```
lkingland, line 632: This method will actually be refactored to instead invoke scaffolding from its own package in a forthcoming commit such that the oci builder can invoke it directly without creating a cyclical dependency.

templates/go/scaffolding/README.md
```diff
@@ -0,0 +1,10 @@
...
+(static, instanced etc) and the invocation is the hint of what kind of requests
+the function expects.
```
lance, line 6: This seems like it might be unique to each runtime. For example, with Node.js I don't think separating the instanced path from the static path is necessary. I don't think this is necessarily a problem. Just something to note.

lance (commented):
Ultimately, I assume the templates will change as well. How do you see instanced templates being surfaced there for go?

/lgtm

lance:
lol too late

> /hold

lance:
@lkingland I was thinking of holding this so we could get the scaffolding code moved to knative-sandbox first. We definitely want to do that before this lands in any executable code path. I recommend we rename knative-sandbox/func-go to knative-sandbox/func-runtimes and then a PR to replace what's there with what you have in your personal repo. wdyt?

lkingland:
> 



> Ultimately, I assume the templates will change as well. How do you see instanced templates being surfaced there for go?
> 
> /lgtm

Yes, since the Go templates use the same method signature before and after scaffolding is introduced, they need not change.  The instanced method signature, however, will only be available when using the "host" builder (or if/when we update Pack and S2I to scaffold before building).  The developer will be able to, at their discretion, choose to implement a `func New()...` method instead, and the system will detect and invoke accordingly.

PR #1733, opened 2023-05-10
feat: host-based scaffolded function runner

lkingland:
- :gift: host-based scaffolded function runner

/kind enhancement

### DIFF
--- a/pkg/functions/runner.go
+++ b/pkg/functions/runner.go
@@ -86,12 +86,14 @@ func runFunc(ctx context.Context, job *Job) (runFn func() error, err error) {
 		err = runnerNotImplemented{runtime}
 	case "node":
 		err = runnerNotImplemented{runtime}
+	case "typescript":
+		err = runnerNotImplemented{runtime}
 	case "rust":
 		err = runnerNotImplemented{runtime}
 	case "":
 		err = fmt.Errorf("runner requires the function have runtime set")
 	default:
-		err = fmt.Errorf("runner does not recognized the %q runtime", runtime)
+		err = fmt.Errorf("the %q runtime is not supported", runtime)
 	}
 	return
 }
@@ -101,7 +103,7 @@ type runnerNotImplemented struct {
 }
 
 func (e runnerNotImplemented) Error() string {
-	return fmt.Sprintf("the runtime %q is not supported by the host runner.  Try running containerized.", e.Runtime)
+	return fmt.Sprintf("the %q runtime may only be run containerized.", e.Runtime)
 }
 
 func runGo(ctx context.Context, job *Job) (err error) {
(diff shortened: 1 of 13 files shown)


pkg/functions/runner.go
```diff
@@ -0,0 +1,221 @@
...
+	// Wait for it to become available before returning the metadata.
+	err = waitFor(job, defaultRunTimeout)
+	return
+}
+
+// runFunc returns a function which will run the user's Function based on
+// the jobs runtime.
+func runFunc(ctx context.Context, job *Job) (runFn func() error, err error) {
+	runtime := job.Function.Runtime
+	switch runtime {
```
lance, line 80: Please add `"typescript"` as a case for this switch statement, returning `runnerNotImplemented`.

pkg/functions/runner.go
```diff
@@ -0,0 +1,221 @@
...
+	case "java":
+		err = runnerNotImplemented{runtime}
+	case "node":
+		err = runnerNotImplemented{runtime}
+	case "rust":
+		err = runnerNotImplemented{runtime}
+	case "":
+		err = fmt.Errorf("runner requires the function have runtime set")
+	default:
+		err = fmt.Errorf("runner does not recognized the %q runtime", runtime)
```
lance, line 94: Nit.
```suggestion
		err = fmt.Errorf("the %q runtime is not supported", runtime)
```

pkg/functions/runner.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	return
+}
+
+type runnerNotImplemented struct {
+	Runtime string
+}
+
+func (e runnerNotImplemented) Error() string {
+	return fmt.Sprintf("the runtime %q is not supported by the host runner.  Try running containerized.", e.Runtime)
```
lance, line 104:
```suggestion
	return fmt.Sprintf("the runtime %q may only be run containerized.", e.Runtime)
```
Not sure if this is the best wording. I'm just trying to follow convention that errors start with lowercase and do not contain final punctuation. That seems to imply only a single statement.

lance (commented):
Looks great! Just a few minor comments.

pkg/functions/runner.go
lance: This wording is not specific enough. My intent is to hide the internal runner resource.

lkingland: "runner does not recognized"  :wince:  thanks

pkg/functions/runner.go
lkingland: Yes, thank you very much.  Usually when writing these error messages my mind is mostly fully occupied by the logic in development and I tend to not spend enough time ensuring the wording is grammatically correct and informative.

lkingland:
Thanks for the review @lance.  Your improvements to the PR have been made

PR #1735, opened 2023-05-10
Prepare pipelines as code template for midstream customization

jrangelramos:
# Changes

- :broom: preparing pipelines as code template for midstream customization

With this change downstream builds can customize the task source for pipeline as code template generated by `func config git set`

### DIFF
--- a/pkg/pipelines/tekton/templates.go
+++ b/pkg/pipelines/tekton/templates.go
@@ -20,9 +20,9 @@ const (
 
 	// Tasks references
 	taskGitCloneRef       = "git-clone"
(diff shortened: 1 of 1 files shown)


pkg/pipelines/tekton/templates.go
```diff
@@ -32,6 +32,8 @@ const (
 	defaultPipelinesTargetBranch = "main"
 )
 
+var FuncRepoRef = "knative/func/main"
```
lance, line 35: In midstream this will be patched to `openshift-knative/kn-plugin-func/release-x.x`?

PR #1736, opened 2023-05-10
chore: remove Procfile from Rust templates

matejvasek:
# Changes

- :broom: Removed Procfile from Rust templates. It appears the file is no longer needed.


lance:
Curious why it's not needed any longer.
/lgtm

matejvasek:
> > Curious why it's not needed any longer. /lgtm
> 
> We did testing and it works without it, not sure why.

Probably the need was removed some time ago.

matejvasek:
I guess this is it: https://github.com/paketo-community/cargo/pull/116/

PR #1737, opened 2023-05-10
fix: use full builder for Rust

matejvasek:
# Changes

- :broom: Use `full` builder for Rust. This is need to make some libs work (e.g. openssl).


lance (commented):
/lgtm

PR #1739, opened 2023-05-15
chore: improve error message when no func on path

matejvasek:
# Changes

- :broom: More clear error message when path does not contain initialized function.


lance (commented):
/lgtm

Issue #1744, opened 2023-05-17
The `build` and `deploy` commands give confusing errors if not in project directory


lance:
Most function commands give a reasonable error message if invoked from outside of a function project directory and `FUNC_REGISTRY` is not set in the environment. For example:

```
func run
Error: the given path '' does not contain an initialized function
```

However the `deploy` and `build` commands immediately drop the user into an interactive prompt asking for the registry, and then giving another confusing message because it can't figure out the function name.

```
func deploy
A registry for function images is required. For example, 'docker.io/tigerteam'.
X Sorry, your reply was invalid: invalid registry ["quay.io/lanceball"]: name required
? Registry for function images:
Error: registry required to build function, please set with `--registry` or the FUNC_REGISTRY environment variable
```

This problem is ameliorated somewhat when `FUNC_REGISTRY` is set, but even then the error message could be improved. I think it should say something like the error message above for `func run` mentioning that the path does not contain an initialized function.

```
export FUNC_REGISTRY=quay.io/lanceball

func deploy
   Building function image
Error: name required
```

/kind bug
/cc @ReToCode

PR #1745, opened 2023-05-17
chore: improve error messages for most commands

lance:
# Changes

- :broom: adds a new error type `UninitializedError` to minimize redundancy across commands, and applies it in all of the commands which need an initialized function to operate.

Fixes: https://github.com/knative/func/issues/1744

/kind cleanup


zroubalik (approved):
/lgtm
/approve

PR #1750, opened 2023-05-23
feat: enable scaffolding for host builds

lkingland:
- :gift: creates the scaffolding package
- :gift: OCI builder updated to use scaffolding package and thus function source
- :bug: Fixes a bug where local jobs were not receiving context cancellation
- :bug: Fixes a bug where signatures of instanced HTTP was throwing errors
- :broom: adds a concurrency test which ensures only one build executes per
  process

Connects the scaffolder to function source code by factoring out scaffolding into a package which is used both by the client and by other packages, allowing for the removal of the mocks.

/kind enhancement

```release-note
- Fixes a bug where local jobs were sometimes not canceling immediately
- Startup timeout for local run tasks now configurable
```


pkg/scaffolding/scaffold.go
```diff
@@ -0,0 +1,87 @@
...
+	// a runtime and invocation hint (default "http")
+	s, err := detectSignature(src, runtime, invoke)
+	if err != nil {
+		return err
+	}
+
+	// Path in the filesystem at which scaffolding is expected to exist
+	d := fmt.Sprintf("%v/scaffolding/%v", runtime, s.String()) // fs uses / on all OSs
+	if _, err := fs.Stat(d); err != nil {
+		return fmt.Errorf("no scaffolding found for '%v' signature '%v'. %v.  Note that this may require the filesystem be regenerated (try make first).", runtime, s, err)
```
lance, line 43: The last part of this error seems to be targeted at the `func` developer instead of a functions developer. I am a bit confused by it.

lance (commented):
/lgtm
/hold for others to provide feedback
There's just one question I have regarding an error message that is a little confusing.

lkingland:
Thanks for the review @lance.  The update to the error message you noticed has been made, as well as several other updates to improve error handling and test coverage (PR description updated)

    - configurable platforms for build tasks
    - additional typed errors and reorganization
    - expanded test cases to check typed errors are used
    - configurable global and per-run startup timeout
    - tests run readiness endpoint is considered and respects timeouts
    - adds a concurrency test which ensures only one build executes per
      process
    - passes the --platform flag through as a build option

pkg/oci/containerize_go.go
```diff
@@ -55,6 +54,10 @@ func (c goLayerBuilder) Build(cfg *buildConfig, p v1.Platform) (desc v1.Descript
 }
 
 func goBuild(cfg *buildConfig, p v1.Platform) (binPath string, err error) {
+	if cfg.tester != nil && cfg.tester.emulateSlowBuild {
```
lkingland, line 57: Embedding "if testing do X" in core logic is not idea, but in this case I feel it is warranted since the complexity of adding a mockable builder implementation is so much greater than a simple "if testing slowness, pause here"

pkg/functions/runner_test.go
```diff
@@ -1,62 +1,44 @@
...
 	"testing"
-
-	fn "knative.dev/func/pkg/functions"
-	"knative.dev/func/pkg/oci"
-	. "knative.dev/func/pkg/testing"
 )
 
-// TestRunner ensures that the default internal runner correctly executes
-// a scaffolded function.
-func TestRunner(t *testing.T) {
```
lkingland, line 19: This test was moved to the client test such that this file could be brought inside the functions package, and thereby allowing the TestGetRunFuncErrors test to be writen which unit-tests the private getRunFunc function.

pkg/oci/builder.go
```diff
@@ -33,25 +34,18 @@ var defaultIgnored = []string{ // TODO: implement and use .funcignore
...
-	return fmt.Sprintf("error performing host build. %v", e.Err)
-}
-
 // Builder which creates an OCI-compliant multi-arch (index) container from
 // the function at path.
 type Builder struct {
 	name    string
 	verbose bool
+
+	tester *testHelper
```
lance, line 43: Do we really want a test helper in the main body of the code?

pkg/oci/builder.go
```diff
@@ -63,52 +57,60 @@ func NewBuilder(name string, verbose bool) *Builder {
...
 	// build directory.  Would be better to have a formal build result object
 	// which includes a general struct which can be used by all builders to
 	// communicate to the pusher where the image can be found.
+	// Tests, however, can use a simple channel:
+	if cfg.tester != nil && cfg.tester.notifyDone {
+		if cfg.verbose {
+			fmt.Println("tester configured to notify on done.  Sending to unbuffered doneCh")
+		}
+		cfg.tester.doneCh <- true
+		fmt.Println("send to doneCh complete")
```
lance, line 99: This seems like debugging output or perhaps should be wrapped in `if cfg.verbose`

pkg/oci/builder.go
```diff
@@ -269,3 +267,22 @@ func updateLastLink(cfg *buildConfig) error {
...
+	pausedCh     chan any
+}
+
+func newTestHelper() *testHelper {
+	return &testHelper{
+		continueCh: make(chan any),
+		doneCh:     make(chan any),
+		pausedCh:   make(chan any),
+	}
+}
```
lance, line 288: I think I saw a comment of yours yesterday about adding test structures in the main code. Can you help me understand why this is necessary?

pkg/oci/errors.go
```diff
@@ -0,0 +1,20 @@
...
+func (e BuildErr) Error() string {
+	return fmt.Sprintf("error performing host build. %v", e.Err)
+}
+
+type ErrBuildInProgress struct {
+	Dir string
+}
+
+func (e ErrBuildInProgress) Error() string {
+	return fmt.Sprintf("Build directory already exists for this version hash and is associated with an active PID.  Is a build already in progress? %v", e.Dir)
```
lance, line 19: Nit: is there a way to make this consistent with other error messages? I.e. starting with a lowercase letter and no real punctuation? A suggestion - perhaps it could be better.
```suggestion
	return fmt.Sprintf("a build for this function is associated with an active PID appears to be already in progress %v", e.Dir)
```

pkg/scaffolding/scaffold.go
```diff
@@ -0,0 +1,87 @@
...
+	static, instanced, err := d.Detect(src)
+	if err != nil {
+		return
+	}
+	// Function must implement either a static handler or the instanced handler
+	// but not both.
+	if static && instanced {
+		return s, fmt.Errorf("function may not implement both the static and instanced method signatures simultaneously")
+	} else if !static && !instanced {
+		return s, fmt.Errorf("function does not appear to implement any known method signatures")
```
lance, line 81: Nit - a small simplification
```suggestion
		return s, fmt.Errorf("function does not implement any known method signatures")
```

pkg/scaffolding/scaffold_test.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+// TestWrite ensures that the Write method writes Scaffolding to the given
+// destination.  This is a failry shallow test.  See the Scaffolding and
+// Detector tests for more depth.
+func TestWrite(t *testing.T) {
+	// The filesystem containing scaffolding is expected to conform to the
+	// structure:
+	// /[language]/scaffolding/["instanced"|"static"]-[invocation]
```
lance, line 59: TODO: we will need to update existing language pack documentation with this kind of information

lance (commented):
Looking really good, @lkingland! I just have a few more comments.

aslom:
I have it build locally and tested OK. The code looks like good starting point with tests.

I have also tested that existing functionality is still working oK

```
 ../func-scaffolding-enable/func deploy --builder=pack
   🙌 Function built: docker.io/aslom/wt1:latest
   ⬆️  Deploying function to the cluster
W0606 12:36:00.466532   68896 warnings.go:70] Kubernetes default value is insecure, Knative may default this to secure in a future release: spec.template.spec.containers[0].securityContext.allowPrivilegeEscalation, spec.template.spec.containers[0].securityContext.capabi
   ✅ Function updated in namespace "default" and exposed at URL:
   http://wt1.default.127.0.0.1.sslip.io
```

Looking forward to see more functionality.

LGTM

pkg/functions/client_test.go
```diff
@@ -1709,3 +1710,68 @@ func TestClient_CreateMigration(t *testing.T) {
...
+	// Initialize
+	f, err := client.Init(fn.Function{Root: root, Runtime: "go", Registry: TestRegistry})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Replace the implementation with the test implementation which will
+	// return a non-200 response for the first 10 seconds.  This confirms
+	// the client is waiting and retrying.
+	// TODO: we need an init option which skips writing example source-code.
```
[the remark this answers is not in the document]
lkingland: Yes, I would expect it to work like that; very much like `git init`, which leaves the directory untouched other than creating `.git`

pkg/functions/client_test.go
```diff
@@ -1709,3 +1710,68 @@ func TestClient_CreateMigration(t *testing.T) {
...
+	client := fn.New(fn.WithBuilder(oci.NewBuilder("", true)), fn.WithVerbose(true))
+
+	// Initialize
+	f, err := client.Init(fn.Function{Root: root, Runtime: "go", Registry: TestRegistry})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Replace the implementation with the test implementation which will
+	// return a non-200 response for the first 10 seconds.  This confirms
```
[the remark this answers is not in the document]
lkingland: Yes, as a matter of fact I did reduce it considerably, and updated the comment here as well.

pkg/functions/client_test.go
```diff
@@ -1709,3 +1710,68 @@ func TestClient_CreateMigration(t *testing.T) {
...
+	// Run
+	// The function returns a non-200 from its readiness handler at first.
+	// Since we already confirmed in another test that a timeout awaiting a
+	// 200 response from this endpoint does indeed fail the run task, this
+	// delayed 200 confirms there is a retry in place.
+	job, err := client.Run(ctx, f)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
```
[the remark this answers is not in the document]
lkingland: Yes, that is unnecessary.  My intention here is to illustrate the expectation for how to use the job returned from runner implementations:  to immediately queue a job.Stop in a defer statement, such that it is very likely to happen, even if logic later throws an error (in the case of the host runner, this stop task removes the job's runtime directory).
In this case it is of course unnecessary because there _is_ no logic afterwards that could fail.   If this seemed strange to you, then it would seem strange to others, so I removed it.

pkg/oci/builder.go
lkingland: Yes I completely agree this is not ideal, however when I weighed the simplicity of this against the complexity of creating a mocked internal implementation to create this delay testing scenario (note that this is within a structure which itself is already mocked and behind an interface), I the tradeoff was worthwhile.  I can be convinced otherwise...

pkg/oci/builder.go
lkingland: Yes this only executes if there is a `tester`, so it will only print in the tests.

pkg/oci/builder.go
lkingland: Yes, I did make a comment about this yesterday.   Testing concurrent build arbitration logic requires forcing one of two or more build tasks to go slow enough to cause a collision with the other.  To do this without creating a racing condition is nicely handled using channels.

To do this while keeping everthing test-related out of the core code would, in my opinion, be much more difficult to understand than the test helper with a few channels here.

So, in the end, I decided to try this, favoring simplicity

PR #1752, opened 2023-05-23
Use Contour Ingress instead of Kourier

matejvasek:
# Changes

-  Use Contour Ingress instead of Kourier. Contour will be useful for exposing resources other that knative services (e.g. PaC controller for webhooks).

- User LoadBalancer instead of NodePort. This makes services available on the same port (80) on both:  localhost and docker network.


matejvasek:
@lance please override the shell lint action.

lance (approved):
/lgtm

Issue #1716, opened 2023-05-02
Support tekton v0.44.*+ 


jrangelramos:
We need at some point to support newer versions of Tekton  (v0.44.*+) such as v0.44.2 and v0.47.0.

Currently func remote builds works well till v0.42.*.
The following error is returned on CLI for newer version due Tekton api change

```
$ func deploy --remote
Creating Pipeline resources
W0502 16:40:13.825853  129470 reflector.go:347] github.com/tektoncd/pipeline/pkg/client/informers/externalversions/factory.go:117: watch of *v1.PipelineRun ended with: an error on the server ("unable to decode an event from the watch stream: unable to decode watch event: no kind \"PipelineRun\" is registered for version \"tekton.dev/v1\" in scheme \"github.com/tektoncd/pipeline/pkg/client/clientset/versioned/scheme/register.go:32\"") has prevented the request from succeeding

```

PR #1753, opened 2023-05-24
deps: update tekton to latest versions

lance:
# Changes

Fixes: https://github.com/knative/func/issues/1716

- :broom: Bumps `tekton/cli` and `tekton/pipelines` to 0.31.0 and 0.47.0 respectively.

/cc @jrangelramos


matejvasek:
@lance this probably requires Go higher than 1.18.

jrangelramos:
Hi @lance maybe we could upgrade CI scripts as well and see how it goes... https://github.com/knative/func/blob/main/hack/tekton.sh#L25

jrangelramos:
btw it works fine with OSH Pipelines 1.10

lance:
@knative/func-reviewers ptal

matejvasek:
@lance you might need to override the linter

PR #1761, opened 2023-05-25
feat: adds a `func environment` command

lance:
# Changes

- :gift: Adds `func environment` to print the current function execution environment as JSON.

See: https://github.com/knative/func/discussions/1676

/kind enhancement

**Release Note**

```release-note
Adds `func environment` to print the current function execution environment as JSON.
```

### DIFF
--- a/cmd/environment.go
+++ b/cmd/environment.go
@@ -1,0 +1,129 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/ory/viper"
+	"github.com/spf13/cobra"
+
+	"knative.dev/func/pkg/config"
+	"knative.dev/func/pkg/functions"
+	"knative.dev/func/pkg/k8s"
+)
+
+func NewEnvironmentCmd(newClient ClientFactory, version *Version) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "environment",
+		Short: "Display function execution environment information",
+		Long: `
+NAME
+	{{rootCmdUse}} environment
+
+SYNOPSIS
+	{{rootCmdUse}} environment - display function execution environment information
+
+DESCRIPTION
+	Display information about the function execution environment, including
+	the version of func, the version of the function spec, the default builder,
+	available runtimes, and available templates.
+`,
+		SuggestFor: []string{"env", "environemtn", "enviroment", "enviornment", "enviroment"},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runEnvironment(cmd, newClient, version)
+		},
+	}
+	cfg, err := config.NewDefault()
+	if err != nil {
+		fmt.Fprintf(cmd.OutOrStdout(), "error loading config at '%v'. %v\n", config.File(), err)
+	}
+
+	addVerboseFlag(cmd, cfg.Verbose)
+
+	return cmd
+}
+
+type Environment struct {
+	Version     string
+	Build       string
+	SpecVersion string
+	SocatImage  string
+	TarImage    string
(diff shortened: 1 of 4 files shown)


lance:
The code style errors are catching the `SuggestFor` values which are deliberately misspelled.

/override "style / Golang / Lint"

lkingland:
This /lgtm!

/hold for anyone else to perhaps comment, but seems to me this is a great place to start iterating if we think of anything else to add, etc.

lkingland (approved):

cmd/environment.go
```diff
@@ -0,0 +1,148 @@
...
+	cc, err := k8s.GetClientConfig().ClientConfig()
+	if err != nil {
+		fmt.Printf("Error getting client config %v\n", err)
+	} else {
+		host = cc.Host
+	}
+
+	environment := Environment{
+		Version:     v.String(),
+		Build:       v.Hash,
```
matejvasek, line 97: I would call this field `Git Revision`.
Also I would add `Build Date` field which is more telling to normal user.

jrangelramos:
it looks good to me. Maybe one very small cosmetic suggestion would be not printing null.. i.e. when no func env var is defined.
Currently it prints
```
"Environment": null,
```
what about?
```
"Environment": [],
```
but this is cosmetic..

cmd/environment.go
matejvasek:
```diff
[k8s:kind@kind-func/default] (venv) ~/devel/kn-plugin-func   func-environment ±  git diff | cat                                                                                                              ST 100   func-environment 
diff --git a/cmd/environment.go b/cmd/environment.go
index 8e63bad8..b987f13c 100644
--- a/cmd/environment.go
+++ b/cmd/environment.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/ory/viper"
 	"github.com/spf13/cobra"
@@ -47,7 +48,8 @@ DESCRIPTION
 
 type Environment struct {
 	Version     string
-	Build       string
+	GitRevision string
+	BuildDate   string
 	SpecVersion string
 	SocatImage  string
 	TarImage    string
@@ -92,9 +94,15 @@ func runEnvironment(cmd *cobra.Command, newClient ClientFactory, v *Version) (er
 		host = cc.Host
 	}
 
+	d, err := time.Parse("20060102T150405Z", v.Date)
+	if err != nil {
+		panic(err)
+	}
+
 	environment := Environment{
 		Version:     v.String(),
-		Build:       v.Hash,
+		GitRevision: v.Hash,
+		BuildDate:   d.Local().Format(time.DateTime),
 		SpecVersion: functions.LastSpecVersion(),
 		SocatImage:  k8s.SocatImage,
 		TarImage:    k8s.TarImage,
```

jrangelramos:
It might be a bit late for this PR (maybe for another).. but what about a `-o [json|yaml|human|jsonpath]` option?
That would be cool seeing this in other formats (more human readable) following `kn`, `podman, `oc` and other cli standards..

lance:
@jrangelramos I assumed that would follow in a subsequent PR and have that as a TODO here: https://github.com/knative/func/pull/1761/files#diff-307b508c3eb9b3182cd02cbf3b681ab8f07937ce80be379fdd361bc9c5387d80R154

Seems like it could be a `good-first-issue`, no?

lance:
@jrangelramos @matejvasek thanks for the feedback; I've incorporated it (for additional output formats, I will create an Issue).

matejvasek:
@lance you need to override the linter.

PR #1771, opened 2023-06-05
fix: PVCSize resets to default with futureproofing

lkingland:
Refactors PVCSize to treat the Function member as a custom setting, and the in-code constant as the default.

- :bug: Fixes a bug where the PVCSize would reset to the default on deploy
- :gift: Allows the default to be changed by future func version, only pinning the value if explicitly requested by the user.

/kind bug

```release-note
- Fixes a bug where volume size for remote builds was sometimes not respected
```


zroubalik (approved):
/lgtm

PR #1773, opened 2023-06-05
src: add missing 'omitempty' annotations

lkingland:
- :broom: image and registry missing `omitempty`

The default func.yaml should be as empty as possible, as it is intended as a way to specify overrides to in-code defaults.
By pre-populating the func.yaml with empty strings, it is ambiguous as to the intent:  eg.  Are you requesting I explicitly set the regitry and image to nothing, or do you not care what I set it too? 

In other words, the initial funciton created by `func craete` should have a registry and image `nil`, indicating the user prefers to use whatever is default, not explicitly setting to emtpy string.

This was mostly fixed in PR #1630.  I think these two were just missed.

/kind cleanup

### DIFF
--- a/pkg/functions/function.go
+++ b/pkg/functions/function.go
@@ -56,7 +56,7 @@ type Function struct {
 
 	// Registry at which to store interstitial containers, in the form
 	// [registry]/[user].
-	Registry string `yaml:"registry"`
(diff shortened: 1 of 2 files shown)


lance (commented):
/lgtm

PR #1774, opened 2023-06-05
feat: source-defined buildpacks

lkingland:
- :broom: define default go buildpacks in-code

Allows buildpacks to be updated by updating to a newer version of func.  Previously, buildpacks were fixed at time of function creation without a manual edit of func.yaml.  Currently using Go as the reference implementation in this commit.

/kind cleanup


zroubalik (approved):
/lgtm

matejvasek:
I really dislike this. What if somebody is using custom builder that does not need these additional buildpacks? There is no way to remove them from build. When it's in func.yaml user can alter behaviour.

PR #1775, opened 2023-06-05
feat: add common command aliases

lkingland:
- :gift: add `ls` as an alias for `func list`
- :gift: add `rm` as an alias for `func remove`


zroubalik (approved):
/lgtm

great idea!

PR #1776, opened 2023-06-05
src: remove datestamp from binary

lkingland:
- :broom: removes version from `func version --verbose`

In an effort to move one step closer to reproducible builds, build timestamp is removed from the CLI version subcommand.


zroubalik (approved):
/lgtm

PR #1777, opened 2023-06-05
src: invoke response cleanup

lkingland:
- :broom: Suppress invocation logging

Reduces output from `func invoke` to only print when `--verbose`.

In particular the text `Received Response:` is confusing as to where it originated, so it has been placed in the "pretty-printed" response yielded when invoking in verbose mode (`func invoke --verbose`)

/kind cleanup


zroubalik (approved):
/lgtm

Issue #1760, opened 2023-05-25
Function starting aborted after about 30 seconds.


rabejens:
I have a function template which takes longer than 30 seconds to become ready. Is it possible to increase the timeout in `func.yaml` to mitigate this?

Followup 1:

I configured my `knative-serving` to set the `scale-to-zero-grace-period` to 600s (10 minutes) but this did not change anything. It is still scaling to zero after 30 seconds.

Followup 2:

Even disabling scale-to-zero completely has it killed after about 30 seconds.

Followup 3:

As a workaround, I currently made liveness and readiness instantly return 200 OK even though the function is still initializing. While this works for the time being, this is not optimal.

lance:
I have tried to replicate this using a simple function. I was not able to replicate it with a 60 second delay, but a 2 minute delay resulted in the failure message, "Initial scale was never achieved". Is this what you are seeing too?

<details><summary>Node.js function with liveness and readiness probes with 2 minute delay</summary>

```js

const handle = async (context, body) => {
  return { "message": "Hello world!" };
}

// Wait until the delay completes before returning 200 OK
const readiness = (request, reply) => {
  if (ready) {
    reply.statusCode = 200;
    return 'Yeah boyeee';
  } else {
    reply.statusCode = 503;
    return 'Not ready';
  }
}

const liveness = readiness;

// Utility function to simulate a delay
const sleep = async (ms) => {
  return new Promise(resolve => setTimeout(resolve, ms));
}

let ready = false;
// 120 second delay to simulate a long startup time
sleep(120000).then( _ => { ready = true; }).catch( e => { console.error(e); })

// Export the function
module.exports = { handle, readiness, liveness };
```
</details>

Running this function yields

```console
kn service list
NAME   URL                                     LATEST      AGE   CONDITIONS   READY   REASON
foo    http://foo.default.127.0.0.1.sslip.io   foo-00005   43h   1 OK / 3     False   RevisionFailed : Revision "foo-00006" failed with message: Initial scale was never achieved.
```

PR #1782, opened 2023-06-06
feat: start timeout for host runs

lkingland:
- :gift: start timeout for host runs

Host-runs now support a Start Timeout option.  The setting is added to the function as `f.Run.StartTimeout`.

Related: #1760

/kind enhancement


docs/reference/func_config_git_set.md
```diff
@@ -26,7 +26,7 @@ func config git set
       --gh-webhook-secret string   GitHub Webhook Secret used for payload validation. If not specified, it will be generated automatically.
   -t, --git-branch string          Git revision (branch) to be used when deploying via the Git repository ($FUNC_GIT_BRANCH)
   -d, --git-dir string             Directory in the Git repository containing the function (default is the root) ($FUNC_GIT_DIR)
-      --git-provider string        The type of the Git platform provider to setup webhook. This value is usually automatically generated from input URL, use this parameter to override this setting. Currently supported providers are "github".
+      --git-provider string        The type of the Git platform provider to setup webhook. This value is usually automatically generated from input URL, use this parameter to override this setting. Currently supported providers are "github" and "gitlab".
```
lance, line 29: I suspect this is from @matejvasek's latest gitlab PR which wasn't reflected in https://github.com/knative/func/pull/1769

pkg/docker/runner.go
```diff
@@ -101,7 +101,11 @@ func (n *Runner) Run(ctx context.Context, f fn.Function) (job *fn.Job, err error
 		}
 	}()
 
-	// Start
+	// TODO: use StartTimeout
```
lance, line 104: Presumably this is why it's a work in progress? 😄

pkg/functions/client_test.go
```diff
@@ -723,6 +730,69 @@ func TestClient_Run_DataDir(t *testing.T) {
...
+
+	// Build
+	if f, err = client.Build(ctx, f); err != nil {
+		t.Fatal(err)
+	}
+
+	// Run
+	// with a very short timeout
+	_, err = client.Run(ctx, f, fn.RunWithStartTimeout(100*time.Millisecond))
+	// TODO: replace following check with `!errors.Is(err, fn.ErrRunTimeout)`
```
lance, line 789: Merged in https://github.com/knative/func/pull/1783

lance (commented):
Some minor feedback (nits, really). Looks good so far. There appear to be modifications from some of your other PRs in here which presumably will make this PR smaller when the others are merged

pkg/functions/runner.go
```diff
@@ -156,10 +152,9 @@ func waitFor(job *Job, timeout time.Duration) error {
 	if job.verbose {
 		fmt.Printf("Waiting for %v\n", url)
 	}
-	to := time.After(timeout)
 	for {
 		select {
-		case <-to:
+		case <-time.After(timeout):
```
matejvasek, line 157: bad rebase?

docs/reference/func_config_git_set.md
lkingland: Yep, these are easy to forget.  It would be nice to have a CI task which regens the docs and checks if git shows a diff or some other such failsafe

pkg/docker/runner.go
lkingland: Haha, no no.  The flag --start-timeout was a necessary digression in order to get the host-based runner functioning.  So this PR only implements that path:  a host-based (non-containerized) run.   It might be two good first issues to have the flag also respected for containerized runs, and by the knative deployer.

pkg/functions/runner.go
matejvasek: I still see incorrect version.

matejvasek: https://github.com/knative/func/pull/1804#discussion_r1224500551
https://github.com/knative/func/pull/1804/commits/14c4c49a14bba1abac4b3f1d8f5074be1361840f

lkingland: Either something is cached somewhere, I missed a different bug, or you think the other way was better :)

pkg/functions/runner.go
```diff
@@ -151,44 +147,56 @@ func runGo(ctx context.Context, job *Job) (err error) {
...
-	if verbose {
-		fmt.Printf("Endpoint returned HTTP %v.\n", resp.StatusCode)
-		dump, _ := httputil.DumpResponse(resp, true)
-		fmt.Println(dump)
+		select {
+		case <-to:
+			return ErrRunTimeout{timeout}
+		case <-ctx.Done():
+			return ErrContextCanceled
+		case <-time.After(interval):
```
matejvasek, line 195: This seems to be redundant and racy as the ctx will be cancelled at the same time.

pkg/functions/runner.go
```diff
@@ -151,44 +147,56 @@ func runGo(ctx context.Context, job *Job) (err error) {
...
 		}
-		return false
-	}
-	defer resp.Body.Close()
 
-	if resp.StatusCode == 200 {
-		return true
-	}
+		if res != nil {
+			_ = res.Body.Close()
```
matejvasek, line 187: I would prefer using defer instead, just wrap critical code into `func() {codehere}()`

lkingland: This also went away when extracting the single readiness check to its own function.

lkingland:
Thanks for helping make sure the logic was correct @matejvasek .

@lance do you think the suggestion above will work: allowing this PR through, which implements the start-timeout for host-based runs, and then opening separate issues to track implementing the option for containerized runs and on deployed services?

lance:
> Thanks for helping make sure the logic was correct @matejvasek .
> 
> @lance do you think the suggestion above will work: allowing this PR through, which implements the start-timeout for host-based runs, and then opening separate issues to track implementing the option for containerized runs and on deployed services?

That seems like a reasonable approach to me

lkingland:
> Could we use feature flag (simple environment variable?) to hide options until they are fully working?

I could, yes, if you think that's entirely necessary.  But I would ask;  is there something in particular from this PR which is not working well enough?  It's worth perhaps noting that 1) this is a new code path, so it's not altering existing functionality and 2) it is part of a larger string of PRs, which, once merged I presume will be full-featured enough to be enabled.  You can check out my develop branch to see the final product to see the state if I can get enough reviews to merge these 7 PRs.

aslom:
@lkingland with feature flag we should have more freedom to experiment without worrying that new options in CLI may confuse users? Also make it easier to test if previous behavior is not affected if the code changing it is behind feature flag? I am ok either way.

aslom:
I have verified that make test works and did basic func run and deploy and everything looks good. Great to see run timeout addition. Hopefully soon exposed:

```
../func/func run --start-timeout 10m
Error: the ability to specify the startup timeout for containerized runs is coming soon
```


/lgtm

PR #1787, opened 2023-06-06
chore: default s2i builder as vars

jrangelramos:
# Changes

- :broom: default s2i builder as var for ldflags


jrangelramos:
Maybe we should list these defaults on `func environment` command (or on `func version -v`)

lance:
> Maybe we should list these defaults on `func environment` command (or on `func version -v`)

This is a great suggestion

lance (commented):
/lgtm

PR #1790, opened 2023-06-07
feat: support dependencies in host-built functions

lkingland:
- :gift:  Host-built function dependencies

Host-built functions can now import third-party dependencies.

In particular scaffolded (host-built) functions now run a go get on the function (treating it as a library).

/kind enhancement

### DIFF
--- a/pkg/oci/containerize_go.go
+++ b/pkg/oci/containerize_go.go
@@ -21,7 +21,7 @@ type goLayerBuilder struct{}
 // the statically linked binary in a tarred layer and return the Descriptor
 // and Layer metadata.
 func (c goLayerBuilder) Build(cfg *buildConfig, p v1.Platform) (desc v1.Descriptor, layer v1.Layer, err error) {
-	// Executable
+	// Build the Executable
 	exe, err := goBuild(cfg, p) // Compile binary returning its path
 	if err != nil {
 		return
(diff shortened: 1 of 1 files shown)


pkg/functions/runner.go
```diff
@@ -153,31 +149,27 @@ func runGo(ctx context.Context, job *Job) (err error) {
...
 		fmt.Printf("Waiting for %v\n", url)
 	}
 	for {
 		select {
 		case <-time.After(timeout):
 			return errors.New("timed out waiting for function to be ready")
-		case <-tick.C:
+		case <-time.After(200 * time.Millisecond):
 			resp, err := http.Get(url)
 			defer resp.Body.Close()
```
matejvasek, line 162: This may lead to `Close()` on `nil` body.

matejvasek: Also `defer` in a loop is not good idea.

aslom: Yes that seems to be the problem - I confirmed by moving defer adter error check and make test work https://github.com/knative/func/issues/1801

lkingland: #1801 is now merged, thanks @aslom 
#1804 is now open which incorporates your suggestion here @matej and Lance's suggestion to expand the typed timeout error.

salaboy (approved):
/lgtm

Issue #1801, opened 2023-06-08
make tests hanging, SIGSEGV when interrupted with Ctl-c


aslom:
Running `make tests` hangs for minutes (that is how long I waited)  - I do not think it is normal?

When I press Ctrl-C I get all kinds output including SIGSEGV - not sure how to interpret it?

How do you find out which test is failing and what causing it?

This is in main branch and on Macbook Pro M1 (details below)

```
aslom@m1:~/Documents/awsm/go/src/github.com/knative/func|main⚡ ⇒  make test
# Removing temporary template files
go generate pkg/functions/templates_embedded.go
go test -race -cover -coverprofile=coverage.txt ./...
?   	knative.dev/func/cmd/func	[no test files]
?   	knative.dev/func/cmd/prompt	[no test files]
?   	knative.dev/func/docs/generator	[no test files]
?   	knative.dev/func/generate	[no test files]
?   	knative.dev/func/generate/templates	[no test files]
?   	knative.dev/func/pkg/git/github	[no test files]
?   	knative.dev/func/pkg/git/gitlab	[no test files]
?   	knative.dev/func/pkg/k8s/labels	[no test files]
?   	knative.dev/func/pkg/mock	[no test files]
?   	knative.dev/func/pkg/openshift	[no test files]
?   	knative.dev/func/pkg/pipelines	[no test files]
ok  	knative.dev/func/cmd	8.547s	coverage: 45.5% of statements
ok  	knative.dev/func/cmd/templates	0.701s	coverage: 81.5% of statements
ok  	knative.dev/func/pkg/builders	0.328s	coverage: 88.0% of statements
ok  	knative.dev/func/pkg/builders/buildpacks	1.125s	coverage: 45.7% of statements
ok  	knative.dev/func/pkg/builders/s2i	0.373s	coverage: 71.0% of statements
ok  	knative.dev/func/pkg/config	1.769s	coverage: 88.1% of statements
ok  	knative.dev/func/pkg/docker	6.915s	coverage: 84.3% of statements
?   	knative.dev/func/pkg/pipelines/tekton/pac	[no test files]
?   	knative.dev/func/pkg/progress	[no test files]
?   	knative.dev/func/pkg/random	[no test files]
?   	knative.dev/func/pkg/ssh	[no test files]
?   	knative.dev/func/pkg/testing	[no test files]
?   	knative.dev/func/plugin	[no test files]
?   	knative.dev/func/schema/generator	[no test files]
[...]

PR #1803, opened 2023-06-08
Check error before defer body close, include timeout in error message

aslom:
Fixes https://github.com/knative/func/issues/1801

### DIFF
--- a/pkg/functions/runner.go
+++ b/pkg/functions/runner.go
@@ -164,16 +164,18 @@ func waitFor(job *Job, timeout time.Duration) error {
 	for {
 		select {
 		case <-time.After(timeout):
-			return errors.New("timed out waiting for function to be ready")
+			timeoutErrMsg := fmt.Sprintf("timed out waiting for function to be ready for %s", timeout)
+			return errors.New(timeoutErrMsg)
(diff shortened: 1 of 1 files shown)


pkg/functions/runner.go
```diff
@@ -164,16 +164,18 @@ func waitFor(job *Job, timeout time.Duration) error {
 	for {
 		select {
 		case <-time.After(timeout):
-			return errors.New("timed out waiting for function to be ready")
+			timeoutErrMsg := fmt.Sprintf("timed out waiting for function to be ready for %s", timeout)
+			return errors.New(timeoutErrMsg)
```
lance, line 168: Can you instead just return `ErrRunTimeout` here?

https://github.com/knative/func/blob/18fe6430a185cd21cc1b511720cae40b9eadbbb5/pkg/functions/errors.go#L17

aslom: ErrRunTimeout seems has no option to actually include the timeout as the message is hardcoded? How do you knwo what was the value of timeout parameter to waitFor - that may be very useful when debugging?

lkingland (approved):

lkingland (approved):

PR #1804, opened 2023-06-09
src: cleanup the run timeout loop and check

lkingland:
- :broom: cleans up the run timeout loop and error

Adds the timeout reached to a typed error on timeout, and factors out the liveness check to a function to avoid tail-stacking defer statements in the check loop.

### DIFF
--- a/pkg/functions/errors.go
+++ b/pkg/functions/errors.go
@@ -3,6 +3,7 @@ package functions
 import (
 	"errors"
 	"fmt"
+	"time"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


aslom:
I really like using verbose option. I tested and `make test` still works OK with macos.

/lgtm

matejvasek:
The tests are failing because dapr cannot be installed for some reason.

matejvasek:
> The tests are failing because dapr cannot be installed for some reason.

After I amended the PR the tests passed (including dapr installation). I don't know why.

PR #1805, opened 2023-06-09
src: add dapr version echo to cluster allocation output

lkingland:
- :broom: adds a `dapr version` output to cluster allocation
- :bug: temporarily disables Dapr installation due to installation error
- :broom: upgrades versions of all CI cluster dependencies

/kind cleanup


pkg/functions/client_int_test.go
```diff
@@ -399,6 +399,11 @@ func Handle(ctx context.Context, res http.ResponseWriter, req *http.Request) {
 // TestInvoke_ServiceToService ensures that a Function can invoke another
 // service via localhost service discovery api provided by the Dapr sidecar.
 func TestInvoke_ServiceToService(t *testing.T) {
+	// Temporarily disable tests which require Dapr
+	// See https://github.com/knative/func/issues/1806
+	t.Log("TestInvoke_ServiceToService temporarily disabled as it depends on Dapr runtme which is failing to install")
```
matejvasek, line 404: nit: The `t.Skip()` function is IMO more fitting.

matejvasek:
This probably happens because `docker.io` IP throttling when not logged in. I am trying to solve this with mirroring. The `dapr` CLI even has flags for this `--image-registry` too bad there is a bug in it. It applies the repository only to 4 of 5 deployments so one deployment still fails see [this](https://github.com/dapr/cli/issues/1311) issue.

Now I am trying to set mirroring in `KinD`'s `containerd` config.

matejvasek:
The dapr images are available at `ghcr.io/dapr`.

matejvasek:
https://github.com/knative/func/pull/1809

matejvasek:
btw it looks like the error spontaneously disappeared.

lkingland:
No longer necessary

PR #1807, opened 2023-06-09
Update springboot template pom.xml dependencies

trisberg:
- update to use spring-cloud.version 2022.0.3
- remove springboot.platform.version property, not needed, rely on parent version
- remove import scope and pom type from spring-boot dependencies

# Changes

- :gift: update to use spring-cloud.version 2022.0.3
- :broom: remove springboot.platform.version property, not needed, rely on parent version
- :broom: remove import scope and pom type from spring-boot dependencies

/kind cleanup

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


matejvasek:
This need rebase because of conflict in `generate/zz_filesystem_generated.go`. Also the `generate/zz_filesystem_generated.go` file contains superfluous files like maven wrapper jars and vscode config files.

matejvasek:
@ningyan325 PTAL

trisberg:
Rebased and re-generated `generate/zz_filesystem_generated.go`

Issue #1333, opened 2022-10-13
Better integration with Knative test and release infrastructure


lance:
## Background

The Functions project was transferred to Knative from a small open source project from Red Hat known as Boson (https://github.com/boson-project). At that time, it did not make use of any Knative infrastructure. Over the months, we have slowly moved to adopt more and more of the tooling. And with our move from knative-sandbox to knative organization, we did even more. Now that we are on the path to GA for the Functions project we should fill in the remaining gaps.

I'm using the following as reference material.

- Creating a sandbox repository document: https://github.com/knative/community/blob/main/mechanics/CREATING-A-SANDBOX-REPO.md
- Knative `hack` repository README.md: https://github.com/knative/hack#readme
- Knative `test-infra` guides: https://github.com/knative/test-infra/tree/main/guides

## Current status

- [x] `LICENCE`, `OWNERS`, `CODE-OF-CONDUCT` and `README.md` in place
- [x] Peribolos configured for `func` repository
- [x] Golang knative.dev import redirects
- [x] Prow / testgrid setup
- [x] Peribolos configured for committers, leads, etc

## Remaining needs

After reviewing (while writing up this issue) it seems like we are __really__ close.

- [ ] Add a `hack/presubmit-tests.sh` script that sets up our CI, including a cluster with Knative Serving and Eventing, an in-cluster registry, and an in-cluster git server. We can pull the non-standard bits like the registry and git server from our existing `hack/` directory.
- [ ] Improve our testgrid/prow configuration to include continuous tests (for each commit to `main`). The dashboard for the client looks like this: https://testgrid.k8s.io/r/knative-own-testgrid/client#continuous and the configuration for that looks like this: https://github.com/knative/test-infra/blob/main/prow/jobs_config/knative/client.yaml
- [ ] Audit our existing CI for pull requests and commits to `main` and adjust where needed. I think using presubmit-tests for e2e tests and running them with Knative test infra will greatly simplify our `.github/workflows/ci.yaml`.
- [ ] Integrate with existing Knative image publishing

/cc @upodroid @kvmware 

/kind

lkingland:
Closing as we are close enough with this integration to wait out the larger shift to using GitHub actions, where we have most of our work already

PR #1808, opened 2023-06-09
chore: use a custom presubmit_tests.sh script

lance:
The Knative test infrastructure runs presubmit tests which are defined in vendor/knative.dev/hack/presubmit-tests.sh. This has been running in test and release infrastructure. It is possible to customize the behavior of this script by defining a handful of functions. The function definitions provided here ensure that we are running tasks that are already being run with the default (e.g. verify-codgen.sh) but make use of our existing build and test targets.

(There also seems to be a stray version of this that I committed a couple of years ago. That has been removed as it is in the wrong location for test-infra and was never used anyway.)

Relates to: https://github.com/knative/func/issues/1333

EDIT: This PR focuses entirely on unit tests. A follow-up PR addressing integration tests will come next.

See: https://github.com/knative/infra/pull/115 (configures prow for func unit tests)
See: https://github.com/knative/infra/pull/104 (reverted, but this includes changes required for integration tests)

/kind chore

### DIFF
--- a/test/presubmit-tests.sh
+++ b/test/presubmit-tests.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2021 The Knative Authors
+# Copyright 2023 The Knative Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


hack/presubmit-tests.sh
```diff
@@ -0,0 +1,82 @@
...
+  header "Running integration tests"
+  make test-integration || failed=1
+
+  if (( failed )); then
+    results_banner "Integration tests failed"
+    exit ${failed}
+  fi
+}
+
+main "$@"
```
matejvasek, line 82: nit: missing newline

hack/presubmit-tests.sh
```diff
@@ -0,0 +1,82 @@
...
+  if (( failed )); then
+    results_banner "Template tests failed"
+    exit ${failed}
+  fi
+}
+
+function integration_tests() {
+  local failed=0
+  header "Running integration tests"
+  make test-integration || failed=1
```
matejvasek, line 74: Where will this run? Integration tests requires some set up (e.g. docker daemon, cluster, registry).

matejvasek:
Where is `presubmit-tests.sh` described? I don't see it in client/serving/eventing.

lance:
Not sure why codecov has failed - I have not reduced the total tests. I will look into it, but I think maybe we shouldn't let that hold this up.

lance:
> Not sure why codecov has failed - I have not reduced the total tests. I will look into it, but I think maybe we shouldn't let that hold this up.

Rebased and force pushed to trigger actions/jobs. We'll see if that changes codecov reporting.

lkingland:
The code coverage test doesn't seem to behave deterministically, and we appear to add an exception to omit style-checking the `vendor` directory
/lgtm

lkingland (approved):

Issue #1806, opened 2023-06-09
Dapr installation times out in CI


lkingland:
The installation of the Dapr runtime is failing in GitHub actions, just after "pulling images from docker hub" is printed.

matejvasek:
@lkingland is this because registry outage or some IP throttling?

matejvasek:
I guess this is know issue when GH Action cannot pull from `docker.io` because too many pulls.

matejvasek:
I don't know why but it seems dapr installation just started working on its own.

lkingland:
> I don't know why but it seems dapr installation just started working on its own.

That's good.  I figured it was a transient failure, probably throttling as you guessed as well.

We have some tech debt here:  we're dependent on both docker.io and ttl.sh for our CI.

PR #1809, opened 2023-06-09
src: use ghcr.io for dapr components in GH Action

matejvasek:
# Changes

- :broom:  Use ghcr.io for dapr components in GH Action.

resolves: #1806


lkingland (approved):

matejvasek:
@zroubalik @lance we need to override the linter error here.

PR #1810, opened 2023-06-11
src: more verbose output

matejvasek:
# Changes

- :broom:  Log cluster events and cluster containers logs in GH Action. This will be useful for debugging of failed actions.

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -43,6 +43,16 @@ jobs:
         run: ./hack/patch-hosts.sh
       - name: Integration Test
         run: make test-integration
+      - name: Dump Cluster Logs
+        if: always()
+        run: |
+          echo "::group::cluster events"
+          kubectl get events -A
+          echo "::endgroup::"
+          
+          echo "::group::cluster containers logs"
+          stern '.*' --all-namespaces --no-follow
+          echo "::endgroup::"
       - uses: codecov/codecov-action@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-integration.yaml
```diff
@@ -43,6 +43,16 @@ jobs:
         run: ./hack/patch-hosts.sh
       - name: Integration Test
         run: make test-integration
+      - name: Dump Cluster Logs
+        if: always()
```
matejvasek, line 47: This also could have been `failure()` but I think it shoudln't hurt having `always()`.

lkingland (approved):

matejvasek:
@zroubalik @lance we need to override the linter error here.

hack/binaries.sh
```diff
@@ -72,4 +73,11 @@ install_helm() {
     helm version
 }
 
+install_stern() {
+  echo 'Installing stern...'
+  curl -sSL "https://github.com/stern/stern/releases/download/v1.25.0/stern_1.25.0_linux_amd64.tar.gz" | \
```
lance, line 78: How about adding a `local stern_version=1.25.0` in `main()`?

lance (commented):
looks good - wdyt about making `stern_version` a varaible?

matejvasek:
@lance PTAL

matejvasek:
@lance the linter error has to be overridden.

PR #1815, opened 2023-06-15
adding support for .funcignore file

### LINKED ISSUE #1463: Add support for .funcignore
We were hoping to have a `.funcignore` file which would work much like `.gitignore` but for func-related tasks such as build (and especially `--remote` builds since this could be quite slow if the local directory contains a large amount of files!)

_Originally posted by @lkingland in https://github.com/knative/func/issues/1450#issuecomment-1334788575_

nitishchauhan0022:
# Changes
added the support for funcignore file for excluding the data which don't need to be added to image build

Fixes #1463 

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/pkg/builders/buildpacks/builder_test.go
+++ b/pkg/builders/buildpacks/builder_test.go
@@ -131,7 +131,7 @@ func TestBuild_BuilderImageExclude(t *testing.T) {
 	)
 	funcIgnoreContent := []byte(`#testing comments
 hello.txt`)
-	expected := []string{"#testing comments", "hello.txt"}
+	expected := []string{"hello.txt"}
 
 	tempdir := t.TempDir()
 	f.Root = tempdir
@@ -146,10 +146,7 @@ hello.txt`)
 		if len(opts.ProjectDescriptor.Build.Exclude) != 2 {
 			t.Fatalf("expected 2 lines of exclusions , got %v", len(opts.ProjectDescriptor.Build.Exclude))
 		}
-		if opts.ProjectDescriptor.Build.Exclude[0] != expected[0] {
-			t.Fatalf("expected a comment line %v, got '%v'", expected[0], opts.ProjectDescriptor.Build.Exclude[0])
-		}
-		if opts.ProjectDescriptor.Build.Exclude[1] != expected[1] {
+		if opts.ProjectDescriptor.Build.Exclude[1] != expected[0] {
 			t.Fatalf("expected excluded file to be '%v', got '%v'", expected[1], opts.ProjectDescriptor.Build.Exclude[1])
 		}
 		return nil
--- a/pkg/builders/buildpacks/builder_test.go
+++ b/pkg/builders/buildpacks/builder_test.go
@@ -2,7 +2,6 @@ package buildpacks
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -130,26 +129,28 @@ func TestBuild_BuilderImageExclude(t *testing.T) {
 			Runtime: "go",
 		}
 	)
-	funcIgnoreContent := []byte("hello.txt")
+	funcIgnoreContent := []byte(`#testing comments
+hello.txt`)
+	expected := []string{"#testing comments", "hello.txt"}
+
+	tempdir := t.TempDir()
+	f.Root = tempdir
+
 	//create a .funcignore file containing the details of the files to be ignored
 	err := os.WriteFile(filepath.Join(f.Root, ".funcignore"), funcIgnoreContent, 0644)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	// clean up after the test
-	defer func() {
-		err = os.Remove(filepath.Join(f.Root, ".funcignore"))
-		if err != nil {
-			t.Fatal(err)
-		}
-	}()
-
 	i.BuildFn = func(ctx context.Context, opts pack.BuildOptions) error {
-		expected := "hello.txt"
-		if opts.ProjectDescriptor.Build.Exclude[0] != expected {
-			t.Fatalf("expected excluded file to be '%v', got '%v'", expected, opts.ProjectDescriptor.Build.Exclude[0])
-			fmt.Println(opts.ProjectDescriptor.Build.Exclude[0])
+		if len(opts.ProjectDescriptor.Build.Exclude) != 2 {
+			t.Fatalf("expected 2 lines of exclusions , got %v", len(opts.ProjectDescriptor.Build.Exclude))
+		}
+		if opts.ProjectDescriptor.Build.Exclude[0] != expected[0] {
+			t.Fatalf("expected a comment line %v, got '%v'", expected[0], opts.ProjectDescriptor.Build.Exclude[0])
+		}
+		if opts.ProjectDescriptor.Build.Exclude[1] != expected[1] {
+			t.Fatalf("expected excluded file to be '%v', got '%v'", expected[1], opts.ProjectDescriptor.Build.Exclude[1])
 		}
 		return nil
 	}
--- a/pkg/builders/buildpacks/builder_test.go
+++ b/pkg/builders/buildpacks/builder_test.go
@@ -2,6 +2,9 @@ package buildpacks
 
 import (
 	"context"
+	"fmt"
+	"os"
+	"path/filepath"
 	"reflect"
 	"testing"
 
@@ -107,6 +110,46 @@ func TestBuild_BuilderImageConfigurable(t *testing.T) {
 		expected := "example.com/user/builder-image"
 		if opts.Builder != expected {
 			t.Fatalf("expected builder image for node to be '%v', got '%v'", expected, opts.Builder)
+		}
+		return nil
+	}
+
+	if err := b.Build(context.Background(), f, nil); err != nil {
+		t.Fatal(err)
+	}
+}
+
+// TestBuild_BuilderImageExclude ensures that ignored files are not added to the func
+// image
+func TestBuild_BuilderImageExclude(t *testing.T) {
+	var (
+		i = &mockImpl{} // mock underlying implementation
+		b = NewBuilder( // Func Builder logic
+			WithName(builders.Pack), WithImpl(i))
+		f = fn.Function{
+			Runtime: "go",
+		}
+	)
+	funcIgnoreContent := []byte("hello.txt")
+	//create a .funcignore file containing the details of the files to be ignored
+	err := os.WriteFile(filepath.Join(f.Root, ".funcignore"), funcIgnoreContent, 0644)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// clean up after the test
+	defer func() {
+		err = os.Remove(filepath.Join(f.Root, ".funcignore"))
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	i.BuildFn = func(ctx context.Context, opts pack.BuildOptions) error {
+		expected := "hello.txt"
+		if opts.ProjectDescriptor.Build.Exclude[0] != expected {
+			t.Fatalf("expected excluded file to be '%v', got '%v'", expected, opts.ProjectDescriptor.Build.Exclude[0])
+			fmt.Println(opts.ProjectDescriptor.Build.Exclude[0])
 		}
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


pkg/builders/s2i/builder.go
```diff
@@ -151,6 +151,29 @@ func (b *Builder) Build(ctx context.Context, f fn.Function, platforms []fn.Platf
...
 
+	funcignorePath := filepath.Join(f.Root, ".funcignore")
+	if _, err := os.Stat(funcignorePath); err == nil {
+		funcignore, err := os.Open(filepath.Join(f.Root, ".funcignore"))
+		if err != nil {
+			return fmt.Errorf("error opening funcignore file: %w", err)
+		}
+		defer funcignore.Close()
+
+		s2iignore, err := os.Create(filepath.Join(f.Root, ".s2iignore"))
```
lkingland, line 162: If the process panics, or otherwise fails to execute the `defer` statement's removal of `.s2iignore`, it may be left on disk after the process exits.

We may want this to create a symbolic link to `.funcignore` if it doesn't exist (and if S2I will still respect it), or use `Open` when creating the file with the "truncate" option so that it does't get an error that the file already exists.

lkingland (approved):
Looks good to me so far

/hold for additional 👀

matejvasek:
Some test would be nice.

nitishchauhan0022:
> Some test would be nice.

Thanks @matejvasek for the suggestion, it would be good addition, i can do that.

matejvasek:
> > Some test would be nice.
> 
> Thanks @matejvasek for the suggestion, it would be good addition, i can do that.

It does not have to be an integration test that would build an image.
* For `s2i` Just mock docker client: check that `DockerClient::ImageBuild()` is invoked with tar stream that does not contain undesired files.
* For `pack` chack that `Impl::Build()` is called with correct `pack.BuildOptions`.

pkg/builders/s2i/builder.go
nitishchauhan0022: @lkingland I have updated it with symbolic link, PTAL

nitishchauhan0022:
@matejvasek PTAL

pkg/builders/s2i/builder_test.go
```diff
@@ -150,6 +150,67 @@ func Test_BuilderImageConfigurable(t *testing.T) {
 	}
 }
 
+// Test_BuildImageWithFuncIgnore ensures that ignored files are not added to the func
+// image
+func Test_BuildImageWithFuncIgnore(t *testing.T) {
+
+	funcIgnoreContent := []byte("hello.txt")
```
matejvasek, line 157: I would also try comments:
```go
funcIgnoreContent := []byte(`#files that won't be in final image
hello.txt
`)
```

pkg/builders/s2i/builder_test.go
```diff
@@ -150,6 +150,67 @@ func Test_BuilderImageConfigurable(t *testing.T) {
...
+	f := fn.Function{
+		Runtime: "node",
+		Root:    ".",
```
matejvasek, line 160: You can use something like `filepath.Join(t.TempDir(), "fn")` as `Root`. Then directory removal will done automatically no need for defers.

pkg/builders/buildpacks/builder_test.go
```diff
@@ -116,6 +119,46 @@ func TestBuild_BuilderImageConfigurable(t *testing.T) {
...
 
+// TestBuild_BuilderImageExclude ensures that ignored files are not added to the func
+// image
+func TestBuild_BuilderImageExclude(t *testing.T) {
+	var (
+		i = &mockImpl{} // mock underlying implementation
+		b = NewBuilder( // Func Builder logic
+			WithName(builders.Pack), WithImpl(i))
+		f = fn.Function{
+			Runtime: "go",
```
matejvasek, line 130: You can use something like filepath.Join(t.TempDir(), "fn") as `Root`. Then directory removal will done automatically no need for defers.

pkg/builders/buildpacks/builder_test.go
```diff
@@ -116,6 +119,46 @@ func TestBuild_BuilderImageConfigurable(t *testing.T) {
...
+		}
+	)
+	funcIgnoreContent := []byte("hello.txt")
```
matejvasek, line 133: I would also try comments:
```go
funcIgnoreContent := []byte(`#files that won't be in final image
hello.txt
`)
```

pkg/builders/buildpacks/builder_test.go
```diff
@@ -116,6 +119,46 @@ func TestBuild_BuilderImageConfigurable(t *testing.T) {
...
+	// clean up after the test
+	defer func() {
+		err = os.Remove(filepath.Join(f.Root, ".funcignore"))
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	i.BuildFn = func(ctx context.Context, opts pack.BuildOptions) error {
+		expected := "hello.txt"
```
matejvasek, line 149: Please check size of the slice first.

nitishchauhan0022:
@matejvasek I have made the changes, PTAL

pkg/builders/buildpacks/builder_test.go
```diff
@@ -116,6 +118,48 @@ func TestBuild_BuilderImageConfigurable(t *testing.T) {
...
+		i = &mockImpl{} // mock underlying implementation
+		b = NewBuilder( // Func Builder logic
+			WithName(builders.Pack), WithImpl(i))
+		f = fn.Function{
+			Runtime: "go",
+		}
+	)
+	funcIgnoreContent := []byte(`#testing comments
+hello.txt`)
+	expected := []string{"#testing comments", "hello.txt"}
```
matejvasek, line 134: The comment is not expected to be there.

pkg/builders/s2i/builder_test.go
```diff
@@ -150,6 +150,59 @@ func Test_BuilderImageConfigurable(t *testing.T) {
...
+		Runtime: "node",
+	}
+	tempdir := t.TempDir()
+	f.Root = tempdir
+	//create a .funcignore file containing the details of the files to be ignored
+	err := os.WriteFile(filepath.Join(f.Root, ".funcignore"), funcIgnoreContent, 0644)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
matejvasek, line 170: It might be also interesting to create file named `#testing comments` and see if it is ignored (it should not).

pkg/builders/buildpacks/builder_test.go
matejvasek: This should be just `expected := []string{"hello.txt"}`

nitishchauhan0022:
@matejvasek PTAL

pkg/builders/s2i/builder.go
```diff
@@ -151,6 +151,26 @@ func (b *Builder) Build(ctx context.Context, f fn.Function, platforms []fn.Platf
 	}
 	defer os.RemoveAll(tmp)
 
+	funcignorePath := filepath.Join(f.Root, ".funcignore")
+	if _, err := os.Stat(funcignorePath); err == nil {
+		s2iignorePath := filepath.Join(f.Root, ".s2iignore")
+
+		// If the .s2iignore file exists, remove it
+		if _, err := os.Stat(s2iignorePath); err == nil {
+			err := os.Remove(s2iignorePath)
```
lance, line 160: This seems destructive to me. It's possible that the `.s2iignore` file contains entries that are not in `.funcignore`. Could this rather be renamed to something like `s2iignore.original` or similar?

pkg/functions/client.go
```diff
@@ -1071,6 +1076,41 @@ func ensureRunDataDir(root string) error {
...
+	// Open the file for writing only
+	file, err := os.OpenFile(filePath, os.O_WRONLY|os.O_CREATE, 0644)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	// Write the desired string to the file
+	_, err = file.WriteString(`
+# Function use the .funcignore file to exclude data which should not
```
lance, line 1103: Nit
```suggestion
# Use the .funcignore file to exclude files which should not
```

lance (commented):
Overall, this definitely lgtm. Thanks for the contribution! I have a couple of minor comments, inline. I like that we are recognizing `.s2iignore` but don't think we should fully remove it. I also think we should consider looking for a `project.toml` file for buildpack based projects, since that is the standard way to exclude files using the buildpack specification. https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md#iobuildpacksinclude-optional-and-iobuildpacksexclude-optional

pkg/builders/buildpacks/builder_test.go
```diff
@@ -116,6 +118,45 @@ func TestBuild_BuilderImageConfigurable(t *testing.T) {
...
+	f.Root = tempdir
+
+	//create a .funcignore file containing the details of the files to be ignored
+	err := os.WriteFile(filepath.Join(f.Root, ".funcignore"), funcIgnoreContent, 0644)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	i.BuildFn = func(ctx context.Context, opts pack.BuildOptions) error {
+		if len(opts.ProjectDescriptor.Build.Exclude) != 2 {
```
matejvasek, line 146: Wait, why is this 2 not 1? The comment should be ignored, right?

matejvasek: btw good way to diff two arrays in test is `cmp.Diff()`

pkg/builders/s2i/builder.go
nitishchauhan0022: It would be case if the user enter the details in .s2iignore by first creating it but as we are adding the option to add the files in .funcignore, they should add there only which will be symbolically linked as .s2iignore to provide a standard pattern for all buiilders.

pkg/builders/buildpacks/builder_test.go
nitishchauhan0022: > Wait, why is this 2 not 1? The comment should be ignored, right?

because we are not excluding the comment lines ourselves, we are passing all the lines directly to the exclude option of buildpack which is doing this using https://github.com/sabhiram/go-gitignore/blob/master/ignore.go. So, it will a duplication of code if we does that here as well.

nitishchauhan0022: > btw good way to diff two arrays in test is `cmp.Diff()`

thanks for the info, will keep that in mind while comparing arrays.

pkg/builders/s2i/builder.go
nitishchauhan0022: > Overall, this definitely lgtm. Thanks for the contribution! I have a couple of minor comments, inline. I like that we are recognizing `.s2iignore` but don't think we should fully remove it. I also think we should consider looking for a `project.toml` file for buildpack based projects, since that is the standard way to exclude files using the buildpack specification. https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md#iobuildpacksinclude-optional-and-iobuildpacksexclude-optional

Thanks for lgtm, indeed it is the way to exclude the file when we don't have option to exclude the files through the buildpack build options `Build: types.Build{
				Exclude: excludes, },`

nitishchauhan0022:
@matejvasek why we are having this problem on windows with TempDir
--- FAIL: Test_BuildImageWithFuncIgnore (1.56s)
    testing.go:1206: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\Test_BuildImageWithFuncIgnore4081273292\001\#testingComments.txt: The process cannot access the file because it is being used by another process.

matejvasek:
> @matejvasek why we are having this problem on windows with TempDir --- FAIL: Test_BuildImageWithFuncIgnore (1.56s) testing.go:1206: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\Test_BuildImageWithFuncIgnore4081273292\001#testingComments.txt: The process cannot access the file because it is being used by another process.

@nitishchauhan0022 you need to close the files you have created via `os.Create()`.

pkg/builders/s2i/builder_test.go
```diff
@@ -150,6 +150,69 @@ func Test_BuilderImageConfigurable(t *testing.T) {
...
+	// create a test file that should be ignored
+	_, err = os.Create(filepath.Join(f.Root, "hello.txt"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = os.Create(filepath.Join(f.Root, "#testingComments.txt"))
+	if err != nil {
+		t.Fatal(err)
+	}
```
matejvasek, line 180: Close these files @nitishchauhan0022  (or create them via `os.WriteFile()`).

matejvasek: `os.WriteFile()` creates,writes and closes file.

PR #1819, opened 2023-06-19
chore: test cluster versions update

lkingland:
Updates versions used for the test clutser, and includes a change to use a specific kind node version rather than the kubernetes version per Kind's docs.

- knative_serving v1.10.1
- knative_eventing v1.10.1
- contour v1.10.0
- kubectl v1.27.2
- dapr v1.10.0
- helm v3.12.0
- kind v0.19.0
- kind_node v1.27.1

/kind cleanup

### DIFF
--- a/hack/allocate.sh
+++ b/hack/allocate.sh
@@ -24,10 +24,13 @@ export TERM="${TERM:-dumb}"
 
 main() {
 
-  local kubernetes_version=v1.24.6
-  local knative_serving_version=v1.8.0
-  local knative_eventing_version=v1.8.0
-  local contour_version=v1.8.0
+  local knative_serving_version=v1.10.1
+  local knative_eventing_version=v1.10.1
+  local contour_version=v1.10.0
+
+  # Kubernetes Version node image per Kind releases (full hash is suggested):
+  # https://github.com/kubernetes-sigs/kind/releases
+  local kind_node_version=v1.27.1@sha256:b7d12ed662b873bd8510879c1846e87c7e676a79fefc93e17b2a52989d3ff42b
 
   # shellcheck disable=SC2155
   local em=$(tput bold)$(tput setaf 2)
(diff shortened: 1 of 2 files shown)


hack/allocate.sh
```diff
@@ -298,6 +301,7 @@ EOF
 
 dapr_runtime() {
```
matejvasek, line 302: Has the [dashboard](https://github.com/knative/func/pull/1819/files?diff=split&w=1#diff-3900b46a5058b65369632a94a58bad6cda0e7e9e8bf991c30e20a32d13a0a527L288) image changed between versions?
If so the `fixup_dapr_dashboard_deployment()` function should be updated too.

matejvasek: Now I am trying installing dapr on k8s and it is not setting "dashboard" at all, weird.

matejvasek: They probably removed dashboard from default installation via helm hart maybe? :man_shrugging:

lkingland: OK, thanks for checking.

hack/allocate.sh
```diff
@@ -271,43 +274,13 @@ EOF
...
-EOF
-}
-
 dapr_runtime() {
   echo "${em}⑦ Dapr${me}"
-
-  local dapr_flags=""
-  if [ "${GITHUB_ACTIONS:-false}" = "true" ]; then
-    (sleep 10; fixup_dapr_dashboard_deployment)&
-    dapr_flags="--image-registry=ghcr.io/dapr --log-as-json"
```
matejvasek, line 305: I think using the `ghcr.io` registry and `--log-as-json` is still useful.

matejvasek: at least in GH Action

lkingland: Sure, that part is added back.  Just kept the removal of line 304; the invocation of the dapr dashboard patch.

zroubalik (approved):
/lgtm
/approve

matejvasek:
@zroubalik please override the linter error.

PR #1820, opened 2023-06-21
src: speed up tests by building with a test platform

lkingland:
- :broom: build tests for single platform

Utilize the platforms option to build test for only one platform when multi-arch is unnecessary.  Speeds up tests.

/kind cleanup


zroubalik (approved):
/lgtm

PR #1821, opened 2023-06-21
src: refactor builder concurrency test

lkingland:
- :broom: refactor build concurrency test

Expands the builder to include an optional notification channel such that temporary test tooling can be removed entirely from the builder core and logic kept in the test only.

/kind cleanup


zroubalik (approved):
/lgtm

PR #1823, opened 2023-06-21
feat: scaffolding ca certs

lkingland:
- :gift: Scaffolded functions include root CA certs
- :gift: adds make target for updating embedded certs

All functions which utilize scaffolding (Go for now, but will work for all) will now include root CA certs, as provided by the curl project.


generate/templates/main.go
```diff
@@ -11,6 +11,8 @@ import (
 	"path/filepath"
 )
 
+// MODIFIED
```
matejvasek, line 14: ?

Makefile
```diff
@@ -86,137 +127,179 @@ clean_templates:
 	@rm -rf templates/springboot/http/target
 	@rm -f templates/**/.DS_Store
 
-.PHONY: generate/zz_filesystem_generated.go
 
-generate/zz_filesystem_generated.go: clean_templates
+# the filesysetm file itself is defined as being fresh if all the files in
+# templates are fresh as well as the certificates file.
+generate/zz_filesystem_generated.go: $(TEMPLATES) templates/certs/ca-certificates.crt
```
matejvasek, line 133: I very much on purpose remove similar condition in the past. This can detect that file was modified but it won't for instance detect file deletion.

matejvasek: This should be phony.

matejvasek: The `generate/zz_filesystem_generated.go` recipe takes 200ms. Most of which is actually spend on `go run`'s on the fly compilation. The actual template generation takes 40ms. I don't see why we cannot run it all the time.

pkg/filesystem/filesystem_test.go
```diff
@@ -117,6 +117,11 @@ func loadFS(fileSys filesystem.Filesystem) ([]FileInfo, error) {
 				return err
 			}
 		case fs.ModeSymlink:
+			if runtime.GOOS == "windows" {
+				// TODO I have no idea why, but symlinks are ever so slightly different
+				// in Windows and fail the test.
+				return nil
```
matejvasek, line 123: Please do not just ignore it. Use `filepath.ToSlash()` to fixup the value.

pkg/filesystem/filesystem_test.go
```diff
@@ -157,6 +162,11 @@ func loadLocalFiles(root string) ([]FileInfo, error) {
...
```
matejvasek, line 168: Please do not just ignore it. Use `filepath.ToSlash()` to fixup the value.

pkg/filesystem/filesystem_test.go
matejvasek: Actually at this place it will be always slash separated since `fs.FS` uses just slashes.

matejvasek: Just remove the `if` altogether.

pkg/filesystem/filesystem_test.go
```diff
@@ -117,6 +117,11 @@ func loadFS(fileSys filesystem.Filesystem) ([]FileInfo, error) {
...
+			}
```
matejvasek, line 124:
```suggestion
```

pkg/filesystem/filesystem_test.go
```diff
@@ -157,6 +162,11 @@ func loadLocalFiles(root string) ([]FileInfo, error) {
...
 			t, _ := os.Readlink(path)
 			bs = []byte(t)
```
matejvasek, line 171:
```suggestion
			t, _ := os.Readlink(path)
			bs = []byte(filepath.ToSlash(t))
```

Makefile
```diff
@@ -86,137 +127,179 @@ clean_templates:
...
+.PHONY: certs
+certs: templates/certs/ca-certificates.crt ## Ensure the certs exist
+
+templates/certs/ca-certificates.crt:
+	# Updating root certificates
+	curl --output templates/certs/ca-certificates.crt https://curl.se/ca/cacert.pem
+
+.PHONY: certs-check
+certs-check: certs ## Check that root certificates are up-to-date
+	@if [[ -n $$(git diff -- schema/func_yaml-schema.json) ]]; \
```
matejvasek, line 204: How does this check certs?

Makefile
```diff
@@ -86,137 +127,179 @@ clean_templates:
...
-docs:
-	# Generating command reference doc
-	go run docs/generator/main.go
-
-#############
-##@ Templates
-#############
+.PHONY: templates-check
+templates-check: certs ## Check that the generated embedded filesystems are up-to-date
+	@if [[ -n $$(git diff -- schema/func_yaml-schema.json) ]]; \
```
matejvasek, line 139: How does this check `generate/zz_filesystem_generated.go`?

matejvasek:
On similar note like https://github.com/knative/func/pull/1823#discussion_r1238839860.
I really do not like using `$(CODE)` as prerequisites -- it cannot detect file deletion (only addtition/modification).
Moreover `go build`, `go test`, `golangci-lint` are smart enough to know when and what re-compilation/re-test is required.
Similarly I don't see much value in these stamps, what I see is possible risk of not running recipe when is should be run.
I rather run recipe that takes 500ms all the time  than risking that recipe is not done when it really needed.

Makefile
lkingland: Yes that comment is confusing.  I removed this one and marked the templates target as .PHONY such that it always builds, the other two instances (certs and schema) were renamed to `x-modified-check` and removed the comment such that they are "hidden" from `make help`.  These are used in source control along with a `make clean` to confirm that the files are up-to-date.

Makefile
lkingland: Renamed this `certs-modified-check` and added better comments 👍🏻

matejvasek: I was getting on `schema/func_yaml-schema.json`. Why it is there?

Makefile
matejvasek: I was getting on `schema/func_yaml-schema.json`. Why it is there?

Makefile
lkingland: I have removed all but the minimal makefile updates and will submit them separately for discussion.

Makefile
lkingland: I have removed all but the minimal makefile updates and will submit them separately for discussion.

Makefile
lkingland: I have removed all but the minimal makefile updates and will submit them as a separate PR for discussion

lkingland:
Thanks for the review @matejvasek , I have added `filepath.ToSlash()` to the filesystem symlink test, and removed all but the minimum makefile updates for discussion in a separate PR.

matejvasek:
The windows test are failing on the symlink despite `filepath.ToSlash()`, it makes no sense.

lkingland:
> The windows test are failing on the symlink despite `filepath.ToSlash()`, it makes no sense.

Shucks.  I was hoping the `ToSlash` would do the trick.  I really don't mind if they are a few bytes different, so long as they are both symlinks.

matejvasek:
> > The windows test are failing on the symlink despite `filepath.ToSlash()`, it makes no sense.
> 
> Shucks. I was hoping the `ToSlash` would do the trick. I really don't mind if they are a few bytes different, so long as they are both symlinks.

@lkingland I think I know where the issue might be. The unit test in GH also regenerates zz file on Windows. The zip generate script should also fixup links.

matejvasek:
>actual app do use zz file generated on Linux 

Actually I am not 100% positive about this. I am only sure that we version file generated on Linux.

matejvasek:
@lkingland  tl;dr we should not suppress the tests, it really indicates that something is wrong

We need `filepath.ToSlash()` in:
* [zip generator](https://github.com/knative/func/blob/main/generate/templates/main.go#L83C4-L83C43)
* [os fs](https://github.com/knative/func/blob/main/pkg/filesystem/filesystem.go#L176)
* [fs test](https://github.com/knative/func/blob/main/pkg/filesystem/filesystem_test.go#L161)

I know this is bug in existing code, not it this PR, sill we should fix it.

PR #1828, opened 2023-06-26
src: make test and check go

lkingland:
Creates a `make check-go` target which lints Go scaffolding.

/kind enhancement


lkingland:
Thanks @matejvasek ; looks like the tests cleared this time

PR #1829, opened 2023-06-26
Cluster integration tests cleanup

matejvasek:
# Changes

- :broom: General improvements for `TestUploadToVolume()` and `TestDialInClusterService()`:
   * Better allocation and cleanup of testing resources. All resources names are randomized and the resources are deleted after test.
   * `TestUploadToVolume()` now uses `Deployment` instead of plain `Pod` for the testing service.
   * `TestUploadToVolume()` and `TestDialInClusterService()` are not running in new namespace, instead they are run in current default namespace. Rationale: not all cluster users can create a namespace, this change ensures the tests can be run with non-admin cluster users.
   
- :broom: Refactor `NewInClusterDialer()` now takes `clientcmd.ClientConfig` as parameter, this removed deplendency on global `kubectl` configuration allowing test parallelization.


lkingland (approved):
Nice improvements;
/lgtm

PR #1833, opened 2023-06-27
chore: add presubmit-unit-tests Makefile target

lance:
This commit adds a Makefile target that will simulate the Prow presubmit tests that are run for pull requests on GitHub. It makes use of the same image used by Prow, and runs the ./test/presubmit-tests.sh script so that developers can run the same suite of tests that are run in CI/CD.

/kind chore

Related: https://github.com/knative/func/issues/1333


lkingland (approved):
👍🏻

PR #1837, opened 2023-06-28
feat: default ADDRESS env var on deploy

lkingland:
- :gift: Requests that deployed functions listen on all interfaces for requests.

Runtimes should, by default, only listen on the loopback interface for security (they may be `func run` locally).

The ADDRESS environment variable added here on deploy requests the runtime to listen on all interfaces (0.0.0.0) when deployed, since they will need to listen for client requests and for health readiness/liveness probes.

Additionally, now a user who wishes to securely open their function to only receive requests on a specific interface, such as a WireGuard-encrypted mesh network which presents as a specific interface; that can be achieved by explicitly setting the ADDRESS on their function.

NOTE this env variable is only a suggestion, and it is up to the runtimes to read and use.  This is currently respected by scaffolded Go functions (`func-runtime-go` library), with other runtimes expected to implement as they are updated to support scaffolding.

/kind enhancement


zroubalik (approved):
/lgtm

/hold for others

lance (commented):
Overall lgtm, although I wonder if `ADDRESS` might be too generic and potentially suffer from overlap. Just a thought.

PR #1843, opened 2023-07-04
feat: use Tekton Resolvers for a standard on cluster build

zroubalik:
# Changes

- :gift: Standard Tekton On-cluster builds (non PAC) are using resolvers to get referenced tasks - users don't have to install them
- :broom: PAC related templates are stored in `.tekton` directory in files `pipeline-pac.yaml` and `pipeline-run-pac.yaml` (added the `-pac` suffix to distingush them from standard ones
- :bug: PAC templates contain knative/func related labels

I have changed the way how Pipeline and PipelineRun resources are created - not using go client anymore, but using [Manifestival](https://github.com/manifestival/manifestival) to apply YAML resources - this simplifies templating and follows the pattern used for PAC builds.
Another benefit is, that users don't have preinstall Tasks anymore, they should be resolved by Tekton.

The workflow is following: 
1. user runs `func deploy --remote`
2. if there are files `pipeline.yaml` and/or `pipeline-run.yaml` in the `.tekton` directory in the function project, we use the resources defined there to create the Pipeline and/or PipelineRun
3. if these files are not present there, new resources are created on the fly and applied to the cluster

Ability to save the pipelines to the `.tekton` directory is not implemented yet, it will be in the follow up PR.

Relates https://github.com/knative/func/issues/620


lance (commented):
Looks good @zroubalik! I'd like to understand how we might use this in the openshift dev console as well. It's great that the tasks don't need to be explicitly installed now which will help scenarios like developer sandbox (where users don't have permission to install tasks). Currently dev sandbox does explicitly need to install both the tasks, AND the pipeline (here: https://github.com/openshift-knative/kn-plugin-func/blob/serverless-1.29.0/pkg/pipelines/resources/tekton/pipeline/dev-console/0.1/nodejs-pipeline.yaml). It would be great to eliminate the need for all of that to be installed, which it seems this PR does. We need to communicate with the dev console folks to ensure that we're all on the same page though. Thanks!

zroubalik:
> Looks good @zroubalik! I'd like to understand how we might use this in the openshift dev console as well. It's great that the tasks don't need to be explicitly installed now which will help scenarios like developer sandbox (where users don't have permission to install tasks). Currently dev sandbox does explicitly need to install both the tasks, AND the pipeline (here: https://github.com/openshift-knative/kn-plugin-func/blob/serverless-1.29.0/pkg/pipelines/resources/tekton/pipeline/dev-console/0.1/nodejs-pipeline.yaml). It would be great to eliminate the need for all of that to be installed, which it seems this PR does. We need to communicate with the dev console folks to ensure that we're all on the same page though. Thanks!

Yeah, that's correct. You don't need to install Tasks,Pipeline etc. Does Dev Console solution run `func deploy`? If so, then everything should be set.


The only thing I don't like about this solution is that Tasks needs to be fetched from git (and checking out https://github.com/knative/func.git can take some time). I think that we can workaround this by moving tasks to a separate repo (thus smaller) or to Tekton Hub, which I prefer.

lance:
LOL I was trying to override everything, thinking prow was still balking on stuff - but the problem was simply that it didn't have the lgtm

matejvasek:
@zroubalik does this work on Red Hat Sandbox?

PR #1817, opened 2023-06-19
src: improved version output

lkingland:
- :gift: Knative version now included in func verbose output
- :bug: fixes bug where builds without an associated version number showed `v0.0.0`
- :broom: validates version passed by build is a valid semver

Ensures that `func version` remains consistent; always returning an ascending semver of the version, with the associated knative tag being part of the verbose version information.

Replaces the default v0.0.0 version tag when building from a non-tagged commit to the "human readable" form provided by "git describe".  This consists of the most recently tagged version along with a suffix which consists of the number of commits which have been added since the tagged version, and the hash.

The reasoning behind this change is that we must retain the ability to release versions independently of the knative releases, while at the same time communicating which knative release to for which this version is associated.  The primary version number is the version of this library, and of the binary itself.  At time of this writing, that is `v0.37.0`, and is the most contectually relevant.  The knative version to which this library and cli tool relates most closely is `v1.10`, and is included as metadata, and printed in verbose mode.

Here's how this will look in practice:

The most recent Knative version is v1.10, which was released with Functions v0.37.0.  Checking out v0.37.0 correctly displays the versions:
```shell
❯ git checkout v0.37.0 && make
❯ ./func version -v
Version: v0.37.0
Knative: v1.10.0
Commit: 61179d88
SocatImage: quay.io/boson/alpine-socat:1.7.4.3-r1-non-root
TarImage: quay.io/boson/alpine-socat:1.7.4.3-r1-non-root
```

When building from an unreleased commit, such as the current Function main branch (which is 37 commits ahead of what was released as v0.37.0 in Knative 1.10:
```shell
❯ git checkout main && make
❯ ./func version -v
Version: v0.37.0-77-gcabba3f9
Knative: v1.10.0
Commit: cabba3f9
SocatImage: quay.io/boson/alpine-socat:1.7.4.3-r1-non-root
TarImage: quay.io/boson/alpine-socat:1.7.4.3-r1-non-root
```

Please note that this only applies to builds triggered with `make build`.  Direct builds of the source code do not populate these values:
```shell
❯ go build ./cmd/func
❯ ./func version -v
Version:
Knative:
Commit:
SocatImage: quay.io/boson/alpine-socat:1.7.4.3-r1-non-root
TarImage: quay.io/boson/alpine-socat:1.7.4.3-r1-non-root

PR #1850, opened 2023-07-05
[release-1.10]: src: interstitial commit version (#1817)

lance:
This is a backport of #1817 to the 1.10 branch.

The Knative version is now included in version command verbose output

Building an unreleased version no longer returns v0.0.0, but instead the value provided by `git describe --tags` which is the most recent tagged release with a suffix consisting of the number of commits since that release and the short hash.

Verbose output now always includes the current commit on a dedicated line.

```release-note
Improved `func --version` output
```


zroubalik (approved):
/lgtm

PR #1856, opened 2023-07-11
chore: update tekton client library

matejvasek:
# Changes

- :broom: Update tekton client library. This prevents possible `send on close channel` panic.


lance (commented):
/lgtm

PR #1871, opened 2023-07-13
test: OCI content created by "host" builder

matejvasek:
# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


matejvasek:
PTAL @lkingland

pkg/oci/builder_test.go
```diff
@@ -172,8 +179,82 @@ func validateOCI(path string, t *testing.T) {
...
+		{Path: "/etc/pki/tls/certs/ca-certificates.crt"},
+		{Path: "/etc/ssl/certs/ca-certificates.crt"},
+		{Path: "/func", Type: fs.ModeDir},
+		{Path: "/func/README.md"},
+		{Path: "/func/f", Executable: true},
+		{Path: "/func/func.yaml"},
+		{Path: "/func/go.mod"},
+		{Path: "/func/handle.go"},
+		{Path: "/func/handle_test.go"},
+	}
```
matejvasek, line 255: @lkingland why are the sources in the image? Shouldn't be there only executable and certificates?

lkingland: The third layer is the `data` layer (for example, a repository which is a static HTTP site and a .go function which serves the files on request).  The goal here was to use a `.funcignore` pre-populated with `*.go` when that feature is available.  I will open a separate issue for that

lkingland (approved):
Thanks for tracking this down!

pkg/oci/builder_test.go
matejvasek: @lkingland I am afraid that if you add `*.go`s to `.funcignore` then the source will be also ignored by builders -- it won't compile application at all.

matejvasek:
@lance please override the error.

pkg/oci/builder_test.go
lkingland: It could, depending on how we interpret the .funcignore per builder

I [opened a discussion](https://github.com/knative/func/discussions/1876)

PR #1873, opened 2023-07-17
chore: bump faas-js-runtime and cloudevents for node & TS

lance:
# Changes

The Node.js and TypeScript templates have been bumped to the latest versions of faas-js-runtime (2.2.2) and CloudEvents (7.0.1).

There is an issue with the `expectType` test for TypeScript CloudEvent functions which I can't quite figure out, but the type test is not critical and does not do anything to validate other than whether the function signature is what's expected by faas-js-runtime, and in this case, I think type exported type for faas-js-runtime may actually be the issue here.

/kind chore

**Release Note**

```release-note
* bumps faas-js-runtime to version 2.2.2
* bumps cloudevents to version 7.0.1
```

/cc @matejvasek


matejvasek:
@lance it looks like there is a double definition of types.
There is `node_modules/faas-js-runtime/node_modules/cloudevents/dist/event/cloudevent` and `node_modules/cloudevents/dist/event/cloudevent`. They are not compatible that's why `expectType<>()` is not working I believe.

matejvasek:
@lance what `node_modules/faas-js-runtime/node_modules` even is? I don't know much about node but having node_modules in module seems fishy to me.

matejvasek:
When I delete `node_modules/faas-js-runtime/node_modules` the type-check does work.

matejvasek:
It is weird nonetheless, I thought that `TS` is structurally type (not nominally) so I would expect even double definition to work.

matejvasek:
> @lance it looks like there is a double definition of types. There is `node_modules/faas-js-runtime/node_modules/cloudevents/dist/event/cloudevent` and `node_modules/cloudevents/dist/event/cloudevent`. They are not compatible that's why `expectType<>()` is not working I believe. But it is weird since the definitions are identical.

@lholmquist any idea why this would be happening?

PR #1874, opened 2023-07-18
fix: host builder retry error

### LINKED ISSUE #1703: Scaffolding
This is the umbrella issue for implementation of Function Scaffolding as laid out in the [related discussion](https://github.com/knative/func/discussions/1589).

At a high level this encapsulates creating a function source-code scaffolding engine which supports host-based running, building and pushing without the need for podman or docker, and both instanced and static method signatures.

This is an initial implementation which focuses on the minimum necessary changes to enable an end-to-end experience using function scaffolding for Go functions.  There are several rough edges and missing features needed before this could be considered complete, and of course scaffolding needs to be created for each of the other runtimes.  The intent of this inital implementation is to serve as a a fast-track for a very focused subset of functions, and as a reference implementation for creating scaffolding for the other languages.

## Running Functions Locally

Running locally from _within_ a container remains the default for now, though this should be the default when scaffolding is implemented for the other languages.  To run a Go container locally outside of a container use `--container=false`
```
mkdir f && cd $_
func create -l go
func run --container=false
```
Languages other than Go will yield the error:
```
Error: The ability to run functions outside of a container via 'func run' is coming soon.
```

## Building Scaffolded Functions

Functions built using the --builder=host option (currently not the default) are scaffolded into the directory `.func/builds/$HASH`.  This option is available to both `run` and `deploy`.  This build step builds the OCI-compliant directory structure directly rather than using Podman or Docker.

## Deploying Scaffolded Functions

A function which indicates that its builder is/was "host" will use the inbuilt OCI pusher, which uses the container data built into the `.func/builds/$HASH` directory.

```
mkdir f && cd $_
func create -l go
func deploy --builder=host
```

## Route Selection

As part of this update, mostly with the intent of making a clean end-to-end demo, the `--domain` flag can be specified when deploying a function to indicate which of the configured domains of the cluster should be used.  The default is to use the cluster default.  This requires that the cluster be configured correctly otherwise the setting is ignored.  On the docket for our next discussion is how we can query the underlying platform for this and other settings which could be used informatively for the user.  For the best experience an entire domain would be delegated to the cluster and the namespace removed from the route creation, along with a ClusterIssuer for HTTPS:

```
$ mkdir www && cd $_
$ func create -l go
$ func deploy --builder=host --domain=boson-project.org
Function deployed at https://www.boson-project.org
$ curl https://www.boson-project.org
Hello, world!
```

## Instanced vs Static Method Signatures

In order to write the correct scaffolding, we need to know invocation (cloudevents or http) and language.  This adds a layer of logic inbetween the code and the scaffolding which is the correct place to do more advanced operations.  To illustrate this, the scaffolding code supports both static and "instanced" cloudevent and http functions.  Currently only available for Go with the "host" builder, a function author can now opt to implement a `func New() *MyFunction` constructor which will have its `.Handle` method invoked.  Also supported are overrides for lifecycle methods such as `.Start`, `.Stop`, `.Alive`, and `Ready`.  This is only possible when using an instance rather than a package-static function Handle. 

## Subtasks

- [x] Track Go Sumfiles and Test backward compatibility
        #1688
- [x] Support domain command for selecting route
        #1690
- [x] Update `run` command to support all build options
        (#1693)
- [x] Rename client library `Instances` to `InstaceRefs` to avoid confusion with func runtme library
        (#1694)
- [x] Update Jobs to support Scaffolded and non-scaffolded (local) runs
        (#1697)
- [x] Define and Detect Function Signatures
        (#1699)
- [x] Basic OCI builder with plugable language layer builders
        (#1730)
- [x] Scaffolding Code for both Static and Instanced Go Functions with Writer
        (#1732)
- [x] Container pusher which directly submits an OCI-compliant directory as a tarball to a remote registry
        (#1702)
- [x] localhost default runner which builds using the OCI builder on `func run --container=false`
        (#1733) 
- [X] integrate the scaffolding with the OCI builder and pusher
        (#1750)
- [X] enable scaffolded builder, runner and pusher into CLI (enable)
        (#1748) 
- [X] enable function external dependencies auto-import
        (#1790)
- [X] Multi-OS root SSL certs layer in resultant image
        (#1823)

### Should we prompt persistence of options for `func run`?

The `run` command does not persist configuration parameters to func.yaml because the way one runs the container locally, especially when outside of a container, should not affect how it is run in production (ie affect the source code).
This could very well lead to confusion on the part of the user.  We could leave this as a documentation issue, or we could implement an interactive Prompt to ask the user if they would like to save changes to key attributes such as "builder".

lkingland:
- :bug: error retrying failed host builder builds

The host builder was not properly starting a clean build on retry of a failed build.

This is also the last known bug for the [Scaffolding Subsystem](https://github.com/knative/func/discussions/1589) MVP, closing the Issue-epic #1703

/kind bug
/closes #1703

### DIFF
--- a/pkg/oci/builder.go
+++ b/pkg/oci/builder.go
@@ -70,7 +70,7 @@ func newBuildConfig(ctx context.Context, b *Builder, f fn.Function, platforms []
 func (b *Builder) Build(ctx context.Context, f fn.Function, pp []fn.Platform) (err error) {
 	cfg := newBuildConfig(ctx, b, f, pp)
 
-	if err = setup(cfg); err != nil { // create directories and links
+	if err = setup(cfg); err != nil {
 		return
 	}
 	defer teardown(cfg)
@@ -154,17 +154,20 @@ func (c *buildConfig) blobsDir() string {
 	return path(c.f.Root, fn.RunDataDir, "builds", "by-hash", c.hash(), "oci", "blobs", "sha256")
 }
 
-// setup errors if there already exists a build directory.  Otherwise, it
-// creates a build directory based on the function's hash, and creates
-// a link to this build directory for the current pid to denote the build
-// is in progress.
 func setup(cfg *buildConfig) (err error) {
-	// error if already in progress
 	if isActive(cfg, cfg.buildDir()) {
 		return ErrBuildInProgress{cfg.buildDir()}
 	}
 
-	// create build files directory
+	// create build directory, recreating if it already existed
+	if _, err = os.Stat(cfg.buildDir()); !os.IsNotExist(err) {
+		if cfg.verbose {
+			fmt.Printf("rm -rf %v\n", cfg.buildDir())
+		}
+		if err = os.RemoveAll(cfg.buildDir()); err != nil {
+			return
+		}
+	}
 	if cfg.verbose {
 		fmt.Printf("mkdir -p %v\n", cfg.buildDir())
 	}
--- a/pkg/oci/containerize.go
+++ b/pkg/oci/containerize.go
@@ -28,7 +28,7 @@ var languageLayerBuilders = map[string]languageLayerBuilder{
 }
 
 func layerBuilderNotImplemented(cfg *buildConfig, _ v1.Platform) (d v1.Descriptor, l v1.Layer, err error) {
-	err = fmt.Errorf("%v functions are not yet supported by the host builder.", cfg.f.Runtime)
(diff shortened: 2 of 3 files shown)


lkingland:
Thanks folks, tests are passing now. PTAL!
@matejvasek @lance

lkingland:
@lance this appears to be failing with an error that NPM is not found?  Maybe that's an easy fix if you're in there already

PR #1875, opened 2023-07-18
fix: remove path bacause of bug in paring logic

matejvasek:
# Changes

- :bug: Fix bug caused by  Docker's `ParseHostURL()` -- I believe that function is not working correctly [see](https://github.com/moby/moby/discussions/46015). The error starts showing up when using Go 1.20.6 which does some additional percent escaping of hostname.

### DIFF
--- a/pkg/docker/docker_client.go
+++ b/pkg/docker/docker_client.go
@@ -131,7 +131,7 @@ func NewClient(defaultHost string) (dockerClient client.CommonAPIClient, dockerH
 	dockerClient, err = client.NewClientWithOpts(
 		client.WithAPIVersionNegotiation(),
 		client.WithHTTPClient(httpClient),
-		client.WithHost("http://placeholder/"))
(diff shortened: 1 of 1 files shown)


pkg/docker/docker_client.go
```diff
@@ -131,7 +131,7 @@ func NewClient(defaultHost string) (dockerClient client.CommonAPIClient, dockerH
 	dockerClient, err = client.NewClientWithOpts(
 		client.WithAPIVersionNegotiation(),
 		client.WithHTTPClient(httpClient),
-		client.WithHost("http://placeholder/"))
+		client.WithHost("http://placeholder"))
```
matejvasek, line 134:
```suggestion
		client.WithHost("tcp://placeholder/"))
```

matejvasek:
PTAL @lance

lance:
/lgtm
Such a weird bug...

PR #1882, opened 2023-07-20
feat: support external deps for host runs

lkingland:
:gift: host runs (go) pull dependencies

This is the same logic as what happens during a containerized build, but locally on the host.

`func run --builder=host --container=false`


matejvasek:
@lance please override the failing windows unit test.

PR #1885, opened 2023-07-24
fix: resolve tkn task spec locally

matejvasek:
# Changes

- Remove Git task resolver and resolve the task specification locally. This is more reliable, git resolver is timing out on some clusters because git-clone is slow operation.


matejvasek:
@lance PTAL

matejvasek:
@jrangelramos please test this.

jrangelramos:
I just tried.. and ot works properly! Thanks. it is much quicker than previous approach.

PR #1857, opened 2023-07-11
src: Use jobs not plain pods for auxiliary tasks

matejvasek:
# Changes

- :broom: Use `job`s instead of plain `pod`s for auxiliary tasks.
  Rationale: `job` should have security context set properly automatically.

```release-note
use `job`s instead of plain `pod`s for auxiliary tasks
```

PR #1888, opened 2023-07-25
fix: add pod security context to avoid permission denied (#1857)

grafvonb:
@matejvasek this addition causes that the permission denied error on EKS...

```
❯ ./func deploy --remote -p func-node-2-remote
🕗 Creating Pipeline resources
🕔 25 22:39:59.387273   60639 v2.go:105] write tcp 192.168.178.20:61074->52.57.203.11:443: use of closed network connection
Error: failed to run pipeline: cannot upload sources to the PVC: the command failed: exitcode=1, out="tar: can't create directory 'source': Permission denied\ntar: can't open 'source/.funcignore': No such file or directory"
```
is gone. 

```
❯ ./func deploy --remote -p func-node-2-remote
   ✅ Function deployed in namespace "knative-func" and exposed at URL: 
   http://func-node-2-remote.knative-func.3.121.145.23.sslip.io
```
I believe we should always set the `fsGroup` on pods that access a PV to ensure that the pod can access the volume. Please, check the impact of this change on deployments on OpenShift and KinD.

### DIFF
--- a/pkg/k8s/persistent_volumes.go
+++ b/pkg/k8s/persistent_volumes.go
@@ -117,14 +117,17 @@ func runWithVolumeMounted(ctx context.Context, podImage string, podCommand []str
 
 	const volumeMntPoint = "/tmp/volume_mnt"
 	const pVol = "p-vol"
+	fsGroup := int64(1000)
 	job := &batchV1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: jobName,
 		},
 		Spec: batchV1.JobSpec{
-
 			Template: corev1.PodTemplateSpec{
 				Spec: corev1.PodSpec{
+					SecurityContext: &corev1.PodSecurityContext{
+						FSGroup: &fsGroup,
+					},
 					Containers: []corev1.Container{
 						{
 							Name:       "container",


matejvasek:
I am afraid this will break OpenShift.

matejvasek:
This does not work with OpenShift. What if you set `SupplementalGroups` ?

grafvonb:
> This does not work with OpenShift. What if you set `SupplementalGroups` ?

@matejvasek No, it doesn't work. The `SupplementalGroups` only adds groups, and it doesn't enforce anything like `fsGroups` does. 

I will check this on right now a second cluster to ensure that this issue is not specific to the current cluster state.

By the way, I believe we should reevaluate how our project creates pods and containers when using persistent volumes. To ensure safety, we should always define the Security Context for a pod or container, as described here: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/. 

In fact, I always define parameters such as `runAsNonRoot`, `runAsUser`, `runAsGroup`, `fsGroup`, `allowPrivilegeEscalation`, and so on, to avoid security issues. On AWS, the gp2 and gp3 storage classes (as part of the CSI driver) support all these options.

I am aware that the implementation on OpenShift differs from others. Perhaps we should explicitly consider this and detect the environment dynamically to choose the proper configuration?

matejvasek:
@grafvonb the thing is that OpenShift set these things automatically. If manually it causes troubles.

matejvasek:
In particular uid,gid and fsGroup are set automatically from a range that is unique to each namespace. We cannot set it to 1000.

matejvasek:
We ave some detection logic for OpenShift but it is not reliable.

grafvonb:
Solved by https://github.com/knative/func/pull/1889

PR #1895, opened 2023-07-31
chore: cron task for paketo's jammy builder

matejvasek:
# Changes

- :gift: Periodic GH Action that builds latest paketo jammy builder with two additinal buildpacks: `paketo-community/rust` and `dev.knative-sandbox.go`. The result published it as `ghcr.io/knative/builder-jammy-full`.


matejvasek:
PTAL @lkingland @aslom @salaboy @gauron99

lkingland (approved):

PR #1899, opened 2023-08-02
Rename knative-sandbox to knative-extensions

davidhadas:
This is an automated PR to replace knative-sandbox strings to knative-extensions.

PR #1905, opened 2023-08-04
rename knative-sandbox

lkingland:
Renames `knative-sandbox` to `knative-extensions` in a few remaining places.

Supersedes #1899 

/kind chore

### DIFF
--- a/hack/update-builder.go
+++ b/hack/update-builder.go
@@ -257,7 +257,7 @@ func patchBuilder(config *builder.Config) {
 		},
 		{
 			ModuleInfo: dist.ModuleInfo{
-				ID:      "dev.knative-sandbox.go",
+				ID:      "dev.knative-extensions.go",
 				Version: "0.0.4",
 			},
 			ImageOrURI: dist.ImageOrURI{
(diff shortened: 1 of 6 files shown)


hack/update-builder.go
```diff
@@ -257,7 +257,7 @@ func patchBuilder(config *builder.Config) {
 		},
 		{
 			ModuleInfo: dist.ModuleInfo{
-				ID:      "dev.knative-sandbox.go",
```
matejvasek, line 260: This will actually cause troubles.

hack/update-builder.go
```diff
@@ -285,7 +285,7 @@ func patchBuilder(config *builder.Config) {
 				},
 				{
 					ModuleInfo: dist.ModuleInfo{
-						ID: "dev.knative-sandbox.go",
```
matejvasek, line 288: This will actually cause troubles.

hack/update-builder.go
matejvasek: You need to also switch version 0.0.4 => 0.0.6.

hack/update-builder.go
matejvasek: You need to also switch version 0.0.4 => 0.0.6.

matejvasek:
https://github.com/knative/func/pull/1906

matejvasek:
@lkingland please check the failing integration test, probably just a flake, but maybe not.

PR #1911, opened 2023-08-07
Use custom jammy paketo builder

matejvasek:
# Changes

Use custom jammy paketo builder. The builders are modified so they include custom Go and Rust buildpacks. This enables on cluster builds for Go and Rust functions. Where possible (Go, Quarkus) the "tiny" builder is used, other runtimes use "base" builder.

Since jammy builder uses different uid/gid updated task is required, but for sake of compatibility this is done by creating new task file instead of modifying the existing tekton task yaml. Please see
```
diff pkg/pipelines/resources/tekton/task/func-buildpacks/0.1/func-buildpacks.yaml \ 
     pkg/pipelines/resources/tekton/task/func-buildpacks/0.2/func-buildpacks.yaml
```


matejvasek:
Note this also contain commit from https://github.com/knative/func/pull/1914 .

matejvasek:
PTAL @lkingland @jrangelramos @Shashankft9

lkingland (approved):
/hold for tests

test/e2e/scenario_runtime-http_test.go
```diff
@@ -18,6 +18,7 @@ import (
 var runtimeSupportMap = map[string][]string{
 	"node":       {"pack", "s2i"},
 	"go":         {"pack"},
+	"rust":       {"pack"},
```
matejvasek, line 21: @jrangelramos do you recall why this has not been enabled before?

matejvasek:
@Shashankft9 this should enable build of Go and Rust on cluster.

test/e2e/scenario_runtime-http_test.go
jrangelramos: I think it was just never added since scratch of tests.

pkg/pipelines/tekton/validate.go
```diff
@@ -32,17 +31,6 @@ func validatePipeline(f fn.Function) (string, error) {
 			return "", ErrRuntimeRequired
```
Shashankft9, line 31: think now we can do away with using `warningMsg` in this function? https://github.com/knative/func/blob/b2f9dc538870f552ca2bc7393a88e3660c54a4c3/pkg/pipelines/tekton/validate.go#L27
and also its usage here:
https://github.com/knative/func/blob/b2f9dc538870f552ca2bc7393a88e3660c54a4c3/pkg/pipelines/tekton/pipelines_provider.go#L121
https://github.com/knative/func/blob/b2f9dc538870f552ca2bc7393a88e3660c54a4c3/pkg/pipelines/tekton/pipelines_pac_provider.go#L33

matejvasek:
PTAL @lkingland @jrangelramos @Shashankft9

Shashankft9 (approved):

matejvasek:
@lkingland please re-lgtm

PR #1916, opened 2023-08-09
chore: released binaries refer correct tekton task yamls

matejvasek:
# Changes

- :broom: The `./hack/release.sh` script builds binaries that refer to correct tekton tasks. Without this changes binaries refer task from `main` branch, not the release branch.

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -25,7 +25,7 @@ PIPELINE_ARTIFACTS="pkg/pipelines/resources/tekton/task/func-buildpacks/0.1/func
 
 function build_release() {
   echo "🚧 🐧 Building cross platform binaries: Linux 🐧 (amd64 / arm64 / ppc64le / s390x), MacOS 🍏, and Windows 🎠"
-  ETAG=${TAG} make cross-platform
+  FUNC_REPO_BRANCH_REF="$(git branch --show-current)" ETAG=${TAG} make cross-platform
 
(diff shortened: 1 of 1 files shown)


hack/release.sh
```diff
@@ -25,7 +25,7 @@ PIPELINE_ARTIFACTS="pkg/pipelines/resources/tekton/task/func-buildpacks/0.1/func
 
 function build_release() {
   echo "🚧 🐧 Building cross platform binaries: Linux 🐧 (amd64 / arm64 / ppc64le / s390x), MacOS 🍏, and Windows 🎠"
-  ETAG=${TAG} make cross-platform
+  FUNC_REPO_BRANCH_REF="$(git branch --show-current)" ETAG=${TAG} make cross-platform
```
matejvasek, line 28: Just a question: can I use some envvar set by `../vendor/knative.dev/hack/release.sh` instead of invoking `git`?

matejvasek: How branch name should look like: `release-1.11`.

matejvasek:
Now the `unit-tests_func_main` test is failing because `npm` is not available, why?

hack/release.sh
matejvasek: After discussion it appears `git branch --show-current` is a good solution.

matejvasek:
PTAL @jrangelramos

PR #1918, opened 2023-08-09
chore: refer correct tekton task yamls in tests

matejvasek:
# Changes

- Refer correct tekton task yamls in tests. Without this change tests would refer tekton task yamls from the `main` branch, not from the PR head branch.

-  Fixed some lint issues.

### DIFF
--- a/test/presubmit-tests.sh
+++ b/test/presubmit-tests.sh
@@ -25,7 +25,10 @@ export NODE_DISTRO=linux-x64
 
 export KNATIVE_SERVING_VERSION=${KNATIVE_SERVING_VERSION:-latest}
 export KNATIVE_EVENTING_VERSION=${KNATIVE_EVENTING_VERSION:-latest}
-source $(dirname $0)/../vendor/knative.dev/hack/presubmit-tests.sh
+source "$(dirname "$0")/../vendor/knative.dev/hack/presubmit-tests.sh"
+
+FUNC_REPO_BRANCH_REF="${PULL_PULL_SHA}"
+export FUNC_REPO_BRANCH_REF
 
 function post_build_tests() {
   local failed=0
@@ -71,7 +74,7 @@ function unit_tests() {
   make test || failed=1
   if (( failed )); then
     results_banner "Unit tests failed"
-    exit ${failed}
+    exit "${failed}"
   fi
   template_tests
 }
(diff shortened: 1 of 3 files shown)


test/presubmit-tests.sh
```diff
@@ -25,7 +25,10 @@ export NODE_DISTRO=linux-x64
 
 export KNATIVE_SERVING_VERSION=${KNATIVE_SERVING_VERSION:-latest}
 export KNATIVE_EVENTING_VERSION=${KNATIVE_EVENTING_VERSION:-latest}
-source $(dirname $0)/../vendor/knative.dev/hack/presubmit-tests.sh
+source "$(dirname "$0")/../vendor/knative.dev/hack/presubmit-tests.sh"
+
+FUNC_REPO_BRANCH_REF="$(git branch --show-current)"
```
jrangelramos, line 30: would that work? I mean.. the branch is on source repo.. maybe you should set the source repo as well.. 
Later on tests it shall resolve (for example) to `https://raw.githubusercontent.com/knative/func/test-builder-img/pkg/pipelines/resources/tekton/task/func-buildpacks/0.2/func-buildpacks.yaml` then it would fail

matejvasek: I think you are right.

jrangelramos: or maybe set to commit hash (if reachable) so it would resolve to https://raw.githubusercontent.com/knative/func/b2f9dc538870f552ca2bc7393a88e3660c54a4c3/pkg/pipelines/resources/tekton/task/func-buildpacks/0.2/func-buildpacks.yaml

matejvasek: @upodroid any idea how could I get repository in form of `org-or-user/repository` in the prow job?

matejvasek: So hash works with `raw.githubusercontent.com` even when repository is not correct? Interesting.

matejvasek: I'll use the `PULL_PULL_SHA` envvar.

matejvasek:
PTAL @jrangelramos

jrangelramos:
I think it shall work now.. let's give a it a change
/lgtm

upodroid (approved):
LGTM

PR #1920, opened 2023-08-09
chore: mark test as integration test

matejvasek:
# Changes

Mark test as integration test. The test relies on external service: `https://raw.githubusercontent.com/`.


matejvasek:
PTAL @jrangelramos

PR #1886, opened 2023-07-25
fix: deploy warning about insecure defaults (#1851)

grafvonb:
PR #1851

PR #1921, opened 2023-08-10
feat: run host-built containers as non-root

### LINKED ISSUE #1913: Deploying Container runAsNonRoot Error
When deploying a new Function to a k3s cluster using the host builder, the container is unable to start due to a permissions error as follows:

```Error: container has runAsNonRoot and image will run as root```

we are now[ deliberately using a more secure configuration](#1886) than the default.  This is causing some errors because our containers expect to be able to run as root.   We can hopefully maintain this higher security standard since we are building the containers, and can configure them at build-time to a non-root user

lkingland:
:gift: Containers built by the Host builder run as a non-root user.

Impetus is PR #1886 which hardens pod security for services deployed by the Knative Deployer, which causes #1913, an error running host-built function containers which were still attempting to run as root.

/kind enhancement

(_partially_ fixes #1913)


matejvasek:
Shouldn't we also set group?

PR #1922, opened 2023-08-10
Remove progress package (spinner)

### LINKED ISSUE #1513: Spinner "clock" icon should be removed when operation finishes
```
$ func deploy
   🙌 Function image built: image-registry.openshift-image-registry.svc:5000/demo/test-function:latest
   ✅ Function updated in namespace "demo" and exposed at URL:
🕘 https://test-function-demo.apps.zroubali.serverless.devcluster.openshift.com
```

The `deploy` command finished. There is a clock (🕘) icon in the final message, which is confusing. It should be removed or replaced by another icon marking success.

### LINKED ISSUE #1893: Retiring the Spinner
The progress "spinner" has long outlived its intended use (an early-stage example/demo), and needs to be removed.

We will soon transition to a shared library across the client app and plugins which will provide facilities for logging, interactivity (prompts) and progress reporting.

To prepare for this shift, we need to completely remove the `progress` package, and anywhere it is being used replace it with a simple `print` statement.

gauron99:
# Changes
- :wastebasket: Remove feature or internal logic - remove spinner progress package, convert to printing to stderr instead as a first step to changing to a new logging package.

/kind removal
Fixes #1893 
Fixes #1513


gauron99:
cc @lkingland

matejvasek:
@gauron99 please try running `TestOnClusterBuild/pack` locally.

pkg/pipelines/tekton/pipelines_integration_test.go
```diff
@@ -114,35 +113,6 @@ func (p pl) log(args ...any) {
...
-func (p pl) Increment(message string) {
-	p.log("ProgressListener::Increment: ", message)
-	if strings.Contains(message, "URL:") {
-		parts := strings.Split(message, "URL:")
-		if len(parts) < 2 {
-			p.log("bad output message: %q", message)
-			return
-		}
-		u := strings.TrimSpace(parts[1])
-		p.urlChan <- u
```
matejvasek, line 130: Removal of this causes timeout of test since it never gets message that function has been deployed. The are additional changes required to make test work again.

matejvasek: For instance we could add `io.Reader`'s for error/normal outputs into the `PipelinesProvider` structure and print to them instead to `os.Std[out|err]`. The test then could pass pipe when constructing `PipelinesProvider` and inspect output for occurrence of `exposed at URL:\n(\w)`.

wdyt @lkingland

matejvasek: @gauron99 other maybe cleaner way would be to also return `fn.Instance` from the `Run()` method.

matejvasek: The point is that `Run()` should somehow return url of deployed function. Be it via return value or output.

matejvasek: So you can try invoking it.

lkingland (commented):
Looks good so far!

gauron99:
@matejvasek @lkingland PTAL guys, removed the goroutines and added the return value for URL as string. Not sure how to handle the structs in other files, where Run method is used so please let me know on that. I ran the test individually and collectively with `make test` and both seem to work fine

pkg/pipelines/tekton/pipelines_integration_test.go
```diff
@@ -57,18 +55,22 @@ func TestOnClusterBuild(t *testing.T) {
...
 			pp := tekton.NewPipelinesProvider(
 				tekton.WithCredentialsProvider(credentialsProvider),
-				tekton.WithNamespace(ns),
-				tekton.WithProgressListener(pl{urlChan: urlChan}))
+				tekton.WithNamespace(ns))
 
 			f := createSimpleGoProject(t, ns)
 			f.Build.Builder = test.Builder
 
 			go func() {
```
matejvasek, line 63: Give we do not retrieve asynchronously we could simplify this code and get rid of channels and goroutines.

matejvasek: Otherwise looks good.

PR #1923, opened 2023-08-10
doc: Go and Rust on cluster build is supported

matejvasek:
# Changes

Update doc: Go and Rust on cluster build is supported.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -8,6 +8,10 @@ To build the core project, run `make` from the repository root.  This will resul
 
 To remove built artifacts, use `make clean`.
 
+### Build affecting environment variables
+* `FUNC_REPO_REF` affects which github repo will be used to fetch tekton tasks for on cluster build. Default: `knative/func`.
+* `FUNC_REPO_BRANCH_REF` affects which github branch will be used to fetch tekton tasks for on cluster build. Default: `main`.
+
 ## Testing
 
 To run core unit tests, use `make test`.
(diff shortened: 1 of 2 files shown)


matejvasek:
PTAL @jrangelramos

Shashankft9:
this can be removed, since its not relevant anymore: https://github.com/knative/func/blob/76f09c6f44b919ef2c891d3dda37d25c219837cb/docs/building-functions/on_cluster_build.md?plain=1#L39-L43

matejvasek:
> would it make sense to add some documentation around why we are using knative custom jammy builders instead of paketo base? And maybe in the `func` development guide to make use of those two env variables if the tekton tasks are changed?

Jammy is now **the** paketo builder. The previous one has been deprecated.

matejvasek:
> And maybe in the func development guide to make use of those two env variables if the tekton tasks are changed?

What do you mean? @Shashankft9

matejvasek:
> > would it make sense to add some documentation around why we are using knative custom jammy builders instead of paketo base? And maybe in the `func` development guide to make use of those two env variables if the tekton tasks are changed?
> 
> Jammy is now **the** paketo builder. The previous one has been deprecated.

What I mean is https://github.com/paketo-buildpacks/base-builder:
>This repository has been archived by the owner on Jul 20, 2023. It is now read-only.

matejvasek:
I don't think we need to justify our builder IMO.

matejvasek:
PTAL @jrangelramos

matejvasek:
PTAL @Shashankft9

Shashankft9 (approved):

PR #1924, opened 2023-08-10
chore: update release generation

matejvasek:
# Changes

* Minimize release binaries size by stripping debug information.
* Release latest version of buildpack tekton task.


matejvasek:
PTAL @jrangelramos

PR #1925, opened 2023-08-10
feat: tekton task urls in the `environment` sub-command output

matejvasek:
# Changes

- :gift: Print tekton tasks urls in output of `func environment`.


matejvasek:
PTAL @jrangelramos @gauron99

Shashankft9 (approved):

PR #1937, opened 2023-08-17
chore: faster `make build`

matejvasek:
# Changes

- :broom: Faster `make build`.
- :bug: The `zz_filesystem_generated.go` file must be build **after** CA bundle download.


matejvasek:

lkingland (approved):

PR #1942, opened 2023-08-24
chore: align dependencies with Knative

matejvasek:


go.mod
```diff
@@ -2,6 +2,21 @@ module knative.dev/func
 
 go 1.20
 
+replace (
+	// Version used in Eventing
+	github.com/cloudevents/sdk-go/sql/v2 => github.com/cloudevents/sdk-go/sql/v2 v2.0.0-20220930150014-52b12276cc4a
```
dsimansk, line 7: IMO that's probably covered on the Eventing side and the versions bubbled correctly. Then the replace shouldn't be needed anymore.

go.mod
```diff
@@ -2,6 +2,21 @@ module knative.dev/func
...
+
+	// Use older version since it checks ~/.docker/config.json for credentials as opposed to later versions
+	github.com/containers/image/v5 => github.com/containers/image/v5 v5.23.1
+
+	// Tekton CLI uses newer version without currently used constructor code
+	github.com/hinshun/vt10x => github.com/hinshun/vt10x v0.0.0-20180809195222-d55458df857c
+	// Pin k8s.io dependencies to align with Knative and Tekton needs
+	k8s.io/api => k8s.io/api v0.25.4
+	k8s.io/apimachinery => k8s.io/apimachinery v0.25.4
+	k8s.io/client-go => k8s.io/client-go v0.25.4
```
dsimansk, line 17: Depending on `Tekton` versions, the `k8s.io` could be at `0.26.x` to align with rest of Knative.

dsimansk: Both my comments are OK to follow-up in single-purpose isolated PRs.

matejvasek:
PTAL @dsimansk

go.mod
```diff
@@ -2,6 +2,18 @@ module knative.dev/func
 
 go 1.20
 
+replace (
+	// Use older version since it checks ~/.docker/config.json for credentials as opposed to later versions
+	github.com/containers/image/v5 => github.com/containers/image/v5 v5.23.1
+
+	// Tekton CLI uses newer version without currently used constructor code
+	github.com/hinshun/vt10x => github.com/hinshun/vt10x v0.0.0-20180809195222-d55458df857c
```
matejvasek, line 10: Could this be removed too @dsimansk (provided the code compiles)?

dsimansk: Definitely, I recall the library was used in a test setup, mock terminal or something like that. That might have been refactored since then.

dsimansk: E.g. here: https://github.com/knative/func/blob/main/cmd/prompt/prompt_test.go#L27. 

Some of the later revision don't have such constructor function.

go.mod
```diff
@@ -2,6 +2,18 @@ module knative.dev/func
...
+	// Pin k8s.io dependencies to align with Knative and Tekton needs
+	k8s.io/api => k8s.io/api v0.26.5
+	k8s.io/apimachinery => k8s.io/apimachinery v0.26.5
+	k8s.io/client-go => k8s.io/client-go v0.26.5
```
dsimansk, line 14: Would the `k8s.io` group jump further ahead without pinning them to aligned version?

matejvasek: The k8s.io/client-go dep would be `v1.5.2`. The other two would remain the same -- `v0.26.5`.
Would jump of k8s.io/client-go to `v1.5.2` be a problem? @dsimansk

matejvasek:
PTAL @dsimansk @lkingland @gauron99 @jrangelramos

matejvasek:
plz lgtm @dsimansk

PR #1949, opened 2023-08-28
chore: use local registry in test

matejvasek:
# Changes

Use local registry instead of ttl.sh.

### DIFF
--- a/.github/workflows/test-e2e-oncluster-runtime.yaml
+++ b/.github/workflows/test-e2e-oncluster-runtime.yaml
@@ -16,6 +16,8 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
       - uses: imjasonh/setup-ko@v0.6
+      - name: Patch S2I Task
+        run: ./hack/patch-s2i-task.sh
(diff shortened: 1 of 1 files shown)


lkingland:
/lgtm
I am really looking forward to this change!
/hold for the E2E OnCluster RT test, which looks like a flake on the s2i subtests (rerun triggered)

matejvasek:
> I believe the test fail is legitimate. The thing is the s2i is using the internal registry as `https`. The pack task works all right because it consider `*.local` as http (not https) repo.

Actually maybe not, I forgot that I applied patch to the s2i task allowing push to the registry as insecure.

matejvasek:
> > I believe the test fail is legitimate. The thing is the s2i is using the internal registry as `https`. The pack task works all right because it consider `*.local` as http (not https) repo.
> 
> Actually maybe not, I forgot that I applied patch to the s2i task allowing push to the registry as insecure.

Actually^2
Well that patch probably does nothing since tasks are downloaded from PR branch that does not contain local changes. I came up with different solution in different PR.

PR #1958, opened 2023-08-31
fix: direct upload OCP ppc64le, x390x

matejvasek:
# Changes

Fixing direct-upload on-cluster-build for OCP ppc64le, x390x.


matejvasek:
PTAL @jrangelramos

jrangelramos:
seems that will impact not only OCP for ppc64le, s390x.

PR #1962, opened 2023-09-04
Cleanup orphaned func, new image-name calculation

### LINKED ISSUE #1953: Cleanup orphaned Function instance on namespace change
When deploying a function in namespace `one` and later on changing context to namespace `two` and deploying same function again with specified `--namespace` flag will result in 2 deployed function in namespace `one` and `two`. 

The function source code should be matched with the cluster Function deployed only, to prevent leaving dangling Function on cluster if user desired a new namespace for it.

Therefore we should delete the Function on cluster in namespace `one` as now the user wants to have it deployed in namespace `two`.

gauron99:
# Changes
- :broom: Delete old image on ns change
-  Add new image field in deploy section - recalculate image name unless it was explicitly given

## Whats new
1. In case user explicitly specifies `--namespace` in a subsequent `func deploy`, the "old" func deployed will be deleted after the new one is deployed successfully.
2. New field added as `.Build.Image` which will serve as the new `.Image` currently with few changes. -> this serves as further following the idea that `Func.yaml` shouldnt be possibly manipulated with at all (`.Build` fields shouldnt be manipulated with now same as `.Deploy`).
- Instead the `.Image` now is empty UNLESS its specified by user using `--image` flag in which case it will be used in all subsequent builds because by specifying this flag user says "I know what Im doing" and no image name calculation will be performed.
3. new field in runtime metadata directory `.func/` called `built-image` now serves as runtime last built image name. Once pushed this also contains a sha (This is temporary, either will be fixed in this PR or separately. -- the issue is that sha is fetched from Push action instead of Build, where it should be possible to get it)
---
4. Namespace structure updated: Now `f.Namespace` describes *namespace TO BE deployed in (user defined)* and `f.Deploy.Namespace` describes *namespace Function WAS deployed in*; in other words, one is the desired namespace and the other is the current  status, where the Function is already deployed.
### additional
new error definitions
test updates and fixups to fit new structures

## TODO
- [x] fix new errors in tests

/kind cleanup

Fixes #1953


pkg/knative/integration_test.go
```diff
@@ -127,10 +127,11 @@ func TestIntegration(t *testing.T) {
...
 		//     * files under /etc/cm and /etc/sc.
 		//   * application also prints the same info to stderr on startup
-		Image:       "quay.io/mvasek/func-test-service",
-		ImageDigest: "sha256:2eca4de00d7569c8791634bdbb0c4d5ec8fb061b001549314591e839dabd5269",
-		Created:     now,
+		Created: now,
 		Deploy: fn.DeploySpec{
+			// TODO: gauron99 - is it okay to have this explicitly set to deploy.image already?
+			// With this I skip the logic of setting the .Deploy.Image field but it should be fine for this test
+			Image:     "quay.io/mvasek/func-test-service@sha256:2eca4de00d7569c8791634bdbb0c4d5ec8fb061b001549314591e839dabd5269",
```
gauron99, line 134: opening this TODO for comments

cmd/deploy.go
```diff
@@ -302,18 +348,38 @@ func runDeploy(cmd *cobra.Command, newClient ClientFactory) (err error) {
...
+	// TODO: gauron99 - This checks whether a registry has changed on a subsequent build
+	// decision (Like on func deploy --registry=X when already built image is Y)
+	// This can probably be added into f.Built() function for simpler and more global
+	// implementation
+	registryChangedSubsequently := false
+	if !strings.Contains(f.Build.Image, f.Registry) && f.Image == "" && f.Build.Image != "" {
+		registryChangedSubsequently = true
+	}
+
+	build, _ := strconv.ParseBool(flag)
```
lkingland, line 361: I agree with your comment here.

Let's see if I can remember/understand the logic:   We want to rebuild if the user has changed the desired registry.  Another way to think of it being:  If the function's metadata indicates that the _next_ build of the function would result in a different image than is currently listed as the last one built (f.Build.Image), then it must be because the user has changed either f.Registry.

So yes, I think this could probably go in f.Built().  Doing so might also allow dropping the `&& f.Image == "" && f.Build.Image != ""` additional conditions:

The conditional `&& f.Image == ""` I think may not need to be included, because we want to treat the function as unbuilt if the registry changed since last build, even if no explicit image was provided as an override (the purpose of `f.Image`).

The subsequent conditional `&& f.Build.Image != ""` may also not need to be included, because the function should be considered unbuilt if there has been no previous build ever, which is covered by the primary case `!strings.Contains(f.Build.Image, f.Registry)`....  That will return true (function unbuilt), if `f.Build.Image` is empty (never been built before).

Perhaps try, and see if the tests pass :)

cmd/deploy.go
```diff
@@ -626,8 +681,8 @@ func (c deployConfig) Validate(cmd *cobra.Command) (err error) {
 
 	// Check Image Digest was included
 	// (will be set on the function during .Configure)
-	var digest string
-	if digest, err = imageDigest(c.Image); err != nil {
+	var digest bool
+	if digest, err = hasDigest(c.Image); err != nil {
```
lkingland, line 685: minor nit:  we called this `digested` above, so probably worth updating to match. It's easy to mistake this for the digest value itself (which is what it was originally named for)

cmd/deploy.go
```diff
@@ -672,54 +729,22 @@ func (c deployConfig) Validate(cmd *cobra.Command) (err error) {
...
 	targetNamespace := cfg.Namespace
 
-	// If potentially creating a duplicate deployed function in a different
-	// namespace.  TODO: perhaps add a --delete or --force flag which will
-	// automagically delete the deployment in the "old" namespace.
+	// If creating a duplicate deployed function in a different
+	// namespace.
 	if targetNamespace != currentNamespace && currentNamespace != "" {
-		fmt.Fprintf(out, "Warning: function is in namespace '%s', but requested namespace is '%s'. Continuing with deployment to '%v'.\n", currentNamespace, targetNamespace, targetNamespace)
+		fmt.Fprintf(out, "Info: chosen namespace has changed from '%s' to '%s'. Undeploying function from '%s' and deploying new in '%s'.\n", currentNamespace, targetNamespace, currentNamespace, targetNamespace)
```
lkingland, line 747: Such a huge improvement 👍🏻

cmd/root.go
```diff
@@ -214,46 +214,6 @@ func deriveNameAndAbsolutePathFromPath(path string) (string, string) {
 	return pathParts[len(pathParts)-1], absPath
 }
 
-// deriveImage returns the same image name which will be used.
```
lkingland, line 217: Removing these 40 lines (along with the others like them) is a great sign we're headed in the right direction

pkg/functions/client_test.go
```diff
@@ -21,6 +22,7 @@ import (
 	"time"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
+	cmd "knative.dev/func/cmd"
```
lkingland, line 25: We only need to use the commands in the tests from within that package.  This package can test the `Client` directly

pkg/functions/client_test.go
```diff
@@ -1888,3 +1904,132 @@ func TestClient_RunRediness(t *testing.T) {
...
+	client := fn.New()
+	// initialize function with namespace defined as nsone
+	_, err := client.Init(fn.Function{Runtime: "go", Root: root,
+		Deploy: fn.DeploySpec{Namespace: nsOne}})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// 	// set deploy command
+	cmd := cmd.NewDeployCmd(cmd.NewTestClient(
```
lkingland, line 1984: You can just remove the `cmd` here and use the client instance above directly.  This is because we are in the client package, testing only the client.

pkg/functions/client_test.go
```diff
@@ -1888,3 +1904,132 @@ func TestClient_RunRediness(t *testing.T) {
...
+	root, cleanup := Mktemp(t)
+	defer cleanup()
+
+	nsOne := "nsone"
+	nsTwo := "nstwo"
+
+	// A mock remover which fails if the name from the func.yaml is not received.
+	remover := mock.NewRemover()
+
+	client := fn.New()
```
lkingland, line 1975: If you update this client to use the mock deployer, mock remover and test registry, you can remove the `cmd` below and use the client directly

pkg/functions/client_test.go
```diff
@@ -1888,3 +1904,132 @@ func TestClient_RunRediness(t *testing.T) {
...
+	}
+
+	// 	// set deploy command
+	cmd := cmd.NewDeployCmd(cmd.NewTestClient(
+		fn.WithDeployer(mock.NewDeployer()),
+		fn.WithRegistry(TestRegistry),
+		fn.WithRemover(remover),
+	))
+
+	// set arguments for deploy command and execute the command
```
lkingland, line 1990: What you'd do here is instantiate `f` from `root` that was initialized above, then deploy it, update namespace, then deploy again (using `client.Deploy` directly)

pkg/functions/function.go
```diff
@@ -67,9 +75,6 @@ type Function struct {
 	// "Registry+Name:latest" to derive the Image.
 	Image string `yaml:"image,omitempty"`
 
-	// ImageDigest is the SHA256 hash of the latest image that has been built
-	ImageDigest string `yaml:"imageDigest,omitempty"`
-
```
lkingland, line 72: 🥂

pkg/knative/integration_test.go
lkingland: I would prefer this test used the complete path, and the `fn` API, rather than hard-coding the function struct, but it appears to perhaps be a test performance enhancement.

Perhaps leave it as-is (as you have done here) for this PR and we can confirm with Matej on our next WG call.

lkingland (commented):
Looking great, and all the pertinent tests are passing! (a couple of timeouts notwithstanding).

I requested a couple small changes moving logic around, but generally looks GTG to me! 🥳

cmd/deploy_test.go
```diff
@@ -1533,3 +1668,156 @@ func Test_ValidateBuilder(t *testing.T) {
...
+
+	cmd.SetArgs([]string{"--registry=" + newRegistry, "--build=false"})
+
+	// Second: Deploy with different registry and expect 'not built' error because
+	// registry has changed but build is disabled
+	if err := cmd.Execute(); err == nil {
+		t.Fatal(err)
+	}
+
+	// ASSERT
```
lkingland, line 1760:
```suggestion
	// ASSERT the failure to build did not result in an altered function on disk
```

cmd/deploy_test.go
```diff
@@ -1533,3 +1668,156 @@ func Test_ValidateBuilder(t *testing.T) {
...
+	}
+
+	if !strings.Contains(expectF.Build.Image, newRegistry) {
+		t.Fatalf("expected built image '%s' to contain new registry '%s'\n", expectF.Build.Image, newRegistry)
+	}
+}
+
+// TestReDeploy_OnRegistryChangeWithBuildFalse should fail with function not
+// being built because the registry has changed
+func TestReDeploy_OnRegistryChangeWithBuildFalse(t *testing.T) {
```
lkingland, line 1723:
```suggestion
// TestDeploy_RegistryUpdateRequiresBuild should fail with function not
// being built because the registry has changed
func TestDeploy_RegistryUpdateRequiresBuild(t *testing.T) {
```
(I'm making these name change suggestions simply so that all of the deploy tests sort alphanumerically and can be run with `go test -run TestDeploy_`, even the ReDeploys)

cmd/run.go
```diff
@@ -158,20 +157,6 @@ func runRun(cmd *cobra.Command, args []string, newClient ClientFactory) (err err
 		return
 	}
 
-	// TODO: this is duplicate logic with runBuild and runRun.
-	// Refactor both to have this logic part of creating the buildConfig and thus
-	// shared because newRunConfig uses newBuildConfig for its embedded struct.
-	if f.Registry != "" && !cmd.Flags().Changed("image") && strings.Index(f.Image, "/") > 0 && !strings.HasPrefix(f.Image, f.Registry) {
```
lkingland, line 164: 🥂

pkg/functions/client.go
```diff
@@ -737,6 +760,30 @@ func (c *Client) Deploy(ctx context.Context, f Function, opts ...DeployOption) (
 		return f, err
 	}
 
+	// If Redeployment to NEW namespace was successful -- undeploy dangling Function in old namespace.
+	// On forced namespace change (using --namespace flag)
+	if f.Namespace != "" && f.Namespace != f.Deploy.Namespace && f.Deploy.Namespace != "" {
+		// TODO: when new prompt is implemented, add here a "are you sure?" check possibly
```
lkingland, line 766: Conceptually, yes, but since we are in the client, and the client is expected to be a library not a CLI, what would probably need to happen here for us to implement a confirmation prompt is:  expose a method which would allow the _CLI_ to do the check `if moving(f) { Prompt("This will remove the old one, groovy? (Y/n):") }`

pkg/functions/client.go
```diff
@@ -737,6 +760,30 @@ func (c *Client) Deploy(ctx context.Context, f Function, opts ...DeployOption) (
...
+		}
+
+		// c.Remove removes a Function in f.Deploy.Namespace which removes the OLD Function
+		// because its not updated yet (see few lines below)
+		err = c.Remove(ctx, f, true)
+
+		// Warn when service is not found and set err to nil to continue. Function's
+		// service mightve been manually deleted prior to the subsequent deploy or the
+		// namespace is already deleted therefore there is nothing to delete
+		if ErrFunctionNotFound != err {
```
lkingland, line 778: IMHO this is going to really help users trust the code; little warnings like this.

pkg/functions/client_test.go
```diff
@@ -926,27 +927,89 @@ func TestClient_Deploy_RegistryUpdate(t *testing.T) {
...
 	}
-	if f, err = client.Deploy(context.Background(), f); err != nil {
+
+	expected = "example.com/bob/f:latest"
+	if f.Build.Image != expected {
+		t.Errorf("expected image name to change to '%v', but got '%v'", expected, f.Build.Image)
+	}
+}
+
+// TestClient_Deploy_NamespaceUpdate ensures that namespace deploymen has
```
lkingland, line 966:
```suggestion
// TestClient_Deploy_NamespaceUpdate ensures that namespace deployment has
```

pkg/functions/client_test.go
```diff
@@ -1888,3 +1998,147 @@ func TestClient_RunRediness(t *testing.T) {
 		t.Fatalf("err on job stop. %v", err)
 	}
 }
+
+// TestClient_BuildCleanFingerprint ensures that when building a Function the
+// source controlled state is not modified (git would show no unstaged changes).
+// For example, the image name generated when building should not be stored
+// in function metadata that is checked into source control (func.yaml).
+func TestClient_BuildCleanFingerprint(t *testing.T) {
```
lkingland, line 2006: This might just be my favorite bonus from this PR

pkg/functions/client_test.go
```diff
@@ -1888,3 +1998,147 @@ func TestClient_RunRediness(t *testing.T) {
...
+		t.Fatal("just building a Function resulted in a dirty function state (fingerprint changed)")
+	}
+}
+
+// Test_RemoveInvokedOnOldFunction checks that Remover was invoked after a
+// subsequent redeploy to a new namespace.
+// specifically: deploy to 'nsone' -> simulate change of namespace with change to
+// f.Namespace -> redeploy to that namespace and expect the remover to be invoked
+// for old Function in ns 'nsone'.
+func TestClient_RemoveInvokedOnOldFunction(t *testing.T) {
```
lkingland, line 2059:
```suggestion
// TestClient_DeployRemoves ensures that the Remover is invoked when a
// function is moved to a new namespace.
// specifically: deploy to 'nsone' -> simulate change of namespace with change to
// f.Namespace -> redeploy to that namespace and expect the remover to be invoked
// for old Function in ns 'nsone'.
func TestClient_DeployRemoves(t *testing.T) {
```

lkingland (changes requested):
Looks great!
I just made a few minor nitpicks about tests and added some suggested name changes etc.  I think this one is basically GTG 👍🏻

cmd/deploy.go
```diff
@@ -234,34 +234,37 @@ func runDeploy(cmd *cobra.Command, newClient ClientFactory) (err error) {
...
+	// If using Openshift registry AND redeploying Function, update image registry
+	if f.Namespace != "" && f.Namespace != f.Deploy.Namespace && f.Deploy.Namespace != "" {
+		// when running openshift, namespace is tied to registry, override on --namespace change
+		// The most default part of registry (in buildConfig) checks 'k8s.IsOpenShift()' and if true,
+		// sets default registry by current namespace
+
+		// If Function is being moved to different namespace in Openshift -- update registry
+		if k8s.IsOpenShift() {
+			// this name is based of k8s package
+			f.Registry = "image-registry.openshift-image-registry.svc:5000/" + f.Namespace
```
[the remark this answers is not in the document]
gauron99: Theres a few bugs on this, I will create a separate PR for OC & registry related stuff so we can merge this

lkingland (approved):

lkingland (approved):
🎉

PR #1969, opened 2023-09-12
fix: use overlay storage for s2i

matejvasek:
# Changes

- :gift: Use overlay storage for `buildah` instead of vfs in `s2i` on cluster build. It is much more efficient where vfs takes 5.5G overlay takes just 1.2G.

```release-note
Use overlay storage driver instead of vfs driver in s2i tekton task.
```


matejvasek:
/hold
we must make sure this works on variety of clusters.

matejvasek:
@grafvonb Hi, if you had some spare time would you please verify this works on EKS?

matejvasek:
@trisberg Hi, if you had some spare time would you please verify this works on GKE?

matejvasek:
**Note:**  when testing this locally you need to set
```go
	FuncRepoRef       = "matejvasek/faas"
	FuncRepoBranchRef = "try-overlay"
```
in https://github.com/knative/func/blob/67517b738e1e1393c2f5deb98904b730257c3140/pkg/pipelines/tekton/templates.go#L91-L92

matejvasek:
The thing is: overlays storage may not work on some cluster.

jrangelramos:
Hi @matejvasek it works on OCP.

matejvasek:
@jrangelramos While testing did you update `FuncRepoRef` and `FuncRepoBranchRef`?

jrangelramos:
> @jrangelramos While testing did you update `FuncRepoRef` and `FuncRepoBranchRef`?

Yes! I did

PR #1971, opened 2023-09-13
Use our own s2i image

matejvasek:
# Changes

- :gift: Use our own s2i image. It is much more recent (the previous one is 2 years old) but more importantly the new image is multi-arch.


matejvasek:
PTAL @jrangelramos

matejvasek:
PTAL @gauron99

PR #1974, opened 2023-09-13
fix: use current func image in tests

matejvasek:
# Changes

- :gift: Use a different approach to pass current func image to the tests, this approach does not require modification of task yaml nor access to outer internet (ttl.sh). The important thing is that now we are refering tekton tasks from github --  they are not preinstalled meaning any local patches to task have no effect.

### DIFF
--- a/hack/create-testing-func-image.sh
+++ b/hack/create-testing-func-image.sh
@@ -4,12 +4,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-KO_DOCKER_REPO="ttl.sh/$(head -c 128 </dev/urandom | LC_CTYPE=C tr -dc 'a-z0-9' | fold -w 8 | head -n 1)"
+KO_DOCKER_REPO="localhost:50000/knative/func"
 export KO_DOCKER_REPO
 
-REF_FILE=$(mktemp)
-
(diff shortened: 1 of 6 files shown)


matejvasek:
PTAL @jrangelramos

lance:
/override "On Cluster RT Test (1.20.2, ubuntu-latest)"

hack/create-testing-func-image.sh
```diff
@@ -4,12 +4,7 @@ set -o errexit
...
 
-KO_DOCKER_REPO="ttl.sh/$(head -c 128 </dev/urandom | LC_CTYPE=C tr -dc 'a-z0-9' | fold -w 8 | head -n 1)"
+KO_DOCKER_REPO="localhost:50000/knative/func"
 export KO_DOCKER_REPO
 
-REF_FILE=$(mktemp)
-
-ko build --image-refs "${REF_FILE}" --tags "30m" -B ./cmd/func
-
-yq -Y -i ".spec.steps[0].image = \"$(cat "${REF_FILE}")\"" \
```
matejvasek, line 14: The thing is this patch does not work anymore. The task is not loaded from local FS but from PR branch so local modifications have no effect.

PR #1976, opened 2023-09-14
chore: use local registry in test 

matejvasek:
# Changes

- :gift: Use local registry for e2e on-cluster-build tests.
- :bug: Increased some timeouts values in `hack/allocate.sh`.
- :bug: Fixed the way we to make `registry.default.svc.cluster.local:5000` insecure in s2i build task.
- :bug: Provisional fix for bug in "In Cluster Dialer" causing error when connection is intentional prematurely closed.


hack/patch-s2i-task.sh
```diff
@@ -1,21 +0,0 @@
-#!/usr/bin/env bash
-
-# This script patches the s2i Tekton task, so it recognizes registry.default.svc.cluster.local:5000 as insecure.
-
-echo "Patching s2i Tekton task to use 'registry.default.svc.cluster.local:5000' as an insecure registry."
-
-patch pkg/pipelines/resources/tekton/task/func-s2i/0.1/func-s2i.yaml <<EOF
```
matejvasek, line 7: Note: this patch was not working anymore. We are pulling tekton tasks directly from PR branch via `https://raw.githubusercontent.com/` -- local changes have no effect.

hack/create-testing-func-image.sh
```diff
@@ -1,10 +0,0 @@
-#!/usr/bin/env bash
```
matejvasek, line 1: This was just moved but git does not detect this as a rename.

matejvasek:
@lkingland @jrangelramos @gauron99

lance:
/override "On Cluster RT Test (1.20.2, ubuntu-latest)"

PR #1979, opened 2023-09-15
WIP: Adding super simple langchain template

matzew:
# Changes

- adding super simple `langchain` template (`func create -l python -t langchain` works, when building the `func` locally)

We need to see if that structure makes sense. Will investigate more

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/templates/python/langchain/Procfile
+++ b/templates/python/langchain/Procfile
@@ -1,0 +1,1 @@
+web: python -m parliament .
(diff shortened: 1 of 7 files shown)


matzew:
/close
replace w/ https://github.com/knative-extensions/func-tastic/pull/47

PR #1991, opened 2023-09-21
test: run "on cluster RT" tests in parallel

matejvasek:
# Changes

- Run on cluster RT tests in parallel.


jrangelramos:
Hmm.. do you need additional improvement w/ this change? It took 34m to run in parallel.
On my previous [PR](https://github.com/knative/func/pull/1977/) it took 33m (not in paralell).
Maybe GH is just busy and it shall run faster some other PR or other moment..

The only caveat is it mix test log output which may turn a bit dificult to troubleshoot whenever something fails..  but I'm ok with that as long as it speed up tests.

matejvasek:
@jrangelramos if tests use `t.Log()` not raw output then output is not really mixed, well maybe a little but in a way that can be demuxed. For instance if you run test in IDE it will split it for you.

This improves speed at least when testing locally. On GH not so much.

PR #1992, opened 2023-09-21
chore: parallelize cluster allocate script

matejvasek:
# Changes

Parallelised cluster allocate script.


matejvasek:
@jrangelramos this makes cluster setup 2x fast.

matejvasek:
PTAL @jrangelramos

PR #2001, opened 2023-09-29
Adding support for `func subscribe` for creating mutiple triggers, based on event filters

matzew:
Little demo: https://youtu.be/O6M7mfR6JfM 

# Changes

- - :gift: Add new `subscribe` command to create Knative Triggers for event routing

/kind <kind>

Fixes #

**Release Note**

```release-note
kn func subscribe will allow you to create Knative Eventing Triggers for improved event processing for kn func
```

**Docs**

```docs

```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -76,6 +76,7 @@ Learn more about Knative at: https://knative.dev`, cfg.Name),
 				NewDeployCmd(newClient),
 				NewDeleteCmd(newClient),
 				NewListCmd(newClient),
+				NewSubscribeCmd(),
 			},
 		},
 		{
--- a/pkg/knative/deployer.go
+++ b/pkg/knative/deployer.go
@@ -314,6 +314,9 @@ func createTriggers(ctx context.Context, f fn.Function, err error, client client
 		err = fmt.Errorf("knative deployer failed to get the Service for Trigger: %v", err)
 		return err
 	}
+
+	fmt.Fprintf(os.Stderr, "🎯 Creating Triggers on the cluster\n")
+
 	for i, sub := range f.Subscription {
 		// create the filter:
 		attributes := make(map[string]string)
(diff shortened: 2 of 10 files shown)


pkg/knative/deployer.go
```diff
@@ -189,6 +196,40 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+			for i, sub := range f.Subscription {
+
+				// create the filter:
+				attributes := make(map[string]string)
+				for key, value := range sub.Filters {
+					attributes[key] = value
+				}
+
+				err = eventingClient.CreateTrigger(ctx, &eventingv1.Trigger{
+					ObjectMeta: metav1.ObjectMeta{
```
lkingland, line 208: I love how straightforward the API call is here

lkingland (commented):
Looking good so far!

matzew:
/hold

the filtering needs to be a bit different

jrangelramos:
Hi @matzew . If possible, could you add some test pls?

lkingland:
> Hi @matzew . If possible, could you add some test pls?

We _do_ install Eventing in our test cluster, so an integration test in https://github.com/knative/func/blob/main/pkg/functions/client_int_test.go might be fairly straightforward

cmd/root.go
```diff
@@ -76,6 +76,7 @@ Learn more about Knative at: https://knative.dev`, cfg.Name),
 				NewDeployCmd(newClient),
 				NewDeleteCmd(newClient),
 				NewListCmd(newClient),
+				NewSubscribeCmd(),
```
lkingland, line 79: For testing support, this should pass through the `newClient` client factory; but I don't think that should hold up the PR.

lkingland (approved):

cmd/root.go
matzew: I will add additional PR. thanks for the review on this

PR #2012, opened 2023-10-09
test: optimise disk usage

matejvasek:
This should save some intermediary steps. With the previous solution the image was stored at three places: docker daemon, local registry and in the KinD cluster node. With this new approach the image should be stored only in the node.


lkingland (approved):

matejvasek:

PR #2031, opened 2023-10-17
chore: Add go setup as composite action

dsimansk:
# Changes

- chore: Add go setup as composite action

This is a very opinionated proposal. I can just `sed` all the version otherwise. :)

There are several go version matrix params, but specifying single value only. I wonder if something more centralized could be a bit easier to maintain. 

/cc @lkingland @matejvasek 

/kind <kind>

Fixes #

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.github/composite/go-setup/action.yaml
+++ b/.github/composite/go-setup/action.yaml
@@ -1,0 +1,9 @@
+name: "Golang setup"
+
+runs:
+  using: "composite"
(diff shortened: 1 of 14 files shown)


lkingland:
Well I like it, and any attention paid to our tooling is appreciated

/lgtm
/approve

PR #2033, opened 2023-10-18
fix: replace deprecated function

matejvasek:


jrangelramos:
Hi @matejvasek I added them on PR before.. https://github.com/knative/func/pull/2032 .. take a look

PR #2034, opened 2023-10-18
test: better clean up after test

matejvasek:


jrangelramos:
shouldn't it delete the NS after test as tear down?

PR #2050, opened 2023-10-24
test: verify embedded templates

matejvasek:


matejvasek:

PR #2055, opened 2023-10-25
Add limit to number of manifests

AdamKorcz:
# Changes
A malicious image can include a high number of manifests and which makes the loop time controllable by an attacker. This can result in a DoS of func through an endless-data attack vector.

This PR sets a limit that is far more than a legitimate image would require, but that would still be fast to loop though.

/kind bug

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


lkingland (approved):
Thanks for the contribution @AdamKorcz

PR #2071, opened 2023-11-06
fix: release uses correct version tags

matejvasek:


hack/release.sh
```diff
@@ -25,7 +25,10 @@ PIPELINE_ARTIFACTS="pkg/pipelines/resources/tekton/task/func-buildpacks/0.2/func
 
 function build_release() {
   echo "🚧 🐧 Building cross platform binaries: Linux 🐧 (amd64 / arm64 / ppc64le / s390x), MacOS 🍏, and Windows 🎠"
-  FUNC_REPO_BRANCH_REF="$(git branch --show-current)" ETAG=${TAG} make cross-platform
+
+  local go_module_version
+  go_module_version="v0.$(( $(minor_version "$TAG") + 27 )).$(patch_version "$TAG")"
```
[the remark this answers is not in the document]
matejvasek: resolved

matejvasek:
PTAL @jrangelramos @lkingland

lkingland (approved):
Nice find; thanks @matejvasek !

hack/release.sh
```diff
@@ -25,7 +25,17 @@ PIPELINE_ARTIFACTS="pkg/pipelines/resources/tekton/task/func-buildpacks/0.2/func
...
-  FUNC_REPO_BRANCH_REF="$(git branch --show-current)" ETAG=${TAG} make cross-platform
+
+  local go_module_version
+  local knative_version
+  if (( TAG_RELEASE )); then
+    knative_version="${TAG}"
+    go_module_version="v0.$(( $(minor_version "$TAG") + 27 )).$(patch_version "$TAG")"
+  else
+    knative_version="$(git describe --tags --match 'knative-*')"
+    go_module_version="$(git describe --tags --match 'v*')"
```
lkingland, line 36: This else satatement may not be necessary, since the values calculated are the defaults for those two values when running Make

matejvasek: I do prefer setting these so there is only single one place where we call `make cross-platform`.
Otherwise there would be either:
   two call to `make cross-platform` (with potential for combinatoric explosion for multiple envvars),
   or modify Makefile so it treats unset and empty envvar the same way.

matejvasek: Just to be clear: `VERS="" KVER="" make cross-platform` will **not** set the version but leaves them empty!

PR #2090, opened 2023-11-20
Log if image is referenced by tag

AdamKorcz:
# Changes
Adds a logging statement if the image is referenced by tag to inform the user of the risk.

/kind enhancement

**Release Note**

```release-note
Log if image is referenced by tag in s2i builder.
```

### DIFF
--- a/pkg/builders/s2i/builder.go
+++ b/pkg/builders/s2i/builder.go
@@ -356,6 +356,9 @@ func s2iScriptURL(ctx context.Context, cli DockerClient, image string) (string, 
 			if err != nil {
 				return "", fmt.Errorf("cannot parse image name: %w", err)
 			}
+			if _, ok := ref.(name.Tag); ok {
+				fmt.Println("image referenced by tag which is discouraged: Tags are mutable and can point to a different artifact than the expected one")
+			}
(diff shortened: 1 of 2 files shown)


pkg/builders/s2i/builder.go
```diff
@@ -356,6 +356,9 @@ func s2iScriptURL(ctx context.Context, cli DockerClient, image string) (string,
 			if err != nil {
 				return "", fmt.Errorf("cannot parse image name: %w", err)
 			}
+			if _, ok := ref.(name.Tag); ok {
```
matejvasek, line 359: Could we ignore this for default (Red Hat) builder images?
It's not good look if there is warning in default configuration.

matejvasek:

pkg/builders/s2i/builder.go
AdamKorcz: Yes, I can add this check today.

matejvasek:
@ReToCode @lance please override the `Check Source (ubuntu-latest)` error.

pkg/builders/s2i/builder.go
matejvasek: I already added it.

PR #2094, opened 2023-11-28
chore: update node builder version

matejvasek:


matejvasek:
Hmm these are not just flakes. Image build on never image does not handle signals properly. We must inquiry why.

matejvasek:
There is a bug in `npm` that is in nodejs-18 image. It does not propagate signals to child processes.

dsimansk:
Rebased to rerun the CI on current version.

pkg/builders/s2i/builder.go
```diff
@@ -139,7 +139,7 @@ func (b *Builder) Build(ctx context.Context, f fn.Function, platforms []fn.Platf
 	// Build Config
 	cfg := &api.Config{}
 	cfg.Quiet = !b.verbose
-	cfg.Tag = f.Build.Image
+	cfg.Tag = f.Image
```
dsimansk, line 142: Is it changing the output of si2 build right?

dsimansk:
There's a error in E2E tests even in unrelated runtimes, e.g. python. That seems like a genuine missing `namespace` value:

```
    cmd.go:128: exit status 1
    knative.go:35: services.serving.knative.dev "http-function-python-s2i" not found
    cmd.go:83: /home/runner/work/func/func/func delete --path /tmp/TestFunctionHttpTemplatepython_s2i_test187929058/001/http-function-python-s2i
    cmd.go:127: Error: namespace required
```

pkg/builders/s2i/builder.go
lkingland: The build config's `cfg.Tag` should be set to `f.Image` if it is defined (this indicates a request to build exactly this image using, for example `--image=` flag via the CLI.  If that is not defined, it should use `f.Build.Image`  (the last built image) if it exists, which is essentially rebuilding.  It seems like there might be a little missing logic here to choose one or the other?  Am I seeing this correctly @gauron99 ?

pkg/builders/s2i/builder.go
```diff
@@ -319,7 +319,7 @@ func (b *Builder) Build(ctx context.Context, f fn.Function, platforms []fn.Platf
 	}()
 
 	opts := types.ImageBuildOptions{
-		Tags:       []string{f.Build.Image},
+		Tags:       []string{f.Image},
```
lkingland, line 322: Same comment here, I presumed we'd have a check early in the function which uses `f.Image` if defined, defaulting to `f.Build.Image`, with an error if neither are defined.

pkg/builders/s2i/builder.go
dsimansk: Thanks for the explanation! 
So I'd say it's a change unrelated to image version update then. Probably easier to address in an isolated issue/PR? Unless it was trying to solve something else during testing with NodeJS 18 image @matejvasek?

gauron99: client builder should have set the `f.Build.Image` to whatever the final resolved image is supposed to be (using forced image via `f.Image` or resolved via registry name etc.), so Im pretty sure the `f.Build.Image` is correct here

gauron99:
> There's a error in E2E tests even in unrelated runtimes, e.g. python. That seems like a genuine missing namespace value:

> ```    cmd.go:128: exit status 1
    knative.go:35: services.serving.knative.dev "http-function-python-s2i" not found
    cmd.go:83: /home/runner/work/func/func/func delete --path /tmp/TestFunctionHttpTemplatepython_s2i_test187929058/001/http-function-python-s2i
    cmd.go:127: Error: namespace required```
  
  I think there is an error during deploy actually, previous to this error there is `Error: Error response from daemon: No such image: localhost:50000/user/cloudevents-function-python-s2i:latest` during the pushing phase looks like, therefore the function is not deployed and namespace not defined during removal (nothing no remove anyways)

dsimansk:
> How that image thingy even got here? I don't know what's right but IMO it belongs to separate PR.

Might be my not very well done rebase then?

lkingland:
> > How that image thingy even got here? I don't know what's right but IMO it belongs to separate PR.
> 
> Might be my not very well done rebase then?

It's very possible this is just a change which sneaked through a rebase, yes, since those lines were recently updated by @gauron99 to refer to `f.Build.Image`.  I think those two changes could be removed, and that might clear up the E2E.

PR #2112, opened 2024-01-10
fix: GitLab test

matejvasek:


jrangelramos:
/lgtm
.. as long as Gitlab test pass

PR #2117, opened 2024-01-15
chore: use Quarkus BP instead of explicit envvars

matejvasek:
The envvars do not have to be set manually anymore since Quarkus BP will set them for us.


matejvasek:
PTAL @lkingland @gauron99

PR #2121, opened 2024-01-18
Adding local field to function for handling transient spec (Local.Remote)

vyasgun:
Added a `Local` field to represent the local spec (such as `Remote`)

# Changes

- 🎁 Added a new field `Local.Remote` to the `function` spec
- 🎁 Modified `func.write` to write it only to `.func/local.yaml` so it is not persisted across clones but only locally
- 🎁 Added unit tests to ensure the expected behaviour

/kind enhancement

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/pkg/functions/function.go
+++ b/pkg/functions/function.go
@@ -22,8 +22,17 @@ const (
 
 	// RunDataDir holds transient runtime metadata
 	// By default it is excluded from source control.
-	RunDataDir = ".func"
+	RunDataDir       = ".func"
+	RunDataLocalFile = "local.yaml"
 )
+
+// Local represents the transient runtime metadata which
+// is only relevant to the local copy of the function
+type Local struct {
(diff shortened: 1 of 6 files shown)


pkg/functions/function.go
```diff
@@ -22,9 +22,18 @@ const (
 
 	// RunDataDir holds transient runtime metadata
 	// By default it is excluded from source control.
-	RunDataDir = ".func"
+	RunDataDir       = ".func"
+	RunDataLocalFile = "local.yaml"
 )
 
+// Local represents the transient runtime metadata which
+// is only relevant to the local copy of the function
```
lkingland, line 30: Good explanation

lkingland (approved):
This is looking very good.  Thanks for the contribution @vyasgun !

PR #2126, opened 2024-01-19
Springboot bumps

matzew:
# Changes

- :broom: Update JDK version to 21 and go with suggested newer versions.

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


matzew:
/assign @trisberg 
please review

/hold

templates/springboot/cloudevents/pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
-    <version>3.1.5</version>
+    <version>3.2.1</version>
```
trisberg, line 7: Just FYI, there is a brand new 3.2.2 release of Spring Boot now

templates/springboot/http/pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
-    <version>3.1.5</version>
+    <version>3.2.1</version>
```
trisberg, line 7: Just FYI, there is a brand new 3.2.2 release of Spring Boot now

templates/springboot/cloudevents/pom.xml
matejvasek: We might keep this to check if auto-updater changes this.

templates/springboot/http/pom.xml
matejvasek: We might keep this to check if auto-updater changes this.

PR #2132, opened 2024-01-23
fix: springboot version auto-updater

matejvasek:


matejvasek:
@matzew @lkingland PTAL

matzew (approved):
/lgtm
/approve

PR #2135, opened 2024-01-24
Remove toolchain directive

dsimansk:
# Changes

- Remove toolchain directive

We need to drop this at least temporarily until release is done, because one of our tools performing auto-update deps is not working with `toolchain` directive yet.

/cc @knative/knative-release-leads 
/cc @matejvasek @lkingland 

/kind <kind>

Fixes #

**Release Note**

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module knative.dev/func
 
 go 1.21
 
-toolchain go1.21.4
-
 // Use older version since it checks ~/.docker/config.json for credentials as opposed to later versions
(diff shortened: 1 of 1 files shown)


creydr (commented):
/lgtm

dsimansk:
> I thought that the directive was added by codegen 😄 .

Well, the chain looks like the following. Firstly, it's added through `go mod tidy` whenever go 1.21 module is pulled. `go mod tidy` is executed after `bouy` recognize new knative version thingy. Then our homegrown `bouy` tool stops actually working. I.e. silenty preventing any more auto updates.

PR #2153, opened 2024-02-09
chore: updated BP builder updater

matejvasek:
Newer version of pack does support adding custom labels to the builder so the Dockerfile workaround is not needed anymore.
Also the Dockerfile workaround was failing recently due to high number of layers in full builder.


matejvasek:

lkingland:
/lgtm
/approve
(Presuming checks complete....)

PR #2156, opened 2024-02-12
feat: host builder nonregular file support

lkingland:
# Changes

- :gift:  Host builder nonregular file support (ex: symlinks)

/kind enhancement

Fixes #

**Release Note**

```release-note
The prototype host builder now supports nonregular files such as symlinks
```

**Docs**

```docs

```


pkg/oci/containerize.go
```diff
@@ -150,27 +151,26 @@ func newDataTarball(source, target string, ignored []string, verbose bool) error
 			}
 		}
 
-		header, err := tar.FileInfoHeader(info, info.Name())
-		if err != nil {
-			return err
+		var link string
+		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
+			if link, err = os.Readlink(path); err != nil {
```
matejvasek, line 156: Could you check that the link does not escape project's root?

matejvasek:
Would it be possible to add some test?

pkg/oci/containerize.go
lkingland: Sure thing.  I will add a check

pkg/oci/containerize.go
```diff
@@ -150,27 +152,32 @@ func newDataTarball(source, target string, ignored []string, verbose bool) error
...
+				return err
+			}
+			ok, err := isWithin(source, link)
```
matejvasek, line 160: I believe this would only work for absolute links (eww). For relative you need to evaluate the link relative to the link file parent dir first.

Consider:
```
source
├── a_directory
│   └── a.lnk -> ../a.txt
└── a.txt

```
The link here does not escape source but I believe this could would evaluate it as escaping.

matejvasek: And this is why I dislike test on some auxiliary functions. There should be a test rather checking whole tar output.

matejvasek: We need something more like:
```go
if isRelative(link) {
    ok, err := isWithin(source, join(parentOf(path), link))
} else {
    ok, err := isWithin(source, link)
}
```

pkg/oci/containerize.go
```diff
@@ -150,27 +152,32 @@ func newDataTarball(source, target string, ignored []string, verbose bool) error
...
 		if err != nil {
 			return err
 		}
 
-		header.Name = slashpath.Join("/func", filepath.ToSlash(relPath))
-		// TODO: should we set file timestamps to the build start time of cfg.t?
-		// header.ModTime = timestampArgument
-
-		if err := tw.WriteHeader(header); err != nil {
+		header.Name = filepath.Join(baseDir, strings.TrimPrefix(path, source))
```
matejvasek, line 173: Why you use `filepath` instead of `slashpath` here?

matejvasek: And why you use `strings.TrimPrefix` instead of `filepath.Rel`?

matejvasek: And why is not path slashified by `filepath.ToSlash`?

pkg/oci/containerize.go
matejvasek: Also if we do allow absolute symliknks (eww). then you need to force relativise them.

pkg/oci/containerize.go
matejvasek: https://github.com/knative/func/actions/runs/7886751749/job/21524127711?pr=2156 -- this is not a flake.

matejvasek:
We already have some complex test here: https://github.com/knative/func/blob/24fe6d36fa07787df3962c40bb299a1aab8f441c/pkg/oci/builder_test.go#L30
Please consider adding symlink there.

pkg/oci/containerize.go
lkingland: Good thing we have the Windows test, sometimes forget the nuances of being Windows-friendy.  Yes, line 173 should have remained the same.

pkg/oci/containerize.go
lkingland: Thanks for the review!  I will resubmit something more compartmentalized and tested

> I believe this would only work for absolute links
The `isWithin` function does resolve the link to its absolute path, so it should work on symlinks.
However, I agree in principle and will add a higher-level test which actually uses symlinks as you suggested to make sure it works as expected.

> consider: 
> source
> ├── a_directory
> │   └── a.lnk -> ../a.txt
> └── a.txt
> The link here does not escape source but I believe this code would falsely evaluate it as escaping

This resolves to (for example if we're in `/var` ) `parent="/var/source/" child = "/var/source/a.txt".
When I tried this case with the included `isWithin` test cases, it does seem to correctly return `true`.
Again, my symlink-tar-fu is weak, so I will add a higher-level test and maybe package which uses actual links as you suggested.

> if we do allow absolute symlinks [..] relativise them

IIUC this is an error on the part of the user (path within a container will usually not match that of the host).  However, rather than alter what they are adding, I would prefer it simply not work as expected, giving them an error, with us at most issuing a warning.  I tend to prefer informing a user of incorrect input rather than trying to fix it for them.  However, I could be convinced otherwise if you feel strongly about rewriting a user's links.

As stated above, I will add a higher-level test which actually includes symlinks.  Are there no packages which handle this?  This is utter _minutia_ when all I want is a simple tarball of a directory with some sane defaults.

matejvasek: >This resolves to (for example if we're in /var ) parent="/var/source/" child = "/var/source/a.txt".

Does it though @lkingland ?

I believe that `child = ../a.txt` on input of`isWithin()` , then inside the `isWithin()` it is absolutizsd relative to __CWD__ (not parent of the link as it should).
So if we are in `/var` then `child = /a.txt` if we are in `/var/source` then `child = /var/a.txt` both is something we do not want?
Also this means that if you invoke `func` CLI outside project dir using `--path` flag you get even worse results.

matejvasek: You want something like
```diff
diff --git a/pkg/oci/containerize.go b/pkg/oci/containerize.go
index 0d812c42e..b7d8c3c6c 100644
--- a/pkg/oci/containerize.go
+++ b/pkg/oci/containerize.go
@@ -158,11 +158,16 @@ func newDataTarball(source, target string, ignored []string, verbose bool) error
 			if err != nil {
 				return err
 			}
-			ok, err := isWithin(source, link)
+			if filepath.IsAbs(link) {
+				return errors.New("absolute links are not permited")
+			}
+			lnkTgt := filepath.Join(filepath.Dir(path), link)
+			relLnkTgt, err := filepath.Rel(source, lnkTgt)
 			if err != nil {
-				return errors.Wrap(err, "error checking if link is outside function root")
-			} else if !ok {
-				return fmt.Errorf("links outside Function root disallowed.  Link %v references %v", path, link)
+				return fmt.Errorf("cannot evaluate relative path of link target: %w", err)
+			}
+			if strings.HasPrefix(relLnkTgt, ".."+string(filepath.Separator)) {
+				return errors.New("link escapes project's root")
 			}
 		}
 
```

pkg/oci/containerize.go
```diff
@@ -185,6 +197,27 @@ func newDataTarball(source, target string, ignored []string, verbose bool) error
...
+		return false, err
+	}
+	// Calculate the relative path between them
+	relPath, err := filepath.Rel(parent, child)
+	if err != nil {
+		return false, err
+	}
+	// Child is not within parent if the relative path from parent to child
+	// requires ..
+	return !strings.HasPrefix(relPath, ".."), nil
```
matejvasek, line 218: What is return value of `isWithin("/adir", "/adir/...weird_but_valid_name")`

pkg/oci/containerize.go
lkingland: Thanks for the suggestion, I've incorporated it into an update along with expanding the test as mentioned earlier

pkg/oci/containerize.go
```diff
@@ -185,6 +188,46 @@ func newDataTarball(source, target string, ignored []string, verbose bool) error
...
+	}
+
+	// Absolute links will not be correct when copied into the runtime
+	// container, because they are placed into path into '/func',
+	if filepath.IsAbs(tgt) {
+		return errors.New("project may not contain absolute links")
+	}
+
+	// Calculate the actual target of the link
+	// (relative to our current working directory)
```
matejvasek, line 214: relative to ~our current working directory~ partntal directory of the symlink

pkg/oci/containerize.go
```diff
@@ -150,27 +152,28 @@ func newDataTarball(source, target string, ignored []string, verbose bool) error
 			}
 		}
 
+		// Check for invalid links (absolute, outside of function root, etc)
+		if err := validateLink(root, path, info); err != nil {
+			return err
+		}
+
 		header, err := tar.FileInfoHeader(info, info.Name())
```
matejvasek, line 160: @lkingland  Shouldn't be the second argument of `tar.FileInfoHeader()` link target?
Again this would have been caught if testing was done at least at level of `newDataTarball()` or better yet [builder.Build()](https://github.com/knative/func/blob/24fe6d36fa07787df3962c40bb299a1aab8f441c/pkg/oci/builder_test.go#L30) level, not at level of some measly auxilary function.

pkg/oci/containerize_test.go
```diff
@@ -0,0 +1,43 @@
...
+	root := "testdata/test-links"
+
+	tests := []struct {
+		path  string // path of the file within test project root
+		valid bool   // If it should be considered valid
+		name  string // descriptive name of the test
+	}{
+		{"a.txt", true, "do not evaluate regular files"},
+		{"a.lnk", true, "do not evaluate directories"},
+		{"absoluteLink", false, "disallow absolute-path links"},
```
matejvasek, line 19: Hmm, it appears this is not working on Windows. I am not sure why. Maybe the link must start with a drive? Or maybe the link on Window cannot be absolute? Worst case: just skip this specific test on Windows.

pkg/oci/containerize_test.go
```diff
@@ -0,0 +1,43 @@
...
+func Test_validateLink(t *testing.T) {
+	root := "testdata/test-links"
+
+	tests := []struct {
+		path  string // path of the file within test project root
+		valid bool   // If it should be considered valid
+		name  string // descriptive name of the test
+	}{
+		{"a.txt", true, "do not evaluate regular files"},
+		{"a.lnk", true, "do not evaluate directories"},
```
matejvasek, line 18: Shouldn't be there a directory name, not link name?

pkg/oci/containerize_test.go
lkingland: I read through the docs and there is a note in there that even paths beginning with a slash are indeed considered relative in windows.  I'll try adding the drive-letter.

pkg/oci/containerize.go
lkingland: Sure; please see the latest addition.   As I mentioned above in our first set of comments, I agree that a high-level test should be added.  Now, I do think there also needs to be a case added which specifically looks for this condition.  That new test in `builder_test.go` is IMHO _in addition_ to the very targeted unit test here confirming the `validateLink` method works as expected.   More the merrier!

pkg/oci/builder_test.go
```diff
@@ -247,19 +313,13 @@ func validateOCI(path string, t *testing.T) {
...
-	if diff := cmp.Diff(expectedFiles, files); diff != "" {
+	if diff := cmp.Diff(expected, files); diff != "" {
 		t.Error("files in oci differ from expectation (-want, +got):", diff)
 	}
 }
+
+type fileInfo struct {
+	Path       string
+	Type       fs.FileMode
+	Executable bool
```
matejvasek, line 324: I guess it might be a time to add `Contetnt []byte` member (like [here](https://github.com/knative/func/blob/19509e5053a1ef41c485da2bb988cb7e49dc74d7/pkg/filesystem/filesystem_test.go#L34)). For regular files it would be a file content, for symlinks it would be symlink target. Maybe we could skip content for regular files for now.

pkg/oci/containerize_test.go
lkingland: Sorry that was just a typo while refactoring.

pkg/oci/builder_test.go
lkingland: Sounds good.  There's now a member for the header's `Linkname` member, which should verify the link was added correctly.

pkg/oci/containerize_test.go
```diff
@@ -0,0 +1,71 @@
...
+		path := "testdata/test-links/absoluteLinkWindows"
+		info, err := os.Lstat(path)
+		if err != nil {
+			t.Fatal(err)
+		}
+		_, err = validatedLinkTarget(root, path, info)
+		if err == nil {
+			t.Fatal("absolute path should be invalid on windows")
+		}
+	}
```
matejvasek, line 56: Couldn't this be part of the tests above?
`{"absoluteLink", false, "disallow absolute-path links"},` =>
`{windows ?  "absoluteLink" : "absoluteLinkWindows", false, "disallow absolute-path links"},`
OK Go does not have a ternary but you know what I mean.

lkingland: Yes, it could.  I left it separate such that OS-specific tests are separateed from shared.  Do you like it better together or separate?

pkg/oci/builder_test.go
```diff
@@ -46,7 +46,69 @@ func TestBuilder_Build(t *testing.T) {
...
+
+		if err := os.Symlink("a.txt", "a.lnk"); err != nil {
+			t.Fatal(err)
+		}
+	} else {
+		// Windows: create a copy
+		if err := os.WriteFile("a.lnk", []byte("file a"), 0644); err != nil {
+			t.Fatal(err)
+		}
+	}
```
matejvasek, line 89: Why is this needed? Doesn't GH Action run with privileged user?

pkg/oci/containerize_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	tests := []struct {
+		path     string   // path of the file within test project root
+		valid    bool     // If it should be considered valid
+		target   string   // optional test of the returned value (target)
+		name     string   // descriptive name of the test
+		runtimes []string // only apply this test to the given runtime(s)
+	}{
+		{"absoluteLink", false, "", "disallow absolute-path links on linux", []string{"linux"}},
+		{"absoluteLinkWindows", false, "", "disallow absolute-path links on windows", []string{"windows"}},
```
matejvasek, line 38: I am confused. I still do not understand why this is written like this.

Why not just have single test:
```go
absoluteLink := "absoluteLink"
if runtime.GOOS == "windows" {
    absoluteLink = "absoluteLinkWindows"
}
tests := []struct {
		path     string   // path of the file within test project root
		valid    bool     // If it should be considered valid
		target   string   // optional test of the returned value (target)
		name     string   // descriptive name of the test
	}{
		{absoluteLink, false, "", "disallow absolute-path links on linux"},
```

pkg/oci/builder_test.go
lkingland: I wanted tests to be runnable on windows hosts other than CI

pkg/oci/containerize_test.go
lkingland: Because this way we can simply list all the tests and optionally provide a set of platforms for which they apply.  With your example, there's a stack of if-statements above override tests implementaitons.

With this method, all tests are listed in the same way, and those with specific OS conditions can clearly state as such in their name.

This is stylistic preference favoring the simplicity of the test slice, not an architectural or logical error.

pkg/oci/containerize.go
```diff
@@ -185,6 +191,40 @@ func newDataTarball(source, target string, ignored []string, verbose bool) error
 	})
 }
 
+// validatedLinkTarget returns the target of a given link or an error if
+// that target is either absolute or outside the given project root.
+func validatedLinkTarget(root, path string, info os.FileInfo) (tgt string, err error) {
```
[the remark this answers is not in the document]
lkingland: Ah yes, now that we've fixed the link content being the target, this param is no longer used.  Now shouldn't a linter catch that!

pkg/oci/builder_test.go
matejvasek: But then you are not really testing symlink logic on Windows.

pkg/oci/containerize_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Test_validatedLinkTaarget ensures that the function disallows
+// links which are absolute or refer to targets outside the given root, in
+// addition to the basic job of returning the value of reading the link.
+func Test_validatedLinkTarget(t *testing.T) {
+	root := "testdata/test-links"
+
+	allRuntimes := []string{} // optional list of runtimes a test applies to
+
+	testAppliesToCurrentRuntime := func(testRuntimes []string) bool {
```
matejvasek, line 18: Couldn't we just use `slices.Contains()`.

pkg/oci/containerize_test.go
```diff
@@ -0,0 +1,73 @@
...
+		{"b/linkOutsideRootsParent", false, "", "disallow links outside project root and its parent", allRuntimes},
+		{"b/c/linkToParent", true, "", " allow links up, but within project", allRuntimes},
+		{"a.lnk", true, "./a.txt", "spot-check link target on linux", []string{"linux"}},
+		{"a.lnk", true, ".\\a.txt", "spot-check link target on windows ", []string{"windows"}},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if !testAppliesToCurrentRuntime(tt.runtimes) {
+				return // The test has a runtime filter defined
```
matejvasek, line 54: Wouldn't be more appropriate call `t.Skip()` ?

pkg/oci/builder_test.go
matejvasek: maybe use `if runtime.GOOS != "windows" && notAdmin {` or something like that.

pkg/oci/containerize_test.go
matejvasek: OK, personally do not find it more readable, but I won't argue about it.

pkg/oci/containerize.go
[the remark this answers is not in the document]
matejvasek: Linter is busy caching irrelevant stuff :smile:

pkg/oci/builder_test.go
matejvasek: Othrwise you may as well have
```go
if runtime.GOOS != "windows" {
    t.Skip()
}
```
On top of the test.

matejvasek: Ok that's little exaggerated, it still checks regular files/dirs.

matejvasek: I think I am okay with this.
Resolved.

pkg/oci/containerize_test.go
lkingland: Sure, you could.  That would have the nice effect of showing it as skipped in the test logs

pkg/oci/containerize_test.go
lkingland: Ah yes, I forgot this method exists.  I am so used to Go forcing things be done the long way.

PR #2182, opened 2024-02-26
Update gitignore for instanced CE module

matzew:
# Changes

- as per title

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /templates/typescript/cloudevents/build
 /templates/typescript/http/build
 /templates/go/cloudevents/go.sum
+/templates/go/.instanced-cloudevents/go.sum
 /coverage.out
 /coverage.txt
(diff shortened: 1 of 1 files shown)


lkingland (approved):

lkingland (changes requested):
I don't think we want to remove these, actually

.gitignore
```diff
@@ -5,6 +5,7 @@
 /templates/typescript/cloudevents/build
 /templates/typescript/http/build
 /templates/go/cloudevents/go.sum
+/templates/go/.instanced-cloudevents/go.sum
```
lkingland, line 8: Upon closer look, I should have said to just check in the sumfile.  Line 7 here will be removed when we replace the middleware with `func-go`

Issue #2185, opened 2024-02-27
gitlab test often fails when unrelated to the PR changes


gauron99:
gitlab test fails in PR actions for integration tests even when unrelated to the PR change at hand.

fails with timeouts

test is skipped here: https://github.com/knative/func/pull/2186 for now until its resolved

PR #2186, opened 2024-02-27
unreliable gitlab test skip

gauron99:
skip gitlab test because it often fails with unrelated issue to the PR changes

issue https://github.com/knative/func/issues/2185 is opened for tracking


lkingland (approved):
Putting this on the agenda for today's meeting

PR #2187, opened 2024-02-28
fix: test and namespace isolation

lkingland:
# Changes

- :broom: removes now-unnecessary Namespace member from structs
- :bug: several small bugfixes
- :broom: general cleanup

## tl;dr

Namespace as a struct member only works under fairly tight circumstances.  Recent functional additions invalidate this assumption.  This PR therefore removes namespace as a structure state member, adding it to method signatures where appropriate.  This keeps things flat and stupid-simple.

## Background

The initial design was to encapsulate the effective namespace within a fn.Client instance  (and concrete implementations of the constituent interfaces like Deployer, Lister, Remover etc) by providing it as a constructor argument.  All operations conducted via the client instance would then be targeted at that namespace.  Clients, being lightweight structures, can be instantiated easily, one for each namespace.  This allowed the target namespace to only be a concern during initial instantiation, with all following operations being able to ignore this internal state.  However, this assumption turned out to be incorrect for a couple of reasons:  cross-namespace operations and developer familiarity.

As new methods and options were added to the Client, there came to exist a few operations whose target namespace was either ambiguous or undefined.  For example deploying a function to a new namespace requires removing it from the old.  What, then, is the internal state of the namespace member?  This broke the encapsulation of that complexity.

Second, and perhaps more practical, is the expectation by users of the client instance (most of whom are at least familiar with the underlying Kubernetes implementation) expect namespace to be a primary concern.

These two conspired to result in several operations which expected namespace as a per-request argument.  This of course caused all logic to then treat the struct member as more of a fallback or default.

This is unnecessary complexity.  If a member like namespace can be completely internalized and removed from all method arguments, only being considered during instantiation, then it has succeeded in reducing complexy.  If this member "leaks" into any of the struct methods, then it is no longer assisting,but instead causing more complexity than just adding it to the set of expected arguments.

Therefore, this PR accepts that namespace is an expected argument in most cases, removing it from constructors and any state.

/kind bug
/kind cleanup

Fixes #

**Release Note**

```release-note
Fixed Function namespace resolution in some edge cases.
```

**Docs**

```docs

```


pkg/testing/testing.go
```diff
@@ -214,6 +214,45 @@ func RunGitServer(root string, t *testing.T) (url string) {
 	return "http://" + url
 }
 
+// FromTempDirectory moves the test into a new temporary directory and
+// clears all known interfereing environment variables.  Returned is the
+// path to the somewhat isolated test environment.
+// Note that KUBECONFIG is also set to testdata/default_kubeconfig whcih can
```
matzew, line 220: typo complaint

pkg/testing/testing.go
```diff
@@ -214,6 +214,45 @@ func RunGitServer(root string, t *testing.T) (url string) {
 	return "http://" + url
 }
 
+// FromTempDirectory moves the test into a new temporary directory and
+// clears all known interfereing environment variables.  Returned is the
```
matzew, line 218: typo complaint

gauron99 (commented):
lgtm after removing the left over "withDeployerNamespace" in tests and typos.
One less namespace-resolution complication :tada:

cmd/config_git_remove.go
```diff
@@ -25,7 +25,7 @@ func NewConfigGitRemoveCmd(newClient ClientFactory) *cobra.Command {
 	such as local generated Pipelines resources and any resources generated on the cluster.
 	`,
 		SuggestFor: []string{"rem", "rmeove", "del", "dle"},
-		PreRunE:    bindEnv("path", "namespace", "delete-local", "delete-cluster"),
+		PreRunE:    bindEnv("path", "delete-local", "delete-cluster"),
```
matzew, line 28: the ns comes from?

cmd/delete.go
```diff
@@ -28,7 +28,7 @@ No local files are deleted.
 {{rootCmdUse}} delete
 
 # Undeploy the function 'myfunc' in namespace 'apps'
-{{rootCmdUse}} delete -n apps myfunc
+{{rootCmdUse}} delete myfunc --namespace apps
```
matzew, line 31: does this need release notes? Mirgation? or the like?

matzew (approved):
/lgtm

/hold
for other reviewers

cmd/deploy_test.go
```diff
@@ -1795,103 +1795,103 @@ func TestReDeploy_OnRegistryChange(t *testing.T) {
...
+			// or its namespace had been manually removed
+			return fn.ErrFunctionNotFound
+		}
+		return nil
 	}
+	clientFn := NewTestClient(
+		fn.WithDeployer(mock.NewDeployer()),
+		fn.WithRemover(remover),
+		fn.WithPipelinesProvider(mock.NewPipelinesProvider()),
+	)
```
gauron99, line 1864: is the pipelines provider necessary here?

cmd/root_test.go
```diff
@@ -271,6 +271,75 @@ func TestRoot_effectivePath(t *testing.T) {
...
+	// if config.DefaultNamespace() != "default" {
+	// 	t.Fatalf("did not receive expected default namespace 'default', got '%v'", config.DefaultNamespace())
+	// }
+	//
+	// // should be "func" when active k8s namespace is "func"
+	// kubeconfig := filepath.Join(cwd, "testdata", "TestDefaultNamespace", "kubeconfig")
+	// t.Setenv("KUBECONFIG", kubeconfig)
+	// if config.DefaultNamespace() != "func" {
+	// 	t.Fatalf("expected default namespace of 'func' when that is the active k8s namespace.  Got '%v'", config.DefaultNamespace())
+	// }
```
gauron99, line 340: is this related to future to-be-added cases or can be deleted?

pkg/config/config_test.go
```diff
@@ -242,9 +221,9 @@ func TestApply(t *testing.T) {
 	if cfg.Language == "" {
 		t.Error("empty f.Runtime should not be mapped")
 	}
-	if cfg.Namespace == "" {
-		t.Error("empty f.Namespace should not be mapped")
-	}
+	// if cfg.Namespace == "" {
+	// 	t.Error("empty f.Namespace should not be mapped")
+	// }
 	if cfg.Registry == "" {
```
gauron99, line 227: to be deleted?

pkg/pipelines/tekton/pipelines_integration_test.go
```diff
@@ -19,81 +22,125 @@ import (
...
+
+	. "knative.dev/func/pkg/testing"
+)
+
+var testCP = func(_ context.Context, _ string) (docker.Credentials, error) {
+	return docker.Credentials{
+		Username: "",
+		Password: "",
+		// Username: "alice",
+		// Password: "alice-registry-token", Careful not to commit this.
```
gauron99, line 42: alice invasion commited?

gauron99 (commented):
amazing stuff! just a few comments

matzew:
@lkingland any updates?

cmd/config_git_remove.go
lkingland: This is just cleanup.  That command does not use namespace.  This command removes git settings from the function's metadata.  This is not `func delete` which actually removes the running service (and requires namespace)

cmd/delete.go
lkingland: These two commands are equivalent.  I was just making the documentation use the long-form name for (hopefully) better clarity.

cmd/deploy_test.go
lkingland: Good call.  No; that should not be necessary.  Let's give it a try with that removed...

cmd/root_test.go
lkingland: Should be able to be deleted.  My guess is it was test test code when putting together the above cases.  Thanks for noticing the cruft

lkingland:
> @lkingland any updates?

Could use another /LGTM presuming the tests pass!

PR #2189, opened 2024-02-28
chore: Update setup E2E to match latest Knative release

dsimansk:
# Changes

- :broom: Update or clean up current behavior

/cc @matejvasek 
/wip

Let's see if it works with current dependencies, or we need to bump those first. 

Fixes #

**Release Note**

```release-note
NONE
```

### DIFF
--- a/hack/allocate.sh
+++ b/hack/allocate.sh
@@ -25,13 +25,13 @@ export TERM="${TERM:-dumb}"
 
 main() {
 
-  local knative_serving_version=v1.10.1
-  local knative_eventing_version=v1.10.1
-  local contour_version=v1.10.0
+  local knative_serving_version=v1.13.0
+  local knative_eventing_version=v1.13.0
+  local contour_version=v1.13.0
 
   # Kubernetes Version node image per Kind releases (full hash is suggested):
   # https://github.com/kubernetes-sigs/kind/releases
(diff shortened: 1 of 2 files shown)


hack/allocate.sh
```diff
@@ -25,13 +25,13 @@ export TERM="${TERM:-dumb}"
 
 main() {
 
-  local knative_serving_version=v1.10.1
-  local knative_eventing_version=v1.10.1
-  local contour_version=v1.10.0
+  local knative_serving_version=v1.13.0
+  local knative_eventing_version=v1.13.0
+  local contour_version=v1.13.0
```
dsimansk, line 30: Not sure if there's any benefit to test with Contour as networking-layer for Serving. I'd switch to Kourier if possible. In a follow-up PR.

hack/allocate.sh
```diff
@@ -25,13 +25,13 @@ export TERM="${TERM:-dumb}"
...
 
   # Kubernetes Version node image per Kind releases (full hash is suggested):
   # https://github.com/kubernetes-sigs/kind/releases
-  local kind_node_version=v1.27.1@sha256:b7d12ed662b873bd8510879c1846e87c7e676a79fefc93e17b2a52989d3ff42b
+  local kind_node_version=v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822606ea5ca245
```
dsimansk, line 34: I assume it's locking kind on certain kube version. But we lock `kind` binary on certain release version anyway. I wonder if their default isn't enough. At least for current release it matches 1.29.2.

PR #2205, opened 2024-03-06
src: remove redundant workflows

lkingland:
# Changes

- :broom: removes redundant build and check workflows

The build and check workflows are now redundant with those imported from the knative tooling.  Removing them as a small part of an ongoing effort to clean up and speed up our CI

/kind cleanup


dsimansk (approved):
/approve
/lgtm

PR #2207, opened 2024-03-07
fix: Fix missing bash function in allocate.sh script

dsimansk:
# Changes

- fix: Fix missing bash function in allocate.sh script

/cc @lkingland 

**Release Note**

```release-note
NONE
```


lkingland:
Thanks for the quick fixup!

/lgtm
/approve

PR #2212, opened 2024-03-11
Drop replace directive

dprotaso:


dprotaso:
Looks like `Func E2E Test / E2E Test` is broken/flakey. It's unable to create/setup the cluster

lkingland (approved):
Looks good to me.
Very glad we can finally `go install`. 
/hold for you to PTAL @matejvasek, since this affects code you mostly wrote

PR #2234, opened 2024-03-16
Introduce a --registry-insecure flag

dprotaso:
- **fix env var casing otherwise flags with hyphens will continue to have hyphens**
- **Introduce an --registry-insecure flag**

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -169,6 +169,9 @@ func bindEnv(flags ...string) bindFunc {
 				return
 			}
 		}
+		viper.AutomaticEnv()       // read in environment variables for FUNC_<flag>
+		viper.SetEnvPrefix("func") // ensure that all have the prefix
+		viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 		return
 	}
 }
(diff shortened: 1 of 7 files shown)


dprotaso:
/hold

Deploy is still broken.

Tthough I'm noticing 'insecure' in oci.Pusher code is being confused with `anonymous` user so I'm unsure what's best to do here

dprotaso:
All good now - anon is still disabled by default (like it was before).

Insecure flag changes the registry scheme

cmd/root.go
```diff
@@ -58,6 +58,7 @@ Learn more about Knative at: https://knative.dev`, cfg.Name),
 	// a version prefixed by "FUNC_"
 	viper.AutomaticEnv()       // read in environment variables for FUNC_<flag>
 	viper.SetEnvPrefix("func") // ensure that all have the prefix
+	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
```
lkingland, line 61: Good stuff!  I am surprised this isn't the default behavior

cmd/root.go
```diff
@@ -170,6 +171,7 @@ func bindEnv(flags ...string) bindFunc {
 		}
 		viper.AutomaticEnv()       // read in environment variables for FUNC_<flag>
 		viper.SetEnvPrefix("func") // ensure that all have the prefix
+		viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
```
lkingland, line 174: And this duplicate directive is another reason we have a flag refactor/cleanup in the works

lkingland (approved):

PR #2236, opened 2024-03-18
Run func without HOME defined/ unaccessible .config dir

### LINKED ISSUE #2150: Do not fail if ~/.config is not writable
If `HOME` is not set and/or is not writable the func CLI is not really usable.
At least some basic functionality should work without `~/.config/func`  existing.

gauron99:
# Changes
Use func without the need for HOME definition or accessible ~/.config with slightly limited functionality (like docker creds or local templates/repos or the pack builder)

When running `func deploy` without `HOME` or `XDG_CONFIG_HOME` specified, deployment is possible if the correct credentials are provided via user prompt (if needed). Additionally, pack builder requires some HOME path for its dot directory where it stores data hence it needs HOME to work. S2i builder is non-problematic. 

/kind enhancement
/kind bug

Fixes #2150

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -159,9 +159,6 @@ func runBuild(cmd *cobra.Command, _ []string, newClient ClientFactory) (err erro
 		cfg buildConfig
 		f   fn.Function
 	)
-	if err = config.CreatePaths(); err != nil { // for possible auth.json usage
-		return
-	}
 	if cfg, err = newBuildConfig().Prompt(); err != nil { // gather values into a single instruction set
 		return
 	}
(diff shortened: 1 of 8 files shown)


cmd/build.go
```diff
@@ -159,9 +159,6 @@ func runBuild(cmd *cobra.Command, _ []string, newClient ClientFactory) (err erro
 		cfg buildConfig
 		f   fn.Function
 	)
-	if err = config.CreatePaths(); err != nil { // for possible auth.json usage
-		return
-	}
```
lkingland, line 164: Glad to see these indeed are no longer necessary!

lkingland (approved):
🎉

PR #2242, opened 2024-03-21
Basic auth

lkingland:
This PR adds HTTP basic auth via CLI arguments to the "host" builder.

Because the Host builder is still behind a feature flag as it is experimental, the newly added flags are currently hidden.

The purpose of this addition is to enable a common use-case of using a third-party credentials management system rather than relying on those provided by the respective registries (docker credentials manager, AWS, etc) or the system.  This is especially useful when considering that the "host" builder does not require docker nor podman be installed.

For example the venerable `pass` command line tool, or the recent Bitwarden, 1Password, etc.

## Simple Example

Here is an example of how this works using the Bitwarden `bw` command, which will prompt the user for the repository master password on each deploy.

```
func deploy --username=func --password=$(bw get password registry.example.com)
```

This workflow is roughly the same for the CLI password management apps.

## Caching Credentials Example

While the following is Bitwarden-specific, it may serve to be illustrative on how this paradigm can be used to cache credentials.  The above command will prompt the user interactively for the master password.  To "cache" the credentials one would us `bw unlock` followed by providing the resultant session key to the `bw get` command.

```
bw unlock --raw > ~/.bw-session
alias bw='bw --session=$(<~/.bw-session)'
```
Now the same deploy command will have credentials cached:
```
func deploy --username=func --password=$(bw  get password registry.example.com)
```

## Future Work

There is a small amount of overlap between the docker pusher's credentials management and this implemention.  I would hope to factor out some of these primitives into a shared `creds` package or similar.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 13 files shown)


matzew:
close/reopen, as the CI seemed to be stuck in nirvana

PR #2245, opened 2024-03-25
make was not called #2234

matzew:
# Changes

- the `make` seems to be forgotten often, running it after 2234 PR

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/docs/reference/func_build.md
+++ b/docs/reference/func_build.md
@@ -9,9 +9,10 @@ NAME
 	func build - Build a function container locally without deploying
 
 SYNOPSIS
-	func build [-r|--registry] [--builder] [--builder-image] [--push]
+	func build [-r|--registry] [--builder] [--builder-image]
+		         [--push] [--username] [--password] [--token]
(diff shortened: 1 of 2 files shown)


matzew:
@lkingland @matejvasek  can we make this more intuitive ? Either by a CI linter, or something?

lkingland:
> @lkingland @matejvasek can we make this more intuitive ? Either by a CI linter, or something?

Yes, we can, and we should 👍🏻   A little background:

The actual root problem is that the task which updates `docs` is not properly clearing its environment, so if (like me) you have `FUNC_ENABLE_HOST_BUILDER=true`, the docs are all updated on `make` to include the "host" builder.  This means I am perpetually confronted with unstaged changes in `./docs`, and forget that sometimes the changes are valid.

The ideal fix here is to 1) fix the docs generator to clear its environment before generating docs and then 2) have a CI task which (like we do for `zz_filesystem_generated`) fail in CI if there are unstanged changes in ./docs after building.

lkingland (approved):
/lgtm

PR #2254, opened 2024-03-27
Fixing Rust version to 1.76.0

matejvasek:
The version 1.77.0 is not compatible with paketo buildpack.

* Updated version of Rust buildpack 0.35.0 => 0.45.0, version 0.45.0 includes Rust 1.76.0.

* Disabling rustup buildpack so latest Rust (1.77.0) is not used, instead Rust 1.76.0 from rust-dist buildpack is used.


matzew (approved):
/lgtm
/approve

PR #2256, opened 2024-03-27
func invoke handles json properly

### LINKED ISSUE #2241: Func invoke encodes json data differently than curl
When sending events with json data with `func invoke`, the way in which the data is encoded in the event behaves differently than when using curl to call the function, which is confusing.

To reproduce:

1. Clone https://github.com/Cali0707/func-invoke-confusion
2. Run the function with `func run`
3. Send an event to the function with curl:
```bash
curl -X POST http://localhost:8080/ \
 -H "ce-id: 12345" \
 -H "ce-source: curl" \
 -H "ce-type: this.will.work" \
 -H "ce-specversion: 1.0" \
 -H "Content-Type: application/json" \
 -d '{"customerId": "0123456", "productId": "6543210"}'
```
4. Send an event to the function with `func invoke`:
```bash
func invoke --data '{"customerId": "0123456", "productId": "6543210"}'
```
5. Check the logs from the function, they will look like:
```
listening on http port 8080
Running on host port 8080
Received event
Context Attributes,
  specversion: 1.0
  type: this.will.work
  source: curl
  id: 12345
  datacontenttype: application/json
Data,
  {
    "customerId": "0123456",
    "productId": "6543210"
  }

Successfully decoded purchase: &{CustomerId:0123456 ProductId:6543210}Received event
Context Attributes,
  specversion: 1.0
  type: boson.fn
  source: /boson/fn
  id: 0385ddeb-043a-472d-93a9-e7b281ba319a
  time: 2024-03-20T20:53:50.985375048Z
  datacontenttype: application/json
Data,
  "{\"customerId\": \"0123456\", \"productId\": \"6543210\"}"

Failed to decode event:  [json] found bytes ""{\"customerId\": \"0123456\", \"productId\": \"6543210\"}"", but failed to unmarshal: json: cannot unmarshal string into Go value of type function.Purchase
```

I'm not sure if this is the expected behaviour or not, but if it is it might be worth mentioning this somewhere, as I expected that it would behave the same way as curl and send json, rather than a json encoded string.

Cali0707:
# Changes

- :bug: func invoke sets the data in the event after unmarshalling the string, so that the event does not treat the data as a json string
-
-

/kind bug

Fixes #2241 

**Release Note**

```release-note
func invoke now correctly handles data with contenttype application/json
```

### DIFF
--- a/pkg/functions/invoke.go
+++ b/pkg/functions/invoke.go
@@ -3,6 +3,7 @@ package functions
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -152,7 +153,17 @@ func sendEvent(ctx context.Context, route string, m InvokeMessage, t http.RoundT
 	event.SetID(m.ID)
 	event.SetSource(m.Source)
 	event.SetType(m.Type)
-	if err = event.SetData(m.ContentType, m.Data); err != nil {
+	if m.ContentType == "application/json" {
+		var d interface{}
+		err = json.Unmarshal([]byte(m.Data), &d)
+		if err != nil {
+			return
+		}
+		err = event.SetData(m.ContentType, d)
+		if err != nil {
+			return
+		}
+	} else if err = event.SetData(m.ContentType, m.Data); err != nil {
 		return
 	}
 


Cali0707:
/cc @lkingland @matejvasek 

Let me know if you think it would be better to make this work more generically for other contenttypes/if you have an idea how to do that. For now, this fix is specific to application/json

Leo6Leo:
I can testify that the fix works on my end.

lkingland (approved):
Thank you for the contribution @Cali0707

/lgtm
/approve

pierDipi:
@lkingland @Cali0707 should this be backported for a patch release?

lkingland:
> @lkingland @Cali0707 should this be backported for a patch release?

Thanks for asking.

This is a feature/enhancement, so I see no need to backport.

PR #2258, opened 2024-03-28
Fix utils image build

matejvasek:


lkingland (approved):
/lgtm

(retried failed tests)

PR #2261, opened 2024-04-02
Update deps

matejvasek:


dsimansk:
@matejvasek I'll do a follow-up to update `client-pkg` there's a new import path now. That breaks/blocks auto-update PRs.

PR #2270, opened 2024-04-10
Clean up e2e oncluster test

jrangelramos:
# Changes

- :broom: Clean up e2e oncluster test

/kind cleanup


matejvasek:
Looks good to me, just a question could be `GetGitServer()` called by multiple goroutine at once?

matejvasek:
Nevermind. It probably does not matter.

PR #2277, opened 2024-04-19
chore: make tkn taks urls setable via linker flags

matejvasek:


lkingland (approved):

matejvasek:
@ReToCode @dsimansk would you please override the three skipped style checks?

PR #2284, opened 2024-04-30
Use image index not just image

matejvasek:
* Use image index aka manifest list instead of just plain image. This makes our image "multi-arch" with single architecture. It may seems weird but on multi-arch cluster this helps cluster to allocate pods on node with matching architecture.
* Refactored `Pusher` tests -- merged two tests into one table driven test.
* Pusher now always tries "daemon push" instead of guessing whether "daemon push" is viable by calling `net.DefaultResolver.LookupHos()`.


matejvasek:

lkingland (approved):
Nice!

PR #2289, opened 2024-05-06
chore: test suite refactor

lkingland:
This PR provides the general structure for an upgrade to the testing system (both integration and E2Es) relying on the newly created Host builder for core functionalit, along with general simplification.

Currently a WIP.

/kind cleanup


.github/workflows/ci.yaml
```diff
@@ -10,36 +10,16 @@ jobs:
...
+      - name: Lint Core
+        run: make check
+      - name: Lint Templates
+        run: make check-templates
+      - name: Check 'func.yaml' schema
+        run: make check-schema
+      - name: Check embedded FS
+        run: make check-embedded-fs
 
-  test-unit:
```
matejvasek, line 20: I would not recommend deleting these.

.github/workflows/test-templates.yaml
```diff
@@ -1,10 +1,10 @@
-name: Func Unit Test
```
matejvasek, line 1: I do not get this rename. This is foremost "unit test". It also runs "template test" but that is IMO secondary.

.github/workflows/ci.yaml
matejvasek: Will `make test-all` run really all tests? What about coverage have you ensured that there are no conflicts?

Makefile
```diff
@@ -196,23 +210,25 @@ templates/certs/ca-certificates.crt:
...
+	go test -ldflags "$(LDFLAGS)" -cover -tags "integration" -timeout 30m --coverprofile=coverage.txt ./...
 
-.PHONY: func-instrumented
+.PHONY: test-e2e
+test-e2e: func-instrumented ## Run unit+E2E tests (cluster required)
+	go clean -testcache
+	go test -ldflags "$(LDFLAGS)" -cover -tags "e2e" -timeout 30m --coverprofile=coverage.txt ./...
+
+.PHONY: test-e2e
+test-all: func-instrumented ## Run unit+integration+E2E tests (cluster required)
```
matejvasek, line 224: Does this properly gather coverage from the instrumented func?

Makefile
```diff
@@ -196,23 +210,25 @@ templates/certs/ca-certificates.crt:
...
+	go clean -testcache
+	go test -ldflags "$(LDFLAGS)" -cover -tags "integration e2e" -timeout 60m --coverprofile=coverage.txt ./...
```
matejvasek, line 226: does not this miss "on cluster build" tag?

PR #2293, opened 2024-05-07
Remove axios dependency 

matejvasek:
# Changes

- Removed dependency on `axios` by using `fetch()`.
- Fixed `getLatestPlatform()` by using the `tag_name` property instead of the `name` property.


matejvasek:
@trisberg  Why the never Spring Boot works only with Java 17 not 21?

trisberg:
> @trisberg Why the newer Spring Boot works only with Java 17 not 21?

It should work with Java 21, I use that all the time. What is the error and the failing config?

lkingland (approved):

matejvasek:
PTAL @lkingland

matejvasek:
I have idea why the coverage fails. @dsimansk @ReToCode please override it.

matejvasek:
Why the **** is there `codecov/project` twice?

matejvasek:
I just created completely new PR.

PR #2295, opened 2024-05-11
fix: pin octokit npm module version to 3

matejvasek:
The newer version of `octokit` (v4) uses some different module type and `require()` does not work nice with it.

This is required to fix currently failing periodic quarkus/sprint-boot version update.


matejvasek:
PTAL @lance

matzew (approved):
/lgtm
/approve

matejvasek:
@dsimansk would you please override `codecov/project`, it seems to be stuck for some reason.

dsimansk:
@matejvasek did a quick dirty close/open.

PR #2299, opened 2024-05-13
chore: ignore test dir in coverage

matejvasek:


matejvasek:
Nevermind it's already ignored. Or should be.

PR #2301, opened 2024-05-13
chore: upload coverage of unittests together

matejvasek:
Upload results of unit tests on various operating systems at once as opposed to uploading them individually.

### DIFF
--- a/.github/workflows/test-unit.yaml
+++ b/.github/workflows/test-unit.yaml
@@ -41,7 +41,7 @@ jobs:
         with:
           name: coverage-${{ matrix.os }}
           path: ./coverage.txt
-          retention-days: 7
+          retention-days: 1
   upload-coverage:
     needs: [test]
     name: "Upload coverage"
--- a/.github/workflows/test-unit.yaml
+++ b/.github/workflows/test-unit.yaml
@@ -36,7 +36,23 @@ jobs:
           FUNC_REPO_BRANCH_REF: ${{ github.head_ref }}
       - name: Template Unit Tests
         run: make test-templates
+      - name: "Archive code coverage results"
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-${{ matrix.os }}
+          path: ./coverage.txt
+          retention-days: 7
+  upload-coverage:
+    needs: [test]
+    name: "Upload coverage"
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: Download Coverage
+        run: |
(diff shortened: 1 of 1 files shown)


matejvasek:
PTAL @dsimansk

.github/workflows/test-unit.yaml
```diff
@@ -36,7 +36,23 @@ jobs:
           FUNC_REPO_BRANCH_REF: ${{ github.head_ref }}
       - name: Template Unit Tests
         run: make test-templates
+      - name: "Archive code coverage results"
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-${{ matrix.os }}
+          path: ./coverage.txt
+          retention-days: 7
```
dsimansk, line 44: It seems a bit long. I'd go with hours, e.g. 8h given it's per PR artifact and might iterate quickly.

matejvasek: Would 7 days hurt anything?

matejvasek: Also minimum is probably 1 day.

dsimansk: not sure who pays for storage :)

I'm not really, it's very small file. But 1 day is plenty.

matejvasek: >not sure who pays for storage :)

Uncle Bill?

dsimansk (approved):

dsimansk:
@matejvasek wrt/ Codecov ~6% drop. Isn't unified uploader missing some reports?

matejvasek:
> @matejvasek wrt/ Codecov ~6% drop. Isn't unified uploader missing some reports?
> 
> Update, actually it's stated in the dashboard. Hence different count of reports shouldn't be a problem, if we have same content, right?

Yes, there should be less reports since three reports were merged into one. I have no idea why the coverage decreased.

dsimansk:
> > @matejvasek wrt/ Codecov ~6% drop. Isn't unified uploader missing some reports?
> > Update, actually it's stated in the dashboard. Hence different count of reports shouldn't be a problem, if we have same content, right?
> 
> Yes, there should be less reports since three reports were merged into one. I have no idea why the coverage decreased.

Strangely, it reports head 14 uploads vs. now 6 uploads. It's really off. 
Anyway, I'd suggest to keep monitoring the PR wrt/ codecov API rate limiting.  WDYT?

dsimansk (approved):

PR #2337, opened 2024-06-04
Update pack dependency

matejvasek:
* Updated pack dependency


matejvasek:
PTAL @dsimansk @lkingland @gauron99

lkingland (approved):

go.mod
```diff
@@ -1,36 +1,38 @@
 module knative.dev/func
 
-go 1.21
+go 1.22
```
dsimansk, line 3: Do we really need to jump on 1.22? Is it convenience or forced by the dependencies?

matejvasek: `go mod tidy` did this.

go.mod
```diff
@@ -1,36 +1,38 @@
...
+
+replace github.com/openshift/source-to-image => github.com/matejvasek/source-to-image v1.3.2-0.20240604071927-a5ace19e3f71
```
dsimansk, line 5: Can we wait for the `source-to-image` PR is merged if it makes sense timewise?

go.mod
dsimansk: Well, it's min version of required Go APIs used, not necessarily go-toolchain used for building it. It may create a downstream issues, because of slow Golang updates in distros like RHEL.

dsimansk: > `go mod tidy` did this.

`go mod tidy` is accurate to bump min go, usually. I guess some of the modules now declares itself as `1.22` then.

go.mod
matejvasek: fixed

matejvasek:
PTAL @dsimansk @lkingland @gauron99

go.mod
matejvasek: @dsimansk also I cannot even downgrade it manually, I believe `./hack/verify-codegen.sh` would failed.

PR #2340, opened 2024-06-05
chore: update tekton,PaC

matejvasek:


go.mod
```diff
@@ -1,17 +1,19 @@
 module knative.dev/func
 
-go 1.22
+go 1.22.0
```
matejvasek, line 3: @dsimansk  Now `./hack/update-deps.sh` did this. I have no idea why.

go.mod
```diff
@@ -1,17 +1,19 @@
...
+
+replace k8s.io/client-go => k8s.io/client-go v0.30.1
```
matejvasek, line 5: @dsimansk for some reason I had to pin, k8s.io/client-go otherwise I got errors like:
```
go: knative.dev/func/pkg/k8s imports
	k8s.io/client-go/dynamic: package k8s.io/client-go/dynamic provided by k8s.io/client-go at latest version v0.30.1 but not at required version v1.5.2
go: knative.dev/func/pkg/k8s imports
	k8s.io/client-go/kubernetes: package k8s.io/client-go/kubernetes provided by k8s.io/client-go at latest version v0.30.1 but not at required version v1.5.2
go: knative.dev/func/pkg/k8s imports
	k8s.io/client-go/kubernetes/scheme: package k8s.io/client-go/kubernetes/scheme provided by k8s.io/client-go at latest version v0.30.1 but not at required version v1.5.2
go: knative.dev/func/pkg/k8s imports
```

go.mod
dsimansk: It'll some default behavior change / more strict rules from Go 1.22, I guess.

go.mod
dsimansk: Tekton CLI seems to be fine: https://github.com/tektoncd/cli/blob/release-v0.37.0/go.mod#L41.

Is it some other package then?

matejvasek: Its github.com/openshift-pipelines/pipelines-as-code, I added also comment about it above the replace.

matzew (approved):
/lgtm
/approve

lkingland (approved):

PR #2341, opened 2024-06-06
fix: PaC build

matejvasek:
* Fixed the issues where PaC configuration failed under unprivileged user.
```release-note
fix: Pipelines as Code setup failure under unprivileged user
```


lkingland (approved):

matejvasek:
The `Integration Test (ubuntu-latest)` test failed several times because of low disk space. I don't think it's related to this PR.

@dsimansk please override it

PR #2342, opened 2024-06-07
fix: error when setting gitlab settings in test

matejvasek:
* Fixed the issue where gitlab setting modification returned error for no good reason. It appears that gitlab returns field `container_registry_import_created_before` as an empty string `""` and JSON deserializer rejects such a value for date/time. I fixed this by injecting custom `RoundTripper` that removes that field in case it's has value of empty string.
* Remove hardcoded sleep in favor of active condition checking.

Related:
https://github.com/xanzy/go-gitlab/issues/1952
https://gitlab.com/gitlab-org/gitlab/-/issues/466132

### DIFF
--- a/pkg/pipelines/tekton/gitlab_test.go
+++ b/pkg/pipelines/tekton/gitlab_test.go
@@ -4,6 +4,7 @@
 package tekton_test
 
 import (
+	"bytes"
 	"context"
 	"crypto/ecdsa"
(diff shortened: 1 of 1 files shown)


matejvasek:

lkingland:
Just...  wow
#facepalm

lkingland (approved):

Issue #2347, opened 2024-06-07
Using Docker Desktop socket causes bind-mount to fail


vpsx:
I am trying to follow the Knative Quickstart tutorial and do `func run` or `func build`, but func seems to have trouble connecting with the docker daemon: I am running Docker Desktop for Linux, and depending on where I point `DOCKER_HOST`, things will fail at a different stage. The default `/var/run/docker.sock` does not work because there is no actual socket there (on the host), so it can't pull the builder; on the other hand, only `/var/run/docker.sock` can be bind-mounted in a Docker Desktop setup [(I believe)](https://github.com/docker/docs/issues/20143), so setting `DOCKER_HOST` to the DD socket `unix:///home/vpsx/.docker/desktop/docker.sock` will cause "connection refused" later, in the lifecycle. 

```
~% docker context ls
NAME                TYPE                DESCRIPTION                               DOCKER ENDPOINT                                 KUBERNETES ENDPOINT   ORCHESTRATOR
default             moby                Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                                           
desktop-linux *     moby                Docker Desktop                            unix:///home/vpsx/.docker/desktop/docker.sock    
```

```
% unset DOCKER_HOST
% func run
Building function image
Error: cannot create docker client: docker/podman API not available
Possible causes:
  The docker/podman daemon is not running.
  The DOCKER_HOST environment variable is not set.
```
```
% export DOCKER_HOST=unix:///var/run/docker.sock 
% func run                                      
Building function image

2024/06/07 13:03:27.928105 DEBUG:  Pulling image ghcr.io/knative/builder-jammy-base:0.4.283

Error: failed to build the function: failed to fetch builder image 'ghcr.io/knative/builder-jammy-base:0.4.283': Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

```
% export DOCKER_HOST=unix:///home/vpsx/.docker/desktop/docker.sock

% func run                                                                                                        
Building function image                                                                                                        
                                                                                                                               
2024/06/07 13:05:23.079035 DEBUG:  Pulling image ghcr.io/knative/builder-jammy-base:0.4.283
0.4.283: Pulling from knative/builder-jammy-base                                                                               
Digest: sha256:ffe1df595b4a4dc334f03a55fa8458cfdbca7bc495b4a97c69410cdfc2d17cd5     
Status: Image is up to date for ghcr.io/knative/builder-jammy-base:0.4.283  
[...]

PR #2350, opened 2024-06-10
fix: mount of Docker API socket from Docker Desktop

matejvasek:
* Fixed mountpoint of docker socket for `Docker Desktop`.

fixes: #2347

```release-note
fix: buildpack build failure caused by wrong socket mount-point when using Docker Desktop
```

### DIFF
--- a/pkg/docker/docker_client.go
+++ b/pkg/docker/docker_client.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/docker/cli/cli/config"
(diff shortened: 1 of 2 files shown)


matejvasek:
PTAL @lkingland @gauron99

vpsx:
I just tried it; it works for me! (Docker Desktop for Linux; Ubuntu)

(Not sure if worth mentioning, but just in case: DOCKER_HOST does still need to be explicitly set to the Docker Desktop socket for the first part/pulling builder image to work; it does not look at the docker context)

Thanks!!

lkingland (approved):

matejvasek:
PTAL @lkingland

matejvasek:
PTAL @matzew @gauron99

gauron99 (approved):

PR #2356, opened 2024-06-12
test: increased wait period

matejvasek:
Increased wait period for Gitlab settings change to kick in.


matejvasek:
PTAL @matzew @gauron99

gauron99 (approved):

PR #2359, opened 2024-06-12
Replace container to due errors

### LINKED ISSUE #2358: Local registry (re)creation issues
Running the `allocate` script, gets me this for the registry:

```
reg Creating Registry
reg Error: creating container storage: the container name "func-registry" is already in use by 07506ff0de0c74f53cc26786815d874274b59b042e04dc083426fc172855bac9. You have to remove that container to be able to reuse that name: that name is already in use, or use --replace to instruct Podman to do so.
```

matzew:
# Changes

Fixes #2358 

- Replace container to due errors

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


matejvasek:
I think it might be better to just check if the container already exist and if it does then skip creation.

matejvasek:
ping @matzew

PR #2361, opened 2024-06-13
fix: HOME with different permissions

gauron99:
Make sure standard functionality works even when HOME has different permissions
- added new tests

/kind enhancement

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -84,9 +84,11 @@ func NewDefault() (cfg Global, err error) {
 		// config file is not required
 		if os.IsNotExist(err) {
 			err = nil
-			// } else if os.IsPermission(err) {
-			// 	fmt.Printf("Warning: Insufficient permissions to read config file at '%s' - continuing without it\n", cp)
-			// err = nil // HOME perms are insufficient for config file -> warning
+		} else if os.IsPermission(err) {
+			err = nil // insufficient perms for config file -> warning
+			// TODO: gauron99 - this gets printed out for lot of commands so its a bit
+			// spammy, perhaps dont print here?
+			fmt.Fprintf(os.Stderr, "Warning: Insufficient permissions to read config file at '%s' - continuing without it\n", cp)
 		}
 		return
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -81,8 +81,12 @@ func NewDefault() (cfg Global, err error) {
 	cp := File()
 	bb, err := os.ReadFile(cp)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/config/config.go
```diff
@@ -81,8 +81,14 @@ func NewDefault() (cfg Global, err error) {
...
+		// config file is not required
 		if os.IsNotExist(err) {
-			err = nil // config file is not required
+			err = nil
+		} else if os.IsPermission(err) {
+			err = nil // insufficient perms for config file -> warning
+			// TODO: gauron99 - this gets printed out for lot of commands so its a bit
+			// spammy, perhaps dont print here?
+			fmt.Fprintf(os.Stderr, "Warning: Insufficient permissions to read config file at '%s' - continuing without it\n", cp)
 		}
```
gauron99, line 92: this branch is new addition of this PR -- allowing to continue when permission error occurs for the func config file. This results in a printing of the newly added warning. It gets printed out many times during init when the program goes trough init phase (since most commands call  function `NewDefault()`). Could be a bit spammy for users but removing it might not give enough information. Just something to consider from my self-review.

lkingland: I agree with your logic here, but printing this warning multiple times to stderr is going to be quite annoying, and will not inspire confidence.  Maybe we can do something half-way:

The fact this situation even exists leads me to believe we have an opportunity to improve the logical flow of how the system initializes its config.  Perhaps a refactor is too deep for this PR, so we could do a shorter workaround: either the `cmd` package's `root` (or perhaps somewhere else) could perform a single early check to check this condition and print a warning to the user that the process will continue as though there were no config file due to insufficient permission to read path X.  Then here, we would squelch the error as being a nonfatal condition.

If we really wanted to be thorough with that "half baked" temporary solution, we'd give the config package a verbosity toggling flag (defaulted to off).  Then we'd print warnings to log if verbose.

Ideally, we'd refactor such that this check need not be run more than once for any given instance of the client, but that's a larger task I think.

lkingland (approved):
/lgtm

/hold for potential squelching of the permission error

pkg/config/config.go
gauron99: Sounds good to me. I will look for a simple workaround for this PR but keep the larger issue at hand looking to follow up or discuss with more details.

gauron99:
The main unit test is failing because "knative" user (GH actions/ something similar) has greater permissions, I suspect. Will have to update the new test.

gauron99:
@lkingland PTAL

lkingland (approved):
Looking good to me!

PR #2363, opened 2024-06-13
test: revert error suppression

matejvasek:


matejvasek:
PTAL @gauron99 @jrangelramos

gauron99 (approved):

Issue #2331, opened 2024-05-29
Regarding the issues of compilation and deployment. knative 1.14


Taylorzhao:
When deploying with func deploy, there is a func.yaml file under the directory, and parameters such as the deployed image have already been described in the file. However, when deploying, it does not read the image in the func.yaml, but reads the image of build-image under.func. When I delete the.func directory under the directory, the deployment will report an error, as shown in the figure.
<img width="1130" alt="image" src="https://github.com/knative/func/assets/13677508/06494414-2e32-42d6-9feb-e32c2674fa07">
However, in versions 1.12 and earlier of Knative, this issue did not exist.
A simple deployment method should only require one yaml file to be sufficient.

my fun.yaml.
```yaml
specVersion: 0.36.0
name: instantiate
runtime: go
registry: 192.168.100.167:80
image: 192.168.100.167:80/serverless/flow-model-gvcbkl-instantiate:v3
namespace: flow-model-gvcbkl
created: 2024-05-28T10:39:25.562440871Z
build:
  builder: pack
run:
  envs:
  - name: FUNC_ID
    value: 5fd83203-aeaf-48c8-88d2-7fa3fc07aa90
  - name: FUNC_TYPE
    value: http
  - name: FUNC_NAME
    value: instantiate
deploy:
  namespace: flow-model-gvcbkl
  image: 192.168.100.167:80/serverless/flow-model-gvcbkl-instantiate:v3
  options:
    scale:
      min: 1
      max: 5
      metric: concurrency
      target: 75
      utilization: 75
    resources:
      requests:
        cpu: 10m
        memory: 128Mi
      limits:
        cpu: 100m
        memory: 256Mi
        concurrency: 100
```

There is still one question. In the.func directory, in the built-hash and built-image files, there are only the corresponding string contents, but there is execution permission. The file permission should be read-only.

braunsonm:
This also affects `func deploy --build=false --image <image>` and `func run` where you do not build with func

Taylorzhao:
> This also affects `func deploy --build=false --image <image>` and `func run` where you do not build with func

I have already compiled it in advance and pushed the compiled image to the remote repository. In version 1.2, under the directory, there is func.yaml. Executing ```func deploy --push=false --build=false``` can be executed successfully, but in version 1.14, this problem will occur.

9numbernine9:
[...]

PR #2390, opened 2024-06-21
fix: Allow undigested images to be deployed directly 

gauron99:
allow tagged images to be deployed directly when specifying --build & --push as false
- with extra tests

fixes https://github.com/knative/func/issues/2331

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -308,17 +308,19 @@ func runDeploy(cmd *cobra.Command, newClient ClientFactory) (err error) {
 
 		// if not digested, image might be provided with a tag
 		if !digested {
-			var tagged bool
-			tagged, err = isTagged(cfg.Image)
+			var undigestedValid bool
+			undigestedValid, err = isValidUndigestedImage(cfg.Image)
 			if err != nil {
 				return
 			}
-			if tagged {
+			if undigestedValid {
 				// this gets overridden when build&push=enabled with built (digested) image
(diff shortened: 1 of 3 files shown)


cmd/deploy.go
```diff
@@ -766,6 +785,32 @@ func printDeployMessages(out io.Writer, f fn.Function) {
...
+		return
+	} else if len(vv) > 2 {
+		err = fmt.Errorf("image '%v' contains an invalid tag (extra ':')", v)
+		return
+	}
+	tag := vv[1]
+	if tag == "" {
+		err = fmt.Errorf("image '%v' has an empty tag", v)
+		return
+	}
```
lkingland, line 808: Do you think this is an error, or should just return false?  For example, is a trailing `:` just another way to be "untagged" (i.e. the colon is superfulous)

lkingland (approved):
Looks good!  I especially like the test suite so we can be sure things continue to work as expected when we refactor the build process to be the one setting the image value

cmd/deploy.go
gauron99: This was my dilemma. Do we want to ease on these restrictions and allow users to do "more" because using `--image` implies "I know what Im doing" or should we still kind of help them along? maybe even in this case it could append ":latest" at the end or something for simplified input.

lkingland (approved):
Looking good to me!  Let's give it a try.

PR #2396, opened 2024-06-26
feat: embed tkn tasks in func binary

matejvasek:
# Changes

- Embedded Tekton tasks in the `func` binary. This removed dependency of files hosted in Github, making func more viable for air-gaped envs. For now I embedded them as plain string, in future I would like to change them into Go struct literals.
- Since the tasks are no longer present as yaml files in GH I added new hidden sub-command `func tkn-tasks` that lists these tasks. This is good for information purposes  or in cases when the task must be installed in cluster for e.g. OCP Web Console build.

/kind enhancement

```release-note
feat: embedded Tekton task definitions in the binary
```


matejvasek:
PTAL @matzew @lkingland @gauron99

pkg/pipelines/tekton/tasks.go
```diff
@@ -1,5 +1,14 @@
----
-apiVersion: tekton.dev/v1
+package tekton
+
+import (
+	"fmt"
+	"strings"
+)
+
+var DeployerImage = "ghcr.io/knative/func-utils:latest"
```
matejvasek, line 8: This variable would be set via linker flags in productized version.

@dsimansk where do we set the linker flags for `kn`?

dsimansk: The soonest midstream, the latest possible is internal build. I'll follow-up on chat.

dsimansk: Let's add a placeholder ldflag to `Makefile`. Nothing fancy, just current default value. For a demonstration and "contract definition" purposes.

pkg/pipelines/tekton/tasks.go
```diff
@@ -228,3 +237,186 @@ spec:
       emptyDir: {}
     - name: layers-dir
       emptyDir: {}
+`
+}
+
+func getS2ITask() string {
+	return `apiVersion: tekton.dev/v1
```
dsimansk, line 244: At first I was worried that `Task` CRD may evolve overtime and break at some point. But on a second though it should be stable enough. It doesn't seem to leverage any exotic parts. :)

dsimansk: Also I wonder if we could reduce dependency on Tekton libraries to bare minimum. That's not directly related to the PR, rather an afterthought.

cmd/tkn_tasks.go
```diff
@@ -0,0 +1,27 @@
...
+	"knative.dev/func/pkg/pipelines/tekton"
+)
+
+func NewTektonClusterTasksCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "tkn-tasks",
+		Short: "List tekton cluster tasks as multi-document yaml",
+		Long: `This command prints tekton tekton task embed in the func binary.
+Some advanced functionality like OpenShift's Web Console build my require installation of these tasks.
+Installation: func tkn-tasks | kubectl apply -f -
```
dsimansk, line 17: Is manual installation always required? Or the most common use case is covered by creating those resources?

matejvasek: AFAIK we only need it when OpenShift's Web Console.

matejvasek:
PTAL @gauron99

cmd/tkn_tasks.go
matejvasek: Only when you create a function in OCP WebUI you need cluster tasks installed.

lkingland (approved):
Looks good to me.  Apart from the large embedded yaml, it's a nice cleanup as well.

matejvasek:
> Looks good to me. Apart from the large embedded yaml, it's a nice cleanup as well.

@lkingland I plan to translate these strings to Go struct literal for type-safety in future.

PR #2401, opened 2024-06-26
Adding GO s2i to e2e test suite

jrangelramos:
# Changes

- Add e2e tests for Go functions using s2i builder (**except** for OnCluster builds scenarios)

### DIFF
--- a/test/e2e/scenario_runtime-http_test.go
+++ b/test/e2e/scenario_runtime-http_test.go
@@ -17,7 +17,7 @@ import (
 
 var runtimeSupportMap = map[string][]string{
 	"node":       {"pack", "s2i"},
-	"go":         {"pack"},
+	"go":         {"pack", "s2i"},
 	"rust":       {"pack"},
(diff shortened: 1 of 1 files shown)


matejvasek:
@lkingland I think these test failures might be caused by actual bug.

matejvasek:
It appears that on cluster build is still broken.
Maybe enable only local build tests for now?

jrangelramos:
yea.. it is failling oncluster build for Go S2i. I would prefer to keep this test until this get sorted out as well.

jrangelramos:
Apparently the same work for scaffold the go s2i func is required to be done during the oncluster build (perhaps in some task in between the tekton git-source/push and build tasks). I'm just leaving the oncluster test coverage apart and will add it later when oncluster build approach for go is sorted out

jrangelramos:
Hi @matejvasek coud you please re-LTGM

matzew (approved):
/lgtm
/approve

PR #2429, opened 2024-07-18
Fix PaC detection for Red Hat's Sandbox cluster

matejvasek:


matejvasek:
@lkingland @gauron99 @matzew PTAL

gauron99 (approved):

PR #2433, opened 2024-07-21
fix: springboot updater missing EOL

matejvasek:


matejvasek:

matzew (approved):
/lgtm
/approve

PR #2440, opened 2024-07-23
Fix failure in GHA caused by missing cargo bin

matejvasek:
It appears that the cargo in no longer part of GHA image at least for macOS.

# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


dsimansk:
@matejvasek what's up with codecov coverage? flake?

PR #2441, opened 2024-07-23
Include go.sum in the Go CloudEvent template

matejvasek:
The go.sum must be in place otherwise "go list -f {{.Incomplete}}" returns true which breaks s2i build.


matejvasek:
@jrangelramos this should fix https://github.com/knative/func/pull/2401

matejvasek:
@gauron99 @matzew @jrangelramos @lkingland  PTAL

PR #2469, opened 2024-08-27
Updated git-clone task

matejvasek:
# Changes

- Updated git fetch task to newer version that is now rootless.

This changes are also already included in the "Go s2i on cluster build" PR.

### DIFF
--- a/pkg/pipelines/tekton/tasks.go
+++ b/pkg/pipelines/tekton/tasks.go
@@ -66,7 +66,7 @@ spec:
       default: "1001"
     - name: GROUP_ID
       description: The group ID of the builder image user.
-      default: "0"
+      default: "65532"
       ##############################################################
       #####  "default" has been changed to "0" for Knative Functions
     - name: PLATFORM_DIR
@@ -107,6 +107,8 @@ spec:
               chmod 775 "$(workspaces.source.path)"
           fi
         done
+
+        chmod -R g+w "$(workspaces.source.path)"
 
         echo "> Parsing additional configuration..."
         parsing_flag=""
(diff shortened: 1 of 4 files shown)


matejvasek:
PTAL @lkingland @matzew @gauron99

pkg/pipelines/tekton/tasks.go
```diff
@@ -66,7 +66,7 @@ spec:
       default: "1001"
     - name: GROUP_ID
       description: The group ID of the builder image user.
-      default: "0"
+      default: "65532"
```
lkingland, line 69: What happens if you just _remove_ this default setting, and `runAsGroup` below, and the `fsGroup` setting?  Is it possible that setting the default to run as group 0 is what caused the problem in the first place?

matejvasek: The primary problem is that that uid is changed, not gid. This causes that git-fetch task on second run cannot clean up the volume. I fixed it by setting gid to match gid of git-fetch task (and also fsGroup) and by setting write access to said group on line 111.

TBH I have no idea why the gid here was set to 0 here. Zbynek implemented it that way long ago.
IMO the gid should be 1000 to match buildpack builder, but it works also with 65532.

I suspect I could just s/65532/1000/g on this PR and it still would work.

matejvasek: This whole ceremony here can be ignored on OCP. OCP automatically sets fsGroup and also adds write perms for the group.

matejvasek:
@lkingland we may also opt not to do this update and keep rootfull git-clone. We just will have to also make func-utils image rootfull to.

pkg/pipelines/tekton/tasks.go
matejvasek: >The primary problem is that that uid is changed, not gid.

This and the fact that set gid works oddly or not at all.

lkingland: Thanks for the clarification.  Let me know if this, or the rootfull git-clone works out better for you and I'll review/approve

PR #2473, opened 2024-08-28
Add s2i to the func-utils image

matejvasek:
# Changes

- Incorporate `s2i`into the func-utils image.

### DIFF
--- a/Dockerfile.utils
+++ b/Dockerfile.utils
@@ -21,7 +21,8 @@ RUN apk add --no-cache socat tar
 
 COPY --from=builder /workspace/func-util /usr/local/bin/
 RUN ln -s /usr/local/bin/func-util /usr/local/bin/deploy && \
-    ln -s /usr/local/bin/func-util /usr/local/bin/scaffold
+    ln -s /usr/local/bin/func-util /usr/local/bin/scaffold && \
+    ln -s /usr/local/bin/func-util /usr/local/bin/s2i
 
 LABEL \
    org.opencontainers.image.description="Knative Func Utils Image" \
--- a/pkg/pipelines/tekton/tasks.go
+++ b/pkg/pipelines/tekton/tasks.go
@@ -241,7 +241,7 @@ spec:
 }
 
 func getS2ITask() string {
-	return `apiVersion: tekton.dev/v1
+	return fmt.Sprintf(`apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   name: func-s2i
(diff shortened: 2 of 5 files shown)


Dockerfile.utils
```diff
@@ -21,7 +21,8 @@ RUN apk add --no-cache socat tar
 
 COPY --from=builder /workspace/func-util /usr/local/bin/
 RUN ln -s /usr/local/bin/func-util /usr/local/bin/deploy && \
-    ln -s /usr/local/bin/func-util /usr/local/bin/scaffold
+    ln -s /usr/local/bin/func-util /usr/local/bin/scaffold && \
+    ln -s /usr/local/bin/func-util /usr/local/bin/s2i
```
matzew, line 25: how to we ensure this is up-to-date?

Is there any "harm" vendoring this extra binary?

pkg/pipelines/tekton/tasks.go
```diff
@@ -303,7 +303,7 @@ spec:
       description: Digest of the image just built.
   steps:
     - name: generate
-      image: quay.io/boson/s2i:latest
```
matzew, line 306: ugh, how old was that? :sweat_smile:

Dockerfile.utils
matejvasek: Actually this ensures the s2i in the binary and the util image is the same which is nice. If there is any updated required (e.g. ther is CVE) we do it just in go.mod and it updates it both. What are disadvantages?

pkg/pipelines/tekton/tasks.go
matejvasek: Quite. Still much up to date that we use now in midstream.

Dockerfile.utils
matzew: I was worried for carrying on CVEs, but looks like it is addressed

matejvasek: Well we basically already ship s2i binary. We use it as an library in `func` CLI.

matejvasek: >Well we basically already ship s2i binary. We use it as an library in func CLI.

So we are already on hook for keeping it up to date.

pkg/pipelines/tekton/tasks.go
matejvasek: `quay.io/boson/s2i:latest` is our image we build manually once in blue moon.

lkingland (approved):

PR #2477, opened 2024-09-04
migrate client-pkg to client/pkg

gauron99:


matejvasek:
just a nit: I would not group `knative.dev/client` with `knative.dev/func`.

gauron99:
Do you want to group it with github & k8s or should i put it separately?

matejvasek:
> Do you want to group it with github & k8s or should i put it separately?

Yes group it with github.

I think that right now we do following grouping:
* Go stdlib
* external deps
* our own packages (knative.dev/func).

cardil:
Looks good to me, but let maintainers approve.

PR #2488, opened 2024-09-09
Improved ldflags setting in Makefile

matejvasek:
# Changes

- Improved ldflags setting in Makefile


matejvasek:
PTAL @gauron99 @matzew @lkingland @jrangelramos

PR #2506, opened 2024-09-17
Improve socat connection success detection using KMP

matejvasek:
# Changes

It appears that Go's `MatchReader()` returns `true` only if two characters are written to the writer after the searched word. Thta's mostly all right because `socat` actually writes several more lines to stderr. However this new implementation is better it should signal immediately when searched patter is written to the writer.

See: https://pkg.go.dev/github.com/agext/regexp#section-documentation
>Note that regular expression matches may need to examine text beyond the text returned by a match, so the methods that match text from a RuneReader may read arbitrarily far into the input before returning.

See also: https://stackoverflow.com/questions/76559206/how-to-match-a-regular-expression-on-a-slow-hanging-io-reader-in-go

KMP: https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm

### DIFF
--- a/pkg/k8s/dialer.go
+++ b/pkg/k8s/dialer.go
@@ -1,7 +1,6 @@
 package k8s
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"errors"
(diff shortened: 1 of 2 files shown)


matejvasek:
@dsimansk @ReToCode please override `codecov/project`.

matejvasek:
Also why the hell is `codecov/project` twice there?

lkingland:
> Also why the hell is `codecov/project` twice there?

That might be a GitHub flake, because it only appears twice here in the summary, but once on the "Checks" screen 🤷🏻

lkingland (approved):

lkingland:
Looks like the codecov double-result is still holding up the merge. Going to retest...

ReToCode:
> @dsimansk @ReToCode please override codecov/project.

Sorry lost that power again... And David is on PTO. @matzew you still an admin here?

matejvasek:
@dsimansk @matzew @lkingland  PTAL

matejvasek:
PTAL @gauron99

gauron99 (approved):
/lgtm

PR #2510, opened 2024-09-20
fix: make image digest check more permissive

### LINKED ISSUE #2508: func deploy doesn't support registry port
I try to run a knative function on a docker desktop k8s cluster.
For the registry, I use registry:2 on docker desktop with forwarded port to 5001:

```sh
$ docker ps | grep registry:2
04a2c59975c7   registry:2   "/entrypoint.sh /etc…"   18 minutes ago   Up 18 minutes   0.0.0.0:5001->5000/tcp   registry
```

Sample func:

```yaml
specVersion: 0.36.0
name: demo
runtime: go
registry: registry.192.168.2.99.sslip.io:5001
image: registry.192.168.2.99.sslip.io:5001/demo:latest
created: 2024-08-02T10:48:42.077191+02:00
invoke: cloudevent
build:
  builder: host
  buildEnvs:
  - name: FUNC_ENABLE_HOST_BUILDER
    value: "true"
  - name: FUNC_BUILDER
    value: host
  - name: FUNC_CONTAINER
    value: "true"
  - name: LOG_LEVEL
    value: error
run:
  envs:
  - value: '{{ secret:demo }}'
deploy:
  namespace: cs
  image: registry.192.168.2.99.sslip.io:5001/demo:latest
  options:
    scale:
      min: 0
      max: 1
  subscriptions:
  - source: default
    filters:
      type: demo.event
```

The build of the function, even with image upload works as expected:

```sh
$ func build -u --registry-insecure
Building function image
   f.linux.amd64
go: module ./f requires go >= 1.22.4; switching to go1.22.7
go: upgraded go 1.21 => 1.22.4
go: added toolchain go1.22.7
go: upgraded github.com/rs/zerolog v1.32.0 => v1.33.0
go: upgraded golang.org/x/sys v0.18.0 => v0.22.0
   f.linux.arm64
   f.linux.arm.v7
🙌 Function built: registry.192.168.2.99.sslip.io:5001/demo:latest

$ curl -k registry.192.168.2.99.sslip.io:5001/v2/demo/tags/list
{"name":"demo","tags":["latest"]}
```

But the deployment command doesn't support image-names with a port in it:

```sh
$ func deploy -i registry.192.168.2.99.sslip.io:5001/demo:latest
Warning: namespace chosen is 'cs', but currently active namespace is 'default'. Continuing with deployment to 'cs'.
Error: image 'registry.192.168.2.99.sslip.io:5001/demo:latest' contains an invalid tag (extra ':')
```

Please fix that validation, so that local registries are allowed to test knative functions.

lkingland:
Makes the image digest check and validation a bit more permissive.

Fixes #2508

/kind bug

```release-note
Fixes a bug where registries could sometimes not specify port
```

### DIFF
--- a/cmd/deploy_test.go
+++ b/cmd/deploy_test.go
@@ -756,34 +756,34 @@ func TestDeploy_ImageWithDigestErrors(t *testing.T) {
 		image     string // value to provide as --image
 		build     string // If provided, the value of the build flag
 		push      bool   // if true, explicitly set argument --push=true
-		errString string // the string value of an expected error
+		errPrefix string // the string value of an expected error
 	}{
 		{
 			name:  "correctly formatted full image with digest yields no error (degen case)",
-			image: "example.com/myNamespace/myFunction:latest@sha256:7d66645b0add6de7af77ef332ecd4728649a2f03b9a2716422a054805b595c4e",
+			image: "example.com/namespace/myfunction@sha256:7d66645b0add6de7af77ef332ecd4728649a2f03b9a2716422a054805b595c4e",
 			build: "false",
 		},
 		{
 			name:      "--build forced on yields error",
-			image:     "example.com/myNamespace/myFunction:latest@sha256:7d66645b0add6de7af77ef332ecd4728649a2f03b9a2716422a054805b595c4e",
+			image:     "example.com/mynamespace/myfunction@sha256:7d66645b0add6de7af77ef332ecd4728649a2f03b9a2716422a054805b595c4e",
 			build:     "true",
-			errString: "building can not be enabled when using an image with digest",
+			errPrefix: "building can not be enabled when using an image with digest",
 		},
 		{
 			name:      "push flag explicitly set with digest should error",
-			image:     "example.com/myNamespace/myFunction:latest@sha256:7d66645b0add6de7af77ef332ecd4728649a2f03b9a2716422a054805b595c4e",
+			image:     "example.com/mynamespace/myfunction@sha256:7d66645b0add6de7af77ef332ecd4728649a2f03b9a2716422a054805b595c4e",
 			push:      true,
-			errString: "pushing is not valid when specifying an image with digest",
+			errPrefix: "pushing is not valid when specifying an image with digest",
 		},
 		{
 			name:      "invalid digest prefix 'Xsha256', expect error",
-			image:     "example.com/myNamespace/myFunction:latest@Xsha256:7d66645b0add6de7af77ef332ecd4728649a2f03b9a2716422a054805b595c4e",
-			errString: "image digest 'Xsha256:7d66645b0add6de7af77ef332ecd4728649a2f03b9a2716422a054805b595c4e' requires 'sha256:' prefix",
+			image:     "example.com/mynamespace/myfunction@Xsha256:7d66645b0add6de7af77ef332ecd4728649a2f03b9a2716422a054805b595c4e",
+			errPrefix: "could not parse reference",
 		},
 		{
 			name:      "invalid sha hash length(added X at the end), expect error",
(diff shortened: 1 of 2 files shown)


gauron99 (approved):
/lgtm 
/hold @lkingland sorry didnt realize ports can be added to the name! thanks.

cmd/deploy_test.go
```diff
@@ -2140,3 +2140,39 @@ func TestDeploy_CorrectImageDeployed(t *testing.T) {
...
+		Error    bool
+	}{
+		{Name: "simple", Image: "alpine", Digested: false, Error: false},
+		{Name: "tagged", Image: "alpine:latest", Digested: false, Error: false},
+		{Name: "domain", Image: "registry.example.com/alice/alpine", Digested: false, Error: false},
+		{Name: "domain-tagged", Image: "registry.example.com/alice/alpine:latest", Digested: false, Error: false},
+		{Name: "domain-port", Image: "registry.example.com:5001/alice/alpine", Digested: false, Error: false},
+		{Name: "domain-port-tagged", Image: "registry.example.com:5001/alice/alpine:latest", Digested: false, Error: false},
+		{Name: "domain-port-tagged-digested", Image: "registry.example.com:5001/alice/alpine@sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855", Digested: true, Error: false},
+		{Name: "user-domain-port-tagged-digested", Image: "user@registry.example.com:5001/alice/alpine@sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855", Digested: true, Error: false},
```
matejvasek, line 2159: Is this even valid input?

matejvasek: Because I do not think so.

matejvasek: Neither does `github. com/google/go-containerregistry/pkg/name.ParseReference()` nor `github.com/distribution/reference.Parse()`.

matejvasek:
Couldn't we implement `isDigested()` as
```go
func isDigested(v string) (validDigest bool, err error) {
        ref, err := name.ParseReference(v)
        if err != nil {
	        return false, err
        }
        _, ok := ref.(name.Digest)
        return ok, nil
}
```
or
```go
func isDigested(v string) (validDigest bool, err error) {
        ref, err := reference.Parse(v)
        if err != nil {
	        return false, err
        }
        _, ok := ref.(reference.Digested)
        return ok, nil
}
```
?

cmd/deploy_test.go
```diff
@@ -2140,3 +2140,39 @@ func TestDeploy_CorrectImageDeployed(t *testing.T) {
...
+		{Name: "error-missing-prefix", Image: "registry.example.com/alice/alpine@e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855", Digested: false, Error: true},
+		{Name: "error-invalid-length", Image: "registry.example.com/alice/alpine@sha256:invalid-hash-value", Digested: false, Error: true},
```
matejvasek, line 2161: What about image name like: `registry.example.com:5001/alice/alpine@sha512:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`

lkingland:
> Couldn't we implement `isDigested()` as
> 
> ```go
> func isDigested(v string) (validDigest bool, err error) {
>         ref, err := name.ParseReference(v)

Yes, I prefer to use an extant implementation when available.  See udpate

cmd/deploy_test.go
lkingland: Unnecessary now that we're using the parsing implementation in go-containerregistry 👍🏻

PR #2513, opened 2024-09-23
Remove unnecessary clause from Dockerfile

matejvasek:


matejvasek:
@dsimansk plz override `codecov/project`

matejvasek:

gauron99 (approved):

PR #2523, opened 2024-09-26
fix: allow sub-pkgs in Go functions

matejvasek:
# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


matejvasek:
@lkingland at three places in the code `go get f(unction)` is/was called, what is reason for that? Does not `go build` automatically pulls modules? Also if antying shouldn't we rather call `go mod download` so all deps are downloaded?

matejvasek:
PTAL @lkingland

lkingland (approved):

PR #2574, opened 2024-11-06
Add Validation check to see if func is already initialized

tarunsunny3:
# Changes

This PR addresses an issue where calling func build without prior function initialization resulted in an irrelevant error message. Now, if func build is invoked before initialization, an appropriate error is shown, guiding users to initialize the function first.

/kind bug

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -303,6 +303,13 @@ func (c buildConfig) Prompt() (buildConfig, error) {
 	// value and will always use the value from the config (flag or env variable).
 	// This is not strictly correct and will be fixed when Global Config: Function
 	// Context is available (PR#1416)
+	hasFunc, err := fn.IsFunctionInitialized(c.Path)
+	if err != nil {
+		return c, err
+	}
+	if !hasFunc {
+		return c, fmt.Errorf("no function has been initialized in the current directory. Please initialize a function by running either:\n- func init --language <your language>\n- func create <function name> --language <your_language>")
+	}
 	f, err := fn.NewFunction(c.Path)
 	if err != nil {
 		return c, err
--- a/pkg/functions/client.go
+++ b/pkg/functions/client.go
@@ -1276,6 +1276,9 @@ func isEffectivelyEmpty(path string) (bool, error) {
 	}
 	return true, nil
 }
+func IsFunctionInitialized(path string) (bool, error) {
+	return hasInitializedFunction(path)
+}
 
 // returns true if the given path contains an initialized function.
 func hasInitializedFunction(path string) (bool, error) {


cmd/build.go
```diff
@@ -303,6 +303,13 @@ func (c buildConfig) Prompt() (buildConfig, error) {
...
 	// This is not strictly correct and will be fixed when Global Config: Function
 	// Context is available (PR#1416)
+	hasFunc, err := fn.IsFunctionInitialized(c.Path)
+	if err != nil {
+		return c, err
+	}
+	if !hasFunc {
+		return c, fmt.Errorf("no function has been initialized in the current directory. Please initialize a function by running either:\n- func init --language <your language>\n- func create <function name> --language <your_language>")
+	}
 	f, err := fn.NewFunction(c.Path)
```
matejvasek, line 313: Should not this already return `ENOENT`?

matejvasek: I think we could simply check `f.Initialized()` instead of introducing new `IsFunctionInitialized()` function.

matejvasek: TBH I am surprised that the function returns `err==nil` when there is no function. Why it does that @lkingland ? https://github.com/knative/func/blob/e6fa020f7833e62c29364902da04e7cc6f7dab60/pkg/functions/function.go#L257-L265

cmd/build.go
```diff
@@ -303,6 +303,13 @@ func (c buildConfig) Prompt() (buildConfig, error) {
 	// value and will always use the value from the config (flag or env variable).
 	// This is not strictly correct and will be fixed when Global Config: Function
 	// Context is available (PR#1416)
+	hasFunc, err := fn.IsFunctionInitialized(c.Path)
```
matejvasek, line 306: I would avoid introducing this new function. Just call `f.Initialized()` below.

cmd/build.go
matejvasek: @lkingland also if you mean zero-value by "default function" , then you also need to zero the path of the f.

lkingland: Yes, good question.  A Function is first a structure in memory, which can be serialized to disk.  The path to where the Function is, or will be when initialized, is the only value provided in the constructor, and thus is the only field that is populated with user data when creating a new function from scratch.


Like you say.  Just check if `f.Initialized()`.

cmd/build.go
```diff
@@ -303,6 +303,13 @@ func (c buildConfig) Prompt() (buildConfig, error) {
...
+	if err != nil {
+		return c, err
+	}
+	if !hasFunc {
```
lkingland, line 310:
```suggestion
        f, err := fn.NewFunction(c.Path)
        if err != nil {
          return c, err
        }
	if !f.Initialized() {
```

cmd/build.go
```diff
@@ -303,6 +303,13 @@ func (c buildConfig) Prompt() (buildConfig, error) {
...
+		return c, fmt.Errorf("no function has been initialized in the current directory. Please initialize a function by running either:\n- func init --language <your language>\n- func create <function name> --language <your_language>")
```
lkingland, line 311:
```suggestion
		return c, fmt.Errorf("no function has been initialized in %q. Please initialize a function by running:\n- func init --language <your language>\n", c.Path)
```

pkg/functions/client.go
```diff
@@ -1276,6 +1276,9 @@ func isEffectivelyEmpty(path string) (bool, error) {
 	}
 	return true, nil
 }
+func IsFunctionInitialized(path string) (bool, error) {
+	return hasInitializedFunction(path)
+}
```
lkingland, line 1281:
```suggestion
```

lkingland (commented):
Thanks for finding out where the check needs to be!

I made a couple of suggestions to 1) use the pre-existing `.Initialzied()` method and 2) print the path of the function in the error message (it's not always in the current directory)

Thanks for the contribution.

cmd/build.go
matejvasek: I still do not like, but can live with it. Just the function docstring says:
>NewFunction from a given path. Invalid paths, or no function at path are errors.

matejvasek: I bet that if you asked 10 programmers what `NewFunction(pathWithNoFunction)`  returns at least 9 would tell it's an error.

lkingland:
Gentle ping @tarunsunny3 .  We would love to merge your PR whenever you're ready to make those minor changes.  👍🏻

tarunsunny3:
Sorry I have been busy with my coursework lately
I just tried it with just f.Initialized() but for some reason c.Path is giving empty output as supposed to give the current directory's path if I am correct

Is it fine if it gives an empty output?
image

PR #2581, opened 2024-11-19
Improved build caching for s2i local build

matejvasek:
# Changes

- :gift: Improve caching for local `s2i` builds, by employing `RUN --mount=type=cache...`.

/kind enhancement

```release-note
improved caching for local s2i builds
```

### DIFF
--- a/pkg/builders/s2i/builder.go
+++ b/pkg/builders/s2i/builder.go
@@ -343,6 +343,7 @@ func (b *Builder) Build(ctx context.Context, f fn.Function, platforms []fn.Platf
 	opts := types.ImageBuildOptions{
 		Tags:       []string{f.Build.Image},
 		PullParent: true,
+		Version:    types.BuilderBuildKit,
 	}
 
 	resp, err := client.ImageBuild(ctx, pr, opts)


matejvasek:
PTAL @lkingland @gauron99 @rhuss

lkingland (approved):
Very nice 👍

rhuss:
Nice finde about `RUN --mount ...` (didn't know about that). Does it also work with Podman ?

matejvasek:
> Nice finde about `RUN --mount ...` (didn't know about that). Does it also work with Podman ?

@rhuss 
I works at least in podman +v4.0, it might not work on older versions.
It works on Docker only if BuildKit is enabled.

rhuss:
Looks good to me, however I wonder whether we should hardcode the cache directory to `/tmp/artefacts`. 

* From the outside there is not indication that it belongs to `func`.
* It's shared across multiple function projects, which might or might not be desired
* Different OSs have different standard location for caches, like on MacOs its `~/Library/Caches`. We should respect this.
* It should be configurable
* `/tmp` is often a memory filesystem, so it might be too small for caching large dependencies.

matejvasek:
The  `/tmp/artefacts` dir is `s2i` specific, I did not made that up.

matejvasek:
To be clear `/tmp/artefacts` is path in container not in host.

matejvasek:
PTAL @gauron99 @lkingland @rhuss

aslom:
I thought one of bots was doing tests and created func binaries?
I see it only runs tests and only keep test results? https://github.com/knative/func/actions/runs/11955247512/job/33327188182?pr=2581 ?

matejvasek:
@aslom  [func.gpg.gz](https://github.com/user-attachments/files/17849380/func.gpg.gz)

matejvasek:
This PR should at least fix java maven dep caching @aslom . It probably want change much for Go. I just wonder whether this would work on your setup.

matejvasek:
@aslom There are also some nightly builds in knative infra, but again there are on `main` not on every PR.

PR #2587, opened 2024-11-22
Fix missing workspaces decl in the scaffold task

matejvasek:
# Changes

- :bug: Fix missing `source` workspaces declaration in the `scaffold` task. This led to TaskRun ending with `TaskRunValidationFailed: workspace binding "source" does not match any declared workspace`.
This behaviour was observed by @alanreynosov on his k3s cluster, we have not seen it so far.

/kind bug

```release-note
fix: workspace binding "source" does not match any declared workspace for scaffold task
```

### DIFF
--- a/pkg/pipelines/tekton/tasks.go
+++ b/pkg/pipelines/tekton/tasks.go
@@ -440,6 +440,9 @@ spec:
     - name: path
       description: Path to the function project
       default: ""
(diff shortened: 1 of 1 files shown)


matejvasek:
cc @alanreynosov

matejvasek:

Issue #2562, opened 2024-11-04
Wrong error message when login to OCI registry interactively.


rhuss:
When using a default registry with 

```
export FUNC_REGISTRY=docker.io/invaliduser
```

and then do a `func deploy` without being logged in you get:

```
10:10 ➜ kn func deploy
function up-to-date. Force rebuild with --build
Please provide credentials for image registry (index.docker.io).
? Username: rhuss
? Password: ************************************
Incorrect credentials, please try again.
? Username:
```

although the password is correct (verified with `podman login`). The reason is probably that for the authentication still `invaliduser` is used. 

I suggest to _not_ asked for the username here, but just reuse the configured one (either via env or via cli option), and just print this out to the user when asking for the password.

matejvasek:
@rhuss 

>The reason is probably that for the authentication still invaliduser is used

I do not think so. I believe the correct username is used, but the user is not authorized on the specific namespace; i.e *authentication* succeeds but *authorization* does not.

>I suggest to not asked for the username here

I am not sure what you mean.
IMO we should ask for username. Imagine you want to push image into some orgnatization.

```
export FUNC_REGISTRY=docker.io/our-org
```
You still want to user your personal login (e.g. `my-user-name`) which belongs to the org, right?

However I agree the error messages could be better.

rhuss:
You could at least repeat the username or org for which authentication is done. And I don't think it's always about authorization, as the `invalid user` does not exist at Docker Hub, including much more context around the error. Of course, the user or orga _could_ exist, too, so it's then about authorization. But still, it's then not an _invalid credentials_ error, that clearly indicates an authentication error (it would be _no access_ kind of message for authorization errors).

At least that error took me some hours to figure out that I had a typo in the env's repo name. Yes, there was a warning, but it fades out quickly.

matejvasek:
> And I don't think it's always about authorization, as the invalid user does not exist at Docker Hub

Yes, but I suspect that API would still return authorization error not 404, so I am not sure if we would be able to differentiate. Some APIs return authorization error even for non-existent entries for sake of security or something.

Again I agree the error should be better.

matejvasek:
@gauron99 would you look into this?

matejvasek:
Basically [CheckAuth()](https://github.com/knative/func/blob/92856e56277bcb4f1a4dd578aa7483fe5a974890/pkg/docker/creds/credentials.go#L51) have to return two distinct errors; *authentication* error and *authorization* error.

[...]

Issue #2595, opened 2024-11-25
Further optimize the credentials process (followup)


gauron99:
**Followup to https://github.com/knative/func/issues/2562**

This issues' aim is to note that the credentials & prompt combo needs further optimization, wrt new prompt implementation that is in process and the linked issue above -- possibly differentiating between authentication and authorization errors etc.

lkingland:
This one could use additional specification — what specifically about the credentials/prompt flow needs optimizing? Concrete repro or pain point would help an implementer.

gauron99:
The idea of this issue was to investigate the options to & subsequently implement an additional cases for errors.

AFAIR we currently still combine 401 and 403 errors, though there was some issues with reliably differentiating between these from different registries (different registries might return different errors, does it work with trying to access private repos? -- not consistent, not sure if still the case -- investigation needed)

So 401 & 403 produce the same error. We could differentiate and improve the registry setup flow.
If bad creds, re-prompt. If valid creds but not valid permissions, dont re-prompt (improve error messages there) <-- this is just an idea from my head.

### key files to touch (might not be all)
`pkg/creds/credentials.go` & `cmd/prompt/prompt.go`

This would ideally make a nicer flow, better messages for user and better visibility of what is happening on bad access / what has gone wrong

Itx-Psycho0:
Hi @gauron99, I'd like to take this on. From your explanation, the goal is to differentiate between 401 and 403 errors from the registry and handle them separately, re-prompting for credentials on 401 but improving the error message on 403 without re-prompting. I'll start by investigating how pkg/creds/credentials.go currently handles these cases and whether the registry response behavior is consistent enough to reliably distinguish them. Could you assign this to me?

PR #2596, opened 2024-11-25
add repository to the credentials prompt

gauron99:
Added `repository` value (now it is `registry/repository`) to prompt for credentials where its used for printing out tto terminal "during the prompt" to make it easier for the user to see what he is trying to provide credentials for.

Additionally I have created a [new issue](https://github.com/knative/func/issues/2595) that is a followup to this and its' purpose is to further optimize this "credentials" process

fix https://github.com/knative/func/issues/2562

### DIFF
--- a/pkg/docker/creds/credentials.go
+++ b/pkg/docker/creds/credentials.go
@@ -236,7 +236,6 @@ func (c *credentialsProvider) getCredentials(ctx context.Context, image string) 
 	}
 
 	registry := ref.Context().RegistryStr()
-
 	for _, load := range c.credentialLoaders {
 
 		result, err = load(registry)
@@ -263,8 +262,14 @@ func (c *credentialsProvider) getCredentials(ctx context.Context, image string) 
 		return docker.Credentials{}, ErrCredentialsNotFound
 	}
 
+	// this is [registry] / [repository]
+	// this is  index.io  / user/imagename
+	registryWithRepository := registry + "/" + ref.Context().RepositoryStr()
+
+	// the trying-to-actualy-authorize cycle
 	for {
-		result, err = c.promptForCredentials(registry)
+		// use repo here to print it out in prompt
+		result, err = c.promptForCredentials(registryWithRepository)
 		if err != nil {
 			return docker.Credentials{}, err
 		}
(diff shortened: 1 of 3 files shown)


pkg/docker/creds/credentials.go
```diff
@@ -23,9 +23,10 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 
 	"knative.dev/func/pkg/docker"
+	fn "knative.dev/func/pkg/functions"
 )
 
-type CredentialsCallback func(registry string) (docker.Credentials, error)
+type CredentialsCallback func(f fn.Function, registry string) (docker.Credentials, error)
```
matejvasek, line 29: @gauron99  I do not see any reason for adding `fn.Function` here. The `registry` parameter is completely sufficient.

matejvasek: Just use registry at https://github.com/knative/func/pull/2596/files#diff-753067f7ed9e5f834cd3d8219369a40264abf6ade1665a12e680430f3986c675R27

matejvasek:
```go
fmt.Fprintf(out, "Incorrect credentials for registry '%s'. Please make sure the registry is correct and try again.\n", registry)
```

gauron99: it is so that I can provide the image name and print it out, making it more obvious what are the credentials "used for"

matejvasek: Hmm I thought that registry also included project/namespace something like `docker.io/johndoe`.

gauron99: I will clean it up a bit, wait. I will ping you when its ready

gauron99:
@matejvasek what about this. I have changed the sole argument that needed it. Added a repository to it so now it is displayed when the prompt asks you for your credentials when trying to 'push' WITH the actual 'registry/repository'
Althought I had to change the prompt tests to reflect this as well

matejvasek:
@gauron99 
>what about this. I have changed the sole argument that needed it.

This seems reasonable. Just call it `repository` instead of overly verbose `registryWithRepo`.

IMO that how our flag should have been called: instead of `--registry` we should have had `--repository` or `--img-reposotry`.

gauron99:
@matejvasek @lkingland PTAL

PR #2604, opened 2024-11-29
Update pack/s2i/docker deps

matejvasek:
# Changes

- :broom: update buildpack dependency
- :broom: update source-to-image dependency
- :broom: update required Go version to 1.23


matejvasek:
I am afraid this dep bump would require knative-wide Go update to 1.23.

go.mod
```diff
@@ -1,6 +1,6 @@
 module knative.dev/func
 
-go 1.22.7
+go 1.23
```
matzew, line 3: should we check w/ the rest of upstream, regarding this bump?

CC @dsimansk

dsimansk: In theory, we would eventually move to toolchain `1.23.z` regardless. Personally I don't like `go.mod` minVersion bumps unless warranted by other dependencies or specific go feature from newer version.

matejvasek:
PTAL @gauron99 @lkingland @dsimansk

go.mod
matejvasek: The latest buildpack requires go1.23. Also if I just wanted to update source-to-image we are at the same place: the latest source-to-image requires docker v0.27.0 and only latest buildpacks work with that.

gauron99 (approved):

PR #2608, opened 2024-12-02
Fix some linter errors in integration tests

matejvasek:
# Changes

- :broom: Fix some linter errors in integration tests

### DIFF
--- a/pkg/docker/runner_int_test.go
+++ b/pkg/docker/runner_int_test.go
@@ -50,6 +50,9 @@ func TestRun(t *testing.T) {
 	}
 
 	f, err = client.Build(ctx, f)
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	// Run the function using a docker runner
 	var out, errOut bytes.Buffer
@@ -125,6 +128,9 @@ func TestRunDigested(t *testing.T) {
 
 	// prebuild default image
 	f, err = client.Build(ctx, f)
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	// simulate passing image from --image flag since client.Run just sets
 	// a timeout and simply calls runner.Run.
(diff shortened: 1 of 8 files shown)


pkg/docker/runner_int_test.go
```diff
@@ -125,6 +128,9 @@ func TestRunDigested(t *testing.T) {
 
 	// prebuild default image
 	f, err = client.Build(ctx, f)
+	if err != nil {
```
matejvasek, line 131: Apparently this fails. Any ideas @lkingland

gauron99: @matejvasek because registry is empty

gauron99: `f := fn.Function{Runtime: "go", Root: root, Registry: "docker.io/4141gauron3268"}` this should work

matejvasek: Does that operation have any effect? What does "prebuild default image" mean?

matejvasek: is `f, err = client.Build(ctx, f)` really needed?

gauron99: the point of the test is to "override" a prebuilt image with a digested image. `f build` -> `f run --image=X` since the --image should take precedence

matejvasek: But the operation is seemingly NOOP. Neither the func structure nor the data on filesystem are changed. It looks like it does nothing.

matejvasek: I think it just crashes.

matejvasek: Maybe it would do something if we applied the changes you mentioned above?

lkingland: Yes, I think that if you add a registry (any registry) to the Function definition as David mentions, the initial build will complete.  Then, the second build will confirm that setting `.Image` will override the image which was derived during the first run.

matejvasek:
PTAL @gauron99 @lkingland @matzew

PR #2612, opened 2024-12-13
use knative go setup action

lkingland:
Keeps the Go version used by Knative projects in sync with each-other and tooling.

/kind cleanup


dprotaso:
/lgtm
/approve

there's a github warning about another outdated action - I'd suggest a follow up PR.

PR #2617, opened 2024-12-17
Provide tiny buildpack builder image for ARM64

matejvasek:
Enable arm64 support for tiny builder (Java,Go).
This PR actually enables only Java since some additional work has to be done for Go because upstream paketo buildpack do not support Go fully yet.


matejvasek:
PTAL @lkingland @gauron99 @jrangelramos @matzew

lkingland (approved):

PR #2622, opened 2024-12-18
Fix authorization for buildpack builder periodic task

matejvasek:
Fixed authorization by setting keychain for the pack client.
This is required since recently we set pull policy to "always" recently.


matzew (approved):
/lgtm
/approve

PR #2647, opened 2025-01-09
Remove gauron image references in code

gauron99:
- Im removing the whole test here since I found a better way to do it (possibly without the hardcoded image reference) and will do so in a followup PR.
- Simple rename to not confuse anyone because its just an example.


cmd/deploy_test.go
```diff
@@ -842,7 +842,7 @@ func TestDeploy_ImageWithDigestErrors(t *testing.T) {
 func TestDeploy_ImageWithDigestDoesntPopulateBuild(t *testing.T) {
 	root := FromTempDirectory(t)
 	// image with digest (well almost, atleast in length and syntax)
-	const img = "docker.io/4141gauron3268@sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+	const img = "example.com/gauron@sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
```
matzew, line 845: I still see a `gauron` here. Not sure if that should be removed (see title of PR)

gauron99: aha yes thanks youre right, i was trying to do 2 things at once with the title and the changes

gauron99:
PTAL @matzew

PR #2650, opened 2025-01-15
test for func run with digested image override

gauron99:
add a test for "Func run with digested image when function is already built, expecting to override said built image"

### DIFF
--- a/cmd/run_test.go
+++ b/cmd/run_test.go
@@ -396,7 +396,7 @@ func TestRun_DirectOverride(t *testing.T) {
 
 	runner.RunFn = func(_ context.Context, f fn.Function, _ time.Duration) (*fn.Job, error) {
 		if f.Build.Image != overrideImage {
-			return nil, fmt.Errorf("Expected image to be overriden with '%v' but got: '%v'", overrideImage, f.Build.Image)
+			return nil, fmt.Errorf("Expected image to be overridden with '%v' but got: '%v'", overrideImage, f.Build.Image)
 		}
 		errs := make(chan error, 1)
 		stop := func() error { return nil }
--- a/cmd/run_test.go
+++ b/cmd/run_test.go
@@ -386,3 +386,79 @@ func TestRun_CorrectImage(t *testing.T) {
 		})
 	}
 }
+
+// TestRun_DirectOverride tests that an --image passed after a function has
+// already been build, the given --image with digest will override built function
+func TestRun_DirectOverride(t *testing.T) {
+	const overrideImage = "registry/myrepo/myimage@sha256:0000000000000000000000000000000000000000000000000000000000000000"
+	root := FromTempDirectory(t)
+	runner := mock.NewRunner()
+
+	runner.RunFn = func(_ context.Context, f fn.Function, _ time.Duration) (*fn.Job, error) {
+		if f.Build.Image != overrideImage {
+			return nil, fmt.Errorf("Expected image to be overriden with '%v' but got: '%v'", overrideImage, f.Build.Image)
+		}
+		errs := make(chan error, 1)
+		stop := func() error { return nil }
+		return fn.NewJob(f, "127.0.0.1", "8080", errs, stop, false)
+	}
+
+	builder1 := mock.NewBuilder()
+
+	// SETUP THE ENVIRONMENT & SITUATION
+	// create function
+	_, err := fn.New().Init(fn.Function{Root: root, Runtime: "go"})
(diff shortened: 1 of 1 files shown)


cmd/run_test.go
```diff
@@ -386,3 +386,79 @@ func TestRun_CorrectImage(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	runErrCh := make(chan error, 1)
+	go func() {
+		_, err := cmd.ExecuteContextC(ctx)
+		if err != nil {
+			runErrCh <- err // error was not expected
+			return
+		}
+
+		// Ensure invocation doesnt happen for the second time as the image was
```
matejvasek, line 451: typo

cmd/run_test.go
```diff
@@ -386,3 +386,79 @@ func TestRun_CorrectImage(t *testing.T) {
...
+			return nil, fmt.Errorf("Expected image to be overridden with '%v' but got: '%v'", overrideImage, f.Build.Image)
+		}
+		errs := make(chan error, 1)
+		stop := func() error { return nil }
+		return fn.NewJob(f, "127.0.0.1", "8080", errs, stop, false)
+	}
+
+	builder1 := mock.NewBuilder()
+
+	// SETUP THE ENVIRONMENT & SITUATION
```
matejvasek, line 408: typo => "SET UP"

cmd/run_test.go
```diff
@@ -386,3 +386,79 @@ func TestRun_CorrectImage(t *testing.T) {
...
+		fn.WithBuilder(builder2),
+		fn.WithRegistry("ghcr.com/reg"),
+	))
+	cmd.SetArgs([]string{fmt.Sprintf("--image=%s", overrideImage)})
+
+	// run function with above argument
+	ctx, cancel := context.WithCancel(context.Background())
+	runErrCh := make(chan error, 1)
+	go func() {
+		_, err := cmd.ExecuteContextC(ctx)
```
matejvasek, line 445: This probably could be just `err := cmd.ExecuteContext(ctx)`

cmd/run_test.go
```diff
@@ -386,3 +386,79 @@ func TestRun_CorrectImage(t *testing.T) {
...
+		fn.WithRunner(runner),
+		fn.WithBuilder(builder2),
+		fn.WithRegistry("ghcr.com/reg"),
+	))
+	cmd.SetArgs([]string{fmt.Sprintf("--image=%s", overrideImage)})
+
+	// run function with above argument
+	ctx, cancel := context.WithCancel(context.Background())
+	runErrCh := make(chan error, 1)
+	go func() {
```
matejvasek, line 444: I think we could get rid of this asynchronicity by closing err channel in the runner:
```patch
diff --git a/cmd/run_test.go b/cmd/run_test.go
index 7340ca59f..88130ea83 100644
--- a/cmd/run_test.go
+++ b/cmd/run_test.go
@@ -398,7 +398,8 @@ func TestRun_DirectOverride(t *testing.T) {
 		if f.Build.Image != overrideImage {
 			return nil, fmt.Errorf("Expected image to be overridden with '%v' but got: '%v'", overrideImage, f.Build.Image)
 		}
-		errs := make(chan error, 1)
+		errs := make(chan error)
+		close(errs)
 		stop := func() error { return nil }
 		return fn.NewJob(f, "127.0.0.1", "8080", errs, stop, false)
 	}
@@ -440,25 +441,16 @@ func TestRun_DirectOverride(t *testing.T) {
 
 	// run function with above argument
-	runErrCh := make(chan error, 1)
-	go func() {
-		_, err := cmd.ExecuteContextC(ctx)
+	defer cancel()
+
+	err = cmd.ExecuteContext(ctx)
 	if err != nil {
-			runErrCh <- err // error was not expected
-			return
+		t.Fatal(err)
 	}
 
 	// Ensure invocation doesnt happen for the second time as the image was
 	// provided with a digest (should not build)
 	if builder2.BuildInvoked {
-			runErrCh <- fmt.Errorf("Function was not expected to build again but it did")
-		}
-
-		close(runErrCh) // release the waiting parent process
-	}()
-	cancel() // trigger the return of cmd.ExecuteContextC in the routine
-	<-ctx.Done()
-	if err := <-runErrCh; err != nil { // wait for completion of assertions
-		t.Fatal(err)
+		t.Fatal("Function was not expected to build again but it did")
 	}
 }

```

matejvasek:
Some nits, but otherwise looks good.

PR #2652, opened 2025-01-16
Update deps to match latest Knative versions

dsimansk:
# Changes

- Update deps to match latest Knative versions
- Update openshift-pipelines/pipelines-as-code to latest version v0.31.0
  - This change enabled drop of `client-go` replace, but added replace for different library it requires. Anyway, still a bit clean wrt/ `k8s.io` deps. 

/cc @matejvasek 

/kind <kind>

Fixes #

**Release Note**

```release-note
Update go modules to match Knative versions
```


go.mod
```diff
@@ -3,7 +3,7 @@ module knative.dev/func
 go 1.23
 
 // this is required bacause of bad dep in github.com/openshift-pipelines/pipelines-as-code
-replace k8s.io/client-go => k8s.io/client-go v0.29.3
+replace github.com/imdario/mergo => dario.cat/mergo v1.0.1
```
matejvasek, line 6: Is the comment above still valid?

dsimansk: Yes, it's used in and due to `pipeline-as-code`. 
ref: https://github.com/openshift-pipelines/pipelines-as-code/blob/v0.31.0/go.mod#L67

PR #2653, opened 2025-01-16
Remove usage of ttl.sh

matejvasek:
* Removed usage of ttl.sh registry in the tests.
* Fixed detection of hostname resolution failure.

### DIFF
--- a/.github/workflows/test-e2e-oncluster.yaml
+++ b/.github/workflows/test-e2e-oncluster.yaml
@@ -30,6 +30,7 @@ jobs:
       - name: E2E On Cluster Test
         env:
           E2E_RUNTIMES: ""
+          E2E_REGISTRY_URL: registry.default.svc.cluster.local:5000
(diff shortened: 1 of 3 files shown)


matejvasek:
PTAL @jrangelramos

matejvasek:
It looks like `Func E2E OnCluster Test / On Cluster Test (ubuntu-latest) (pull_request)` unfortunately does run also one *non-on-cluster* test which fails with the registry visible only from within the cluster.

jrangelramos:
@matejvasek The test mentioned is the "TestFromCliBuildLocal". It cover the case which func.yaml indicates the build should be on-cluster but the user wants a local build and override this beavior from CLI `--remote false` (or ommited). An option would be to transfer this test to regular e2e test by replacing the build tag on the test file.

lkingland (approved):
/lgtm
(assuming the e2e works, which I retried)

matejvasek:

matejvasek:
PTAL @jrangelramos

matzew (commented):
/lgtm

PR #2654, opened 2025-01-16
Remove dependency on docker hub

matejvasek:
Removed dependency on 
* `index.docker.io/library/golang:1.23-alpine3.20` and
* `index.docker.io/library/alpine:latest`.

For future I consider making the image `scratch` or at least `distroless` but for now it's still `alpine`.
Also we might consider skipping the `builder` part -- we could build the binaries locally and then use `COPY`.

Also I hate the fact that the `ADD` clause do not have `--from` flag, because of that I have to do fetching outside of the `Dockerfile`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -188,6 +188,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and push
         run: |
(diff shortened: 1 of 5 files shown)


matejvasek:
@lkingland @matzew @dsimansk PTAL

lkingland (approved):
/lgtm

PR #2664, opened 2025-01-22
fix: capitalise kind broker and bump versions in script

gauron99:
beginning with eventing 1.17, tests now fail -> kind type of eventing CRD was never capitalised (how did this work?) and eventing now requires 1.30+ k8s version -> bumping

### DIFF
--- a/hack/allocate.sh
+++ b/hack/allocate.sh
@@ -26,7 +26,7 @@ source "$(dirname "$(realpath "$0")")/common.sh"
 set_versions() {
   # Note: Kubernetes Version node image per Kind releases (full hash is suggested):
   # https://github.com/kubernetes-sigs/kind/releases
-  kind_node_version=v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822606ea5ca245
+  kind_node_version=v1.32.0@sha256:c48c62eac5da28cdadcf560d1d8616cfa6783b58f0d94cf63ad1bf49600cb027
   knative_serving_version="v$(get_latest_release_version "knative" "serving")"
   knative_eventing_version="v$(get_latest_release_version "knative" "eventing")"
   contour_version="v$(get_latest_release_version "knative-extensions" "net-contour")"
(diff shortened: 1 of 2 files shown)


hack/allocate.sh
```diff
@@ -305,7 +305,7 @@ namespace() {
   # Default Broker
   $KUBECTL apply -f - <<EOF
   apiVersion: eventing.knative.dev/v1
-  kind: broker
+  kind: Broker
```
matzew, line 308: This should have never been `broker` :sweat_smile: 

Is this now failing, b/c a newer Kube version is less lenient, and hence failing?

matejvasek: It most likely fails because of new eventing release.

hack/install-binaries.sh
```diff
@@ -71,7 +71,7 @@ warn_architecture() {
 
 install_kubectl() {
     echo '=== kubectl'
-    curl -sSLo "${bin}"/kubectl "https://storage.googleapis.com/kubernetes-release/release/v$kubectl_version/bin/linux/${ARCH}/kubectl"
+    curl -sSLo "${bin}"/kubectl "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
matejvasek, line 74: This does not respect kubectl_version and ARCH envvars.

gauron99: im just testing if this works at all just in case

hack/install-binaries.sh
```diff
@@ -71,7 +71,7 @@ warn_architecture() {
 
 install_kubectl() {
     echo '=== kubectl'
-    curl -sSLo "${bin}"/kubectl "https://storage.googleapis.com/kubernetes-release/release/v$kubectl_version/bin/linux/${ARCH}/kubectl"
+    curl -sSLo "${bin}"/kubectl "https://dl.k8s.io/v${kubectl_version}/bin/linux/${ARCH}/kubectl"
```
gauron99, line 74: higher versions of kubectl are now in different place

PR #2672, opened 2025-01-27
Go tar

matejvasek:
# Changes

* Get rid of dependency on alpine's `sh`/`tar` by implementing extraction in func-util binary. This will allow us to make func-util image `FROM scratch` or base it on google's distroless image.
* Fixed bug in `podReady()` where the function did not detect premature exit of a container.


matejvasek:

lkingland:
> * This will allow us to make func-util image `FROM scratch` or base it on google's distroless image.

🎉 
/lgtm

lkingland (approved):

PR #2674, opened 2025-01-29
[WIP] Make func-util image "FROM scratch"

matejvasek:


matejvasek:
Some more work needs to be done here!

matejvasek:
I'll have to reimplement https://github.com/knative/func/blob/8815c53e79d44ec9e6c50eb2e4edb1afdb42aa2d/pkg/pipelines/tekton/tasks.go#L310-L338 in go binary.

PR #2675, opened 2025-01-29
Set platform of builder image

matejvasek:
Previously I thought the platform does not matter for 'scratch'. However the image build is failing in CI.
It worked on my machine with both podman and moby.


lkingland (approved):
/lgtm

PR #2677, opened 2025-01-29
Have static knative component versions in hack/, update in automatic PR

gauron99:
## Whats changing
- Remove the functionality that components are always fetched as `latest` and instead have a static version that is periodically checked via Cron job in Workflow and in case a new latest exists == create a PR with that change against `main branch` programmatically. 
- Creating a new `component-versions.sh` file that will hold all of these versions that we need in repo (in `hack/`) and can be this file can be changed via the workflows.
- creating a subdir `hack/cmd` that now holds all the go scripts and creates the structure moving forward 
structure is `cmd/utility/main.go` where utility is `update-builder` or `update-knative-components` etc.
Scripts are to be run with `go run` if being run locally, otherwise they are mainly used in workflows programatically

- minor changes regarding the `go-github` version to `v68` to to update deprecated functions
- Tests  https://github.com/knative/func/issues/2687 for this current change

## Followup intention
- New workflow that runs with nightly eventing and serving so we can check against latest earlier and catch possible errors (such as version missmatches etc.) sooner.

fixes https://github.com/knative/func/issues/2687


hack/update-knative-components.go
```diff
@@ -0,0 +1,275 @@
...
+			return false, fmt.Errorf("error while updating file: %s", err)
+		}
+		return true, nil
+	}
+	return false, nil
+}
+
+// prepare branch for PR via git commands
+func prepareBranch(branchName string) error {
+	fmt.Print("> prepare branch...")
```
matejvasek, line 114: Plz do this via single exec.

gauron99:
PTAL @lkingland @matejvasek

hack/update-knative-components.go
```diff
@@ -0,0 +1,258 @@
...
+	if v == "" {
+		return "", fmt.Errorf("internal error: returned latest release name is empty for '%s'", repo)
+	}
+	fmt.Println("done")
+	return v, nil
+}
+
+// Read the file where components versions are located. Fetch their version
+// and return them in 'v1.23.0' format (unquoted).
+func getVersionsFromFile() (srv string, evt string, ctr string, err error) {
```
lkingland, line 46: One of the benefits of being in Go now is that we can use builtin structure serialization:

```Go
type Versions struct {
	Eventing string
	Serving  string
	Contour  string
}

func ReadVersions(filename string) (v Versions, err error) {
	data, err := os.ReadFile(filename)
	if err != nil {
		return
	}
	err = json.Unmarshal(data, &v)
	return
}

```

versions.json:
```json
{
    "Eventing": "v1.2.3",
    "Serving": "v2.2.3",
    "Contour": "v3.2.3"
}
```

hack/update-knative-components.go
```diff
@@ -0,0 +1,258 @@
...
+		// if all values are acquired, no need to continue
+		if srv != "" && evt != "" && ctr != "" {
+			break
+		}
+	}
+	return
+}
+
+// try updating the version of component named by "repo" via 'sed'
+func tryUpdateFile(prefix, newV, oldV string) (bool, error) {
```
lkingland, line 92: Another benefit of moving to Go is that we can treat the source code as the canonical source of truth from which other files are derived rather than having to parse through and munge them:

```Go
const versionsScriptTemplate = `#!/usr/bin/env bash

# AUTOGENERATED FILE.

set_versions() {
  knative_serving_version="{{.Serving}}"
  knative_eventing_version="{{.Eventing}}"
  contour_version="{{.Contour}}"
}
`

func WriteVersionsScript(versions Versions, filename string) error {
	tmpl, err := template.New("versions").Parse(versionScriptTemplate)
	if err != nil {
		return err
	}

	f, err := os.Create(filename)
	if err != nil {
		return err
	}
	defer f.Close()

	if err := tmpl.Execute(f, versions); err != nil {
		return err
	}
	if err := os.Chmod(VersionsScript, 0755); err != nil {
		return fmt.Errorf("failed to make %v executable: %w", filename, err)
	}
	return nil
}
```

hack/update-knative-components.go
```diff
@@ -0,0 +1,258 @@
...
+			prefix = knSrvPrefix
+		case "eventing":
+			oldV = oldEvt
+			prefix = knEvtPrefix
+		case "net-contour":
+			oldV = oldCntr
+			prefix = knCtrPrefix
+		}
+		// try and overwrite the file with new versions
+		isNew, err := tryUpdateFile(prefix, newV, oldV)
```
lkingland, line 233: Here, the versions struct would be updated with the latest versions, and then serialized out (overwriting versions.json), followed by regenerating the shell script.

lkingland (commented):
Looking good!  See comments inline where I think perhaps leaning on Go's struct serialization and templating could remove some complexity of file parsing and sed-fu, respectively.

hack/update-knative-components.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	return false, nil
+}
+
+// ----------------------------------------------------------------------------
+// ----------------------------------- MAIN -----------------------------------
+// ----------------------------------------------------------------------------
+
+// entry function -- essentially "func main() for this file"
+func updateComponentVersions() error {
```
lkingland, line 170: Putting a "main" at the bottom of a file is a relic of a bygone era where symbols had to be defined before they were used, essentially forcing a file to be written upside-down.

In compiled languages, it is not necessary, allowing us to write code more like an essay, with the opening paragraph being the highest-level instruction set (or table of contents, if you like), which is then expanded upon in a depth-first-search way as the file continues, with ancillary/misc/util functions relegated to the bottom.

For readability, putting this function first in the file (just after  constants and global variables) is a great way to introduce a reader to the purpose of the code, slowly resolving into more detail as one reads on.

hack/update-knative-components.go
```diff
@@ -0,0 +1,258 @@
...
+	prTitle := "chore: Update components' versions to latest"
+	ctx := context.Background()
```
matejvasek, line 172: ctx should  be set as in updateBuilder

https://github.com/knative/func/pull/2677/files#diff-2383360e2fb8ba21204dc95c1ac97c10b88cb8e8eacaaa22bfa88ba139a4483aR50-R60

hack/update-knative-components.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+		return true, nil
+	}
+	return false, nil
+}
+
+// prepare branch for PR via git commands
+func prepareBranch(branchName string) error {
+	fmt.Println("> preparing branch")
+	cmd := exec.Command("bash", "-c", fmt.Sprintf(`
```
matejvasek, line 115: Might be useful to set command stdio
```go
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
```

matejvasek: for debugging purposes if anything fails.

hack/update-knative-components.go
```diff
@@ -0,0 +1,258 @@
...
+		return err
+	}
+	fmt.Println("ready")
+	return nil
+}
+
+// returns true when PR with given title already exists in knative/func repo
+// otherwise false. Returns an error if occurred, otherwise nil.
+func prExists(ctx context.Context, c *github.Client, title string) (bool, error) {
+	opt := &github.PullRequestListOptions{State: "open"}
```
matejvasek, line 151: Technically you should iterate over multiple "pages", but chances are the PR is in the first "page" (first 30 entries).

gauron99: exactly I dont think we are going to run into a 'second page PR' with this since we will be merging it 90% of the time without fail and the rest with some version bumps probably

hack/update-knative-components.go
```diff
@@ -0,0 +1,276 @@
...
+	Serving  string
+	Eventing string
+	Contour  string
+}
+
+const versionsScriptTemplate = `#!/usr/bin/env bash
+
+# AUTOGENERATED FILE
+
+set_versions() {
```
matejvasek, line 31: What about the KinD version?

gauron99: thanks for pointing it out! I will add it accordingly

gauron99:
Ill add a new line at the end of Marshalling the json so reviewdog doesnt return an error

hack/component-versions.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
lkingland, line 1: Maybe re-generate this file using the Go code, that way it has the "generated file" warning here.

hack/update-knative-components.go
```diff
@@ -0,0 +1,282 @@
...
+// all the components that are kept up to date
+type Versions struct {
+	Serving  string
+	Eventing string
+	Contour  string
+}
+
+const versionsScriptTemplate = `#!/usr/bin/env bash
+
+# AUTOGENERATED FILE - edit versions in ./component-versions.yaml.
```
lkingland, line 29: This would be:
```suggestion
# AUTOGENERATED FILE - edit versions in ./component-versions.json.
```

hack/update-knative-components.go
```diff
@@ -0,0 +1,282 @@
...
+		newV, err := getLatestVersion(ctx, client, p.owner, p.repo)
+		if err != nil {
+			return fmt.Errorf("error while getting latest v of %s/%s: %v", p.owner, p.repo, err)
+		}
+
+		// try to overwrite with possibly new versions
+		if updateVersion(&v, p.repo, newV) {
+			// if any of the files are updated, set true
+			updated = true
+		}
```
lkingland, line 95: This is quite pedantic, but what the hay!

It's generally suggested as a Go idiom to return updated structures rather than mutate a reference unless there's a good reason for a ptr.

for example we `myslice = append(myslice, something)` rather than `append(&myslice, something)`.  Like so:

```suggestion
	v, updated = updateVersion(v, p.repo, newV)
```
(and the `if` can be collapsed into an assignment)

hack/update-knative-components.go
```diff
@@ -0,0 +1,282 @@
...
+	"os/exec"
+	"time"
+
+	github "github.com/google/go-github/v68/github"
+)
+
+var (
+	fileScript = "hack/component-versions.sh"
+	fileJson   = "hack/component-versions.json"
+)
```
lkingland, line 18: These should probably be constants

hack/update-knative-components.go
```diff
@@ -0,0 +1,282 @@
...
+
+// all the components that are kept up to date
+type Versions struct {
```
lkingland, line 21: I think this struct needs `KindNode` as well

hack/update-knative-components.go
```diff
@@ -0,0 +1,282 @@
...
+	// gets updated programatically via workflow -> PR creation
+	knative_serving_version="{{.Serving}}"
+	knative_eventing_version="{{.Eventing}}"
+	contour_version="{{.Contour}}"
+}
+`
+
+// entry function -- essentially "func main() for this file"
+func updateComponentVersions(ctx context.Context) error {
+	prTitle := "chore: Update components' versions to latest"
```
lkingland, line 47: This might be good as a constant as well

lkingland (commented):
Looking good!
The only bug I see is a missing `KindNode` member of the struct.  Other comments are just suggestions 👍🏻

hack/update-knative-components.go
gauron99: not sure how yaml got there, thanks

hack/update-knative-components.go
gauron99: ok and here I thought doing it "without the extra return" would be the better way :smile: 
thanks

.github/workflows/update-knative-hack.yaml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-go@v5
+      with:
+        go-version: 1.23.4
+    - name: Run script
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      run: cd hack && go run ./cmd/update-knative-components/main.go
```
lkingland, line 23: If youhave this run the make target, it reinforces that workflow

Makefile
```diff
@@ -269,3 +269,14 @@ schema-check: ## Check that func.yaml schema is up-to-date
 	(echo "\n\nFunction config schema 'schema/func_yaml-schema.json' is obsolete, please run 'make schema-generate'.\n\n"; rm -rf schema/func_yaml-schema-previous.json; exit 1)
 	rm -rf schema/func_yaml-schema-previous.json
 
+######################
+##@ Hack scripting
+######################
+
+.PHONY: regenerate-kn-components
+regenerate-kn-components: # Regenerate hack/component-versions(.json/.sh) locally
+	cd hack && go run ./cmd/update-knative-components/main.go "generate"
```
lkingland, line 278: You can omit the `/main.go` suffix I believe.

Makefile
```diff
@@ -269,3 +269,14 @@ schema-check: ## Check that func.yaml schema is up-to-date
 	(echo "\n\nFunction config schema 'schema/func_yaml-schema.json' is obsolete, please run 'make schema-generate'.\n\n"; rm -rf schema/func_yaml-schema-previous.json; exit 1)
 	rm -rf schema/func_yaml-schema-previous.json
 
+######################
+##@ Hack scripting
+######################
+
+.PHONY: regenerate-kn-components
+regenerate-kn-components: # Regenerate hack/component-versions(.json/.sh) locally
```
lkingland, line 277: We have a make target `help` which prints an index of targets.  Entries are added by the double-hash prefix on the comment `## Regenerate hack/component-versions`.  Perhaps worth adding the extra hash so that the help index shows this new target

lkingland (commented):
A couple more tiny suggestions

lkingland (approved):
Such a good improvement

lkingland (approved):

PR #2681, opened 2025-01-31
Create test symlinks dynamically

matejvasek:
Some tools dislike having questionable symlinks in git repository so we must create this symlinks dynamically in the test instead of having it committed into the repository.


matejvasek:
PTAL @lkingland @gauron99

gauron99 (approved):

PR #2685, opened 2025-02-04
Python Middleware v2 Scaffolding

lkingland:
# Changes

- :gift: Host builder now supports Python Functions
- :gift: Can now `func run` Python Functions locally outside of a container (no container runtime required)
- :gift: Python Functions now suppport instanced method signatures
- :gift: Python Functions now support lifecycle events (start, stop, readiness, liveness)
- :gift: Updates Python Template Tests to use pytest
- :gift: Python Functions now use the ASGI Spec and run using an ASGI optimized server
- :gift: adds support for base images to the OCI builder

- :broom: Cleans up .gitignore
- :broom: Cleans up and simplifies the OCI builder
- :broom: Cleans up and simplifies the repository management code
- :broom: removes broken interactive prompts from the `run` command pending forthcoming UX update

/kind enhancement

Fixes #

**Release Note**

```release-note
Python Functions now use the ASGI specification for method signature.
Python Functions now support instances and lifecycle events.  See the new templates for details.
The Host builder now can build and run Python functions locally without a container.

```

**Docs**

```docs

```


pkg/functions/runner.go
```diff
@@ -165,6 +167,106 @@ func runGo(ctx context.Context, job *Job) (err error) {
 	return
 }
 
+func runPython(ctx context.Context, job *Job) (err error) {
+	if job.verbose {
+		fmt.Printf("cd %v\n", job.Dir())
+	}
+
+	// TODO: should we allow the python binary location to be specified with
```
matzew, line 175: how was that done before?

matzew:
nice for getting this started

pkg/functions/runner.go
lkingland: These commands presume the binaries are available in `PATH`, and use the first match.  This is common.  However, a nice feature offered by some is to be able to override this on a case-by-case basis by setting environment variables such as `MYAPP_PYTHON=/path/to/python`.   We also support this in some cases, but not consistently throughout.  This might be a #good-first-issue

pkg/oci/builder_test.go
```diff
@@ -44,9 +44,33 @@ func TestBuilder_Build(t *testing.T) {
...
+
+// TestBuilder_BuildPython ensures that, when given a Python Function, an
+// OCI-compliant directory structure is created on .Build in the expected path.
+func TestBuilder_BuildPython(t *testing.T) {
+	root, done := Mktemp(t)
+	defer done()
+
+	client := fn.New(fn.WithVerbose(true))
+
+	f, err := client.Init(fn.Function{Root: root, Runtime: "python"})
```
matzew, line 60: Looks like the unit test is failing on this

```
--- FAIL: TestBuilder_BuildPython (0.72s)
    builder_test.go:68: exec: "python": executable file not found in $PATH
```

I guess the CI server does need updates?

templates/python/cloudevents/pyproject.toml
```diff
@@ -0,0 +1,26 @@
+[project]
```
matzew, line 1: this is for `poetry`, right? 

Would it make sense to have a `requirements.txt` as well - for pip? 
Or would this not work? 

I am not really an expert there, but would perhaps like in some section mentioned why `pyproject` / `poetry` vs `pip` ?

matzew:
ON, https://github.com/lkingland/func/tree/scaffolding-python/templates/python 
can we have a `README.md` ? with a bit of overview?

.github/workflows/test-e2e-runtime.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     continue-on-error: true
     strategy:
       matrix:
-        runtime: ["node", "go", "python", "quarkus", "springboot", "typescript", "rust"]
```
matejvasek, line 15: Why has been python removed here?

templates/python/cloudevents/pyproject.toml
lkingland: Yes this fully works, and is supported by `pip`.  This method, using a pyproject.toml appears to be the widely supported current standard amongst build systems.  Also, while the `func-pyuthon` repository (middleware) uses poetry (so we can take advantage of some of its more advanced features) the template written out for Function authors, defined here, uses the more widely used (and less complex to setup) hatchling build system.  This can be swapped out by the Function author by editing the `build-system` section of the pyproject.toml.  IIUC this is pretty widely understood by python developers.

.github/workflows/test-e2e-runtime.yaml
lkingland: Because the Pack builder used in the E2Es is not supported in the new middleware.  I suspect we'll have back-portable support added in the next few days.

pkg/builders/s2i/assemblers.go
```diff
@@ -55,11 +82,156 @@ else
...
+// such that the the script executes from subdirectory .s2i/builds/last
+// (where main resides) rather than the root, and indicates the main is
+// likewise in .s2i/builds/last/service/main.py via Procfile.  See the comment
+// inline on line 50 of the script for where the directory change instruction
+// was added.
+const PythonAssembler = `
+#!/bin/bash
+
+function is_django_installed() {
+  python -c "import django" &>/dev/null
```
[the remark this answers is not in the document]
lkingland: This is because this assembler script is a copy of the one from the official S2I image.  If rewritten, even if much smaller, it would still be "new code" and require more testing.  This copy, however, is identical with the only difference being adding a generated app.sh and a change of active working directory in the section marked `MODIFICATIONS FOR FUNCTIONS`.  This should also make it easy to keep fresh.

gauron99 (approved):
/lgtm

pkg/oci/go_builder.go
```diff
@@ -45,20 +63,23 @@ func buildGoLayer(cfg *buildConfig, p v1.Platform) (desc v1.Descriptor, layer v1
 	desc.Platform = &p
 
 	// Blob
-	blob := path(cfg.blobsDir(), desc.Digest.Hex)
+	blob := filepath.Join(cfg.blobsDir(), desc.Digest.Hex)
 	if cfg.verbose {
 		fmt.Printf("mv %v %v\n", rel(cfg.buildDir(), target), rel(cfg.buildDir(), blob))
 	}
 	err = os.Rename(target, blob)
```
matejvasek, line 70: @lkingland you are ignoring the `err` here.

PR #2686, opened 2025-02-04
Optimise func-utils image

matejvasek:
# Changes

* The `func-utils` image is now based on `scratch`. There is only our binary (and symlinks to it) in the image.
* Adjusted several tekton task to work *without* `/bin/sh` present in the `func-utils` image.
  * Replaced `script` with `command` in the tasks using the `func-utils` image.
  * Implemented new `s2i-generate` command that encompasses functionality previously implemented in `script` section of task specification.
* Changed default tag from `func-utils:latest` to `func-utils:v2` for sake of backward compatibility.
We do not want to break existing `func` installations that refer to `func-utils:latest` and do require `/bin/sh` being present there. Alternatively we could have new image `func-utils-v2:latest`.

You may find reviewing better on commit by commit basis.


matejvasek:
PTAL @lkingland @matzew @gauron99

gauron99 (approved):
this looks good to me but will
/hold for more qualified folks to take a look

lkingland (approved):
/lgtm
/approve

matejvasek:
@dsimansk could you try override `codecov/project` please?

matzew (approved):
/lgtm
/approve

PR #2701, opened 2025-02-18
fix: Update depracted function to fix failing CI-check

gauron99:


gauron99:
@matzew PTAL

matejvasek:
hmm tide says this is in merge pool.

PR #2713, opened 2025-02-24
fix: updates node template to ignore node_modules in .gitignore

### LINKED ISSUE #2557: Node template should ignore `node_modules` in `.gitignore`
A new Node function doesn't include `node_modules` in `.gitignore`

KapilSareen:
## Changes

- Updated the `ensureRunDataDir` function in `client.go` to check for  runtime and conditionally add `node_modules` to `.gitignore`.

/kind enhancement

Fixes #2557


KapilSareen:
> IMO it would be better that just put .gitignore into the template under ./templates directory.

That makes sense, but I wasn't sure if it would be optimal since it would require maintaining separate `.gitignore` files in two `node` template folders. Let me know if you think this approach is better, I’ll refactor accordingly.

lkingland:
I agree. Let's discuss a bit further on the Functions call tonight. Perhaps we append. 

/lgtm
/approve

lkingland (approved):

lkingland:
Oops, I was a bit too quick on the "approve"

gauron99:
@KapilSareen did you try this?
> To test, you'll need to rebuild the embedded filesystem, which happens when you run make

I believe you just need to run `make` in the root dir

KapilSareen:
> @KapilSareen did you try this?
> 
> > To test, you'll need to rebuild the embedded filesystem, which happens when you run make
> 
> I believe you just need to run `make` in the root dir You should have a new "zz" file with changes in your local repository that will need to be included in this PR

Yes, I did try that, but the PR showed conflicts, so I wasn’t sure if it was the right approach.

Anyway, I’m pushing it again so you can check. Let me know if anything needs to be fixed!

gauron99:
@KapilSareen You can simply remove the zz file, regenerate it again (using `make`) and there should be no conflicts. There was probably a different PR merged with a zz file (like automatic dependency version bumps etc.) So you just need to regenerate it again to match your code now

PR #2715, opened 2025-02-24
Bug: Fixes Incorrect Implementation of --file Option

### LINKED ISSUE #2545: Bug: Incorrect Implementation of `--file` Option in `kn func invoke`
#### Summary
The implementation of the `--file` option in the `kn func invoke` command encodes the file content as a base64 string, but it is not decoded before being sent, resulting in errors when invoking functions via HTTP or CloudEvents.

#### Steps to Reproduce
1. Create a JSON file, e.g., `test.json`, with the following content:
    ```json
    {
        "id": 123
    }
    ```
2. Invoke the function with the command:
    ```bash
    kn func invoke --file ./test.json --content-type application/json --format http -v
    ```
3. The output shows:
    ```
    Sending values
      ID: 
      Source: /boson/fn
      Type: boson.fn
      ContentType: application/json
      Data: ewogICAgImlkIjogMTIzCn0=
    ```
4. Similarly, invoke the function using CloudEvents format:
    ```bash
    kn func invoke --file ./test.json --content-type application/json --format cloudevent -v
    ```
5. The resulting error is:
    ```
    Error: invalid character 'e' looking for beginning of value
    ```

#### Analysis
The sequence of operations in `func/cmd/invoke.go` shows that when `--file` is specified, the file content gets read and encoded as a base64 string (`m.Data`), but this encoded data is not being decoded before being sent in the request. This inconsistency leads to the errors observed.

Relevant code snippet from `func/cmd/invoke.go`:
```go
if cfg.File != "" {
    content, err := os.ReadFile(cfg.File)
    if err != nil {
        return err
    }
    m.Data = base64.StdEncoding.EncodeToString(content)
}
```

The data is sent without decoding in the `sendEvent` and `sendPost` functions located in `func/pkg/functions/invoke.go`. For example, in the `sendEvent` function, the decoding code is as follows:

```go
err = json.Unmarshal([]byte(m.Data), &d)
```

At this point, `m.Data` is still a base64 string (e.g., `ewogICAgImlkIjogMTIzCn0=`), which cannot be decoded into JSON.

#### Expected Behavior
When using the --file option, the command should read the content of the specified file and send it in the same format as if the --data option had been provided. 

#### Environment
- Knative version: 
```bash
* Serving
  - serving.knative.dev/v1 (knative-serving v0.42.1-0.20240820122005-5f5f6d820b03)
* Eventing
  - sources.knative.dev/v1 (knative-eventing v0.42.1-0.20240828134450-34f9cd384dea)
  - eventing.knative.dev/v1 (knative-eventing v0.42.1-0.20240828134450-34f9cd384dea)
 ```

### PR DESCRIPTION
**Pull request by KapilSareen**
## Changes

- removes base64 encoding of request in `invoke` command  

/kind bug  

Fixes #2545  

## Testing

<div align="center">
  
  
</div>

### DIFF
--- a/cmd/invoke.go
+++ b/cmd/invoke.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"encoding/base64"
 	"fmt"
 	"os"
 	"strings"
@@ -173,7 +172,7 @@ func runInvoke(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 		if err != nil {
 			return err
 		}
-		m.Data = base64.StdEncoding.EncodeToString(content)
+		m.Data = string(content)
 	}
 
 	// Invoke

### CODE REVIEW

**Comment by lkingland** [comment]
If the message's Data isn't supposed to be base64 encoded, perhaps the fact that the file is being read and base64 encoded is the problem.  Could we just remove the encoding here?
https://github.com/knative/func/blob/main/cmd/invoke.go#L176

**Comment by KapilSareen** [comment]
> If the message's Data isn't supposed to be base64 encoded, perhaps the fact that the file is being read and base64 encoded is the problem. Could we just remove the encoding here? https://github.com/knative/func/blob/main/cmd/invoke.go#L176

I initially thought base64 encoding would help when the request data contains special characters, but I’ve refactored the commit to remove it.

**Comment by KapilSareen** [comment]
@matejvasek any thoughts on this?

**Comment by matejvasek** [comment]
@KapilSareen sorry for late reply.  I think this PR is OK as it is. WRT base64 & non-ascii character: this should be handled by the Go CloudEvent library we use. We should not do it by ourselves.

Outside of scope of this PR: I would say that the `Data` member of  the `InvokeMessage` structure should be `[]byte` not `string`. I believe that with that change  we could remove gymnastic at https://github.com/knative/func/blob/701e258acf7d12e69a838404d364f41886b9c482/pkg/functions/invoke.go#L156-L168 and replace it with simple:
```go
	err = event.SetData(m.ContentType, []byte(m.Data))
	if err != nil {
		return "", fmt.Errorf("cannot set data: %w", err)
	}
```

BTW If you are interested  what is the result of marshaling try this code:
```go
package functions

import (
	"context"
	"fmt"
	"io"
	"net/http"
	"os"
	"strings"
	"testing"
)

func Test_sendEvent(t *testing.T) {
	msg := InvokeMessage{
		ID:          "abc",
		Source:      "xyz",
		Type:        "mnl",
		ContentType: "application/json",
		Data:        `{"message": "hello"}`,
		Format:      "cloudevent",
	}
	resp, err := sendEvent(context.Background(), "localhost:8000", msg, rt{}, false)
	if err != nil {
		t.Fatal(err)
	}
	t.Log(resp)
}

type rt struct {
}

func (r rt) RoundTrip(request *http.Request) (*http.Response, error) {
	_, _ = fmt.Fprintf(os.Stderr, "%+v\n", request.Header)
	_, _ = io.Copy(os.Stderr, request.Body)
	_, _ = fmt.Fprintln(os.Stderr)
	return &http.Response{
		StatusCode: 202,
		Body:       io.NopCloser(strings.NewReader("")),
	}, nil
}
```

matejvasek:
For little bit more high level test:
```go
package functions_test

import (
	"context"
	"fmt"
	"io"
	"net/http"
	"os"
	"strings"
	"testing"

	"knative.dev/func/cmd"
)

func Test_sendEvent(t *testing.T) {
	c := cmd.NewInvokeCmd(func(config cmd.ClientConfig, option ...fn.Option) (*fn.Client, func()) {
		return fn.New(fn.WithTransport(rt{})), func() {}
	})
	c.SetArgs([]string{
		"--target=https://dummy:8080/",
		"--id=my-id",
		"--type=my-type",
		"--source=my-source",
		"--type=application/json",
		`--file=testdata/data.json`,
		"--path=/tmp/tmp.pJNexNLnYL/go-test-private",
	})
	err := c.ExecuteContext(context.Background())
	if err != nil {
		t.Fatal(err)
	}
}

type rt struct {
}

func (r rt) RoundTrip(request *http.Request) (*http.Response, error) {
	_, _ = fmt.Fprintf(os.Stderr, "%+v\n", request.Header)
	_, _ = io.Copy(os.Stderr, request.Body)
	_, _ = fmt.Fprintln(os.Stderr)
	return &http.Response{
		StatusCode: 202,
		Body:       io.NopCloser(strings.NewReader("")),
	}, nil
}
```

matejvasek:
>WRT base64 & non-ascii character: this should be handled by the Go CloudEvent library we use. We should not do it by ourselves.

Also the library does base64 encoding only for "structured JSON encoding" but we use "binary encoding" where data is simple sent as HTTP body "as it is" without any transcoding.

KapilSareen:
> > WRT base64 & non-ascii character: this should be handled by the Go CloudEvent library we use. We should not do it by ourselves.
> 
> Also the library does base64 encoding only for "structured JSON encoding" but we use "binary encoding" where data is simplt sent as HTTP body "as it is" without any transcoding (no matter if it is json/binary/image).

Thanks for the great explaination, it really helped clarify my doubts!

> I would say that the Data member of the InvokeMessage structure should be []byte not string

Regarding this, would you like me to create an issue and start working on this?

PR #2733, opened 2025-03-04
Refactor data member of the InvokeMessage to be []byte

### LINKED ISSUE #2732: `Data` member of  the `InvokeMessage` structure should be `[]byte` not `string`
`Data` member of  the `InvokeMessage` structure should be `[]byte` not `string`. I believe that with that change  we could remove gymnastic at https://github.com/knative/func/blob/701e258acf7d12e69a838404d364f41886b9c482/pkg/functions/invoke.go#L156-L168 and replace it with simple:
```go
	err = event.SetData(m.ContentType, []byte(m.Data))
	if err != nil {
		return "", fmt.Errorf("cannot set data: %w", err)
	}
```

_Originally posted by @matejvasek in https://github.com/knative/func/issues/2715#issuecomment-2686007798_

KapilSareen:
## Changes
Changes the datatype of `Data` member of InvokeMessage to be of type `[]byte` as discussed in https://github.com/knative/func/pull/2715#issuecomment-2686007798

/kind enhancement

Fixes #2732

### DIFF
--- a/pkg/functions/invoke.go
+++ b/pkg/functions/invoke.go
@@ -3,7 +3,6 @@ package functions
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -30,8 +29,8 @@ type InvokeMessage struct {
 	Source      string
 	Type        string
 	ContentType string
-	Data        string
-	Format      string //optional override for function-defined message format
+	Data        []byte
+	Format      string // optional override for function-defined message format
 }
 
 // NewInvokeMessage creates a new InvokeMessage with fields populated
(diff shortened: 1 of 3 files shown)


KapilSareen:
@matejvasek, could you review this to confirm if it aligns with your intent?

pkg/functions/invoke.go
```diff
@@ -19,7 +18,7 @@ const (
 	DefaultInvokeSource      = "/boson/fn"
 	DefaultInvokeType        = "boson.fn"
 	DefaultInvokeContentType = "application/json"
-	DefaultInvokeData        = `{"message":"Hello World"}`
+	DefaultInvokeData        = `{"message": "Hello, World!"}`
```
matejvasek, line 21: Any reason to change this? The thing is that this change also requires to regenerate the generated doc files, that's why the checks are failing here.

matejvasek: Please revert this.

matejvasek:
This looks good.

PR #2739, opened 2025-03-07
Add a retry functionality to workflows

gauron99:
Add a retry functionality to our current allocate script in the workflows and the e2e runtime test which I saw were the most affected by the flakiness of (probably) not enough resources on the host the runners run on. 
This should ensure that less human resources are needed for maintenance of these tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -105,7 +105,24 @@ jobs:
       - name: Install Binaries
         run: ./hack/install-binaries.sh
       - name: Allocate Cluster
-        run: ./hack/allocate.sh
+        run: |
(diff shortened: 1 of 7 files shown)


lkingland (approved):
Seems fine to me.  I might put in a fairly long sleep, like several minutes, because I suspect the flakiness is due to resource exhaustion in the host (shared).  That would help ensure the retries are spread across a larger time, increasing the liklihood it's run during a lull in usage.

/hold for potentially adding a 5m sleep in that loop

lkingland (commented):
/lgtm
/hold for potential Matej commentary
/approve :shipit:

PR #2745, opened 2025-03-14
change k8s security varaibles

### LINKED ISSUE #2237: Cannot write into node_modules directory
When container starts up, we need to generate some additional code for specific library, that's located within `node_modules` (utilizing "prestart" script in `package.json`). 
After version 1.12.0, this is not possible as the directory does not seem writable for the running user. This used to work with version 1.11.0 and bellow.

After some debugging we have found out, that the ownership of the files have changed, which does not allow the running user to write into the directories.

This is output of `id` and `ls -la /layers` commands when using **v1.11.0**:
`uid=1000(cnb) gid=1000(cnb) groups=1000(cnb)`
```
drwxr-xr-x 1 cnb  cnb  4096 Jan  1  1980 .
drwxr-xr-x 1 root root 4096 Mar 19 09:39 ..
drwxr-xr-x 2 cnb  root 4096 Jan  1  1980 config
drwxr-xr-x 3 cnb  root 4096 Jan  1  1980 paketo-buildpacks_ca-certificates
drwxr-xr-x 3 cnb  root 4096 Jan  1  1980 paketo-buildpacks_node-engine
drwxr-xr-x 3 cnb  root 4096 Jan  1  1980 paketo-buildpacks_npm-install
drwxr-xr-x 3 cnb  root 4096 Jan  1  1980 sbom
```

With **v1.12.0** and above, we get the following output with different ownership (and uid of cnb user):
`uid=1002(cnb) gid=1000(cnb) groups=1000(cnb)`
```
drwxr-xr-x 1 1001 cnb  4096 Jan  1  1980 .
drwxr-xr-x 1 root root 4096 Mar 19 09:34 ..
drwxr-xr-x 2 1001 root 4096 Jan  1  1980 config
drwxr-xr-x 3 1001 root 4096 Jan  1  1980 paketo-buildpacks_ca-certificates
drwxr-xr-x 3 1001 root 4096 Jan  1  1980 paketo-buildpacks_node-engine
drwxr-xr-x 3 1001 root 4096 Jan  1  1980 paketo-buildpacks_npm-install
drwxr-xr-x 3 1001 root 4096 Jan  1  1980 sbom
```

It seems that the related changes have been done as part of https://github.com/knative/func/pull/1911.

We have also tried to utilize KNative security context and `"runAsUser": 1001`, but didn't work either. Is this expected behaviour? Is there anything else we are missing in our configuration? Thanks.

PinotNoir04:
# Changes

- Updated the values of runAsUser and runAsGroup to 1000

/kind bug

Fixes #2237

### DIFF
--- a/pkg/docker/creds/auth.json
+++ b/pkg/docker/creds/auth.json
@@ -1,0 +1,3 @@
+{
+    "credsStore": ""
+}
--- a/pkg/k8s/security_context.go
+++ b/pkg/k8s/security_context.go
@@ -13,8 +13,8 @@ func defaultPodSecurityContext() *corev1.PodSecurityContext {
 	if IsOpenShift() {
 		return nil
 	}
-	runAsUser := int64(1001)
-	runAsGroup := int64(1002)
+	runAsUser := int64(1000)
+	runAsGroup := int64(1000)
 	return &corev1.PodSecurityContext{
 		RunAsUser:  &runAsUser,
 		RunAsGroup: &runAsGroup,
--- a/pkg/oci/testdata/test-links/absoluteLink
+++ b/pkg/oci/testdata/test-links/absoluteLink
@@ -1,0 +1,1 @@
+/var/example/absolute/link
--- a/pkg/oci/testdata/test-links/absoluteLinkWindows
+++ b/pkg/oci/testdata/test-links/absoluteLinkWindows
@@ -1,0 +1,1 @@
+c://some/absolute/path


matejvasek:
What is the associated bug?

pkg/oci/testdata/test-links/absoluteLink
```diff
@@ -0,0 +1 @@
+/var/example/absolute/link
```
matejvasek, line 1: Theses links are deliberately created dynamically during test and not included in repo.

pkg/oci/testdata/test-links/absoluteLinkWindows
```diff
@@ -0,0 +1 @@
+c://some/absolute/path
```
matejvasek, line 1: Theses links are deliberately created dynamically during test and not included in repo.

pkg/docker/creds/auth.json
```diff
@@ -0,0 +1,3 @@
+{
+    "credsStore": ""
+}
```
matejvasek, line 3: Why was this file added?

pkg/oci/testdata/test-links/absoluteLink
PinotNoir04: sorry will remove them

pkg/oci/testdata/test-links/absoluteLinkWindows
PinotNoir04: sorry will remove them

pkg/docker/creds/auth.json
PinotNoir04: I think my lsp accidentally changed the line position or something and this was committed as an updated file as well. sorry will fix this.

pkg/k8s/security_context.go
```diff
@@ -13,8 +13,8 @@ func defaultPodSecurityContext() *corev1.PodSecurityContext {
 	if IsOpenShift() {
 		return nil
 	}
-	runAsUser := int64(1001)
-	runAsGroup := int64(1002)
+	runAsUser := int64(1000)
```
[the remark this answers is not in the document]
PinotNoir04: I checked the buildpack.yaml file and the runAsUser variable was already set to 1000 but the runAsGroup was set to 0 so I let it be. Should I update that too?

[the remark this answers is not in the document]
matejvasek: >I checked the buildpack.yaml

You are probably referring to old yaml file that exists only for backward compatibility reasons.
The current definition is here:
https://github.com/knative/func/blob/9c0c2161933d5387564c6dfcfd4b8eab2f2acd68/pkg/pipelines/tekton/tasks.go#L186-L190


Also check:
```sh
podman run -it --rm --network=host ghcr.io/knative/builder-jammy-tiny:latest id    
uid=1001(cnb) gid=1000(cnb) groups=1000(cnb)
```

matejvasek:
Could you please refer/describe bug you are trying to fix?
When the issue happens? What is your environment? What kind of cluster do you use?

matejvasek:
Is this bug for "on cluster" build or local build?

matejvasek:
Also the bug is quite old. Could you please verify it's still valid @PinotNoir04?

matejvasek:
Please add tests to PR that verify the bug in question was fixed.

matejvasek:
I think this might have been just some transitional issue, that happened when you created pipelines with one version of `func` but then latter re-run it with newer `func`.

pkg/k8s/security_context.go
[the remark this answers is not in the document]
matejvasek: Just to elaborate more: 1001 is paketo user for build image, 1002 is paketo user for run image.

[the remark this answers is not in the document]
matejvasek: Unfortunately I do not recall why the gui==0 in tekton task. Preferably it should be 1002.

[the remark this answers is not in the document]
matejvasek: So your change of gid might be OK. But the uid should stay 1002.

PinotNoir04:
> Could you please refer/describe bug you are trying to fix? When the issue happens? What is your environment? What kind of cluster do you use?

I didn't face it myself. I'm on an Arch system. I have been using a single node cluster through minikube mainly for experimentation. I am a novice.

PinotNoir04:
> Also the bug is quite old. Could you please verify it's still valid @PinotNoir04?

Ok I will ask them

PinotNoir04:
> I think this might have been just some transitional issue, that happened when you created pipelines with one version of `func` but then latter re-run it with newer `func`.
> 
> The issue would have been fixed back then by calling `kubectl delete pipeline --all`.

In that case should I close this MR? or just add the tests?

matejvasek:
Please just close the PR it if it not reproducible. 
If it is reproducible please continue on this and also include test that reproduces it.

PR #2748, opened 2025-03-17
Add testcase for Go private repositories

matejvasek:
Just added tests that are skipped for now. The actual functionality will be added later.


matejvasek:
PTAL @lkingland

PR #2749, opened 2025-03-17
fix: create custom http transport for each run for Auth tests

gauron99:
The unit test got flaky after the auth version bump (im pretty sure) so in attempt to remedy this flakyness I edited the auth test function.

Main changes>
- certificate generation gets its own function
- removed the http default transport hack which got created at the `startServer` and changed it into: create custom http transport on each run and pass it to the called function.

### DIFF
--- a/pkg/docker/creds/credentials_test.go
+++ b/pkg/docker/creds/credentials_test.go
@@ -92,7 +92,7 @@ func TestCheckAuth(t *testing.T) {
 		incorrectPwd = "badpwd"
 	)
 
-	localhost, localhostTLS := startServer(t, uname, pwd)
+	localhost, localhostTLS, cert := startServer(t, uname, pwd)
 
 	_, portTLS, err := net.SplitHostPort(localhostTLS)
 	if err != nil {
@@ -132,7 +132,6 @@ func TestCheckAuth(t *testing.T) {
 			},
 			wantErr: false,
 		},
-
 		{
 			name: "correct credentials non-localhost",
 			args: args{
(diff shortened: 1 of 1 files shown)


pkg/docker/creds/credentials_test.go
```diff
@@ -170,7 +171,30 @@ func TestCheckAuth(t *testing.T) {
...
+
+			transport.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+				h, p, err := net.SplitHostPort(addr)
+				if err != nil {
+					return nil, err
+				}
+				if h == "test.io" {
+					h = "localhost"
+				}
+				return dialer.DialContext(ctx, network, net.JoinHostPort(h, p))
```
gauron99, line 195: create custom transport and ca pool each run instead of mutating the http.defaultTransport

pkg/docker/creds/credentials_test.go
```diff
@@ -179,141 +203,136 @@ func TestCheckAuth(t *testing.T) {
...
-					resp.WriteHeader(http.StatusOK)
-				}
-				return
-			}
-		}
-		resp.WriteHeader(http.StatusUnauthorized)
-	})
-
+// generate Certificate
+func generateCert(t *testing.T) (tls.Certificate, *x509.Certificate) {
```
gauron99, line 214: move certificate gen to separate function and generate tls certificate

pkg/docker/creds/credentials_test.go
```diff
@@ -170,7 +169,30 @@ func TestCheckAuth(t *testing.T) {
 				Username: tt.args.username,
 				Password: tt.args.password,
 			}
-			if err := creds.CheckAuth(tt.args.ctx, tt.args.registry+"/someorg/someimage:sometag", c, http.DefaultTransport); (err != nil) != tt.wantErr {
+			// create trusted certificates pool and add our certificate
+			certPool := x509.NewCertPool()
+			certPool.AddCert(cert)
+
+			// client transport with the certificate
+			transport := &http.Transport{
```
gauron99, line 177: the `startServer` whichi previously used to handle the http transport mutation gets used by one more function `TestCheckAuthEmptyCreds` which still uses the default http transport instead of this one but doesnt seem to cause any problems? 
Perhaps `TestCheckAuthEmptyCreds` needs to have the Transport changed as well

PR #2751, opened 2025-03-19
Test adjustments

matejvasek:
* Use tiny instead of base BP builder (since it has smaller footprint).
* Decrease minimal required Go version (S2I images are on Go 1.22 supported by RH).
* Make the self-signed certificate CA==true so it works with `update-ca-trust` utility.


matejvasek:
PTAL @gauron99

gauron99 (approved):

PR #2753, opened 2025-03-19
Allow host mounts for build phahse

matejvasek:
This is in particular useful for using paket bindings, e.g. to inject git credentials into build process.

# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/pkg/functions/function.go
+++ b/pkg/functions/function.go
@@ -148,6 +148,14 @@ type BuildSpec struct {
 	// Image stores last built image name NOT in func.yaml, but instead
 	// in .func/built-image
 	Image string `yaml:"-"`
+
+	// Mounts used in build phase. This is useful in particular for paketo bindings.
+	Mounts []MountSpec `yaml:"mounts,omitempty"`
+}
+
+type MountSpec struct {
+	Source      string `yaml:"source"`
+	Destination string `yaml:"destination"`
 }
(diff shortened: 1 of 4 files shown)


pkg/functions/function.go
```diff
@@ -148,6 +148,14 @@ type BuildSpec struct {
 	// Image stores last built image name NOT in func.yaml, but instead
 	// in .func/built-image
 	Image string `yaml:"-"`
+
+	// Mounts used in build phase. This is useful in particular for paketo bindings.
+	Mounts []MountSpec `yaml:"mounts,omitempty"`
+}
+
+type MountSpec struct {
```
matejvasek, line 156: Alternative to this new type would be to reuse our `Volume` type and just add `Source *string` member to it.
WDYT @lkingland ?

matejvasek:
PTAL @lkingland @gauron99

matejvasek:
PTAL @lkingland @gauron99

pkg/functions/function.go
matejvasek: It could to a degree help on cluster and local build unification.

gauron99 (approved):
/lgtm
/hold for potential Lukes review

PR #2764, opened 2025-04-02
Use s2i in standard mode, not "as-dockerfile"

matejvasek:
Basically a revert of https://github.com/knative/func/pull/1063.

Today's podman  (+v4.9)  works all right.

### DIFF
--- a/pkg/builders/s2i/builder.go
+++ b/pkg/builders/s2i/builder.go
@@ -1,37 +1,21 @@
 package s2i
 
 import (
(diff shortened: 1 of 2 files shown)


matejvasek:
@jrangelramos if you have a time you could verify this works on RHEL podman, but I am pretty sure it should work.

matejvasek:

lkingland (approved):
/lgtm

PR #2765, opened 2025-04-02
Improvements for Go s2i assembler

matejvasek:
# Changes

- :bug: Return error code when the function was not compiled
- :bug: Run `go mod tidy` on scaffoded code

/kind bug

```release-note
fix: Go s2i build issue with user added dependencies
```


gauron99 (approved):
/lgtm
/hold unhold when you please

PR #2769, opened 2025-04-03
Fix: Go scaffolding uses correct module name

matejvasek:
Using incorrect name works for a functions with flat structure -- no sub-packages. When sub-packages are used we need to refer the user module by its true name.

```release
fix: allow sub-modules for Go functions
```


lkingland (approved):
/lgtm

PR #2773, opened 2025-04-07
Cleanup: Codecov update, unify flags and ignore unwanted files

gauron99:
# Changes
- :broom: Removes unnecessary flags/coverages (runtimes tests)
- :broom: remove unnecessary workaround file
- :broom: ignore dirs that dont contain code we want to cover (`.codecov.yaml` changes)
- Unify the flags so that we have only 3 clear ones -- `unit` `integration` `e2e`

/kind cleanup


.github/workflows/ci.yaml
```diff
@@ -187,6 +187,69 @@ jobs:
           verbose: true
           token: ${{ secrets.CODECOV_TOKEN }}
 
+  e2e-runtime-test:
```
gauron99, line 190: add runtime tests into CI (runs on push to main)

gauron99:
@matejvasek @lkingland PTAL

PR #2777, opened 2025-04-09
Fix buildpacks builder update task

matejvasek:
The paketo project moved some buildpack from gcr.io to docker.io.


matejvasek:
@lkingland override any failing tests please.

PR #2779, opened 2025-04-10
Refactor builder

matejvasek:
Just refactor. For simpler understanding review per commit.


matejvasek:
PTAL @gauron99 @lkingland

gauron99 (approved):

PR #2785, opened 2025-04-14
Refactor if-elif-else => switch

matejvasek:


matejvasek:
PTAL @gauron99 @lkingland

gauron99 (approved):

PR #2792, opened 2025-04-15
adding python ARM64 test on CI

jrangelramos:
# Changes

- :broom:  adding python ARM64 test on CI


matejvasek:
@lkingland when and how we fix this?

lkingland:
The Python runtime currently requires either the Host or S2I builder.

Issue #2516, opened 2024-09-24
func can't build with untrusted builder image


yozel:
We are trying to use our own builder image, and it fails with following error:

```
[detector] ERROR: failed to write group file: open /layers/group.toml: permission denied
```

We noticed that this only happens because our builder image is not trusted, which is a [hardcoded](https://github.com/knative/func/blob/783d04d0b0883463ef150d140d12531fcf257b8c/pkg/builders/buildpacks/builder.go#L47) list.

## Steps to reproduce
Create the function
```
knative-func $ func create -l go
```
It's successfully building with default builder image
```
knative-func $ func build --builder=pack --builder-image=ghcr.io/knative/builder-jammy-tiny:0.0.240
Building function image
Still building
Still building
Yes, still building
🙌 Function built: index.docker.io/tigerteam/knative-func:latest
```
Pull and create a tag for the default builder image
```
knative-func $ docker pull ghcr.io/knative/builder-jammy-tiny:0.0.240
knative-func $ docker tag ghcr.io/knative/builder-jammy-tiny:0.0.240 builder-jammy-tiny:0.0.240-local
```
Run build with the new tag
```
knative-func $ func build --builder=pack --builder-image=builder-jammy-tiny:0.0.240-local
Building function image
Error: executing lifecycle: failed with status code: 1
```
With verbose logging
```
knative-func $ func build --builder=pack --builder-image=builder-jammy-tiny:0.0.240-local -v
Building function image
Pulling image index.docker.io/library/builder-jammy-tiny:0.0.240-local
CheckReadAccess succeeded for the run image index.docker.io/paketobuildpacks/run-jammy-tiny:latest
Selected run image index.docker.io/paketobuildpacks/run-jammy-tiny:latest
Pulling image index.docker.io/paketobuildpacks/run-jammy-tiny:latest with platform linux/amd64
latest: Pulling from paketobuildpacks/run-jammy-tiny
Digest: sha256:fac4a3749284e198247f4ead26fd8ee2816c4db428ebb44fbfd19e6fef6309dc
Status: Image is up to date for paketobuildpacks/run-jammy-tiny:latest
Pulling image docker.io/buildpacksio/lifecycle:553c041 with platform linux/amd64
553c041: Pulling from buildpacksio/lifecycle
Digest: sha256:41ed46de4c426cd8462ae0e6fca8745f71432236f0c6aa6bfaa956b9d1704bcf
Status: Image is up to date for buildpacksio/lifecycle:553c041
[...]

PR #2819, opened 2025-04-29
Fix: do not override uid/git in buildpack builder

matejvasek:
Zero value implies override (to root presumably),
we must set it to negative value.

I really dislike that zero value is not neutral here.

fixes: #2516

```release-note
fix: error "failed to write group file" when using untrusted builder (#2516)
```


matejvasek:

lkingland (approved):
Setting to a negative value seems correct here.

If we wanted to use the pointer method to have nil indicate "use the default", we could, but then we have an explosion of nil checks and airball variable declarations to deal with.

PR #2831, opened 2025-05-14
test: fix python e2e http test

jrangelramos:
# Changes

- :broom: Fix python test.

Default python functions no longer echoes as response what it receives from request. It just returns "OK", so this fix python tests used on midstream


matejvasek:
@jrangelramos any ideas why the test is faling?

matejvasek:
/override "Func E2E OnCluster Test / On Cluster Test (ubuntu-latest)"

matejvasek:
@lkingland could you please override the action?

PR #2837, opened 2025-05-26
More granular image references

matejvasek:
# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/pkg/pipelines/tekton/tasks.go
+++ b/pkg/pipelines/tekton/tasks.go
@@ -5,7 +5,24 @@ import (
 	"strings"
 )
 
-var DeployerImage = "ghcr.io/knative/func-utils:v2"
+var (
+	FuncUtilImage = "ghcr.io/knative/func-utils:v2"
+	DeployerImage string
+	ScaffoldImage string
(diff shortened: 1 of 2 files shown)


pkg/pipelines/tekton/tasks.go
```diff
@@ -5,7 +5,24 @@ import (
 	"strings"
 )
 
-var DeployerImage = "ghcr.io/knative/func-utils:v2"
+var (
+	FuncUtilImage = "ghcr.io/knative/func-utils:v2"
```
matzew, line 9: where is the code for this container?

matejvasek: wdym?
Dockerfile is here: https://github.com/knative/func/blob/main/Dockerfile.utils

matejvasek: This is not a new image.

dsimansk:
/override "Func E2E OnCluster Test / On Cluster Test (ubuntu-latest)"

PR #2838, opened 2025-05-26
Fix quarkus buildpack image reference

matejvasek:
The paketo is moving to docker.io from gcr.io.


matejvasek:
@lkingland @dsimansk could you please override the failing actions? They are not relevant here.

PR #2843, opened 2025-05-26
Fix: build correct func-util binary for give arch

matejvasek:


matejvasek:
PTAL @jrangelramos @lkingland

matejvasek:
I don't know how I could have missed this.

PR #2852, opened 2025-06-01
Changed volume uploader pod image pull policy to "Always"

### LINKED ISSUE #2835: Tekton task "func-buildpacks" prepare/results steps  use non-multi-arch bash:5.1.4 image, causing exec format errors on ARM64
When using buildpack to build Go on an ARM64 host, the Tekton "prepare" and "results" steps fail with:

    Error executing command: fork/exec /tekton/scripts/script-0-....: exec format error

The problem is the image used for both steps:

https://github.com/knative/func/blob/a062cd28728aca02d162c9e1af3f32cf2ad1d5fa/pkg/pipelines/tekton/tasks.go#L87

https://github.com/knative/func/blob/a062cd28728aca02d162c9e1af3f32cf2ad1d5fa/pkg/pipelines/tekton/tasks.go#L193

After changing both images to the floating tag:

      image: docker.io/library/bash:5.1

buildpack work correctly.

luciantin:
# Changes
 
- :bug: Fix stale func-utils:v2 image usage by changing imagePullPolicy to "Always"

/kind bug

Fixes #2835 

Default image pull policy was "IfNotPresent", therefore the image would be pulled only once, unless manually changed, and builds would not pull new updated images.

Problem described in https://github.com/knative/func/issues/2835#issuecomment-2913756811.
Solved by changing the image pull policy of the volume-uploader pod only.

As the volume uploader pod runs before any tasks in https://github.com/knative/func/blob/main/pkg/pipelines/tekton/tasks.go, if the tasks use the default func-utils:v2 image, then those pods will also use the latest image.

**Release Note**

```release-note
Fixes a bug where remote tekton builds would use a stale image to upload the source directory.
```

### DIFF
--- a/pkg/k8s/persistent_volumes.go
+++ b/pkg/k8s/persistent_volumes.go
@@ -128,12 +128,13 @@ func runWithVolumeMounted(ctx context.Context, podImage string, podCommand []str
 			SecurityContext: defaultPodSecurityContext(),
 			Containers: []corev1.Container{
 				{
-					Name:       podName,
-					Image:      podImage,
-					Stdin:      true,
-					StdinOnce:  true,
-					WorkingDir: volumeMntPoint,
-					Command:    podCommand,
+					Name:            podName,
+					Image:           podImage,
+					ImagePullPolicy: corev1.PullAlways,
+					Stdin:           true,
+					StdinOnce:       true,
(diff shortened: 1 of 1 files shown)


lkingland:
Thanks for the contribution!

/ok-to-test

Issue #2806, opened 2025-04-23
AI Integration - MCP Server


lkingland:
Enable AI integration via an MCP server runnable via the command `func mcp`.

This includes full-coverage of Function features via the MCP, as well as context injection via MCP "Prompts", and integration with our official templates repository.

- [x] Python POC
- [x] Func MCP Server
- [x] Basic CRUDL Subcommands
- [x] Labels Support
- [x] Secrets Support
- [x] Volumes Support
- [x] Domain Mapping Support
- [x] Remote Builds
- [x] Model Context with Instructions
- [ ] Cursor Integration
- [ ] Windsurf Integration
- [ ] VSCode Integration
- [ ] Claude Integration
- [ ] Claude Code Integration
- [ ] Upstream Documentation
- [x] Screencast

Tasfia-Ara:
What kind of AI integration is it? I'd like to help!

lkingland:
Thanks Tasfia-Ara,

However, this integration is already actively in progress.  I assume there will be many opportunities to improve/add to it once complete.  Please keep an eye on our open issues!

xyq043170:
@lkingland @jcrossley3 May I ask what can be merged and published for use？

xyq043170:
@lkingland @jcrossley3 May I ask when it can be released for use？

lkingland:
Thanks for the implementation @KapilSareen 
This lays the ground-work for an ongoing investment in the Functions MCP integration

lkingland:
Re-opening beacuse there's a few documentation steps left to complete.

lkingland:
🎉

PR #2859, opened 2025-06-09
mcp: adds create and deploy tools

KapilSareen:
## Changes

This PR aims to add two new tools to `func`'s MCP server.

Related Epic https://github.com/knative/func/issues/2806


gauron99:
hi @KapilSareen, thanks for the new addition of mcp server youre working on!
Could we add some tests for these? Otherwise looks good to me
cc @lkingland

KapilSareen:
Yes, Thanks for pointing it out.
I already had a discussion with @lkingland and we're thinking of adding tests for these in a seperate PR.

gauron99:
ok great. I will leave this to him for review then. Sounds like you two got it covered!

lkingland:
Thanks @gauron99 
I wanted tests as well, but we have a small problem where the implementation here requires a compiled `func` binary, but that would mean the tests _have_ to be E2E tests...
So I am refactoring the E2E suite to support this, and we'll be adding the tests later in the process 👍🏻

PR #2876, opened 2025-06-24
feat: support FUNC_GIT env in Host builder

lkingland:
# Changes

- :gift: Adds FUNC_GIT env support to the Host builder

Allow specifying the path to git when using the Host builder via the FUNC_GIT environment variable.

/kind enhancement

**Release Note**

```release-note
The path to git can be altered from that in PATH by using the FUNC_GIT environment variable when using the host builder.
```

**Docs**

```docs

```


gauron99 (approved):
/lgtm

PR #2879, opened 2025-06-24
feat: add noninteractive path to func config envs remove

lkingland:
/kind enhancement

```release-note
func config remove now supports noninteractive usecases via a --name flag
```


gauron99 (approved):
/lgtm

PR #2880, opened 2025-06-24
feat: add LISTEN_ADDRESS to Python default runner

lkingland:
The "PORT" env is deprecated in favor of LISTEN_ADDRESS.  This change moves us further towards fully implementing that across the runtimes.  PORT is left for now for backwards compatibility.

/kind enhancement

### DIFF
--- a/pkg/functions/runner.go
+++ b/pkg/functions/runner.go
@@ -218,8 +218,9 @@ func runPython(ctx context.Context, job *Job) (err error) {
 	}
 
 	// Run
+	listenAddress := net.JoinHostPort(job.Host, job.Port)
 	if job.verbose {
-		fmt.Printf("PORT=%v ./.venv/bin/python ./service/main.py\n", job.Port)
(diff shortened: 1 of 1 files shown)


pkg/functions/runner.go
```diff
@@ -218,8 +218,9 @@ func runPython(ctx context.Context, job *Job) (err error) {
 	}
 
 	// Run
+	listenAddress := net.JoinHostPort(job.Host, job.Port)
 	if job.verbose {
-		fmt.Printf("PORT=%v ./.venv/bin/python ./service/main.py\n", job.Port)
+		fmt.Printf("PORT=%v LISTEN_ADDRESS=%v ./.venv/bin/python ./service/main.py\n", job.Port, listenAddress)
```
matejvasek, line 223: Will ./.venv/bin/python do all the job? Don't we need ./.venv/bin/activate?

matejvasek: Apparently this works, but how?

matejvasek: .venv/bin/python is merely a symlink to /usr/bin/python.
I guess python must have some logic checking arg[0].

PR #2881, opened 2025-06-24
fix: remove redundant initialized check

lkingland:
This initialized check is redundant, and leads to an error in a forthcoming E2E test case, and is thus removed here preemptively.

/kind cleanup


gauron99 (approved):
\lgtm

PR #2882, opened 2025-06-24
feat: add labels to func describe

lkingland:
Adds a "Labels" section to describe command's output.

/kind enhancement

Note that, as with the other `func config ...` commands, adding unit tests is not possible without a refactor.  The forthcoming E2E tests do test these features, and unit tests will be added when these subcommands are refactored to match the testability of the other commands.

```release-note
Function describe subcommand now includes labels.
```


gauron99 (approved):
/lgtm

PR #2883, opened 2025-06-24
feat: add noninteractive flow to volumes config

lkingland:
Adds flags for manipulating volumes via cli flags

Note that tests are not included because the commands were not developed in a way conducive to testing.  Tests are included in the forthcoming E2E test refactor, and will be included when the various config subcommands are refactored to match the architecture of other commands.

```release-note
Configuring volumes now supports noninteractive CLI flow
```


gauron99 (approved):
/lgtm

PR #2884, opened 2025-06-24
feat: check and warn on local to remote branches mismatch

lkingland:
Warn the user when the local branch is different than the configured remote build target branch (revision).

Example:

A user is working on the main branch locally, but wants to trigger a remote deployment of a specific branch feature-xyz.

$ func deploy --remote --git-branch feature-xyz

Here, a remote deployment is triggered which uses the Git repository and pulls code from feature-xyz

This will still use the func.yaml from the local main branch for function metadata (name, runtime, environment variables, etc.)

This can create a problem when there is a pertinent difference between the two brances' func.yaml.

The longer-term fix for this is to update the pipelines flow to pull the repository and reference indicated on demand.  This warning is a stop-gap.

The warning encourages the user to first checkout the feature-xyz branch locally before triggering the remote deployment, ensuring the function metadata matches the code being deployed.

/kind enhancement

```release-note
User will be warned when their local branch differs from that configured for remote builds.
```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


gauron99 (approved):
/lgtm

PR #2887, opened 2025-06-25
feat: add --address option to func run

lkingland:
Adds the --address flag to func run

/kind enhancement

```release-note
func run now supports the --address flag
```


cmd/run.go
```diff
@@ -68,9 +68,12 @@ EXAMPLES
 
 	o Run the function locally on the host with no containerization (Go only).
```
gauron99, line 69: OT: this can be expanded right?
```suggestion
	o Run the function locally on the host with no containerization (Go & Python only).
```

PR #2890, opened 2025-06-25
mcp: adds resources and prompts for func subcommands

KapilSareen:
## Changes 

This PR adds resources and prompts of the func subcommands `help` for the LLM context

Related Epic #2806


lkingland:
I really like this dynamic way of defining the prompts and resources.

I understand there may be some spotty support in Clients right now, despite it being part of the MCP spec.

If we can get it to work, great.  If not, we may need to do it the long/hard way (for now) and create a separate one for each command.  Perhaps that can be automated.

lkingland (approved):

PR #2897, opened 2025-06-30
feat: macos support for dev cluster scripts

lkingland:
Development cluster setup scripts now support MacOS.

/kind enhancement

```release-note
Local clusters can now be set up on MacOS via the repository's allocate.sh and registry.sh scripts.
```


matejvasek:
Does this mean we could enable integration test on MacOS?

PR #2898, opened 2025-06-30
feat: Python,Go func listen dual-stack

matejvasek:
# Changes

- :gift: Python and Go function listen dualstack.
           On Linux listening on `[::]` also listen on `0.0.0.0`.

/kind enhancement

```release-note
feat: Python and Go function listen dualstack
```


matejvasek:
@jrangelramos could you please test this on OCP?

matejvasek:
PTAL @lkingland

PR #2901, opened 2025-06-30
fix: pass LISTEN_ADDRESS to runner

matejvasek:
* Local (non-containerized) runner can run on various interfaces/addresses (e.g. `[::]:8080`, `10.0.0.1:1234`).
Before it worked only on IPv4 loopback (`127.0.0.1`).
* Fixed incorrect host port joining (causing a bug with IPv6 addresses).


matejvasek:
PTAL @lkingland

lkingland:
/approve
/lgtm

Thanks Matej!

PR #2904, opened 2025-07-01
fix: return better error when using pack+python

matejvasek:


pkg/builders/buildpacks/builder.go
```diff
@@ -120,6 +120,10 @@ var DefaultLifecycleImage = "docker.io/buildpacksio/lifecycle:553c041"
 
 // Build the Function at path.
 func (b *Builder) Build(ctx context.Context, f fn.Function, platforms []fn.Platform) (err error) {
+	if f.Runtime == "python" {
+		return fmt.Errorf("python is not currently supported with pack builder")
```
matzew: let's recommend `host` ? 

I got it working w/ that

matejvasek: updated

PR #2907, opened 2025-07-01
fix: Python local buildpack build

matejvasek:
# Changes

- fix: Python local `pack` build.
Local `pack` build was not working because of absent scaffolding. Fix is done by doing scaffolding in docker client (that is passed into constructor of `pack` client). Basically our custom docker client hijacks tar stream and alters it. It moves content of `/workspace/` into `/workspace/fn` and then puts scaffolding directly into `/workspace`, the scaffolding then refers handler in `./fn`.

So when `pack` clients tires to upload something like:
```
/workspace/func.yaml
/workspace/handler.py
/workspace/pyproject.toml
```
The actual content pushed to container is:
```
/workspace/fn/func.yaml
/workspace/fn/handler.py
/workspace/fn/pyproject.toml
/workspace/main.py           # scaffolding main
/workspace/pyproject.toml    # scaffolding project toml refering ./fn
```
The thing is that the `Python` paketo buildpack (as opposed to `Node` or `Go`) does not allow build form sub-directory, the project **must** be in workspace root.

This is just **temporary** hacky solution, we should come up with something better in future.

- fix: sane `LISTEN_ADDRESS` default for local pack build

/kind bug

```release-note
fix: Python local buildpack build
```


lkingland (approved):
/lgtm !
You say hacky, I say brilliant.  Thanks Matej!

PR #2911, opened 2025-07-03
chore: fix build and test tags

lkingland:
removes !integration build tags

The current build tagging system was predicated on the idea that
integraion, e2e and unit tests are entirely separate.  The correct way
to treat these is as inclusive with unit tests.  Thus this pr removes
the exclusion of unit test from integration tests runs, treating `-tags
integration` as indicating "unit tests plus integration tests".

/kind cleanup


gauron99 (approved):
/lgtm

PR #2918, opened 2025-07-04
Use localtest.me instead of sslip.io

matejvasek:
The `localtest.me` domain is shorter than `127.0.0.1.sslip.io`, but foremostly  is has `AAAA` record.


matejvasek:
PTAL @jrangelramos

matejvasek:
PTAL @lkingland

PR #2924, opened 2025-07-08
fix: refer correct version of schema in func.yaml

matejvasek:
# Changes

- Refer correct version of schema in func.yaml. Before we erroneously referred `specversion`, *not* `git` reference (branch/tag/hash).
- Moved version info into separate package to break cyclical dependency.

/kind bug

```release-note
fix: refer correct version of schema in func.yaml
```


matejvasek:
PTAL @lkingland @gauron99 @jrangelramos

PR #2926, opened 2025-07-08
test: simplify func-git service

matejvasek:
Use k8s primitives pod/svc/ingress instead of knative service.

### DIFF
--- a/hack/install-git-server.sh
+++ b/hack/install-git-server.sh
@@ -17,31 +17,79 @@ set -o nounset
 set -o pipefail
 
 git_server() {
-  echo "Creating Git Server Knative service..."
-  cat << EOF | kubectl apply -f -
-apiVersion: serving.knative.dev/v1
+  echo "Creating Git Server"
+
+  local name="func-git"
+
+  local namespace
+  namespace="$(kubectl config view --minify --output 'jsonpath={..namespace}')"
+  namespace="${namespace:-"default"}"
+
+
+  local ingress_class="contour-external"
+  local cluster_domain="localtest.me"
+  if kubectl api-versions | grep -q openshift.io; then
+    cluster_domain="$(kubectl get ingresses.config/cluster -o jsonpath='{.spec.domain}')"
+    ingress_class="openshift-default"
+  fi
+
+
+  local -r func_git_host="${name}.${namespace}.${cluster_domain}"
+
+  kubectl apply -f - <<EOF
+apiVersion: v1
+kind: Pod
+metadata:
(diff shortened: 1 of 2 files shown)


hack/install-git-server.sh
```diff
@@ -17,31 +17,70 @@ set -o nounset
...
+    - host: "${func_git_host}"
+      http:
+        paths:
+          - backend:
+              service:
+                name: "${name}"
+                port:
+                  number: 80
+            pathType: Prefix
+            path: /
```
matejvasek: Creating ingress like this does not seem to work in OCP. Some other way would have to be used there.
WDYT @jrangelramos

matejvasek:
cc @jrangelramos

test/common/gitserver.go
```diff
@@ -39,71 +36,44 @@ type GitProvider interface {
...
+	cmdResult := g.Kubectl.Exec("exec", podName, "--", "git-repo", "create", repoName)
 	if !strings.Contains(cmdResult.Out, "created") {
 		g.t.Fatal("unable to create git bare repository " + repoName)
 	}
-	namespace, _, _ := k8s.GetClientConfig().Namespace()
 	gitRepo := &GitRemoteRepo{
 		RepoName:         repoName,
-		ExternalCloneURL: g.ServiceUrl + "/" + repoName + ".git",
-		ClusterCloneURL:  "http://func-git." + namespace + ".svc.cluster.local/" + repoName + ".git",
+		ExternalCloneURL: fmt.Sprintf("http://func-git.%s.localtest.me/%s.git", g.ns, repoName),
```
jrangelramos: This should be resolved externally, it may work for kind but not for OCP test.

hack/install-git-server.sh
jrangelramos: it worked for me on OCP 4.19

matejvasek: @jrangelramos did the test you run use just internal name?

matejvasek: Internal name should be ok. I but external would not be, I think.

test/common/gitserver.go
matejvasek: Well actually this could work for OCP too, you would just have to put `func-git.${NS}.localtest.me` into your `/etc/hosts`  to point to the cluster IP.

But i agree that here and also in the install scrip we should use cluster public domain. Question is how to determine it?

matejvasek:
PTAL @lkingland @gauron99 @jrangelramos

PR #2938, opened 2025-07-15
e2e upgrade

lkingland:
Refactors E2E tests and workflows

/kind cleanup
 
Notable Name Changes:  the scripts in ./hack were renamed to be simple nouns:
hack/cluster.sh
hack/binaries.sh
hack/images.sh
hack/gitlab.sh


lkingland:
It ain't perfect, but I would prefer we merge this sooner rather than later as folks need this, and we can open new PRs for any niggling issues.

cmd/describe.go
```diff
@@ -197,7 +196,7 @@ func (i info) JSON(w io.Writer) error {
 }
 
 func (i info) XML(w io.Writer) error {
-	return xml.NewEncoder(w).Encode(i)
+	return errors.New("XML export not currently supported")
```
lkingland: XML breakes trying to export anonymous structs.  And XML is evil.

docs/CONTRIBUTING.md
```diff
@@ -8,10 +8,6 @@ To build the core project, run `make` from the repository root.  This will resul
 
 To remove built artifacts, use `make clean`.
 
-### Build affecting environment variables
-* `FUNC_REPO_REF` affects which github repo will be used to fetch tekton tasks for on cluster build. Default: `knative/func`.
-* `FUNC_REPO_BRANCH_REF` affects which github branch will be used to fetch tekton tasks for on cluster build. Default: `main`.
```
lkingland, line 13: Apparently these are completely unused.

Makefile
```diff
@@ -196,15 +260,20 @@ test-typescript: ## Test Typescript templates
...
 update-runtime-go:
+	@echo "Updating Go runtime..."
 	cd templates/go/scaffolding/instanced-http && go get -u knative.dev/func-go/http
 	cd templates/go/scaffolding/static-http && go get -u knative.dev/func-go/http
 	cd templates/go/scaffolding/instanced-cloudevents && go get -u knative.dev/func-go/cloudevents
 	cd templates/go/scaffolding/static-cloudevents && go get -u knative.dev/func-go/cloudevents
 
+.PHONY: update-runtime-python
+update-runtime-python:
+	# Nothing to update for Python
```
lkingland: Explicitly stating why it's not here.

Makefile
```diff
@@ -72,13 +77,72 @@ test: generate/zz_filesystem_generated.go ## Run core unit tests
 	go test -race -cover -coverprofile=coverage.txt ./...
 
 .PHONY: check
-check: $(BIN_GOLANGCI_LINT) ## Check code quality (lint)
+check: check-lint check-goimports check-misspell check-whitespace check-eof ## Check code quality (comprehensive)
```
lkingland, line 80: Adds checkers to be as close to those from knative automations such that running `make check` locally is a reasonable approximation.

hack/common.sh
```diff
@@ -19,13 +19,13 @@ init() {
...
 
 find_executables() {
-  KUBECTL=$(find_executable "kubectl")
-  KIND=$(find_executable "kind")
-  DAPR=$(find_executable "dapr")
-  HELM=$(find_executable "helm")
-  STERN=$(find_executable "stern")
-  KN=$(find_executable "kn")
-  JQ=$(find_executable "jq")
+  KUBECTL=$(find_executable "kubectl" || true)
```
lkingland, line 23: not all scripts require all binaries, so this is opportunistic; allowing failure to happen at time of use if not found.

.github/workflows/functions.yaml
```diff
@@ -0,0 +1,520 @@
...
+      FUNC_INT_GITLAB_ENABLED: true
+      FUNC_INT_GITLAB_HOSTNAME: gitlab.localtest.me
+      FUNC_TEST_GITLAB_PASS: test-password-123
+      FUNC_INT_PAC_HOST: pac-ctr.localtest.me
+      KUBECONFIG: ${{github.workspace}}/hack/bin/kubeconfig.yaml
+    steps:
+      - uses: actions/checkout@v4
+      - uses: knative/actions/setup-go@main
+
+      - name: Free up disk space
```
twoGiants: Nice one! :1st_place_medal: 

This should be a well maintained Github Action...just searched a bit and [there are a few](https://github.com/marketplace?query=free+disk+space&type=actions), but not all maintained. [This one](https://github.com/marketplace/actions/free-disk-space-ubuntu-runners) looks maintained.

.gitignore
```diff
@@ -36,5 +39,15 @@ __pycache__
 .vscode
 .idea
 
+# Agentfiles
+CLAUDE.md
+
 # Operating system temporary files
 .DS_Store
+
+# TODO: Update this test to be from a temp directory with a hard-coded impl:
```
twoGiants, line 48: I would not commit `# TODO: ....` Better to create an issue and reference it here, so it can be picked up by someone.

```shell
# Issue #1234: ...
``` 

Same for the `TODO` below.

Makefile
twoGiants: Is this desired? Why not remove it?

docs/CONTRIBUTING.md
```diff
@@ -88,7 +84,7 @@ Please note that the version of `yq` required is installed via `pip3 install yq`
 
 ### Allocate
```
twoGiants, line 85: In the `Prerequisites` above it should be added that the `dapr` cli is needed.

> The cluster allocation script requires [jq](https://stedolan.github.io/jq/), [yq](https://github.com/kislyuk/yq), [kubectl](https://kubernetes.io/docs/tasks/tools/), [kind](https://kind.sigs.k8s.io/docs/user/quick-start/), [dapr](https://docs.dapr.io/getting-started/install-dapr-cli/), docker and python.

cmd/describe.go
twoGiants: Maybe it makes sense to remove it here and in `format.go`?

hack/common.sh
twoGiants: A short comment here telling exactly that would be great!

hack/common.sh
```diff
@@ -19,13 +19,13 @@ init() {
 }
 
 find_executables() {
```
twoGiants, line 22:
```suggestion
# Not all scripts require all binaries, so this is opportunistic; allowing failure to happen
# at time of use if not found.
find_executables() {
```

hack/mac-bash.sh
```diff
@@ -0,0 +1,91 @@
...
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+  fi
+
+  # Update Homebrew
+  echo "${blue}Updating Homebrew...${reset}"
+  brew update
+
+  # Install bash 4+ (macOS ships with bash 3.x)
+  echo "${blue}Installing bash...${reset}"
+  brew install bash
```
twoGiants, line 46: Is this really needed on macOS? It ships with [zsh as a default since 2019](https://en.wikipedia.org/wiki/Z_shell#History).

pkg/pipelines/tekton/pipelines_int_test.go
```diff
@@ -192,7 +201,7 @@ func setupNS(t *testing.T) string {
 }
 
 func checkTestEnabled(t *testing.T) {
```
twoGiants: This function is unused. Also `setupNS`, `createSimpleGoProject`. Comment in line 47 can also be removed.

pkg/k8s/dialer_int_test.go
```diff
@@ -129,14 +150,20 @@ func TestDialInClusterService(t *testing.T) {
 	})
 	t.Log("created svc:", svc.Name)
 
-	// wait for service to start
+	// TODO: Replace with proper readiness check. This sleep gives the deployment
```
twoGiants, line 153: I would replace the `TODO` with a Github issue and reference it here.

pkg/functions/client_int_test.go
```diff
@@ -454,6 +454,7 @@ func Handle(res http.ResponseWriter, req *http.Request) {
 // TestInt_Invoke_ServiceToService ensures that a Function can invoke another
 // service via localhost service discovery api provided by the Dapr sidecar.
 func TestInt_Invoke_ServiceToService(t *testing.T) {
+	t.Skip("TODO: dapr appears to be borked")
```
twoGiants: I would add a Github issue and reference it here and replace the `TODO` with it.

pkg/pipelines/tekton/gitlab_int_test.go
```diff
@@ -43,7 +45,21 @@ import (
...
+func TestInt_Gitlab(t *testing.T) {
+	// Skip on ARM64 due to Paketo buildpack limitations
+	if runtime.GOARCH == "arm64" || runtime.GOARCH == "arm" {
+		t.Skip("Paketo buildpacks do not currently support ARM64 architecture. " +
+			"See https://github.com/paketo-buildpacks/nodejs/issues/712")
+	}
+
+	// Skip in CI due to persistent timeout issues regardless of allocated time
+	// Note it does indeed run locally.
+	// TODO: Investigate why GitLab webhook builds are not completing in CI
```
twoGiants, line 57: I would create a Github issue and reference it here and replace the `TODO` with it. Then it can be picked up by someone :smile_cat: .

twoGiants (commented):
Looks pretty awesome! What a great refactoring! :1st_place_medal: :smile_cat: 

Learned a lot about the project reviewing it. And I always like it in PRs when many files are removed and only a few added. It feels more cleaned up then. :broom:  :laughing: :broom: 

I have 3 files to review left. Will continue on Monday. 

Most, if not all comments are non-blocking.

e2e/README.md
```diff
@@ -0,0 +1,213 @@
...
+E2E tests are designed in a way that they can be easily runnable (and thus
+debuggable) locally by a developer, in addition to remotely in CI as
+acceptance criteria for pull requests.
+
+## Runnning E2Es locally: a Quick-start
+
+- `./hack/binaries.sh`          Fetch binaries into `./hack/bin`
+- `./hack/registry.sh`          (once) Configure insecure local registriry
+- `./hack/cluster.sh`           Create a cluster and kube config in `./hack/bin`
+- `make test-all`               Run all tests using these binaries and cluster
```
jrangelramos:
```suggestion
- `make test-full`               Run all tests using these binaries and cluster
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+
+	// DefaultKubeconfig is the default path (relative to this test file) at
+	// which the kubeconfig can be found which was created when setting up
+	// a local test cluster using the cluster.sh script.  This can be
+	// overridden using FUNC_E2E_KUBECONFIG.
+	DefaultKubeconfig = "../hack/bin/kubeconfig.yaml"
+
+	// DefaultNamespace for E2E tests is that used by default in the
+	// CLI being tested.
+	DefaultNamespace = cmd.DefaultNamespace
```
jrangelramos: This resolves to "default" aways. It would be nice if this could be configurable (i.e. FUNC_E2E_NAMESPACE) or resolve to current namespace on cluster

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+
+	// DockerHost - the DOCKER_HOST to use for container operations (not including podman-specific tests)
+	DockerHost = getEnv("FUNC_E2E_DOCKER_HOST", "", "")
+
+	// Gocoverdir - the coverage directory to use while testing the go binary.
+	Gocoverdir = getEnvPath("FUNC_E2E_GOCOVERDIR", "", DefaultGocoverdir)
+
+	// Kubeconfig - the kubeconfig to pass ass KUBECONFIG env to test
+	// environments.
+	Kubeconfig = getEnvPath("FUNC_E2E_KUBECONFIG", "", DefaultKubeconfig)
```
jrangelramos, line 337:
```suggestion
	Kubeconfig = getEnvPath("FUNC_E2E_KUBECONFIG", "KUBECONFIG", DefaultKubeconfig)
```
it could also rely on KUBECONFIG env var (just suggestion)

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+// Always uses the looback interface; OS-chosen port.
+func chooseOpenAddress(t *testing.T) (address string, err error) {
+	t.Helper()
+	var l net.Listener
+	if l, err = net.Listen("tcp", "127.0.0.1:"); err != nil {
+		return "", fmt.Errorf("cannot bind tcp: %w", err)
+	}
+	defer l.Close()
+	return l.Addr().String(), nil
+}
```
jrangelramos, line 1156: Suggestion: what about split this file per test category.
-  e2e_test.go -> consts, init, utils, waitFor funcs,...
-  e2e_core_test.go -> TestCore_* 
-  e2e_metadata_test.go -> TestMetadata_* 
-  e2e_[remote|podman|matrix]_test.go -> and so on...

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	}
+
+	if err := newCmd(t, "deploy").Run(); err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		clean(t, name, DefaultNamespace)
+	}()
+
+	if !waitForEcho(t, fmt.Sprintf("http://%v.default.localtest.me", name)) {
```
jrangelramos: Maybe it could have cluster domain as environment variable either? it would make test more cluster agnostic.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+//
+//	func deploy --remote
+func TestRemote_Deploy(t *testing.T) {
+	name := "func-e2e-test-remote-deploy"
+	_ = fromCleanEnv(t, name)
+
+	if err := newCmd(t, "init", "-l=go").Run(); err != nil {
+		t.Fatal(err)
+	}
+	if err := newCmd(t, "deploy", "--remote", "--builder=pack", "--registry=registry.default.svc.cluster.local:5000/func").Run(); err != nil {
```
jrangelramos: Registry is hardcoded here (there are other occurencies). Maybe this could be a variable (i.e ClusterInternalRegistry) also configurable by an env var?

.gitignore
lkingland: Ideally, yes, but there's an extenuating circumstance:  1) the Knative bot auto-deletes issues after a certain time, and 2) that creates an implicit dependency between our code and the GitHub issues database.

I'm not sure if this is standard practice, but I tend to say:
 `# TODO:`  Nonfatal, to be done in the future, possibly (a known, acceptable capitulation)
 `# FIXME:`  Fatal, should not be committed until handled.

docs/CONTRIBUTING.md
lkingland: Fortunately that's no longer a manual requirement because the `./hack/binaries.sh` script now installs all necessary binaries in an isolated environment. That document needs to be (will be) completely rewritten.

hack/mac-bash.sh
lkingland: I think it boils down to the hashbang:  If we specify zsh, that rules out many linux development machines without zsh (our primary development platform).  If we use bash (the universal default), then we need to ensure it's the updated GNU core utils version or our scripts break.

e2e/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+E2E tests confirm the functionality of the system end-to-end from the
+perspective of a user employing the Functions CLI `func`, either standalone
+or as a plugin to `kn` (`kn func`).
+
+E2E tests are designed in a way that they can be easily runnable (and thus
+debuggable) locally by a developer, in addition to remotely in CI as
+acceptance criteria for pull requests.
+
+## Runnning E2Es locally: a Quick-start
```
twoGiants:
```suggestion
## Running E2Es locally: a Quick-start
```

e2e/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+- `./hack/binaries.sh`          Fetch binaries into `./hack/bin`
+- `./hack/registry.sh`          (once) Configure insecure local registriry
```
twoGiants:
```suggestion
- `./hack/registry.sh`          (once) Configure insecure local registry
```

e2e/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+These tests expect a compiled binary, which will be executed utilizing a
+cluster configured to run Functions, as well as an available and authenticated
+container registry.  These can be configured locally for testing by using
+scripts in `../hack`:
+
+- `binaries.sh`: Installs executables needed for cluster setup and
+  configuration into hack/bin.
+
+- `registry.sh`: Configures the local Podman or Docker to allow unencrypted
```
twoGiants, line 61: Should there be also a `--revert` or `--undo` option for this script?

e2e/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+The suite accepts environment variables which alter the default behavior:
+
+`FUNC_E2E_BIN`: sets the path to the binary to use for the E2E tests.  This is
+by default the binary created when `make` is run in the repository root.
+Note that if providing a relative path, this path is relative to this test
+package, not the directory from which `go test` was run.
+
+`FUNC_E2E_PLUGIN`: if set, the command run by the tests will be
+`${FUNC_E2E_BIN} func`, allowing for running all tests when func is installed
```
twoGiants:
```suggestion
`${FUNC_E2E_BIN} ${FUNC_E2E_PLUGIN}`, allowing for running all tests when func is installed
```

e2e/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+`FUNC_E2E_TOOLS`: specifies the path to supporting tools. Defaults to
+"../hack/bin" relative to the e2e directory.
+
+`FUNC_E2E_TESTDATA`: specifies the path to supporting testdata. Defaults to
+"./testdata" relative to the e2e directory.
+
+## Running
+
+From the root of the repository, run `make test-all`.  This will compile
```
twoGiants: From the root of the repository, run `make test-full`. This will compile

e2e/README.md
```diff
@@ -0,0 +1,213 @@
...
+- TestRemote
+- TestMatrix
+
+## Cleanup
+
+The tests do attempt to clean up after themselves, but since a test failure is
+definitionally the entering of an unknown state, it is suggested to delete
+the cluster between full test runs. To remove the local cluster, use the
+`delete.sh` script described above.  If you do plan to remove the cluster, tests
+can be sped up by disabling cleanup with FUNC_E2E_CLEAN=false
```
twoGiants, line 219: can be sped up by disabling cleanup with `export FUNC_E2E_CLEAN=false`.

e2e/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+## TODO
```
twoGiants: I would put open TODOs in a Github issue.

e2e/README.md
```diff
@@ -0,0 +1,213 @@
...
+- Core tests should also test w/ CloudEvents template
+- Add OpenShift registry detection support (k8s.IsOpenShift() -> k8s.GetDefaultOpenShiftRegistry())
+
+## Migration Notes
+
+Replace script executions with "make test-all" (backwards compatibility for
```
twoGiants:
```suggestion
Replace script executions with `make test-full` (backwards compatibility for
```

hack/test-full.sh
```diff
@@ -0,0 +1,255 @@
...
+# - Podman
+# - Gitlab
+# - Pipelines
+# - etc.
+#
+# (See the environment variables which allow selective overriding.)
+#
+# By running `make test-full`, this script intends to roughly
+# replicate what is run in GitHub CI, but locally and without
+# parellelization.
```
twoGiants:
```suggestion
# parallelization.
```

hack/test-full.sh
```diff
@@ -0,0 +1,255 @@
...
+#
+# This script presumes a local testing environment set up using the
+# helper scripts in ./hack and performs some precondition checks to ensure
+# resources are available for the features enabled (nonexhaustive).
+#     hack/binaries.sh   - Installs necessary binaries in ./hack/bin
+#     hack/cluster.sh    - Start test cluster with Knative Serving/Eventing
+#     hack/registry.sh   - Starts and configures a local container registry
```
twoGiants: Changed the order.

```suggestion
#     hack/registry.sh   - Starts and configures a local container registry
#     hack/cluster.sh    - Start test cluster with Knative Serving/Eventing
```

hack/test-full.sh
```diff
@@ -0,0 +1,255 @@
...
+set -o nounset
+set -o pipefail
+
+source "$(cd "$(dirname "$0")" && pwd)/common.sh"
+
+# Enable Optional Tests
+# ---------------------
+# The defaults in the e2e test implementation are a bit more conservative.
+# Here we toggle on All The Things.  Note that we still allow any settings
+# made explicitly in the current environment to take precidence; just setting
```
twoGiants:
```suggestion
# made explicitly in the current environment to take precedence; just setting
```

hack/test-full.sh
```diff
@@ -0,0 +1,255 @@
...
+export FUNC_CLUSTER_RETRIES="${FUNC_CLUSTER_RETRIES:-5}"
+export FUNC_E2E_MATRIX="${FUNC_E2E_MATRIX:-true}"
+export FUNC_E2E_VERBOSE="${FUNC_E2E_VERBOSE:-true}"
+export FUNC_E2E_PODMAN="${FUNC_E2E_PODMAN:-true}"
+export FUNC_INT_TEKTON_ENABLED="${FUNC_INT_TEKTON_ENABLED:-1}"
+export FUNC_INT_GITLAB_ENABLED="${FUNC_INT_GITLAB_ENABLED:-1}"
+export FUNC_INT_GITLAB_HOSTNAME="${FUNC_INT_GITLAB_HOSTNAME:-gitlab.localtest.me}"
+export FUNC_INT_PAC_HOST="${FUNC_INT_PAC_HOST:-pac-ctr.localtest.me}"
+
+main() {
```
twoGiants, line 56: Beautiful! Very clean :smile_cat: :+1: :1st_place_medal:

hack/test-full.sh
```diff
@@ -0,0 +1,255 @@
...
+    echo "${blue}Preconditions${reset}"
+
+    # Check if binaries are installed
+    if [ ! -d "${PROJECT_ROOT}/hack/bin" ]; then
+        echo "ERROR: hack/bin directory not found!"
+        echo "Please run ./hack/install-binaries.sh first to install required tools."
+        exit 1
+    fi
+    MISSING_BINS=""
+    for bin in kubectl kind; do
```
twoGiants: Are those the only ones required? jq, stern, dapr, help, kn?

hack/test-full.sh
```diff
@@ -0,0 +1,255 @@
...
+# Preconditions
+# -------------
+preconditions() {
+    echo ""
+    echo "${blue}Preconditions${reset}"
+
+    # Check if binaries are installed
+    if [ ! -d "${PROJECT_ROOT}/hack/bin" ]; then
+        echo "ERROR: hack/bin directory not found!"
+        echo "Please run ./hack/install-binaries.sh first to install required tools."
```
twoGiants:
```suggestion
        echo "Please run ./hack/binaries.sh first to install required tools."
```

hack/test-full.sh
twoGiants: Are those the only ones required? jq, stern, dapr, help, kn?

```suggestion
    for bin in kubectl kind jq stern dapr helm kn; do
```

hack/test-full.sh
```diff
@@ -0,0 +1,255 @@
...
+    if [ -n "${MISSING_BINS}" ]; then
+        echo "ERROR: Required binaries not found:${MISSING_BINS}"
+        echo "Please run ./hack/binaries.sh to install required tools."
+        exit 1
+    fi
+
+    # Check if cluster is allocated
+    if [ ! -f "${KUBECONFIG}" ]; then
+        echo "ERROR: KUBECONFIG not found at ${KUBECONFIG}"
+        echo "Please run ./hack/allocate.sh to set up a test cluster."
```
twoGiants:
```suggestion
        echo "Please run ./hack/cluster.sh to set up a test cluster."
```

hack/test-full.sh
```diff
@@ -0,0 +1,255 @@
...
+        exit 1
+    fi
+
+    # Verify cluster connectivity
+    if ! kubectl cluster-info &>/dev/null; then
+        echo "ERROR: Cannot connect to Kubernetes cluster"
+        echo "KUBECONFIG: ${KUBECONFIG}"
+        echo "Please ensure your cluster is running and KUBECONFIG is valid."
+        echo "You may need to run ./hack/allocate.sh"
```
twoGiants:
```suggestion
        echo "You may need to run ./hack/cluster.sh"
```

hack/test-full.sh
```diff
@@ -0,0 +1,255 @@
...
+        kubectl cluster-info
+        kin
```
twoGiants: This calls should not be here right? And it would be `kind` not `kin`.

```suggestion

```

hack/test-full.sh
```diff
@@ -0,0 +1,255 @@
...
+    # Check if GitLab is installed (if GitLab tests are enabled)
+    if [ "${FUNC_INT_GITLAB_ENABLED}" = "1" ]; then
+        if ! kubectl get namespace gitlab &>/dev/null; then
+            echo "ERROR: GitLab namespace not found"
+            echo "Please run ./hack/gitlab.sh to install GitLab"
+            exit 1
+        fi
+    fi
+
+    # TODO: if Podman tests are enabled, check that podman is installed and running
```
twoGiants, line 148: I would create a Github issue for that and reference it here.

```suggestion

```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"knative.dev/func/pkg/k8s"
+)
+
+const (
+	// DefaultBin is the default binary to run, relative to this test file.
+	// This is the binary built by default when running 'make'.
+	// This can be customized with FUNC_E2E_BIN.
+	// NOte this is always relative to this test file.
```
twoGiants:
```suggestion
	// Note this is always relative to this test file.
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	fn "knative.dev/func/pkg/functions"
+	"knative.dev/func/pkg/knative"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"knative.dev/func/pkg/k8s"
+)
+
+const (
```
twoGiants, line 35: All `const` and `var` (further down below) can be package scoped.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	// Instead, this can be set explicitly using FUNC_E2E_KUBECONFIG.
+	Kubeconfig string
+
+	// Matrix indicates a full matrix test should be run.  Defaults to false.
+	// Enable with FUNC_E2E_MATRIX=true
+	Matrix bool
+
+	// MatrixBuilders specifies builders to check during matrix tests.
+	// Can be set with FUNC_E2E_MATRIX_BUILDERS.
+	// MatrixBuilders = []string{"host", "s2i", "pack"}
```
twoGiants:
```suggestion

```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	MatrixBuilders = []string{"host", "s2i", "pack"}
+
+	// MatrixRuntimes for which runtime-specific tests should be run.  Defaults
+	// to all core language runtimes.  Can be set with FUNC_E2E_MATRIX_RUNTIMES
+	MatrixRuntimes = []string{"go", "python", "node", "typescript", "rust", "quarkus", "springboot"}
+
+	// MatrixTemplates specifies the templates to check during matrix tests.
+	// MatrixTemplates = []string{"http", "cloudevents"}
```
twoGiants:
```suggestion

```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	// Can be set with FUNC_E2E_PLUGIN
+	Plugin string
+
+	// PodmanHost is the DOCKER_HOST value to use specifically for Podman tests.
+	// Can be set with FUNC_E2E_PODMAN_HOST.
+	PodmanHost string
+
+	// Registry is the container registry to use by default for tests;
+	// defaulting to the local container registry set up by the allocation
+	// scripts running on localhost:5000.
```
twoGiants:
```suggestion
	// defaulting to the local container registry set up by the cluster.sh
	// script running on localhost:5000.
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	// checked with the Podman container engine.
+	// Set with FUNC_E2E_PODMAN
+	Podman bool = false
+
+	// Verbose mode for all command runs.
+	// Set with FUNC_E2E_VERBOSE
+	Verbose bool
+
+	// Tools is the path to tools which the E2E tests should use with
+	// precidence.  It's a path, and is prepended to PATH.  By default this
```
twoGiants:
```suggestion
	// precedence.  It's a path, and is prepended to PATH.  By default this
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+// sometimes not at all; hence why the actual environment setup is deferred
+// into each test, merely reading them in here during E2E process init.
+func init() {
+	fmt.Fprintln(os.Stderr, "Initializing E2E Tests")
+	fmt.Fprintln(os.Stderr, "----------------------")
+	// Useful for CI debugging:
+	// fmt.Fprintln(os.Stderr, "--  Initial Environment: ")
+	// for _, env := range os.Environ() {
+	// 	fmt.Println(env)
+	// }
```
twoGiants: I would uncomment the code add a check for a `CI_DEBUGGING` environment variable. It could be set in repository environment variables and used in the `functions.yaml`.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+		fmt.Fprintf(os.Stderr, "error creating coverage directory %q: %v\n", Gocoverdir, err)
+	}
+
+	// Version
+	fmt.Fprintln(os.Stderr, "---------------------")
+	fmt.Fprintln(os.Stderr, "Func Version:")
+	printVersion()
+
+	fmt.Fprintln(os.Stderr, "--- init complete ---")
+	fmt.Fprintln(os.Stderr, "") // TODO: there is a superfluous linebreak from "func version".  This balances the whitespace.
```
twoGiants: I would replace the `TODO` with a Github "help wanted" issue. Someone will pick it up then.

e2e/e2e_test.go
twoGiants: I thought that for that case the `FUNC_E2E_KUBECONFIG` should be used.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+
+// newCmd returns an *exec.Cmd
+// bin will be FUNC_E2E_BIN, and if FUNC_E2E_PLUGIN is set, the subcommand
+// will be set as well.
+// arguments set to those provided.
+func newCmd(t *testing.T, args ...string) *exec.Cmd {
+	t.Helper()
+	bin := Bin
+
+	// If Plugin proivided, it is a subcommand so prepend it to args.
```
twoGiants:
```suggestion
	// If Plugin provided, it is a subcommand so prepend it to args.
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+
+// ----------------------------------------------------------------------------
+// Helpers
+// ----------------------------------------------------------------------------
+
+// fromCleanEnv provides a clean environment for a function E2E test.
+func fromCleanEnv(t *testing.T, name string) (root string) {
+	root = cdTemp(t, name)
+	// Deprecated?  We're allowing HOME to stay set for now:
+	// setupHome(t)
```
twoGiants: Remove comments. Seems not to be relevant anymore.

```suggestion

```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	setupEnv(t)
+	return
+}
+
+// cdTmp changes to a new temporary directory for running the test.
+// Essentially equivalent to creating a new directory before beginning to
+// use func.  The path created is returned.
+// The "name" argument is the name of the final Function's directory.
+// Note that this will be unnecessary when upcoming changes remove the logic
```
twoGiants: Maybe an issue reference should be added here which points to the actual implementation issue so that this change is not forgotten.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	warnInterval  time.Duration // only warn this often
+}
+
+func withWaitTimeout(t time.Duration) waitOption {
+	return func(cfg *waitCfg) {
+		cfg.timeout = t
+	}
+}
+
+// waitForEcho returns true if there is service at the given addresss which
```
twoGiants:
```suggestion
// waitForEcho returns true if there is service at the given address which
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	return waitFor(t, address, content, "expected content not found", options...)
+}
+
+// waitFor an endpoint to return an OK response which includes the given
+// content.
+//
+// If the Function returns a 500, it is considered a positive test failure
+// by the implementation and retries are discontinued.
+//
+// TODO:  Implement a --output=json flag on `func run` and update all
```
twoGiants: I would create a Github issue for this `TODO` and reference it here.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	)
+
+	startTime := time.Now()
+	lastWarnTime := time.Time{}
+	attemptCount := 0
+
+	for time.Since(startTime) < cfg.timeout {
+		attemptCount++
+		time.Sleep(cfg.interval)
+		// t.Logf("GET %v\n", address)
```
twoGiants: Remove or uncomment.

```suggestion
          t.Logf("GET %v\n", address)
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	cmd := newCmd(t, "run", "--address", address)
+	if err := cmd.Start(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Wait for echo
+	if !waitForEcho(t, "http://"+address) {
```
twoGiants: Does this run into flakes a lot? If not then all good! 

If yes, then a more specific implementation using `kubectl` which waits for the pod and service availability and then sends an `echo` would be an improvement.

But not here in this PR of course. It can be a **_help-wanted_** issue.

e2e/e2e_test.go
twoGiants: But the cluster created for this e2e tests is using this cluster domain. Not sure if someone would create/use their own custom setup with a different DNS config just for the tests. 

Wdyt?

twoGiants (commented):
Great job :smile_cat: :rocket: :guitar: :1st_place_medal:! 

This is a big improvement to what is there and the structure is nice, clean and clear.

I haven't found anything critical. Just typos, `TODOs` which can be extracted into issues and a few minor _nice-to-haves_.

I am mostly done. The last file I am still reviewing are the `e2e-test.go` which is the biggest one. I will submit what I have so far for it and the rest today/tomorrow.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	name := "func-e2e-test-core-init"
+	root := fromCleanEnv(t, name)
+
+	// Act (newCmd == "func ...")
+	if err := newCmd(t, "init", "-l=go").Run(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Assert
+	f, err := fn.NewFunction(root)
```
twoGiants: I would probably assert a bit more here. Because it's an e2e test I would check that what I expect was actually generated and has what I expect in it.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+		clean(t, name, DefaultNamespace)
+	}()
+
+	if !waitForEcho(t, fmt.Sprintf("http://%v.default.localtest.me", name)) {
+		t.Fatalf("function did not deploy correctly")
+	}
+}
+
+// TestCore_Deploy_Template ensures that the system supports creating
+// functions based off templates in a remote repository.
```
twoGiants:
```suggestion
// functions based of templates in a remote repository.
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+func TestCore_Deploy_Template(t *testing.T) {
+	name := "func-e2e-test-core-deploy-template"
+	_ = fromCleanEnv(t, name)
+
+	// Creates a new Function from the template located in the repository at a
+	// well-known path:  {repo}/{runtime}/{template} where
+	//   repo: github.com/functions-dev
+	//   runtime: go
+	//   template: http (the default.  can be changed with --template)
+	if err := newCmd(t, "init", "-l=go", "--repository=https://github.com/functions-dev/func-e2e-tests").Run(); err != nil {
```
twoGiants: `https://github.com/functions-dev/func-e2e-tests` => hosted on the gitlab instance in the bootstrapped cluster would be nice and more independent :smile_cat: . Maybe also a **_help-wanted_** issue for later.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	}
+
+	// Parse the JSON output
+	var instance fn.Instance
+	if err := json.Unmarshal(out.Bytes(), &instance); err != nil {
+		t.Fatalf("error unmarshaling describe output: %v", err)
+	}
+
+	// Validate that the name matches what we expect
+	if instance.Name != name {
```
twoGiants: Describe shows more than just the name. Maybe we could assert for the entire content?

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	}
+	run := cmd // for the closure
+	defer func() {
+		// ^C the running function
+		if err := run.Process.Signal(os.Interrupt); err != nil {
+			fmt.Fprintf(os.Stderr, "error interrupting. %v", err)
+		}
+	}()
+
+	// TODO: complete implementation of `func run --json` structured output
```
twoGiants: I would put the `TODO` in a Github issue and reference it here.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	cmd = newCmd(t, "invoke", "--data=func-e2e-test-core-invoke-local")
+	out := bytes.Buffer{}
+	cmd.Stdout = &out
+	if err := cmd.Run(); err != nil {
+		t.Fatal(err)
+	}
+	if !strings.Contains(out.String(), "func-e2e-test-core-invoke-local") {
+		t.Logf("out: %v", out.String())
+		t.Fatal("function invocation did not echo data provided")
+	}
```
twoGiants: Put in a function and use here and below to avoid duplication.

```suggestion
	checkInvoke := func(funcName string) {
		cmd = newCmd(t, "invoke", "--data="+funcName)
		out := bytes.Buffer{}
		cmd.Stdout = &out
		if err := cmd.Run(); err != nil {
			t.Fatal(err)
		}
		if !strings.Contains(out.String(), funcName) {
			t.Logf("out: %v", out.String())
			t.Fatal("function invocation did not echo data provided")
		}
	}
	checkInvoke("func-e2e-test-core-invoke-local")
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	cmd = newCmd(t, "invoke", "--data=func-e2e-test-core-invoke-remote")
+	out = bytes.Buffer{}
+	cmd.Stdout = &out
+	if err := cmd.Run(); err != nil {
+		t.Fatal(err)
+	}
+	if !strings.Contains(out.String(), "func-e2e-test-core-invoke-remote") {
+		t.Logf("out: %v", out.String())
+		t.Fatal("function invocation did not echo data provided")
+	}
```
twoGiants:
```suggestion
	checkInvoke("func-e2e-test-core-invoke-remote")
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	// ----------------------------------------
+	// Deploys the function remotely.  `func invoke` will then invoke it
+	// with preference over the (still) running local instance.
+	if err := newCmd(t, "deploy").Run(); err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		clean(t, name, DefaultNamespace)
+	}()
+	if !waitForEcho(t, "http://func-e2e-test-core-invoke.default.localtest.me") {
```
twoGiants: To keep it consistent :smile_cat: 

```suggestion
	if !waitForEcho(t, fmt.Sprintf("http://%v.default.localtest.me", name)) {
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	}
+	defer func() {
+		clean(t, name, DefaultNamespace)
+	}()
+	if !waitForContent(t, fmt.Sprintf("http://%v.default.localtest.me", name), "OK") {
+		t.Fatalf("handler failed")
+	}
+
+	// Set a test Environment Variable
+	// Add
```
twoGiants: Remove comment.

```suggestion

```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	}
+	defer func() {
+		clean(t, name, DefaultNamespace)
+	}()
+	if !waitForEcho(t, fmt.Sprintf("http://%v.default.localtest.me", name)) {
+		t.Fatalf("function did not deploy correctly")
+	}
+
+	// Use the output from "func describe" (json output) to verify the
+	// function contains the both the test labels as expected.
```
twoGiants:
```suggestion
	// function contains both test labels as expected.
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+		t.Fatal(err)
+	}
+
+	var instance fn.Instance
+	if err := json.Unmarshal(out.Bytes(), &instance); err != nil {
+		t.Fatalf("error unmarshaling describe output: %v", err)
+	}
+	if instance.Labels == nil {
+		t.Fatal("No labels returned")
+	}
```
twoGiants: This code is duplicated in the next test method two more times and could be extracted into a helper function.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+
+	if !waitForContent(t, fmt.Sprintf("http://%s.default.localtest.me", name), "OK") {
+		t.Fatalf("function failed after volume removal")
+	}
+}
+
+// TODO: TestMetadata_Subscriptions ensures that function instances can be
+// subscribed to events.
+func TestMetadata_Subscriptions(t *testing.T) {
+	// TODO
```
twoGiants: I would extract `TODOs` into Github issues and reference here.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+
+// ---------------------------------------------------------------------------
+// REMOTE TESTS
+// Tests related to invoking processes remotely (in-cluster).
+// All remote tests presume the cluster has Tekton installed.
+// ---------------------------------------------------------------------------
+
+// TestRemote_Deploy ensures that the default action of running a remote
+// build succeeds:  uploading local souce code to the cluster for build and
+// delpoy in-cluster.
```
twoGiants:
```suggestion
// build succeeds:  uploading local source code to the cluster for build and
// deploy in-cluster.
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+}
+
+// TODO: TestMetadata_Subscriptions ensures that function instances can be
+// subscribed to events.
+func TestMetadata_Subscriptions(t *testing.T) {
+	// TODO
+	// Create a function which emits an event with as much defaults as possible
+	// Create a function which subscribes to those events
+	// Succeed the test as soon as it receives the event
+	t.Skip("Subscritions E2E tests not yet implemented")
```
twoGiants:
```suggestion
	t.Skip("Subscriptions E2E tests not yet implemented")
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	}
+	if _, ok := desc2.Labels["foo"]; !ok {
+		t.Error("Label 'foo' should still exist")
+	}
+	if _, ok := desc2.Labels["foo2"]; ok {
+		t.Error("Label 'foo' was not removed")
+	}
+}
+
+// TestMetadta_Volumes ensures that adding volumes of various types are
```
twoGiants:
```suggestion
// TestMetadata_Volumes ensures that adding volumes of various types are
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	// locally in order to use its name.
+	cmd := exec.Command("git", "clone", "https://github.com/functions-dev/func-e2e-tests", ".")
+	if err := cmd.Run(); err != nil {
+		t.Fatal(err)
+	}
+
+	// IMPORTANT: The local func.yaml must match the one in the target branch.
+	// This is a current limitation where remote builds still require local
+	// source to determine function metadata (name, runtime, etc).
+	// TODO: Remove this checkout once the implementation supports fetching
```
twoGiants: I would add this `TODO` to the actual Github issue for this feature and reference it here. Then it won't be forgotten.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	cmd := exec.Command("git", "clone", "https://github.com/functions-dev/func-e2e-tests", ".")
+	if err := cmd.Run(); err != nil {
+		t.Fatal(err)
+	}
+
+	// IMPORTANT: When using --git-dir, we need to change to that directory locally
+	// to ensure the local func.yaml matches the one that will be used in the remote build.
+	// This is a current limitation where remote builds still require local source to
+	// determine function metadata (name, runtime, etc).
+	// TODO: Remove this cd once the implementation supports fetching function metadata
```
twoGiants: Same comment as above.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+//		OS:       Linux, Mac, Windows (handled at the Git Action level)
+//		Runtime:  Go, Python, Node, Typescript, Quarkus, Springboot, Rust
+//		Builder:  Host, Pack, S2I
+//		Template: http, CloudEvent
+//
+//	 Test it can:
+//	 1.  Run locally on the host (func run)
+//	 3.  Deploy and receive the default response (an echo)
+//	 4.  Remote build and run via an in-cluster build
+// -----------------
```
twoGiants:
```suggestion
// ---------------------------------------------------------------------------
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+// TestPodman_Pack ensures that the Podman container engine can be used to
+// deploy functions built with Pack.
+func TestPodman_Pack(t *testing.T) {
+	name := "func-e2e-test-podman-pack"
+	_ = fromCleanEnv(t, name)
+	if err := setupPodman(t); err != nil {
+		t.Fatal(err)
+	}
+
+	if !Podman {
```
twoGiants: I would put those checks in a function and use before `name := "..."`. Then I you can reuse that function in the next test.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	httpAddress := "http://" + address
+	var ready bool
+	if template == "cloudevents" {
+		ready = waitForCloudevent(t, httpAddress, withWaitTimeout(timeout))
+	} else { // default is http:
+		ready = waitForEcho(t, httpAddress, withWaitTimeout(timeout))
+	}
+
+	if !ready {
+		t.Fatalf("service does not appear to have started correctly.")
```
twoGiants:
```suggestion
		t.Fatal("Service does not appear to have started correctly.")
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+//
+//	 Test it can:
+//	 1.  Run locally on the host (func run)
+//	 3.  Deploy and receive the default response (an echo)
+//	 4.  Remote build and run via an in-cluster build
+// -----------------
+
+// TestMatrix_Run ensures that supported runtimes and builders can run both
+// builtin templates locally.
+func TestMatrix_Run(t *testing.T) {
```
twoGiants: Nicely done! I like the matrix implementation :smile_cat: :+1:

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+			for _, template := range MatrixTemplates {
+				name := fmt.Sprintf("func-e2e-matrix-%s-%s-%s-run", runtime, builder, template)
+				// Test Running Locally
+				// --------------------
+				t.Run(name, func(t *testing.T) {
+					doMatrixRun(t, name, runtime, builder, template)
+				})
+			}
+		}
+	}
```
twoGiants: This repeats in all matrix tests and could be extracted into a helper to reduce duplication. Not a must though!

Something like that:

```go
func TestMatrix_Run(t *testing.T) {
	runMatrixTests(t, doMatrixRun, "run")
}

type matrixTest func(t *testing.T, name string, runtime string, builder string, template string)

func runMatrixTests(t *testing.T, runOneMatrixTest matrixTest, testType string) {
	t.Helper()
	if !Matrix {
		t.Skip("Matrix tests not enabled. Enable with FUNC_E2E_MATRIX=true")
	}
	for _, runtime := range MatrixRuntimes {
		for _, builder := range MatrixBuilders {
			for _, template := range MatrixTemplates {
				name := fmt.Sprintf("func-e2e-matrix-%s-%s-%s-%s", runtime, builder, template, testType)
				t.Run(name, func(t *testing.T) {
					runOneMatrixTest(t, name, runtime, builder, template)
				})
			}
		}
	}
}
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	httpAddress := fmt.Sprintf("http://%v.default.localtest.me", name)
+	var ready bool
+	if template == "cloudevents" {
+		ready = waitForCloudevent(t, httpAddress, withWaitTimeout(timeout))
+	} else {
+		ready = waitForEcho(t, httpAddress, withWaitTimeout(timeout))
+	}
+
+	if !ready {
+		t.Fatalf("function did not deploy correctly")
```
twoGiants:
```suggestion
		t.Fatal("function did not deploy correctly")
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	functionURL := fmt.Sprintf("http://%v.default.localtest.me", name)
+	var ready bool
+	if template == "cloudevents" {
+		ready = waitForCloudevent(t, functionURL, withWaitTimeout(timeout))
+	} else {
+		ready = waitForEcho(t, functionURL, withWaitTimeout(timeout))
+	}
+
+	if !ready {
+		t.Fatalf("function did not deploy correctly")
```
twoGiants:
```suggestion
		t.Fatal("function did not deploy correctly")
```

e2e/e2e_test.go
twoGiants: This block repeats in all matrix tests and could also be a helper.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	// Buildpacks do not currently support ARM
+	if builder == "pack" && (runtime.GOARCH == "arm64" || runtime.GOARCH == "arm") {
+		t.Skip("Paketo buildpacks do not currently support ARM64 architecture. " +
+			"See https://github.com/paketo-buildpacks/nodejs/issues/712")
+	}
+
+	// Python Special Treatment
+	// --------------------------
+	// Skip Pack builder (not supported)
+	// TODO: Remove when pack support is added
```
twoGiants: I would extract `TODO` into a Github issue and reference here.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+insecure=true
+`
+	cfgPath := filepath.Join(t.TempDir(), "registries.conf")
+	if err := os.WriteFile(cfgPath, []byte(cfg), 0644); err != nil {
+		return fmt.Errorf("failed to create registries.conf: %v", err)
+	}
+	os.Setenv("CONTAINERS_REGISTRIES_CONF", cfgPath)
+
+	// Podman Info
+	// May be useful when debugging:
```
twoGiants: I would activate it via an environment variable in the CI which is set/changed in the Github repository settings.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	}
+	t.Logf("Podman registry proxy enabled: %s", strings.TrimSpace(string(output)))
+
+	return nil
+}
+
+// newCmd returns an *exec.Cmd
+// bin will be FUNC_E2E_BIN, and if FUNC_E2E_PLUGIN is set, the subcommand
+// will be set as well.
+// arguments set to those provided.
```
twoGiants:
```suggestion
// will be set as well. Arguments set to those provided.
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	}
+	base := filepath.Base(bin)
+	t.Logf("$ %v %v\n", base, b.String())
+
+	cmd := exec.Command(bin, args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd
+
+	// TODO: create an option to only print stdout and stderr if the
```
twoGiants: I would extract the `TODO` into a Github issue and reference here. And remove the commented code.

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	return
+}
+
+// isAbnormalExit checks an error returned from a cmd.Wait and returns true
+// if the error is something other than a known exit 130 from a SIGINT.
+func isAbnormalExit(t *testing.T, err error) bool {
+	if err == nil {
+		return false // no error is not an abnormal error.
+	}
+	t.Helper()
```
twoGiants:
```suggestion
	t.Helper()
	if err == nil {
		return false // no error is not an abnormal error.
	}
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	} else {
+		t.Fatalf("Function errored during execution. %v", err)
+		return true
+	}
+	return false
+}
+
+// setSecret creates or replaces a secret.
+func setSecret(t *testing.T, name, ns string, data map[string][]byte) {
+	ctx := context.Background()
```
twoGiants, line 850:
```suggestion
	t.Helper()
	ctx := context.Background()
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	}
+	_, err = clientset.CoreV1().Secrets(ns).Create(ctx, secret, metav1.CreateOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+// setConfigMap creates or replaces a configMap
+func setConfigMap(t *testing.T, name, ns string, data map[string]string) {
+	ctx := context.Background()
```
twoGiants, line 873:
```suggestion
	t.Helper()
	ctx := context.Background()
```

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+// clean up by deleting the named function (via defers)
+func clean(t *testing.T, name, ns string) {
+	t.Helper()
+	if !Clean {
+		return
+	}
+	// There is currently a bug in delete which hangs for several seconds
+	// when deleting a Function. This adds considerably to the test suite
+	// execution time.  Tests are written such that they are not dependent
+	// on a clean exit/cleanup, so this step is skipped for speed.
```
twoGiants: The comment is a bit misleading:  `..., so this step is skipped for speed`. Where is this step skipped?

e2e/e2e_test.go
```diff
@@ -0,0 +1,2610 @@
...
+	if imagesRemoved > 0 {
+		logImageStats(t, name, "post-cleanup")
+	}
+}
+
+// logImageStats logs Docker/Podman storage statistics for debugging disk usage
+func logImageStats(t *testing.T, name string, phase string) {
+	t.Helper()
+
+	// Get image size for this specific function
```
twoGiants:
```suggestion
	// Get image name for this specific function
```

twoGiants (commented):
Hi @lkingland! Finished the `e2e_test.go`. That was the biggest one! :sweat_smile: 

Looks good to me! 

I have only found some typos and propose to cleanup a bit here and there, extract issues from `TODOs` and reduce duplication. Can all be done in a follow up, except the `TODOs`, those I would remove now and create issues. But you decide, its all good :smile_cat:

e2e/e2e_test.go
twoGiants: Agree, this would be great!

e2e/e2e_test.go
jrangelramos: Yes these tests are strict to the Kind cluster created for CI, but I would (I probably will) run these test set with OCP.

e2e/e2e_test.go
jrangelramos: It could be even be hosted on the git-server available for remote tests.

.gitignore
twoGiants: Not a fan tbh. It "never" gets picked up and it lacks clarity. Is someone working on it? Should I resolve it? Do I leave it? Whats the priority?

- Can the Knative bot be configured not to do that if the reference in the code is there?
- What issue do you see with having this dependency in the code to the Github issues (a tool)?
- We could also remove the `TODO` and have an issue in our internal JIRA. Then it won't be deleted and triaged at some point.

.github/workflows/functions.yaml
lkingland: 👍🏻

.gitignore
lkingland: The purpose of a TODO is to acknowledge the reality that no code is ever complete.  It's a note to future developers that a certain code path was deliberately skipped, as this may be useful when understanding the limits of the system's logic during future maintenance.  FIXME's are notes of things which must be completed before the code goes live.

Makefile
lkingland: Removed.

The reason, fyi, is to communicate to future readers there is no need to explicitly update Python middleware.

The process:  I spent a bit of time looking this up thinking this update target was needed, only to realize later it was pulled on-demand... twice.  Thus, this target with an explicit "it's not needed" comment is a note to future readers to save them the same waste of time.  Code is for agents fist, computers second.

e2e/README.md
lkingland: That would be nice, and could be a feature of our forthcoming Go-based cluster configuration system, which will replace these scripts.  In the mean-time, there is a clear note in the output from `delete.sh` that this change is left in-place.

In case you're interested, I checked with AI if the cost/benefit was worth it; here's the response:

 Moderate to High Complexity due to:

  1. Multiple configuration formats:
    - JSON for Docker (needs jq manipulation)
    - TOML for Podman (two different formats: v1 and v2)
    - Need different parsing strategies for each
  2. State tracking challenges:
    - No record of what existed before modification
    - Can't distinguish between:
        - Entries added by this script vs. manually added
      - Files created by script vs. pre-existing
    - Would need to implement backup mechanism or state file
  3. Edge cases:
    - Partial failures during original setup
    - Multiple runs of the script
    - Manual edits between runs
    - Other tools modifying same configs
  4. Implementation requirements:
    - Backup original configs before modification
    - Store metadata about changes made
    - Parse and modify JSON (Docker) safely
    - Parse and modify TOML (Podman) safely
    - Handle sudo requirements correctly
    - Find and kill SSH forwarding processes

⏺ Worth/Value Assessment

  Limited Value for the following reasons:

  1. Low frequency of use:
    - This is typically a one-time setup per development environment
    - Developers rarely need to undo this configuration
  2. Minimal negative impact:
    - Adding localhost:50000 as insecure doesn't pose security risks (it's local-only)
    - Doesn't interfere with other registries or normal operations
    - Easy to manually remove if needed
  3. Alternative solutions exist:
    - Manual removal is straightforward (edit one line in config files)
    - Complete reset possible by deleting config files (Docker/Podman recreate defaults)
    - Documentation could provide manual undo steps
  4. Cost-benefit analysis:
    - High implementation complexity vs. low usage frequency
    - Risk of bugs in undo logic could be worse than leaving config as-is
    - Maintenance burden for edge cases

⏺ Recommendation

  Not worth implementing a revert/undo capability for registry.sh.

e2e/README.md
lkingland: I wish we could just give exact commands to run!  I use `fish`, for example, which is `set -x FUNC_E2E_CLEAN false`, and Windows has several different ways depending on scope.  Therefore I just stick to the pseudocode.

e2e/README.md
lkingland: Ahhhh, nice catch; that

e2e/README.md
lkingland: Thanks; these can be removed.  The cloudevents templates are now test in the Matrix tests, and the OpenShift QE team uses their own E2E test suite now 👍🏻

e2e/e2e_test.go
lkingland: I may be misunderstanding, but these are indeed package-scoped, and by being exported they appear in godoc

e2e/e2e_test.go
lkingland: This is now configurable

e2e/e2e_test.go
lkingland: interesting `misspell` doesn't have that in its common misspellings list.

e2e/e2e_test.go
lkingland: fix made in a local branch; will submit PR

e2e/e2e_test.go
lkingland: This explicit separation was made because even a single inadvertent run of `make test-full` when connected to a production cluster could cause quite some problems for a developer.  Instead, there is complete isolation where the default kubeconfig is set to the one placed in hack/bin/kubeconfig.yaml upon creation of the test cluster with `cluster.sh`.  Also, it "just works" if they run that allocation script followed by an e2e test without altering any system settings.

e2e/e2e_test.go
lkingland: So the "degenerate case" was a bit... too much so?  😄 
I added a couple more, but the truth is that the _only_ absolutely necessary setting upon initialization is language runtime; this simplicity is by design.

e2e/e2e_test.go
lkingland: Yes, they are actually that unreliable in CI, but more importantly, this is measuring observed behavior, and is thus usable for both cluster services as well a locally-started services started via "func run"

e2e/e2e_test.go
lkingland: The cluster domain is now configurable
This keeps the _default_ being the cluster set up using the CI scripts without needing additional configuration, but the ability to run against any cluster with necessary configuration via environment variables.

e2e/e2e_test.go
lkingland: This one's intended

e2e/e2e_test.go
lkingland: I decided to use this so that the E2E tests didn't have a dependency on the cluster having a git server installed, and to test "real world" expected behavior more closely; however I am happy to revisit these assumptions on the call if you'd like and open a ticket to refactor!

e2e/e2e_test.go
lkingland: Rationale:  E2E tests should verify that "describe describes".  Unit tests verify that "what describe describes".

e2e/e2e_test.go
lkingland: This is now configurable

e2e/e2e_test.go
lkingland: I concur this is critical 🫡

hack/test-full.sh
lkingland: Just spot-checking that it ran; didn't want to tightly couple, but enumerating all is fine too

docs/CONTRIBUTING.md
```diff
@@ -8,10 +8,6 @@ To build the core project, run `make` from the repository root.  This will resul
 
 To remove built artifacts, use `make clean`.
 
-### Build affecting environment variables
-* `FUNC_REPO_REF` affects which github repo will be used to fetch tekton tasks for on cluster build. Default: `knative/func`.
```
gauron99, line 12: there is still reference to this variable in `test/presubmit-tests.sh`

e2e/e2e_matrix_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+// matrixExceptionsShared are exceptions to the full matrix which are shared
+// between both local (run, build, deploy) and remote (deploy --remote)
+func matrixExceptionsShared(t *testing.T, initArgs []string, funcRuntime, builder, template string, waitTimeout time.Duration) ([]string, time.Duration) {
+	t.Helper()
+
+	// Buildpacks do not currently support ARM
+	if builder == "pack" && (runtime.GOARCH == "arm64" || runtime.GOARCH == "arm") {
+		t.Skip("Paketo buildpacks do not currently support ARM64 architecture. " +
+			"See https://github.com/paketo-buildpacks/nodejs/issues/712")
```
gauron99, line 229: i dont think this is the right link? it links to general github-config update (go bump and version change); its from 2 years ago

gauron99 (commented):
some comments

.github/workflows/functions.yaml
```diff
@@ -0,0 +1,461 @@
...
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        run: |
+          for a in amd64 arm64 ppc64le s390x; do
+            CGO_ENABLED=0 go build -o "func-util-$a" -trimpath -ldflags '-w -s' ./cmd/func-util
```
matejvasek, line 441: How did `GOARCH="$a"` disappear here?

PR #2942, opened 2025-07-22
feat: func invoke get requests

gauron99:
# Changes

- :gift: GET requests with `func invoke --request-type=GET`

/kind enhancement

**Release Note**

```release-note
feat: run 'func invoke --request-type=GET' for invoking GET request
```

### DIFF
--- a/cmd/invoke.go
+++ b/cmd/invoke.go
@@ -99,9 +99,14 @@ EXAMPLES
 	o Allow insecure server connections when using SSL
 		$ {{rootCmdUse}} invoke --insecure
 
+	o In case you need to specifically send GET request
+		$ {{rootCmdUse}} invoke --request-type=GET
 `,
-		SuggestFor: []string{"emit", "emti", "send", "emit", "exec", "nivoke", "onvoke", "unvoke", "knvoke", "imvoke", "ihvoke", "ibvoke"},
-		PreRunE:    bindEnv("path", "format", "target", "id", "source", "type", "data", "content-type", "file", "insecure", "confirm", "verbose"),
+		SuggestFor: []string{"emit", "emti", "send", "emit", "exec", "nivoke",
+			"onvoke", "unvoke", "knvoke", "imvoke", "ihvoke", "ibvoke"},
+		PreRunE: bindEnv("path", "format", "target", "id", "source", "type",
+			"data", "content-type", "request-type", "file", "insecure",
+			"confirm", "verbose"),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runInvoke(cmd, args, newClient)
 		},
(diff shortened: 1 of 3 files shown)


cmd/invoke.go
```diff
@@ -162,6 +168,7 @@ func runInvoke(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 		Source:      cfg.Source,
 		Type:        cfg.Type,
 		ContentType: cfg.ContentType,
+		RequestType: strings.ToUpper(cfg.RequestType),
```
gauron99, line 171: Perhaps we accept strict upper case instead?
```suggestion
		RequestType: cfg.RequestType,
```

pkg/functions/invoke.go
```diff
@@ -79,19 +89,27 @@ func invoke(ctx context.Context, c *Client, f Function, target string, m InvokeM
...
 	switch format {
 	case "http":
-		return sendPost(ctx, route, m, c.transport, verbose)
+		return sendHttp(ctx, route, m, c.transport, verbose)
 	case "cloudevent":
-		// CouldEvents return a string which always includes a fairly verbose
-		// summation of fields, so metadata is not applicable
-		meta := make(map[string][]string)
-		body, err = sendEvent(ctx, route, m, c.transport, verbose)
-		return meta, body, err
```
gauron99, line 90: Since return type and value is set in definition, im pretty sure we dont need this `meta` and return statements

pkg/functions/invoke.go
```diff
@@ -79,19 +89,27 @@ func invoke(ctx context.Context, c *Client, f Function, target string, m InvokeM
...
+		switch m.RequestType {
+		case "POST":
+			body, err = sendEvent(ctx, route, m, c.transport, verbose)
+		case "GET":
+			// Construct a special CloudEvents GET request.
+			// This will be used most likely only for very special cases
+			body, err = sendGetEvent(ctx, route, m, c.transport, verbose)
+		}
```
gauron99, line 108: The functions `sendEvent` and `sendGetEvent` are very similar yet also distinct, instead of making bunch of if statements in the function itself -> this switch

pkg/functions/invoke.go
```diff
@@ -176,30 +194,84 @@ func sendEvent(ctx context.Context, route string, m InvokeMessage, t http.RoundT
...
+			"Data":        {string(m.Data)},
+		}
 		fmt.Println("Sending values")
 		for k, v := range values {
 			fmt.Printf("  %v: %v\n", k, v[0]) // NOTE len==1 value slices assumed
 		}
 	}
 
-	req, err := http.NewRequestWithContext(ctx, "POST", route, bytes.NewReader(m.Data))
+	req, err := http.NewRequestWithContext(ctx, m.RequestType, route, bytes.NewReader(m.Data))
```
gauron99, line 270: For simplicity I pretty much leave the HTTP request unchanged. Should I make a more proper GET request by excluding data,contentType etc ? This works fine.

cmd/invoke.go
lkingland: Usually best to be strict, but I think this auto upper casing is a reasonable capitulation.

pkg/functions/invoke.go
lkingland: It works fine :shipit:

lkingland (approved):

PR #2950, opened 2025-07-24
fix: linter reporting xml unsupported type

gauron99:
linter failing in main https://github.com/knative/func/actions/runs/16489108500/job/46619720738#step:4:12
- Labels (map[string]string) type is unsupported in xml, exclude it

### DIFF
--- a/pkg/functions/client.go
+++ b/pkg/functions/client.go
@@ -181,7 +181,7 @@ type Instance struct {
 	Image         string            `json:"image" yaml:"image"`
 	Namespace     string            `json:"namespace" yaml:"namespace"`
 	Subscriptions []Subscription    `json:"subscriptions" yaml:"subscriptions"`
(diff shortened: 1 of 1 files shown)


pkg/functions/client.go
```diff
@@ -181,7 +181,7 @@ type Instance struct {
 	Image         string            `json:"image" yaml:"image"`
 	Namespace     string            `json:"namespace" yaml:"namespace"`
 	Subscriptions []Subscription    `json:"subscriptions" yaml:"subscriptions"`
-	Labels        map[string]string `json:"labels" yaml:"labels"`
+	Labels        map[string]string `json:"labels" yaml:"labels" xml:"-"`
```
matejvasek, line 184: Why only Labels require this?

gauron99: `trying to marshal unsupported type map[string]string, via x.Labels (staticcheck)`

because of the map type, im assuming by the linter message

PR #2953, opened 2025-07-26
fix: update console output to show both host and port 

### LINKED ISSUE #2900: func run --address does not work
It appears that the `--address` flag on `func run` does nothing.

RayyanSeliya:
# Changes
fixes #2900 

- Updated the non-JSON console output in [cmd/run.go](cci:7://file:///c:/Users/RAYYAN/Desktop/func/cmd/run.go:0:0-0:0) to use `net.JoinHostPort` for better host:port formatting
- The console output now shows both host and port in a more reliable format

## Why

The previous implementation only showed the port number in the console output, which could be confusing when running in different environments. The new implementation:
- Uses the standard library's `net.JoinHostPort` for consistent host:port formatting

## Testing

The changes were tested by:
1. Running the function locally and verifying the new output format
see the output of  testing 
```txt
rayyan@rayyan-seliya:/mnt/c/Users/RAYYAN/Desktop/func/myfunction$ ../func run --address=127.0.0.1:8081
{"level":"debug","time":1753544409,"message":"func runtime creating function instance"}
{"level":"debug","address":"127.0.0.1:8081","time":1753544409,"message":"function starting"}
{"level":"debug","time":1753544409,"message":"function does not implement Start. Skipping"}
{"level":"debug","time":1753544409,"message":"waiting for stop signals or errors"}
Function running on 127.0.0.1:8081
```

## Release Note

```release-note
Improved function run output to show both host and port when running locally

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"os"
 	"strconv"
 	"time"
@@ -282,7 +283,7 @@ func runRun(cmd *cobra.Command, newClient ClientFactory) (err error) {
 		}
 		fmt.Fprintln(cmd.OutOrStdout(), string(jsonData))
 	} else {
-		fmt.Fprintf(cmd.OutOrStderr(), "Running on host port %v\n", job.Port)
+		fmt.Fprintf(cmd.OutOrStderr(), "Function running on %s\n", net.JoinHostPort(job.Host, job.Port))
 	}
 
 	select {
--- a/test/e2e/scenario_extended_flow_test.go
+++ b/test/e2e/scenario_extended_flow_test.go
@@ -67,9 +67,18 @@ func TestFunctionExtendedFlow(t *testing.T) {
 					}
 					return
 				}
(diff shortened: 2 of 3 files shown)


RayyanSeliya:
hey @gauron99  i was getting fail on e2e test specifying a mismatch on the text display like before  we have like Running on host port... and now we have Function running on ... so the test was failing can retrigger the test ! thx ! Sorry still failing I think more logic is required separate extraction of host and port coz it is getting stuck on url thing as I can see on the test any suggestions?

matejvasek:
@RayyanSeliya you also need to update test, https://github.com/knative/func/blob/3641bfe5d03af71fcbfaa3065b90198da13646c3/test/e2e/scenario_extended_flow_test.go#L70-L69. You need to update the regex there.

RayyanSeliya:
> @RayyanSeliya you also need to update test, https://github.com/knative/func/blob/3641bfe5d03af71fcbfaa3065b90198da13646c3/test/e2e/scenario_extended_flow_test.go#L70-L69. You need to update the regex there.

Thx @matejvasek for helping me out i will update the regex in that test file  you mentioned

PR #2959, opened 2025-07-31
cleanup: Simplify hack/ components

gauron99:
Simplify components kept up to date in `/hack`
- now runs as part of update-deps.sh
- only need to modify json and autogen script
1) add/modify .json 
note: if a repo doesnt have owner/repo fields filled, dont auto-bump it
2) add the autogenerated file
3) celebrate with a beer

PS: tekton is not auto-updated because their release version fetched from github contains other text-string instead of just the clean "v<version>" ... need to update the script
PSS: Will update the other scripts as I go (makefile w springboot/typescript bumps etc.)

fixes https://github.com/knative/func/issues/2960


gauron99:
I forgot to update the test itself, let me do that tomorrow

gauron99:
OH I thought there is a GH token in the verify deps action `client without token` interesting

but I guess not

hack/cmd/components/main.go
```diff
@@ -0,0 +1,241 @@
...
+		cancel()
+		<-sigs
+		os.Exit(130)
+	}()
+
+	client := github.NewClient(nil)
+	if os.Getenv("GITHUB_TOKEN") != "" {
+		fmt.Println("using gh client with token")
+		client = client.WithAuthToken(os.Getenv("GITHUB_TOKEN"))
+	}
```
gauron99: Im adding a little if here so you dont always have to have a token and try without it if you need to run once after your local change
although now It seems that the action update deps actually doenst use this token either so might need to update later. Maybe lets see how the rate limiting is?

gauron99:
this has the same error as in https://github.com/knative/func/pull/2954#issuecomment-3144052833
lets see if actually a flake

Makefile
```diff
@@ -308,23 +308,17 @@ schema-check: ## Check that func.yaml schema is up-to-date
...
+## This is used by workflows
 
+## TODO: gauron99 -- include this in update-deps as well
 .PHONY: update-builder
-wf-update-builder: # Used in automation
-	cd hack && go run ./cmd/update-builder
-
-### end of automation section
+__update-builder: # Used in automation
+	cd hack && go run ./cmd/builder
```
matejvasek: Why have this in update-deps? Makes no sense to me.

gauron99: to unify having all our deps in one flow and one PR pretty much

gauron99:
@dsimansk can you PTAL if this looks good for the `hack/update-deps.sh` pls?

Makefile
matejvasek: bit update-builder does not change any source, it produces no PR

dsimansk:
> @dsimansk can you PTAL if this looks good for the `hack/update-deps.sh` pls?

Sure, it seems to work as intended.

lkingland (approved):
Looks good!

/hold until we can resolve Matej's question.  We can discuss it on this week's call if that's easier.

Makefile
gauron99: Ah i see, i mustve forgot to remove that comment for myself...

gauron99:
I wanted to update the name to `update-builder` dir in hack/cmd but I see theres a conflict so let me not complicate things further :D

PTAL

gauron99:
The Codecov was showing not uploaded coverages against base, but also I wasnt rebased on main and was 8 commits behind. Rebasing to see if this helps.

gauron99:
@lkingland @matejvasek LGTM pls

PR #2961, opened 2025-08-01
Fix for arm64 buildpacks

matejvasek:
There is a bug paketo-buildpacks/packit/issues/650. This commits mitigates the issue by explicitly setting the arch.


matejvasek:
PTAL @gauron99 @lkingland @jrangelramos

lkingland (approved):

Issue #2964, opened 2025-08-04
Host builder `host` and `base image` environment mismatch


matejvasek:
When local builder creates an image it includes locally populated venv directory.
The directory contains files that depend on local environment.
For example local version of the standard `C` library or local version of `Python`

These artifact might not work within base image.

Example: somebody tried to build a function that uses `numpy` on `Ubuntu 24.04.2 LTS`,  but the image would not work because of this mismatch (probably version of `Python`).

I am not sure how to mitigate this. We could build multiple base images with based on various distributions and various Python versions.
The host builder then could choose the appropriate base-image by looking at local environment.

lkingland:
Thanks Matej.

Just to clarify the impact: This affects Functions which 1) have libraries with C dependencies and 2) are deploying using the Host builder and 3) have a differing version of Python from that used by the runtime image.

PR #2965, opened 2025-08-04
fix: Python host builder base image version

matejvasek:
# Changes

related to #2964

- :bug: Host builder picks base-image with the same version of Python that is used locally.

/kind bug

```release-note
fix: host builder uses base-image with correct version of Python
```


lkingland (approved):
/approve

Nice call!

PR #2966, opened 2025-08-04
fix: force container=true for pack/s2i builders when not explicitly set

### LINKED ISSUE #2955: bug: function runs on host even when specified --builder
Running 

`func run --verbose --builder=pack --build=1` 

in python function actually runs with host because container is false by default

Contrary running 

`func run --verbose --builder=pack --build=1 --container=1`

 runs as expected

I suspect a simple check /validation somewhere is missing for the combination of builder container etc. Maybe it's  build config value is not checked ( result of config, ENVs, flags) etc.

/assign
/bug

RayyanSeliya:
- Added logic to automatically set `container=true` for containerized builders (pack/s2i) when `--container` flag is not explicitly provided by user
- Added smart auto-selection to choose host builder when `--container=false` is specified without explicit builder
- Fixes issue where `func run --builder=pack` would incorrectly run on host instead of container
- Solution checks if builder is pack/s2i and `--container` flag wasn't changed by user
- Resolves #2955

# Changes

- :bug: Fix pack/s2i builders incorrectly running on host instead of container
- :sparkles: Add smart auto-selection for --container=false to choose host builder
- :test_tube: Add comprehensive unit tests (`TestCtrBuilder`, `TestCtrBuilderConflict`, `TestSmartBuilderSelection`)
- Add automatic container=true enforcement for containerized builders when flag not explicitly set
- Preserve user choice when --container flag is explicitly provided

/kind bug

Fixes #2955

**Release Note**

```release-note
Enhanced builder/container flag handling: pack/s2i builders now default to container mode, --container=false automatically selects host builder, with validation for incompatible combinations.
```

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -179,12 +179,12 @@ func runRun(cmd *cobra.Command, newClient ClientFactory) (err error) {
 
 	// Smart auto-fix logic for builder/container compatibility
 	// This fixes the original bug where --builder=pack doesn't default to container=true
-	
+
 	// Case 1: Containerized builders (pack/s2i) should force container=true when not explicitly set
 	if (f.Build.Builder == "pack" || f.Build.Builder == "s2i") && !cfg.Container && !cmd.Flags().Changed("container") {
 		cfg.Container = true
 	}
-	
+
 	// Case 2: container=false should auto-select host builder when no builder explicitly set
 	if !cfg.Container && cmd.Flags().Changed("container") && !cmd.Flags().Changed("builder") {
 		f.Build.Builder = "host"
--- a/cmd/run_test.go
+++ b/cmd/run_test.go
@@ -705,6 +705,4 @@ func TestSmartBuilderSelection(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	
-	
 }


gauron99:
@RayyanSeliya Hi! thanks for the PR, much appreciated. I think this is a good addition, however Im not sure it fixes the issue completely (Thats my mistake for not mentioning it clearly in the issue). There is also another case, see below:
Running
`❯ func run --builder=pack --container=false`
tries to build with a `host` builder, which instead I think should error for an impossible match of flags because both `builder` and `container` were explicitly given.

cmd/run.go
```diff
@@ -180,6 +180,12 @@ func runRun(cmd *cobra.Command, newClient ClientFactory) (err error) {
 		return
 	}
 
+	// Force container=true for containerized builders if not explicitly set
+	// This fixes the bug where --builder=pack doesn't default to --container=true
+	if (f.Build.Builder == "pack" || f.Build.Builder == "s2i") && !cmd.Flags().Changed("container") {
```
gauron99: actually now, looking at this: We can also set container via `FUNC_CONTAINER` therefore doing
`FUNC_CONTAINER=false func run --builder=pack` should fail, giving an error for a missmatch between builder and container.

Running the above command with your changes builds with pack which I think should instead give an error as well.

RayyanSeliya:
> @RayyanSeliya Hi! thanks for the PR, much appreciated. I think this is a good addition, however Im not sure it fixes the issue completely (Thats my mistake for not mentioning it clearly in the issue). There is also another case, see below: Running `❯ func run --builder=pack --container=false` tries to build with a `host` builder, which instead I think should error for an impossible match of flags because both `builder` and `container` were explicitly given.

most welcome @gauron99  no worries for not mentioning i have inculcated the changes  in the new commit !! thx

cmd/run.go
RayyanSeliya: yeah i have taken care of this case too

RayyanSeliya:
hey @gauron99  i have also tested all the possibilities and tested can see the output here little bit messy !

```txt
rayyan@rayyan-seliya:/mnt/c/Users/RAYYAN/Desktop/func/test-fix$ 
# TEST 1: Original bug - should work now (force container=true)
echo "🧪 TEST 1: Pack builder auto-forces container mode"
../func-fixed run --builder=pack --build --registry=localhost:5000
# Expected:  Container operations (Docker pulls, etc.)


# TEST 2: Explicit container=false should ERROR  
echo "🧪 TEST 2: Pack + explicit container=false should error"
../func-fixed run --builder=pack --container=false --build --registry=localhost:5000
# Expected:  Error: "builder "pack" requires container mode but --container=false was explicitly set"

# TEST 3: Environment variable should ERROR
echo "🧪 TEST 3: FUNC_CONTAINER=false should error" 
FUNC_CONTAINER=false ../func-fixed run --builder=pack --build --registry=localhost:5000
# Expected:  Same error as above


# TEST 4: Host builder allows container=false (control test)
echo "🧪 TEST 4: Host builder allows container=false"
../func-fixed run --builder=host --container=false
# Expected:  Container operationscontainer=true --build --registry=localhost:5000


🧪 TEST 1: Pack builder auto-forces container mode
Building function image
Still building
Still building
^C^C
🧪 TEST 2: Pack + explicit container=false should error
Error: builder "pack" requires container mode but --container=false was explicitly set

🧪 TEST 3: FUNC_CONTAINER=false should error
Error: builder "pack" requires container mode but --container=false was explicitly set

🧪 TEST 4: Host builder allows container=false
Error: exec: "python": executable file not found in $PATH

🧪 TEST 5: Pack + explicit container=true works
Building function image
Still building
```

cmd/run.go
```diff
@@ -180,6 +181,22 @@ func runRun(cmd *cobra.Command, newClient ClientFactory) (err error) {
 		return
 	}
 
+	// Check for explicit container=false before applying auto-fix
+	envValue := os.Getenv("FUNC_CONTAINER")
+	containerExplicitlySetToFalse := cmd.Flags().Changed("container") ||
+		(envValue != "" && strings.ToLower(envValue) == "false")
```
gauron99: Theres no need to check the actual ENV value, this is actually handled by the Viper package. You can check this by putting a print statement after the run config is first created.
```go
	cfg = newRunConfig(cmd) // Will add Prompt on upcoming UX refactor
	fmt.Printf("container is %v\n", cfg.Container)
``` 
We already have the "final" container value.

gauron99 (commented):
Thanks for the quick followup! I appreciate your time on this. I have a suggestion for a nicer solution using the Viper package (which takes care of the value fetching!) and you get to work with the final value already.

I suspect that a simple missmatch "if" in the
`func (c runConfig) Validate(cmd *cobra.Command, f fn.Function) (err error) {` function should be sufficient.

cmd/run.go
gauron99: see newer review

cmd/run.go
gauron99: Note: We want to do it this way also to keep the priorities, for example:
```bash
FUNC_CONTAINER=false func run --container=true
```
We have specified container 2 different ways but one has prio over the other (check the output)

matejvasek:
@RayyanSeliya  I would add tests:

```go
// in cmd/run_test.go
func TestCtrBuilder(t *testing.T) {
	var runner = mock.NewRunner()
	var builder = mock.NewBuilder()
	tmp := t.TempDir()

	fnPath := filepath.Join(tmp, "fn")
	f, err := fn.New().Init(fn.Function{Root: fnPath, Runtime: "python"})
	if err != nil {
		t.Fatal(err)
	}
	err = f.Write()
	if err != nil {
		t.Fatal(err)
	}
	t.Chdir(fnPath)

	cmd := NewRunCmd(NewTestClient(
		fn.WithRunner(runner),
		fn.WithBuilder(builder),
		fn.WithRegistry(TestRegistry),
	))
	cmd.SetArgs([]string{"--builder=pack", "--build=1"})
	time.AfterFunc(time.Second, func() {
		cancel()
	})
	err = cmd.ExecuteContext(ctx)
	if err != nil {
		t.Fatal(err)
	}
	// verify that builder was called since `--builder=pack` was provided
	if !builder.BuildInvoked {
		t.Error("builder has not been invoked")
	}
}

func TestCtrBuilderConflict(t *testing.T) {
	var runner = mock.NewRunner()
	var builder = mock.NewBuilder()
	tmp := t.TempDir()

	fnPath := filepath.Join(tmp, "fn")
	f, err := fn.New().Init(fn.Function{Root: fnPath, Runtime: "python"})
	if err != nil {
		t.Fatal(err)
	}
	err = f.Write()
	if err != nil {
		t.Fatal(err)
	}
	t.Chdir(fnPath)

	cmd := NewRunCmd(NewTestClient(
		fn.WithRunner(runner),
		fn.WithBuilder(builder),
		fn.WithRegistry(TestRegistry),
	))
	cmd.SetArgs([]string{"--builder=pack", "--build=1", "--container=0"})
	time.AfterFunc(time.Second, func() {
		cancel()
	})
	err = cmd.ExecuteContext(ctx)
        // since explicit flag `--builder=pack` and `--container=0` are contradiction we want error
	if err == nil {
		t.Error("error expected but got <nil>")
	}
}

```

test/e2e/scenario_no_container_test.go
```diff
@@ -28,7 +28,7 @@ func TestFunctionRunWithoutContainer(t *testing.T) {
 	var funcPath = filepath.Join(t.TempDir(), funcName)
 
 	knFuncTerm1 := common.NewKnFuncShellCli(t)
-	knFuncTerm1.Exec("create", "--language", "go", "--template", "http", funcPath)
+	knFuncTerm1.Exec("create", "--language", "go", "--template", "http", "--builder", "host", funcPath)
```
matejvasek: Shouldn't be host builder default here? We should not explicitly set it IMO.
wdyt @lkingland @gauron99

RayyanSeliya:
> @RayyanSeliya I would add tests:
> 
> ```go
> // in cmd/run_test.go
> func TestCtrBuilder(t *testing.T) {
> 	var runner = mock.NewRunner()
> 	var builder = mock.NewBuilder()
> 	tmp := t.TempDir()
> 
> 	fnPath := filepath.Join(tmp, "fn")
> 	f, err := fn.New().Init(fn.Function{Root: fnPath, Runtime: "python"})
> 	if err != nil {
> 		t.Fatal(err)
> 	}
> 	err = f.Write()
> 	if err != nil {
> 		t.Fatal(err)
> 	}
> 	t.Chdir(fnPath)
> 
> 	cmd := NewRunCmd(NewTestClient(
> 		fn.WithRunner(runner),
> 		fn.WithBuilder(builder),
> 		fn.WithRegistry(TestRegistry),
> 	))
> 	cmd.SetArgs([]string{"--builder=pack", "--build=1"})
> 	ctx, cancel := context.WithCancel(context.Background())
> 	time.AfterFunc(time.Second, func() {
> 		cancel()
> 	})
> 	err = cmd.ExecuteContext(ctx)
> 	if err != nil {
> 		t.Fatal(err)
> 	}
> 	// verify that builder was called since `--builder=pack` was provided
> 	if !builder.BuildInvoked {
> 		t.Error("builder has not been invoked")
> 	}
> }
> 
> func TestCtrBuilderConflict(t *testing.T) {
> 	var runner = mock.NewRunner()
> 	var builder = mock.NewBuilder()
> 	tmp := t.TempDir()
> 
> 	fnPath := filepath.Join(tmp, "fn")
> 	f, err := fn.New().Init(fn.Function{Root: fnPath, Runtime: "python"})
> 	if err != nil {
> 		t.Fatal(err)
> 	}
> 	err = f.Write()
> 	if err != nil {
> 		t.Fatal(err)
> 	}
> 	t.Chdir(fnPath)
> 
> 	cmd := NewRunCmd(NewTestClient(
> 		fn.WithRunner(runner),
> 		fn.WithBuilder(builder),
> 		fn.WithRegistry(TestRegistry),
> 	))
> 	cmd.SetArgs([]string{"--builder=pack", "--build=1", "--container=0"})
> 	ctx, cancel := context.WithCancel(context.Background())
> 	time.AfterFunc(time.Second, func() {
> 		cancel()
> 	})
> 	err = cmd.ExecuteContext(ctx)
>         // since explicit flag `--builder=pack` and `--container=0` are contradiction we want error
> 	if err == nil {
> 		t.Error("error expected but got <nil>")
> 	}
> }
> ```

thx @matejvasek  for this test suggestion i have added this in my recent commit !!

test/e2e/scenario_no_container_test.go
RayyanSeliya: yeah @matejvasek! ,
You are  absolutely right ,forcing users to explicitly specify `--builder host` when they want `--container=false ` feels clunky. I think we could make this much more intuitive by having the CLI auto-select the appropriate builder based on the container flag.
The idea would be:> when someone runs `func run --container=false`, the system automatically picks `host ` builder since it is  the only one that makes sense for non-containerized runs. We would still validate and error out for incompatible combinations like `--container=false  --builder=pack`, but the happy path would just work without users needing to know the "magic combination".
This approach would also make the E2E test work naturally without requiring the explicit `--builder host ` flag , the system would be smart enough to choose the right builder automatically.
what do you think?  worth implementing this smarter UX logic, or should we stick with the explicit approach? @lkingland @gauron99 would love your thoughts on this as well!

gauron99: I would go the other way around, silencing the container flag more in favor of builder. I wasnt going to mention this because its beyond the scope of the issue but since you mentioned this, if you're up for it, you can even probably get rid of the container flag all together :smile: Having the builder would determine if it is a containerized or non-containerized build and you really dont need the container, hence now it really just adds the complexity of 2 flags needing to cooperate.

Since the host builder is always container-less and pack/s2i are always containerized, you determine the containerization via the builder flag alone. And you remove complexity which is a +

If you're not up for this, no worries, lets just get this PR finished for the bug and Im gonna tackle that later. In this case let's favor Matej's approach of keeping the host builder default then - Im guessing all you can do in this case is check the flag override ( what you did initially? :smile: )

RayyanSeliya: Hey @gauron99!
I went ahead and implemented the smart auto-selection approach , now `func run --container=false`  automatically picks the host builder, and` func run --builder=pack` forces container mode. The E2E test works naturally without needing explicit flags.
I love your idea about removing the container flag entirely and just using builder to determine everything , that would definitely be cleaner,  For now this fixes the bug and improves the UX, but for the bigger refactor we can tackle it later i  guess

RayyanSeliya:
> thanks for the contribution! This looks good to me
> /lgtm
> /approve
> /hold for Matej's review

Thx @gauron99 &@matejvasek for being patience throughout the pr !!

PR #2979, opened 2025-08-11
oci: refactors the auth for pusher

KapilSareen:
**Refactor OCI Pusher Authentication**

## Changes

* Reuse `cmd/client`’s `credentialsProvider` instead of relying on the keychain
* Add interactive fallback for the pusher when no credentials are provided by a `credentialsProvider`
* Introduce additional credential loaders (ACR, GCP) for broader registry support

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -394,9 +394,14 @@ func (c buildConfig) clientOptions() ([]fn.Option, error) {
 	o := []fn.Option{fn.WithRegistry(c.Registry)}
 	switch c.Builder {
 	case builders.Host:
+		t := newTransport(c.RegistryInsecure) // may provide a custom impl which proxies
+		creds := newCredentialsProvider(config.Dir(), t)
 		o = append(o,
 			fn.WithBuilder(oci.NewBuilder(builders.Host, c.Verbose)),
-			fn.WithPusher(oci.NewPusher(c.RegistryInsecure, false, c.Verbose)))
+			fn.WithPusher(oci.NewPusher(c.RegistryInsecure, false, c.Verbose,
+				oci.WithCredentialsProvider(creds),
+				oci.WithVerbose(c.Verbose))),
+		)
 	case builders.Pack:
 		o = append(o,
 			fn.WithBuilder(pack.NewBuilder(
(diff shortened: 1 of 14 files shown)


pkg/oci/pusher.go
```diff
@@ -22,29 +21,63 @@ import (
 	fn "knative.dev/func/pkg/functions"
 )
 
+type Credentials struct {
```
KapilSareen, line 24: Transferring some key types from docker package to here, with an idea to slowly replace dependency on docker package.

cmd/client.go
```diff
@@ -100,19 +101,26 @@ func newTransport(insecureSkipVerify bool) fnhttp.RoundTripCloser {
 // newCredentialsProvider returns a credentials provider which possibly
 // has cluster-flavor specific additional credential loaders to take advantage
 // of features or configuration nuances of cluster variants.
-func newCredentialsProvider(configPath string, t http.RoundTripper) docker.CredentialsProvider {
+func newCredentialsProvider(configPath string, t http.RoundTripper) oci.CredentialsProvider {
+	additionalLoaders := k8s.GetOpenShiftDockerCredentialLoaders()
+    // additionalLoaders := append(
```
KapilSareen: I plan to replace the keychain flow from the oci's authOption function with this. (with adding support for ECR and ACR creds as well).

This way we use the already existing support for prompting in case of fallback and openShift as well.

(NOT SURE THO) Also I think it verifies the creds as well now.

cmd/build.go
```diff
@@ -394,9 +394,14 @@ func (c buildConfig) clientOptions() ([]fn.Option, error) {
 	o := []fn.Option{fn.WithRegistry(c.Registry)}
 	switch c.Builder {
 	case builders.Host:
+		t := newTransport(c.RegistryInsecure) // may provide a custom impl which proxies
```
KapilSareen, line 397: This is just a temporary place to configure the oci Pusher with appropriate options, eventually it will replace the original docker pusher in `cmd/client.go` once complete.

KapilSareen: Decided to use Credential provider because it thought it was more robust than existing oci's auth (ctx + keychains) plus prompt for fallback was already implemented there and easy addition of more credential fallbacks is way more natural than the keychain approach (imo, might be wrong).

pkg/oci/pusher.go
lkingland: It's ok if the implementation has a dependency on the Docker package to support the case where the user has docker installed; we just don't want to create a situation where they _have_ to have docker (or podman) installed 👍🏻

Issue #2973, opened 2025-08-07
Improvement: Reduce complexity of container and builder flag


gauron99:
Currently there is an unreasonable complexity with `builder` and `container` flag when using `func run`. Since it is determinable if we want to run the function in a container or not based on the builder, we can remove the container flag completely.

/assign

PR #2983, opened 2025-08-13
Cleanup: Remove container flag & make the builder dynamic (var II)

gauron99:
# Variant II
- This version creates a dynamic builder which prefers host builder instead of pack wherever possible, that means when a runtime has an oci builder.
- These changes are more complex and reaching greater multitude of systems, including changing some default behavior of current implementations and their respective tests in comparison to variant I.
- With this variant, upon creating oci builder for a new runtime would immediately prefer the host builder by default making the adoption and transition easier.
- The current "non standard" implementations could be reversed once the host builder becomes the standard default.

# Changes
- :broom: Remove the `--container` flag
- now the logic in `run` cmd uses the current builder
- assign `host` builder as default (not only for run) if runtime supports it, otherwise `pack`
- builder determination now depends on runtime

/kind cleanup
Fixes https://github.com/knative/func/issues/2973

**Release Note**
```release-note
Remove the --container flag - builds are determined via builder dynamically; Default to 'host' for oci-supported runtimes
```

**Docs**
```docs

```

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -98,7 +98,7 @@ EXAMPLES
 	// Globally-Configurable Flags:
 	// Options whose value may be defined globally may also exist on the
 	// contextually relevant function; sets are flattened above via cfg.Apply(f)
-	cmd.Flags().StringP("builder", "b", cfg.Builder,
+	cmd.Flags().StringP("builder", "b", defaultBuilder(f),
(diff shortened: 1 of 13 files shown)


gauron99:
Closing this in favor of variant I https://github.com/knative/func/pull/2987

PR #2986, opened 2025-08-14
pin knative.dev/pkg so that it compiles

dprotaso:
tektoncd uses an old metrics package we dropped. thus we need to
wait for them to migrate away before resuming these bumps

For now we pin knative.dev/pkg


gauron99:
We can bump the tekton we use I think

dprotaso:
tekton hasn't made the necessary changes yet

dprotaso:
This is a workaround to unblock knative dep updates

PR #2987, opened 2025-08-15
Cleanup: Remove container flag - variant I

gauron99:
# Variant I
- This version simply removes the container flag
- Much simpler solution compared to Variant II
- We lose the host builder "defaultability" - you now have to specify `--builder=host` for all languages to use it.
- 

# Changes
- :broom: Remove the `--container` flag for `func run`
- The containerization is determined via builder now in `func run`

/kind cleanup
Fixes https://github.com/knative/func/issues/2973

**Release notes**
```release-note

```

 **Docs**
```docs

```


lkingland (approved):
Lookin' good to me

PR #2992, opened 2025-08-20
bug: python injector didnt choose signature

gauron99:
# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

PR #2994, opened 2025-08-20
bug: bug: python injector needs the invoke default

gauron99:
# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


gauron99:
This is an immediate fix for https://github.com/knative/func/pull/2992. The previous fixed that python http function built via pack builder didnt work. (cloudevent hard coded)

This fixes the signature choice since we need to set a default here just like the standard scaffolding does.

PR #3004, opened 2025-08-28
fix: hotfix remote python with pack builder deploy

gauron99:
- This is a hotfix for the remote python pack deployment.
- Implements the "python injector" workaround for remote as well.
- Please note that this is bound to change! with a larger scale re-do of the scaffolding implementation (the injector wont be needed at all)


gauron99:
This does fix the immediate issue of direct upload remote deployment ( deploy local code to cluster ) with python + pack but the "hacky" changes do fail the non-direct upload (using the git deployment) therefore instead of digging further into this temporary hack - Im going to start the subsequent work immediately of the scaffolding rework, which will include the python + pack fix.

If you are reading this coming from the issue linked, please feel free to pull this PR branch and use the python+pack remote deployment when using local code in the meantime! Thank you for understanding.

closes issue to track is https://github.com/knative/func/issues/3005

PR #3054, opened 2025-09-29
Improve the error description when no name is provided in `func delete`

### LINKED ISSUE #3053: Improve `func delete` UX: provide clear guidance when function name is missing
## **Current behavior:**
When running `func delete` without a function name, the command shows a cryptic error message that doesn't help users understand what they need to do:

```txt
rayyan@rayyan-seliya:/mnt/c/Users/RAYYAN/Desktop/func$ func delete
Error: name required
```

This forces users to:
1.Guess what's wrong - unclear what "name required" means
2.Look up documentation to understand the command usage
3.Still not understand the available deletion options

## **Proposed improvement:**
Provide clear, educational guidance about how to delete functions:

```txt
Error: function name required for deletion.

Function name is required for deletion.

You can delete functions in two ways:

1. By name (from anywhere):
   func delete myfunction                     Delete function by name
   func delete myfunction --namespace apps    Delete from specific namespace

2. By path (when inside function directory):
   func delete --path /path/to/function       Delete function at specific path

Examples:
   func delete myfunction                     Delete 'myfunction' from cluster
   func delete myfunction --namespace prod    Delete from 'prod' namespace
   func delete --path ./myfunction            Delete function at path
  
For more options, run 'func delete --help'
```

**Solution:**
Implement a 2-layer error system that provides both technical error information and user-friendly guidance.

## **Expected Workflow:**
According to the [official documentation](https://knative.dev/docs/functions/building-functions/), building requires:

1,.By name :` func delete <function-name>`

2.By path : `func delete --path <path>`

The improved error message guides users through both deletion methods with clear examples.

## **User Impact:**
- **Current**: Cryptic error → confusion → need to look up documentation
- **Improved**:Immediate understanding → better user experience

RayyanSeliya:
# Changes
- :broom: Fix confusing UX in `func delete` when no function name provided
- :broom: Replace cryptic "name required" error with clear guidance
- :broom: Show educational error message with deletion workflow examples and help pointer

/kind enhancement

Fixes #3053 

Previously, `func delete` would show a cryptic "name required" error that didn't help users understand how to delete functions. Now it immediately shows clear guidance on the two ways to delete functions (by name or by path) with practical examples.
Follows two layer approach as suggested to do so !

**Release Note**

```release-note
Improve func delete user experience by replacing cryptic "name required" error with clear guidance on how to delete functions using name or path options.
```

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -40,7 +40,7 @@ No local files are deleted.
 			// Layer 2: Catch technical errors and provide CLI-specific user-friendly messages
 			err := runDelete(cmd, args, newClient)
 			if err != nil && errors.Is(err, fn.ErrNameRequired) {
-				return formatDeleteNameRequiredError(err)
+				return NewErrDeleteNameRequired(err)
 			}
 			return err
 		},
@@ -176,20 +176,31 @@ func (c deleteConfig) Prompt() (deleteConfig, error) {
 	return dc, err
 }
 
-// formatDeleteNameRequiredError provides CLI-specific user-friendly error message
-// for delete command when function name is required
-func formatDeleteNameRequiredError(err error) error {
-	return fmt.Errorf(`%v
+// ErrDeleteNameRequired wraps core library errors with CLI-specific context
+// for delete operations that require a function name or path.
+type ErrDeleteNameRequired struct {
+	// Underlying error from the core library (e.g., fn.ErrNameRequired)
+	Err error
+}
+
+// NewErrDeleteNameRequired creates a new ErrDeleteNameRequired wrapping the given error
+func NewErrDeleteNameRequired(err error) ErrDeleteNameRequired {
+	return ErrDeleteNameRequired{Err: err}
+}
+
+// Error implements the error interface with CLI-specific help text
+func (e ErrDeleteNameRequired) Error() string {
+	return fmt.Sprintf(`%v
 
 Function name is required for deletion (or --path not specified).
 
 You can delete functions in two ways:
 
-1. By name :
+1. By name:
    func delete myfunction                     Delete function by name
    func delete myfunction --namespace apps    Delete from specific namespace
 
-2. By path :
+2. By path:
    func delete --path /path/to/function       Delete function at specific path
 
 Examples:
@@ -197,5 +208,5 @@ Examples:
    func delete myfunction --namespace prod    Delete from 'prod' namespace
    func delete --path ./myfunction            Delete function at path
 
-For more options, run 'func delete --help'`, err)
+For more options, run 'func delete --help'`, e.Err)
 }


cmd/delete.go
```diff
@@ -36,7 +37,28 @@ No local files are deleted.
 		PreRunE:           bindEnv("path", "confirm", "all", "namespace", "verbose"),
 		SilenceUsage:      true, // no usage dump on error
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runDelete(cmd, args, newClient)
+			// Layer 2: Catch technical errors and provide CLI-specific user-friendly messages
+			err := runDelete(cmd, args, newClient)
+			if err != nil && errors.Is(err, fn.ErrNameRequiredForDelete) {
```
lkingland: This is good!  I would just use a general `fn.ErrNameRequired` as the error's name.  The fact it's "for delete" is implied by its context here (so you can reuse ErrNameRequired in other places where... a name is required)

cmd/delete.go
```diff
@@ -36,7 +37,28 @@ No local files are deleted.
...
+				return fmt.Errorf(`%v
```
lkingland: This is exactly what I was thinking as well:  Print the error message as it was returned from the client, and then append "cli specific" text to help out more.

Since this text is a bit long, you might want to consider breaking this out into either a separate function dedicated to error formatting, or create an "cli specific" version of the error with the full text, and the layer 1 error as a member:

```
return ErrDeleteNameRequired(err)
```
where err is a fn.ErrNameRequired and ErrDeleteNameRequired returns the CLI-specific error you have here.

Experiment with different patterns/structures to see what flows well.

Good start

RayyanSeliya:
hey @lkingland  @gauron99  i just wanted to tell that in this case  as i have created an explicit function for the typed error coz its a bit lengthy so i think in future we will need to shift this type of implementation coz it will affect the code readability as david sir you mentioned like we can create an explicit file in `cmd/errors.go` or something like that as if i will go forward for implementing the issues i have mentioned in the docs of `func build`  and `func deploy` it will get flooded by the long description in the code it will  be so messy what are your thoughts on this ???

lkingland (commented):
I think keeping the errors in the command's file (`delete.go`) is fine because it's specific to the delete command.  The core errors (like `fn.ErrDeleteRequired`) are shared, so putting them in a dedicated file (pkg/functions/errors.go) seems to make sense.

Here's a suggestion on how to use a custom error types to help make this a bit cleaner:

```Go
// ErrDeleteNameRequired wraps core library errors with CLI-specific context
// for delete operations that require a function name or path.
type ErrDeleteNameRequired struct {
	// Underlying error from the core library (e.g., fn.ErrNameRequired)
	Err error
}

// NewErrDeleteNameRequired creates a new ErrDeleteNameRequired wrapping the given error
func NewErrDeleteNameRequired(err error) *ErrDeleteNameRequired {
	return &ErrDeleteNameRequired{Err: err}
}

// Error implements the error interface with CLI-specific help text
func (e *ErrDeleteNameRequired) Error() string {
	return fmt.Sprintf(`%v
Function name is required for deletion (or --path not specified).
You can delete functions in two ways:
1. By name:
   func delete myfunction                     Delete function by name
   func delete myfunction --namespace apps    Delete from specific namespace
2. By path:
   func delete --path /path/to/function       Delete function at specific path
Examples:
   func delete myfunction                     Delete 'myfunction' from cluster
   func delete myfunction --namespace prod    Delete from 'prod' namespace
   func delete --path ./myfunction            Delete function at path
For more options, run 'func delete --help'`, e.Err)
}
```

The naming convention here would be "Err{Command Name}{Core Error Name}"

This may seem like it's no different than the formatting function you suggested here, but it allows us to do interesting things like this in the future if we need to:

```Go
// Unwrap returns the underlying error for use with errors.Is/As
func (e *ErrDeleteNameRequired) Unwrap() error {
	return e.Err
}

// Is allows this error to match the underlying error type
// This enables errors.Is(err, fn.ErrNameRequired) to work through the wrapper
func (e *ErrDeleteNameRequired) Is(target error) bool {
	return errors.Is(e.Err, target)
}
```
(No need to implement those methods now, however; that's just to give you an example of the flexibility afforded when using types.

RayyanSeliya:
thanks @lkingland  for the clarification  i will follow your suggestion and keep the error in the command file `delete.go` as of now if the code grows too much coz of typed errors we can move it into a separate file  like  `cmd/delete_errors.go` . And for shared/common cases, we’ll keep them in the dedicated locations as  ` pkg/functions/errors.go`

lkingland (approved):
/lgtm

Just need those tests to pass..

gauron99:
It looks like you might need to rebase @RayyanSeliya . This looks like its missing https://github.com/knative/func/pull/3055/files

TLDR: theres been changes in one of the dependencies we use to create clusters for these tests and the instalation changed, hence none of the tests that use a cluster will work

RayyanSeliya:
Sure @gauron99  will update the branch I have seen many  chats in these days on slack 🙂 going on so will do it after my exam today in theate evening I was busy with my exams so was inactive !! Thx !

RayyanSeliya:
hey @gauron99  is this quarkus test related to the changes in the pr i guess no i have updated the branch can have a look now

gauron99 (approved):
/lgtm

PR #3055, opened 2025-09-30
dapr install upgrade

lkingland:
- Update Dapr to 1.16
- Install Dapr's Redis using docs-suggested standard deployment

Fixes some problems with the dapr installation which was causing knative services to never become ready due to their Dapr sidecars' hangups

### DIFF
--- a/hack/allocate.sh
+++ b/hack/allocate.sh
@@ -351,9 +351,54 @@ dapr_runtime() {
   # to only start a single instance rather than four.
   # helm repo add bitnami https://charts.bitnami.com/bitnami
   echo "${blue}- Redis ${reset}"
-  $HELM repo add bitnami https://charts.bitnami.com/bitnami
-  $HELM install redis bitnami/redis --set image.tag=6.2
-  $HELM repo update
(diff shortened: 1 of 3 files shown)


matejvasek:
/override "Func Integration Test / Integration Test (ubuntu-latest)"

matejvasek:
Just override the test @lkingland .

Or you could try
```patch
diff --git a/pkg/pipelines/tekton/gitlab_int_test.go b/pkg/pipelines/tekton/gitlab_int_test.go
index 2f6bd4776..cf321e59d 100644
--- a/pkg/pipelines/tekton/gitlab_int_test.go
+++ b/pkg/pipelines/tekton/gitlab_int_test.go
@@ -503,7 +503,7 @@ func getAPIToken(baseURL, username, password string) (string, error) {
 	}
 
 	data := struct {
-		NewToken string `json:"new_token,omitempty"`
+		NewToken string `json:"token,omitempty"`
 	}{}
 	e := json.NewDecoder(resp.Body)
 	err = e.Decode(&data)
 ```

PR #3058, opened 2025-10-01
Improve func build/deploy UX: validate registry before prompting

### LINKED ISSUE #3057: func `build/deploy` prompts for registry before validation - confusing UX
## **Current behavior:**
When running `func deploy` or `func build` inside a function directory without specifying a registry, the commands prompt for registry information and after doing ctrl + C  it provides the error of registry required it should validate first rather than asking first for the registry , creating a confusing user experience:

```txt
rayyan@rayyan-seliya:/mnt/c/Users/RAYYAN/Desktop/func/myfunction$ func deploy
A registry for function images is required. For example, 'docker.io/tigerteam'.
? Registry for function images: 
Error: registry required to build function, please set with `--registry` or the FUNC_REGISTRY environment variable
```

## **Proposed improvement:**
Check if a function exists before prompting for any configuration and provide clear, educational guidance:

```txt
Error: registry required to build or deploy function, please set with `--registry` or the FUNC_REGISTRY environment variable

Try this:
  func deploy --registry ghcr.io/myuser    Deploy with registry

Or set the FUNC_REGISTRY environment variable:
  export FUNC_REGISTRY=ghcr.io/myuser
  func deploy

Common registries:
  ghcr.io/myuser       GitHub Container Registry
  docker.io/myuser     Docker Hub
  quay.io/myuser       Quay.io

```

```txt
Error: registry required to build or deploy function, please set with `--registry` or the FUNC_REGISTRY environment variable

Try this:
  func build --registry ghcr.io/myuser    Build with registry

Or set the FUNC_REGISTRY environment variable:
  export FUNC_REGISTRY=ghcr.io/myuser
  func build

Common registries:
  ghcr.io/myuser       GitHub Container Registry
  docker.io/myuser     Docker Hub
  quay.io/myuser       Quay.io

Or specify full image name:
  func build --image ghcr.io/myuser/myfunction:latest

For more options, run 'func build --help'



```

**Solution:**
Validate the missing registry error first before asking for a registry prompt ! 
Use a 2 layer approach as discused ! 

## **Expected Workflow:**
According to the official documentation, building and deploying requires:
1. Create a function with `func create --language <lang> <name>`
2. Navigate to the function directory (`cd <name>`) OR use `--path` flag
3. Build with `func build` or `func build --registry <registry>`
4. Deploy with `func deploy` or `func deploy --registry <registry>`

The improved error message guides users through this exact workflow.

## **User Impact:**
- **Current**: Confusing prompts → manual cancellation → unclear error
- **Improved**: Immediate clear guidance → better onboarding experience

RayyanSeliya:
# Changes

- :bug: Fix confusing UX in func build/deploy commands
- :broom: Implement 2-layer error system for better user guidance
- :broom: Validate registry requirement before prompting users

/kind enhancement

Fixes #3057 

**Solution:**
Validate the missing registry error first before asking for a registry prompt! Use a 2-layer approach as discussed!

## **Implementation Details:**
- Refactored `func build` and `func deploy` commands to validate registry requirement before prompting
- Implemented 2-layer error system for both "no function found" and "registry required" scenarios
- Enhanced error messages with actionable guidance and examples
- Improved UX by preventing unnecessary interactive prompts

**Release Note**

```release-note
func build and func deploy commands now provide clearer error messages when registry is missing, with actionable guidance instead of confusing prompts.
```

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -159,6 +159,51 @@ func runBuild(cmd *cobra.Command, _ []string, newClient ClientFactory) (err erro
 		f   fn.Function
 	)
 	if cfg, err = newBuildConfig().Prompt(); err != nil { // gather values into a single instruction set
+		// Layer 2: Catch technical errors and provide CLI-specific user-friendly messages
+		
+		// Check if it's a "not initialized" error (no function found)
+		var errNotInit *fn.ErrNotInitialized
+		if errors.As(err, &errNotInit) {
+			return fmt.Errorf(`%w
+
+To build a function, you need to point to a function directory or use a --path flag approach.
+
+Try this:
+  func create --language go myfunction    Create a new function
+  cd myfunction                          Go into the function directory
+  func build --registry <registry>       Build the function container
+
+Or navigate to an existing function:
+  cd path/to/your/function
+  func build --registry <registry>
+
+Or use --path flag:
+  func build --path /path/to/function --registry <registry>
+
+For more options, run 'func build --help'`, err)
+		}
+		
+		// Check if it's a registry required error (function exists but no registry)
+		if errors.Is(err, fn.ErrRegistryRequired) {
+			return fmt.Errorf(`%w
+
+Try this:
+  func build --registry ghcr.io/myuser    Build with registry
+  
+Or set the FUNC_REGISTRY environment variable:
+  export FUNC_REGISTRY=ghcr.io/myuser
+  func build
+
+Common registries:
+  ghcr.io/myuser       GitHub Container Registry
+  docker.io/myuser     Docker Hub
+  quay.io/myuser       Quay.io
+
+Or specify full image name:
+  func build --image ghcr.io/myuser/myfunction:latest
+
+For more options, run 'func build --help'`, err)
+		}
 		return
 	}
 	if err = cfg.Validate(); err != nil { // Perform any pre-validation
@@ -312,7 +357,21 @@ func (c buildConfig) Prompt() (buildConfig, error) {
 	if err != nil {
 		return c, err
 	}
-	if (f.Registry == "" && c.Registry == "" && c.Image == "") || c.Confirm {
+	
+	// Check if function exists first
+	if !f.Initialized() {
+		// Return a specific error for uninitialized function
+		return c, fn.NewErrNotInitialized(f.Root)
+	}
+	
+	// Check if registry/image is missing BEFORE prompting
+	if (f.Registry == "" && c.Registry == "" && c.Image == "") {
+		// Return error immediately - don't prompt
+		return c, fn.ErrRegistryRequired
+	}
+	
+	// Only prompt if in confirm mode and registry exists
+	if c.Confirm {
 		fmt.Println("A registry for function images is required. For example, 'docker.io/tigerteam'.")
 		err := survey.AskOne(
 			&survey.Input{Message: "Registry for function images:", Default: c.Registry},
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -254,19 +254,29 @@ func runDeploy(cmd *cobra.Command, newClient ClientFactory) (err error) {
 	if !f.Initialized() {
 		if !cfg.Remote || f.Build.Git.URL == "" {
 			// Only error if this is not a fully remote build
-			return fmt.Errorf(`no function found in current directory.
-You need to be inside a function directory to deploy it.
+			// Layer 2: Wrap technical error with CLI-specific guidance
+			var errNotInit *fn.ErrNotInitialized
+			notInitErr := fn.NewErrNotInitialized(f.Root)
+			if errors.As(notInitErr, &errNotInit) {
+				return fmt.Errorf(`%w
+
+To deploy a function, you need to point to a function directory or use a --path flag approach.
 
 Try this:
   func create --language go myfunction    Create a new function
   cd myfunction                          Go into the function directory
   func deploy --registry <registry>      Deploy to the cloud
 
-Or if you have an existing function:
-  cd path/to/your/function              Go to your function directory
-  func deploy --registry <registry>     Deploy the function
+Or navigate to an existing function:
+  cd path/to/your/function
+  func deploy --registry <registry>
 
-For more detailed deployment options, run 'func deploy --help'`)
+Or use --path to deploy from anywhere:
+  func deploy --path /path/to/function --registry <registry>
+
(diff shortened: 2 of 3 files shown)


RayyanSeliya:
PTAL @gauron99  when you have some moments !

pkg/functions/errors.go
```diff
@@ -27,7 +27,7 @@ var (
 	// TODO: change the wording of this error to not be CLI-specific;
 	// eg "registry required".  Then catch the error in the CLI and add the
 	// cli-specific usage hints there
-	ErrRegistryRequired = errors.New("registry required to build function, please set with `--registry` or the FUNC_REGISTRY environment variable")
+	ErrRegistryRequired = errors.New("registry required to build or deploy function, please set with `--registry` or the FUNC_REGISTRY environment variable")
```
gauron99: lets actually change this to be more technical since this is L1 and add the verbosity in the L2.
```suggestion
	ErrRegistryRequired = errors.New("registry required")
```

cmd/deploy.go
```diff
@@ -278,6 +288,24 @@ For more detailed deployment options, run 'func deploy --help'`)
 
 	// Now that we know function exists, proceed with prompting
 	if cfg, err = cfg.Prompt(); err != nil {
+		// Layer 2: Catch technical errors and provide CLI-specific user-friendly messages
+		if errors.Is(err, fn.ErrRegistryRequired) {
+			return fmt.Errorf(`%w
+
+Try this:
+  func deploy --registry ghcr.io/myuser    Deploy with registry
```
gauron99: I would be in favor of a little change here so that the line is easily copied and pasted. Keep the first line only as code since we are adding some suggestions of common registries below
```suggestion
  func deploy --registry ghcr.io/myuser
```

cmd/build.go
```diff
@@ -312,7 +357,21 @@ func (c buildConfig) Prompt() (buildConfig, error) {
...
+	if !f.Initialized() {
+		// Return a specific error for uninitialized function
+		return c, fn.NewErrNotInitialized(f.Root)
+	}
+
+	// Check if registry/image is missing BEFORE prompting
+	if f.Registry == "" && c.Registry == "" && c.Image == "" {
+		// Return error immediately - don't prompt
+		return c, fn.ErrRegistryRequired
+	}
```
gauron99, line 346: I think when a Function is initialized AND registry/image is missing, we can prompt for registry. Because this would effectively make `--registry` a forced flag if you havent specified FUNC_REGISTRY or used the config file.
Something like
```
func init
func deploy  (/build)
``` 
without any prior setup should not fail, rather prompt for the registry since this is the minimal requirement.
I think thats a bit more user friendly

cmd/deploy.go
```diff
@@ -254,19 +254,29 @@ func runDeploy(cmd *cobra.Command, newClient ClientFactory) (err error) {
...
 			// Only error if this is not a fully remote build
-			return fmt.Errorf(`no function found in current directory.
-You need to be inside a function directory to deploy it.
+			// Layer 2: Wrap technical error with CLI-specific guidance
+			var errNotInit *fn.ErrNotInitialized
+			notInitErr := fn.NewErrNotInitialized(f.Root)
+			if errors.As(notInitErr, &errNotInit) {
+				return fmt.Errorf(`%w
+
+To deploy a function, you need to point to a function directory or use a --path flag approach.
```
gauron99: I think this was intended to be 
```suggestion
To deploy a function, you need to be in a function directory or use a --path flag approach.
```

gauron99 (commented):
just a few suggestions

cmd/build.go
```diff
@@ -159,6 +159,52 @@ func runBuild(cmd *cobra.Command, _ []string, newClient ClientFactory) (err erro
...
+Common registries:
+  ghcr.io/myuser       GitHub Container Registry
+  docker.io/myuser     Docker Hub
+  quay.io/myuser       Quay.io
+
+Or specify full image name:
+  func build --image ghcr.io/myuser/myfunction:latest
+
+For more options, run 'func build --help'`, err)
+		}
```
gauron99, line 173: I think it might be already better to introduce a separate functions/ file for these. Now that we have 2 if statements in a row its already taking up 40+ lines

cmd/build.go
```diff
@@ -297,10 +343,6 @@ func (c buildConfig) Configure(f fn.Function) fn.Function {
 // Skipped if not in an interactive terminal (non-TTY), or if --confirm false (agree to
 // all prompts) was set (default).
 func (c buildConfig) Prompt() (buildConfig, error) {
-	if !interactiveTerminal() {
-		return c, nil
-	}
```
gauron99, line 316: why are we removing the check for non interactive terminal? The function itself is called `Prompt()`. If theres functionality that should be performed regardless of interactivity of the terminal, I think it should be done elsewhere. Either it must be moved prior the function call or the function must change.

I see that a check is done later but the function has now become more of a validity function instead which im not sure is the right approach. WDYT @lkingland ?

lkingland: This is correct.

The "Prompt" methods (where used) are always called, for both interactive and noninteractive terminals, in order to keep the calling method clean.  The pattern here is to return having done nothing if the terminal is noninteractive.

Please keep this check in place, or it will break when `func` is used in (for example) a script or CI.

cmd/build.go
```diff
@@ -312,23 +354,38 @@ func (c buildConfig) Prompt() (buildConfig, error) {
...
+	// Check if function exists first
+	if !f.Initialized() {
+		// Return a specific error for uninitialized function
+		return c, fn.NewErrNotInitialized(f.Root)
+	}
+
+	// If function IS initialized AND registry/image is missing
+	if f.Registry == "" && c.Registry == "" && c.Image == "" {
+		// In interactive terminal, prompt for registry (user-friendly workflow)
+		if interactiveTerminal() {
```
lkingland: I see there might be a misunderstaning of what th e`.Prompt` method means:  It's specifically for gathering additional information from a user... which can only happen if it's currently an interactive terminal.

So placing that check at the beginning of the method is fine.

RayyanSeliya: @lkingland Thanks for clarifying! I've kept the `interactiveTerminal()` check at the start of `.Prompt()` as you explained.

gauron99:
There is a very strange IPV6 error which happens across PRs, completely unrelated to this. Overriding the unit test to not block PR
/override unit-tests_func_main
/lgtm
/approve

PR #3066, opened 2025-10-04
 Fix func build/deploy validation: Add early validation for conflicting --image and --registry flags

### LINKED ISSUE #3065: func build/deploy proceeds with build despite conflicting --image and --registry flags
## **Current behavior:**
When running `func deploy` or `func build` with both `--image` and `--registry` flags that specify different registries, the command proceeds to build instead of validating the conflict:

```txt
rayyan@rayyan-seliya:/mnt/c/Users/RAYYAN/Desktop/func/myfunction$ func deploy --image ghcr.io/user/func:latest --registry different-registry.com/user
Building function image
^CError: failed to fetch builder image 'ghcr.io/knative/builder-jammy-tiny:latest': context canceled
```

This forces users to:
1. Wait through the entire build process before realizing the conflict
2. Manually cancel (Ctrl+C) to stop the misleading build
3. Guess what went wrong from a generic "failed to fetch builder image" error
4. Not understand that the issue is with conflicting registry specifications

## **Proposed improvement:**
Provide immediate validation and clear guidance when image and registry flags conflict:

```txt
Error: conflicting image and registry options

You cannot use both --image and --registry together when they specify different registries.

These are two different ways to specify where to push your function image.

Choose ONE of these approaches:

1. Use --image to specify the complete image name:
   func deploy --image <full-image-name>

2. Use --registry and let the system create the image name automatically:
   func deploy --registry <registry>

```

For `func build`:
```txt
Error: conflicting image and registry options

You cannot use both --image and --registry together when they specify different registries.

These are two different ways to specify where to push your function image.

Choose ONE of these approaches:

1. Use --image to specify the complete image name:
   func build --image <full-image-name>

2. Use --registry and let the system create the image name automatically:
   func build --registry <registry>

For more options, run 'func build --help'
```

**Solution:**
Implement a 2-layer error system:
- **Layer 1** (Technical): Return `ErrConflictingImageAndRegistry` from validation in `pkg/functions/errors.go`
- **Layer 2** (CLI): Catch the error in `cmd/build.go` and `cmd/deploy.go` and provide user-friendly guidance

## **Expected Workflow:**
The validation should occur **before** any build attempts:
1. Parse command flags
2. Validate flag combinations for conflicts
3. Return clear error if conflicts detected
4. Only proceed to build if validation passes

## **Compatibility Notes:**
The validation should **allow** compatible combinations:
- ✅ `--registry example.com/alice --image example.com/alice/subnamespace/myfunc` (subnamespace)
- ✅ `--registry example.com/alice --image example.com/alice/func:latest` (exact match)
- ❌ `--registry example.com/alice --image different.com/user/func` (different registry)

## **User Impact:**
- **Current**: Confusing build failure → time wasted waiting → need to investigate → unclear error message
- **Improved**: Immediate validation → clear error explanation → actionable guidance → better developer experience

RayyanSeliya:
## Changes

- :bug: Fix confusing UX when using conflicting --image and --registry flags
- :broom: Implement 2-layer error system for image/registry validation
- :gift: Add clear guidance for resolving flag conflicts

/kind enhancement

Fixes #3065 

## **Solution:**
Implement a 2-layer error system:
- **Layer 1** (Technical): Return `ErrConflictingImageAndRegistry` from validation in `pkg/functions/errors.go`
- **Layer 2** (CLI): Catch the error in `cmd/build.go` and `cmd/deploy.go` and wrap with user-friendly guidance

## **Implementation Details:**
- Add validation logic in `buildConfig.Validate()` and `deployConfig.Validate()` methods
- Use `strings.HasPrefix()` to check if image starts with registry (allows subnamespaces)
- Maintain backward compatibility with existing test cases
- Ensure error messages are consistent between build and deploy commands

**Release Note**

```release-note
func build and func deploy commands now validate conflicting --image and --registry flags before starting the build process, providing clear error messages and actionable guidance instead of misleading build failures.
```

### DIFF
--- a/cmd/deploy_test.go
+++ b/cmd/deploy_test.go
@@ -1368,16 +1368,20 @@ func testRegistry(cmdFn commandConstructor, t *testing.T) {
 			expectedImage:    "registry.example.com/charlie/f:latest",
 		},
 		{
-			// Image flag sets the image directly when no registry is present
+			// Image flag overrides registry when both exist:
+			// Function has a pre-existing registry from previous deploy,
+			// but --image flag is provided to use a different image.
+			// Expected: --image takes precedence, registry field remains unchanged
 			name: "image flag overrides",
 			f: fn.Function{
+				Registry: "registry.example.com/alice", // pre-existing registry
 				Build: fn.BuildSpec{
 					Image: "registry.example.com/bob/f:latest",
 				},
 			},
 			args:             []string{"--image=registry.example.com/charlie/f:latest"},
-			expectedRegistry: "",                                      // no registry set
-			expectedImage:    "registry.example.com/charlie/f:latest", // updated
+			expectedRegistry: "registry.example.com/alice",            // registry unchanged
+			expectedImage:    "registry.example.com/charlie/f:latest", // image updated
 		},
 	}
 	for _, test := range tests {


RayyanSeliya:
I think i am missing some edge cases @gauron99  @lkingland  please point out if any the test are still failing :(

gauron99:
im gonna take a look tomorrow

RayyanSeliya:
Sure no worries !! Good night !

cmd/build.go
```diff
@@ -362,6 +380,20 @@ func (c buildConfig) Validate() (err error) {
 		return
 	}
 
+	// Check for conflicting --image and --registry
+	if c.Image != "" && c.Registry != "" {
+		// Skip validation if image has digest (@sha256:...) - digests are immutable overrides
+		if !strings.Contains(c.Image, "@sha256:") {
+			// Only validate if image has explicit registry (contains "/" or ":")
+			// Simple names like "myimage" are compatible - registry will be prepended
```
gauron99: are they really prepended?

gauron99:
```
❯ func build --image mine --registry quay.io/dfridric
Building function image
Still building
🙌 Function built: mine
```

gauron99 (commented):
perhaps we could simply check that -registry and --image are not given together at all? Image takes precedence now which is correct. But we want to avoid having the need to remove `FUNC_REGISTRY` or your registry in config file when trying to once specify `--image` for a different registry. 
In other words `--image` should just take precedence and work even when `FUNC_REGISTRY` is specified BUT `--image` and `--registry` could just return an error saying "just choose one or the other" for simplicity?

Otherwise you would need to check both registries, determine what parts image contains and prepend (which I dont think works currently? check my other comment) and that would require a bit more work and consideration.

lkingland:
I agree with @gauron99 on this one.  Please simply check if the flags were both provided, and then raise an error that only one or the other should be provided, not both.  👍🏻

cmd/build.go
RayyanSeliya: @gauron99 u are right! The old validation was confusing , your test showed that --image override/takes precedence --registry, which wasn’t clear to users. We’ve fixed this by rejecting both flags when used together and showing a clear error message.

cmd/deploy_test.go
```diff
@@ -1372,18 +1368,15 @@ func testRegistry(cmdFn commandConstructor, t *testing.T) {
 			expectedImage:    "registry.example.com/charlie/f:latest",
 		},
 		{
-			// Image flag takes highest precidence, but is an override such that the
-			// resultant image member is updated but the registry member is not
-			// updated (subsequent builds will not be affected).
+			// Image flag sets the image directly when no registry is present
 			name: "image flag overrides",
```
gauron99, line 1375: I think in this test we want to test exactly the case where image DOES override registry since registry is not being provided via flag but rather its already been given previously.

This test seems to to to cover the case when you 
1. `func deploy --registry ...`
... now `f.registry` exists in `func.yaml` and is not empty 
2. `func deploy --image <different image>`

we want this case to work -> override the image from `registry` thats been used from `1.` with `image` from `2.` 
the `func.yaml` registry is NOT overriden and kept as is from `1.` but the new image is used in deployment because `--image` was provided with different value.

gauron99 (commented):
nice, this already looks cleaner and simpler. I added a comment about one test

cmd/deploy_test.go
RayyanSeliya: thx for the clarification @gauron99  i have updated to included this test too ...

lkingland (approved):
/lgtm
/hold to make sure @gauron99 your OK too

PR #3075, opened 2025-10-10
Add raw deployment deployer

creydr:
# Changes

- :gift: Adds the possibility to deploy the function as a raw kubernetes deployment instead of a knative service
  - Updated the existing integration tests for the deployer, describer, remover and lister and made them generic, so I could reuse them for the raw deployments too
  - Updated clients WithLister(), WithDescriber(), WithRemover(), to accept a variadic argument list. Then letting the coresponding lister/describer/remover handle the action.
  - Added the deployer as a new column in the list command

```
$ func deploy --registry localhost:50000 --deployer raw
Building function image
Still building
🙌 Function built: localhost:50000/go-http-func:latest
Pushing function image to the registry "localhost:50000" using the "" user credentials
🎯 Creating Triggers on the cluster
✅ Function deployed in namespace "default" and exposed at URL: 
   http://go-http-func.default.svc.cluster.local

$ func list                                                       
NAME          NAMESPACE  RUNTIME  DEPLOYER     URL                                             READY
go-http-func  default             raw          http://go-http-func.default.svc                 True
go-http-ksvc  default             knative      http://go-http-ksvc.default.127.0.0.1.sslip.io  True

$ func describe go-http-func                                      
Function name:
  go-http-func
Function is built in image:
  
Function is deployed in namespace:
  default
Routes:
  http://go-http-func.default.svc
Deployer:
  raw
Labels:
  boson.dev/function: true
  function.knative.dev/name: go-http-func
  function.knative.dev/runtime: go

$ k get ksvc
NAME            URL                                               LATESTCREATED         LATESTREADY           READY   REASON
go-http-ksvc    http://go-http-ksvc.default.127.0.0.1.sslip.io    go-http-ksvc-00001    go-http-ksvc-00001    True    

$ k get ksvc go-http-func # <-- no Knative Service deployment for our function !
Error from server (NotFound): services.serving.knative.dev "go-http-func" not found

$ k get deploy
NAME                             READY   UP-TO-DATE   AVAILABLE   AGE
go-http-func                     1/1     1            1           47s

$  k get svc   
NAME                                    TYPE           CLUSTER-IP      EXTERNAL-IP                                         PORT(S)                                              AGE
go-http-func                            ClusterIP      10.96.173.129   <none>                                              80/TCP                                               62s
...
```

/kind enhancement

Fixes: #2804

**Release Note**

```release-note
Add the possibility to deploy a function as raw kubernetes deployment via the --deployer=raw argument
```

### DIFF
--- a/pkg/deployer/testing/integration_test_helper.go
+++ b/pkg/deployer/testing/integration_test_helper.go
@@ -1,5 +1,6 @@
 package testing
 
+//nolint:staticcheck  // ST1001: should not use dot imports
 import (
 	"context"
 	"encoding/json"
--- a/pkg/describer/testing/integration_test_helper.go
+++ b/pkg/describer/testing/integration_test_helper.go
@@ -1,5 +1,6 @@
 package testing
 
+//nolint:staticcheck  // ST1001: should not use dot imports
 import (
 	"context"
 	"testing"
--- a/pkg/lister/testing/integration_test_helper.go
+++ b/pkg/lister/testing/integration_test_helper.go
@@ -1,5 +1,6 @@
 package testing
 
+//nolint:staticcheck  // ST1001: should not use dot imports
 import (
 	"context"
 	"testing"
--- a/pkg/remover/testing/integration_test_helper.go
+++ b/pkg/remover/testing/integration_test_helper.go
@@ -1,5 +1,6 @@
 package testing
 
+//nolint:staticcheck  // ST1001: should not use dot imports
 import (
 	"context"
 	"testing"


creydr:
Rebased after #3100 was in

creydr:
/cc @lkingland @gauron99 @matzew 
Ready for review

creydr:
Rebased to fix merge conflicts

cmd/client.go
```diff
@@ -9,13 +9,23 @@ import (
...
+	"knative.dev/func/pkg/describer"
+	k8sdescriber "knative.dev/func/pkg/describer/k8s"
+	knativedescriber "knative.dev/func/pkg/describer/knative"
 	"knative.dev/func/pkg/docker"
 	fn "knative.dev/func/pkg/functions"
 	fnhttp "knative.dev/func/pkg/http"
 	"knative.dev/func/pkg/k8s"
-	"knative.dev/func/pkg/knative"
+	"knative.dev/func/pkg/lister"
+	k8slister "knative.dev/func/pkg/lister/k8s"
```
lkingland: Let's restructure these packages to be based on dependency name first, type second (I'll explain in further detail via IM)

cmd/deploy.go
```diff
@@ -131,7 +131,7 @@ EXAMPLES
 		PreRunE: bindEnv("build", "build-timestamp", "builder", "builder-image",
 			"base-image", "confirm", "domain", "env", "git-branch", "git-dir",
 			"git-url", "image", "namespace", "path", "platform", "push", "pvc-size",
-			"service-account", "registry", "registry-insecure", "remote",
+			"service-account", "deploy-type", "registry", "registry-insecure", "remote",
```
lkingland: To fit with the nomenclature of the other flag options, this would just be `--deployer`

cmd/client.go
```diff
@@ -58,16 +58,16 @@ func NewClient(cfg ClientConfig, options ...fn.Option) (*fn.Client, func()) {
...
 			fn.WithVerbose(cfg.Verbose),
 			fn.WithTransport(t),
 			fn.WithRepositoriesPath(config.RepositoriesPath()),
 			fn.WithBuilder(buildpacks.NewBuilder(buildpacks.WithVerbose(cfg.Verbose))),
-			fn.WithRemover(knative.NewRemover(cfg.Verbose)),
-			fn.WithDescriber(knative.NewDescriber(cfg.Verbose)),
-			fn.WithLister(knative.NewLister(cfg.Verbose)),
+			fn.WithRemovers(knative.NewRemover(cfg.Verbose), k8s.NewRemover(cfg.Verbose)),
+			fn.WithDescribers(knative.NewDescriber(cfg.Verbose), k8s.NewDescriber(cfg.Verbose)),
+			fn.WithListers(knative.NewLister(cfg.Verbose), k8s.NewLister(cfg.Verbose)),
```
lkingland, line 70: Great!  A Client which now internally is instantiated with support for knative and raw k8s Functions

cmd/client.go
```diff
@@ -135,6 +135,15 @@ func newKnativeDeployer(verbose bool) fn.Deployer {
 	return knative.NewDeployer(options...)
 }
 
+func newK8sDeployer(verbose bool) fn.Deployer {
+	options := []k8s.DeployerOpt{
+		k8s.WithDeployerVerbose(verbose),
+		k8s.WithDeployerDecorator(deployDecorator{}),
```
lkingland, line 141: Hopefully we can make this a bit cleaner than a decorator in the future, but for now it's what we got!  Glad the pattern continues to work.

cmd/deploy.go
```diff
@@ -192,6 +192,8 @@ EXAMPLES
 		"When triggering a remote deployment, set a custom volume size to allocate for the build operation ($FUNC_PVC_SIZE)")
 	cmd.Flags().String("service-account", f.Deploy.ServiceAccountName,
 		"Service account to be used in the deployed function ($FUNC_SERVICE_ACCOUNT)")
+	cmd.Flags().String("deployer", f.Deploy.Deployer,
```
lkingland, line 196: Good; this should keep the pattern of the flag's defaults being dynamically set to the current state of the active Function

cmd/deploy.go
```diff
@@ -192,6 +192,8 @@ EXAMPLES
...
+		fmt.Sprintf("Type of deployment to use: '%s' for Knative Service (default) or '%s' for Kubernetes Deployment ($FUNC_DEPLOY_TYPE)", knative.KnativeDeployerName, k8s.KubernetesDeployerName))
```
lkingland, line 197: Just to reiterate our conversation on the recent call:  we'll eventualy be expanding this to automatically select "knative" vs "raw" in the future using a Cluster Capabilities reporter.  Keeping it "knative" leaves us backwards compatible for now.

cmd/deploy.go
```diff
@@ -233,6 +235,10 @@ EXAMPLES
 		fmt.Println("internal: error while calling RegisterFlagCompletionFunc: ", err)
 	}
 
+	if err := cmd.RegisterFlagCompletionFunc("deployer", CompleteDeployerList); err != nil {
```
lkingland, line 239: Thanks for getting a completion handler in there as well

pkg/deployer/testing/integration_test_helper.go
```diff
@@ -0,0 +1,1155 @@
+//go:build integration
+// +build integration
+
```
lkingland: Not really necessary to have tags in this file, since the test entrypoints are elsewhere, but it doesn't hurt anything other; just confuses some IDEs.

pkg/describer/testing/integration_test_helper.go
```diff
@@ -1,41 +1,44 @@
...
-
 	fn "knative.dev/func/pkg/functions"
-	"knative.dev/func/pkg/knative"
 	"knative.dev/func/pkg/oci"
+	fntest "knative.dev/func/pkg/testing"
+	fnk8stest "knative.dev/func/pkg/testing/k8s"
 )
 
-func TestInt_Labels(t *testing.T) {
+func DescribeIntegrationTest(t *testing.T, describer fn.Describer, deployer fn.Deployer, remover fn.Remover, deployerName string) {
```
lkingland: These need to remain prefixed with `TestInt_` for our selection magic to work properly.

pkg/describer/testing/integration_test_helper.go
```diff
@@ -1,41 +1,44 @@
...
 	"context"
 	"testing"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/rand"
-
 	fn "knative.dev/func/pkg/functions"
-	"knative.dev/func/pkg/knative"
 	"knative.dev/func/pkg/oci"
+	fntest "knative.dev/func/pkg/testing"
```
lkingland: This is "intended" to be imported as "." such that no prefix is necessary.  Not a necessary change, but for consistency that would be best.

pkg/describer/testing/integration_test_helper.go
```diff
@@ -1,41 +1,44 @@
...
+	fnk8stest "knative.dev/func/pkg/testing/k8s"
```
lkingland: If you import the testing package into the global namespace, then this can be simplified to not require the explicit name

pkg/describer/testing/integration_test_helper.go
```diff
@@ -1,41 +1,44 @@
...
 	f, err := client.Init(fn.Function{
 		Root:      root,
 		Name:      name,
 		Runtime:   "go",
 		Namespace: ns,
-		Registry:  registry(),
+		Registry:  fntest.Registry(),
+		Deploy: fn.DeploySpec{
+			Deployer: deployerName,
+		},
```
lkingland: We may need to set up a defaults system where, on first deploy, the deployer used is `deployers[0]`.  I'd not want every Function created programmatically to require setting the deployer.

Even better, we could add an optional argument to `client.Deploy` where the deployer name can be specified.  WDYT?

pkg/functions/client.go
```diff
@@ -139,13 +140,14 @@ type Runner interface {
...
 	// Remove the function from remote.
-	Remove(ctx context.Context, name string, namespace string) error
+	Remove(ctx context.Context, name string, namespace string) (bool, error)
 }
 
 // Lister of deployed functions.
 type Lister interface {
 	// List the functions currently deployed.
-	List(ctx context.Context, namespace string) ([]ListItem, error)
+	// the 2nd argument (bool) indicates it the lister had to handle any function at all
```
lkingland: I am forgetting why the `bool` was necessary?  What is the circumstance in which `ok==true` but `len(items)==0`?

pkg/functions/client.go
```diff
@@ -154,12 +156,13 @@ type ListItem struct {
...
 	URL       string `json:"url" yaml:"url"`
 	Ready     string `json:"ready" yaml:"ready"`
+	Deployer  string `json:"deployer" yaml:"deployer"`
 }
 
 // Describer of function instances
 type Describer interface {
 	// Describe the named function in the remote environment.
-	Describe(ctx context.Context, name, namespace string) (Instance, error)
+	Describe(ctx context.Context, name, namespace string) (*Instance, error)
```
lkingland: We should stick to pass-by-value unless there's a good reason for a pointer.

pkg/functions/client.go
```diff
@@ -987,7 +991,24 @@ func (c *Client) Describe(ctx context.Context, name, namespace string, f Functio
 	// If it has a populated deployed namespace, we can presume it's deployed
 	// and attempt to describe.
 
-	return c.describer.Describe(ctx, f.Name, f.Deploy.Namespace)
+	return c.describeByMatchingDescriber(ctx, f.Name, f.Deploy.Namespace)
+}
+
+// describeByMatchingDescriber iterates over the registered describers and executes them on the given object.
+func (c *Client) describeByMatchingDescriber(ctx context.Context, name, namespace string) (d Instance, err error) {
```
lkingland, line 1017: Good; this seems like just the right place for this complexity.  It might be worth noting in a comment that there should never be a describer which describes a Function of a different deployer, so there's no need here to handle that case (exiting-fast lik you did is correct).  It rests on that assumption, however, so future readers may think it's a missing case.

pkg/functions/client.go
```diff
@@ -996,8 +1017,17 @@ func (c *Client) Describe(ctx context.Context, name, namespace string, f Functio
 // using the current kubernetes context namespace, falling back to the static
 // default "namespace".
 func (c *Client) List(ctx context.Context, namespace string) ([]ListItem, error) {
-	// delegate to concrete implementation of lister entirely.
-	return c.lister.List(ctx, namespace)
+	list := []ListItem{}
```
lkingland, line 1043: Again, a little comment here mentioning how a lister should never list a Function from a mismatched deployer might be nice.  This is an expectation we should place on lister (and describer above) implementations via tests.  That removes this as a potential case to be consdered here; but may look like a missing scenario at first glance.

pkg/functions/function.go
```diff
@@ -208,6 +208,10 @@ type DeploySpec struct {
 	// More info: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
 	ServiceAccountName string `yaml:"serviceAccountName,omitempty"`
 
+	// Deployer specifies the type of deployment to use: "knative" or "raw"
+	// Defaults to "knative" for backwards compatibility
+	Deployer string `yaml:"deployer,omitempty" jsonschema:"enum=knative,enum=raw"`
```
lkingland, line 213: This is OK for now, but it's worth noting that the enum creates an implicit dependency between the core and the deployer implementations (a dependency we try hard to avoid using the interfaces etc).  It's a soft dependency, but one nonetheless.  I'm not sure what a good solution would be, however, because providing accepted enum values is valuable for clients using the schema.

pkg/functions/function_unit_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"gopkg.in/yaml.v2"
+	. "knative.dev/func/pkg/functions"
```
lkingland: We've been naming this import `fn` pretty consistently

pkg/functions/instances_test.go
```diff
@@ -1,11 +1,12 @@
-package functions
+package functions_test
 
 import (
 	"context"
 	"errors"
 	"strings"
 	"testing"
 
+	. "knative.dev/func/pkg/functions"
```
lkingland: Also here, for consistency `fn` might be better.

pkg/functions/job_test.go
```diff
@@ -1,10 +1,11 @@
-package functions
+package functions_test
 
 import (
 	"context"
 	"errors"
 	"testing"
 
+	. "knative.dev/func/pkg/functions"
```
lkingland: Same here

pkg/k8s/deployer_int_test.go
```diff
@@ -0,0 +1,62 @@
...
+package k8s_test
+
+import (
+	"testing"
+
+	deployertesting "knative.dev/func/pkg/deployer/testing"
+	"knative.dev/func/pkg/k8s"
+)
+
+func TestIntegration(t *testing.T) {
```
lkingland: This one will need `TestInt_` prefix; maybe `TestInt_Full`?

pkg/k8s/deployer_test.go
```diff
@@ -1,15 +1,14 @@
-package knative
+package k8s
```
lkingland, line 1: Nice to see this split working out 👍🏻

pkg/k8s/describer.go
```diff
@@ -0,0 +1,69 @@
...
+	return &Describer{
+		verbose: verbose,
+	}
+}
+
+// Describe a function by name. Note that the consuming API uses domain style
+// notation, whereas Kubernetes restricts to label-syntax, which is thus
+// escaped. Therefor as a knative (kube) implementation detail proper full
+// names have to be escaped on the way in and unescaped on the way out. ex:
+// www.example-site.com -> www-example--site-com
```
lkingland: This comment is, unfortunately, no longer accurate.

pkg/k8s/describer.go
```diff
@@ -0,0 +1,69 @@
...
+	service, err := serviceClient.Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("unable to get service for function: %v", err)
+	}
+
+	if !UsesRawDeployer(service.Annotations) {
+		return nil, nil
+	}
+
+	description := &fn.Instance{
```
lkingland: Noted above, but pretty sure there's no reason for a pointer here.

pkg/k8s/describer_int_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"testing"
+
+	describertesting "knative.dev/func/pkg/describer/testing"
+	"knative.dev/func/pkg/k8s"
+)
+
+func TestInt_Describe(t *testing.T) {
+	describertesting.DescribeIntegrationTest(t,
```
lkingland: Shared basic functionality tests like this really help out 👍🏻

pkg/k8s/lister.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil, false, fmt.Errorf("unable to list services: %v", err)
+	}
+
+	listItems := make([]fn.ListItem, 0, len(services.Items))
+	ok := false
+	for _, service := range services.Items {
+		if !UsesRawDeployer(service.Annotations) {
+			continue
+		}
+		ok = true
```
lkingland: So far I am still not quite remembering why we needed this boolean 🤔

pkg/knative/describer.go
```diff
@@ -26,54 +25,60 @@ func NewDescriber(verbose bool) *Describer {
 // escaped. Therefore as a knative (kube) implementation detal proper full
 // names have to be escaped on the way in and unescaped on the way out. ex:
 // www.example-site.com -> www-example--site-com
-func (d *Describer) Describe(ctx context.Context, name, namespace string) (description fn.Instance, err error) {
+func (d *Describer) Describe(ctx context.Context, name, namespace string) (*fn.Instance, error) {
```
lkingland: My initial reaction here is that we should not have the possibility of a nil return, but instead have a typed error such as ErrFunctionNotFound the caller can selectively decide to watch for, returning an empty struct if the given name does not exist in the given namespace.

pkg/knative/logs.go
```diff
@@ -24,126 +17,6 @@ import (
...
-	defer b.mu.Unlock()
-	return b.b.Read(p)
-}
-
-func (b *SynchronizedBuffer) Reset() {
-	b.mu.Lock()
-	defer b.mu.Unlock()
-	b.b.Reset()
+	selector := fmt.Sprintf("serving.knative.dev/service=%s", kServiceName)
+	return k8s.GetPodLogsBySelector(ctx, namespace, selector, "user-container", image, since, out)
```
lkingland, line 21: Nice that this is also now shared; we can use it as a starting point for proper logging facilities through the client (a long-standing feature request!)

pkg/knative/remover.go
```diff
@@ -22,24 +22,32 @@ type Remover struct {
 	verbose bool
 }
 
-func (remover *Remover) Remove(ctx context.Context, name, ns string) (err error) {
+func (remover *Remover) Remove(ctx context.Context, name, ns string) (bool, error) {
```
lkingland: As for this one, I am glad to have a boolean indicating something was removed

pkg/lister/testing/integration_test_helper.go
```diff
@@ -0,0 +1,96 @@
+//go:build integration
```
lkingland: Since this file has no actual test entrypoints, I don't think the tag is necessary.  Doesn't really hurt anything, however.

pkg/lister/testing/integration_test_helper.go
```diff
@@ -0,0 +1,96 @@
...
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/rand"
+	fn "knative.dev/func/pkg/functions"
+	"knative.dev/func/pkg/oci"
+	fntest "knative.dev/func/pkg/testing"
+	fnk8stest "knative.dev/func/pkg/testing/k8s"
+)
+
+func IntegrationTest(t *testing.T, lister fn.Lister, deployer fn.Deployer, describer fn.Describer, remover fn.Remover, deployerName string) {
```
lkingland: I think we should also add a call to `t.Helper()` as the first line of all these abstracted test functions.

pkg/remover/testing/integration_test_helper.go
```diff
@@ -0,0 +1,111 @@
+//go:build integration
```
lkingland: Same comment as the others:  probably not necessary to tag.

pkg/remover/testing/integration_test_helper.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"k8s.io/apimachinery/pkg/util/rand"
+	fn "knative.dev/func/pkg/functions"
+	"knative.dev/func/pkg/oci"
+	fntest "knative.dev/func/pkg/testing"
+	fnk8stest "knative.dev/func/pkg/testing/k8s"
+)
+
+func IntegrationTest(t *testing.T, remover fn.Remover, deployer fn.Deployer, describer fn.Describer, lister fn.Lister, deployerName string) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*10)
```
lkingland, line 19: Same comment as above:  a call to `t.Helper()` would be nice.

pkg/testing/k8s/testing.go
```diff
@@ -0,0 +1,49 @@
+// package testing includes Kubernetes-specific testing helpers.
+package k8s
```
lkingland: Is this intended to be in a `testing` package?

lkingland (commented):
This feature add and generalization of the system is really looking good!  🎉 

See comments inline.

pkg/describer/testing/integration_test_helper.go
creydr: Those are the "helper" methods, which then get called by the actual "TestInt_xyz" test. But I change it, so they align and make it hopefully clearer

pkg/deployer/testing/integration_test_helper.go
creydr: Removed it

pkg/describer/testing/integration_test_helper.go
creydr: Updated

pkg/describer/testing/integration_test_helper.go
creydr: Updated

pkg/functions/function_unit_test.go
creydr: Updated

pkg/functions/instances_test.go
creydr: Updated

pkg/functions/job_test.go
creydr: Updated

pkg/k8s/deployer_int_test.go
creydr: Good catch. Updated

pkg/k8s/describer.go
creydr: True. Copy & Paste :grin:

pkg/lister/testing/integration_test_helper.go
creydr: Removed

pkg/lister/testing/integration_test_helper.go
creydr: Good point. Updated it

pkg/remover/testing/integration_test_helper.go
creydr: Updated

pkg/remover/testing/integration_test_helper.go
creydr: Updated

pkg/k8s/deployer_int_test.go
```diff
@@ -0,0 +1,62 @@
+//go:build integration
+// +build integration
```
gauron99, line 2: note: so far from what ive reviewed only this int test has this `+build...` tag for backward compatibility, others dont. I guess not many people will use `<go1.17` in the current day but there is some discrepancy

pkg/knative/deployer_int_test.go
```diff
@@ -1,858 +1,60 @@
 //go:build integration
+// +build integration
```
gauron99, line 2: there is one again. Is there a reason why only some contain this?

pkg/knative/remover_int_test.go
```diff
@@ -3,106 +3,17 @@
 package knative_test
```
gauron99: this file actually doesnt have go integration build tag?

pkg/lister/testing/integration_test_helper.go
gauron99: why is this not part of the integration tests?

pkg/remover/testing/integration_test_helper.go
```diff
@@ -0,0 +1,111 @@
...
+		if item.Name == f.Name {
+			found = true
+			break
+		}
+	}
+	if found {
+		t.Errorf("function %s was not removed", f.Name)
+	}
+
+	// Remove
```
gauron99: It was probably intended here to remove the function forcefully?

gauron99 (commented):
Thanks for a huge improvement! Others already reviewed so I just found some smaller nits/comments. Also I think you will have to remove all the dot `.` imports for reviewdog to stop screaming.

pkg/remover/testing/integration_test_helper.go
creydr: As we setup the test namespace via k8s.Namespace(), it adds a cleanup function to the test suite, so the namespace and all its components should be cleaned up:

https://github.com/knative/func/blob/a3221ad5895cd51a4f7dd9f5885f9794d1dd437d/pkg/testing/k8s/testing.go#L40-L45

So I _guess_ it's simply a leftover. But I'll remove it

pkg/knative/deployer_int_test.go
creydr: I guess this was simply some copy&paste mistake. Will update it

pkg/k8s/deployer_int_test.go
creydr: Oh good point. I guess this was simply some copy&paste mistake. Will update it

pkg/functions/client.go
creydr: With the update of the Describe and Remove signatures (and methods to return a ErrNotHandled), I updated this as well. So not needed anymore

pkg/functions/client.go
creydr: The reason was, that I wanted some indication if the handler was responsible for a function. But with the refactoring to return an ErrNotHandled, I updated this as well

pkg/functions/client.go
creydr: I added some more comments to this method and also the the interface/method declaration

pkg/functions/client.go
creydr: Added a comment to the method and the lister interface/method declaration

pkg/k8s/describer.go
creydr: Updated with the latest rework with the `ErrNotHandled` error

pkg/k8s/lister.go
creydr: Removed with the latest rework with the `ErrNotHandled` error

pkg/knative/describer.go
creydr: Updated with the latest rework with the `ErrNotHandled` error

pkg/testing/k8s/testing.go
creydr: Good catch. Wrong comment. Updated it

pkg/describer/testing/integration_test_helper.go
creydr: I only needed this in the deployer to setup the annotations correctly. But in the deployer I have this information, so no need for it anymore in the test (not sure, why I added it...).
But: I am not sure if we need this at any other place though :thinking:

Issue #3079, opened 2025-10-13
func build/run doesn't pick up code changes


dprotaso:
func verison
v0.46.3

```
$ func create -l python sentiment-analysis-app
$ cd sentiment-analysis-app
$ func build -v
...
$ func run -v
...
# stop the func run
$ # change some code
$ func run -v
...
$ func invoke ... 
```

When changing the code I updated a bunch of log lines but the changes weren't showing up in the second invocation

dprotaso:
I'm on Mac 15.6  - Docker for Desktop Mac v4.48

gauron99:
`host` and `s2i` builders work as intended, `pack` however fails. Confirmed on 1.19.3 and on main -- on my machine

PR #3080, opened 2025-10-13
fix: python scaffolding

matejvasek:
# Changes

* fix: python scaffolding -- make "this function" dependency editable, otherwise only first build has effect any subsequent build "does nothing".

fixes: #3079

/kind bug

```release-note
fix: Python pack build/run doesn't pick up code changes (#3079)
```


matejvasek:
@dprotaso plz override `unit-tests_func_main`

Issue #3078, opened 2025-10-13
func run doesn't work without running `func build`


dprotaso:
func verions v0.46.3

```
$ func create -l python sentiment-analysis-app
$ cd sentiment-analysis-app
$ func run -v
mkdir -p /var/folders/0x/lsrcyczd1yn3gggn6w9yc68m0000gn/T/tmp.15goFVNxjx/sentiment-analysis-app/.func/runs/64467
cd /var/folders/0x/lsrcyczd1yn3gggn6w9yc68m0000gn/T/tmp.15goFVNxjx/sentiment-analysis-app/.func/runs/64467
python -m venv .venv
Error: exec: "python": executable file not found in $PATH
```

If I run `func build` then `func run` it uses the docker image.

I think i might have expected the func cli to use `python3` as a fallback?

gauron99:
This is trying to run `host` builder as default since that is the default in 1.19.3.
Host builder will use your host resources to build that image. therefore this is intended from func pov

dprotaso:
@gauron99 I guess could we fallback to `python3` executable?

PR #3082, opened 2025-10-13
fix: use python3 when python not present

matejvasek:
# Changes

* fix: use `python3` when `python` not present, sometimes there is missing symlink `python->python3` on some systems.

fixes: #3078

/kind bug

```release-note
fix: fallback to python3 if python not present
```


matejvasek:
@gauron99 override the tests.

PR #3092, opened 2025-10-15
feat(describe): prompt user when no function name provided

intojhanurag:
This PR enhances the func describe command by adding an interactive prompt that asks the user to enter a function name when it is not provided as a command argument.

## Before
Screenshot 2025-10-17 013005

## After
Screenshot 2025-10-17 012926

### DIFF
--- a/cmd/describe.go
+++ b/cmd/describe.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"encoding/json"
 	"encoding/xml"
 	"errors"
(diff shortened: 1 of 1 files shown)


intojhanurag:
Hey @gauron99, I’m just curious, why is it asking for the function name even when no prompt was declared?
The TODO mentioned adding a prompt section to describe the function name.

gauron99:
not sure if I entirely understand what youre saying, but when you run `func describe` by itself in an empty dir (no function) it doesnt ask you anything because the prompt is not implemented; it just fails. Same as you found the `//TODO` in the test, there is one (equivalent) todo in the `runDescribe` function itself so once that is in place, the TODO indicates that this is where you would test it as well.

So this test addition you implemented is only 50% of the intended change, you still need the actual implementation itself :smile:

intojhanurag:
Thanks @gauron99 , I was a bit confused earlier.

PR #3093, opened 2025-10-15
fix: host builder push missing docker error

lkingland:
Fix bug where missing docker causes an error in the oci pusher

### DIFF
--- a/pkg/creds/credentials.go
+++ b/pkg/creds/credentials.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/creds/credentials.go
```diff
@@ -167,6 +167,20 @@ func NewCredentialsProvider(configPath string, opts ...Opt) oci.CredentialsProvi
...
 	var defaultCredentialLoaders = []CredentialsCallback{}
 
+	// Add localhost registry handler first - anonymous auth for localhost registries
+	defaultCredentialLoaders = append(defaultCredentialLoaders,
+		func(registry string) (oci.Credentials, error) {
+			// Check if this is a localhost registry that should use anonymous auth
+			if registry == "localhost" || registry == "127.0.0.1" || 
+			   registry == "localhost:5000" || registry == "localhost:50000" ||
+			   registry == "127.0.0.1:5000" || registry == "127.0.0.1:50000" ||
+			   strings.HasPrefix(registry, "localhost:") || strings.HasPrefix(registry, "127.0.0.1:") {
```
gauron99: should we use a regex here for any port for example?

lkingland: Ugh, good catch.  I was in too much of a hurry.

gauron99 (approved):
/lgtm
/approve

PR #3094, opened 2025-10-15
fix: remote builds defeault to insecure for local registries

lkingland:
Default to insecure connections to local registries for remote builds when registry is a known local name.


gauron99 (approved):
/lgtm
/approve

PR #3095, opened 2025-10-15
fix: pipeline names easily exceed character limit

lkingland:
Hash pipeline names because they quickly exceed character limit and fail.


go.mod
```diff
@@ -1,6 +1,6 @@
 module knative.dev/func
 
-go 1.24.0
+go 1.25.0
```
gauron99: this PR also includes go 1.25 update, I think this was meant to be separate? im assume this breaks some things

gauron99:
seems that some E2Es check for that pipeline name

gauron99 (commented):
/lgtm
/approve

PR #3096, opened 2025-10-15
make typescript template a better echo

lkingland:
typescript template now echoes a bit better.  E2E tests use a GET and expect an echo to include the URL requested.


gauron99 (approved):
/lgtm
/approve

PR #3097, opened 2025-10-15
fix: fallback to anonymous auth in oci pusher

lkingland:
When pushing, the fallback should be to at least attempt anonymous authentication for, for example test clusters with unsecured local registries.


gauron99 (approved):
/lgtm
/approve

PR #3099, opened 2025-10-15
update Rust templates to be a more accurate echo

lkingland:
Default templates should be as accurate of an echo as possible for ease of testing/validation.  This updates the Rust templates thusly


gauron99 (commented):
/lgtm
/approve

PR #3100, opened 2025-10-15
fix: knative deployer should choose last built image by default

lkingland:
The default deploy image should be the last one built if available.


gauron99 (approved):
/lgtm
/approve

gauron99:
this unit test fail seems to be caused by go bump to 1.25; Im fixing this elsewhere

PR #3101, opened 2025-10-15
fix: oci builder should remove itself from active PIDs on completion

lkingland:
Fix OCI builder should remove its own pid link on completion.


gauron99 (approved):
/lgtm
/approve

PR #3102, opened 2025-10-15
refactor and expand knative integration tests

lkingland:
Expand and gently refactor knative integration tests.


pkg/knative/labels_int_test.go
```diff
@@ -0,0 +1,99 @@
...
+		if err != nil {
+			t.Logf("error removing Function: %v", err)
+		}
+	})
+
+	// Wait for function to be ready
+	_, err = client.Describe(ctx, "", "", f)
+	if err != nil {
+		t.Fatal(err)
+	}
```
gauron99: out of curiosity why is this needed?

pkg/knative/labels_int_test.go
```diff
@@ -0,0 +1,99 @@
...
+	// cliSet, err := k8s.NewKubernetesClientset()
+	// if err != nil {
+	// 	t.Fatal(err)
+	// }
+
+	// t.Log("deleting namespace: ", ns)
+	// err = cliSet.CoreV1().Namespaces().Delete(ctx, ns, metav1.DeleteOptions{})
+	// if err != nil {
+	// 	t.Logf("error deleting namespace: %v", err)
+	// }
```
gauron99: leftover comments/ cleanup?

pkg/knative/remover_int_test.go
```diff
@@ -0,0 +1,108 @@
...
+		if item.Name == f.Name {
+			found = true
+			break
+		}
+	}
+	if found {
+		t.Errorf("function %s was not removed", f.Name)
+	}
+
+	// Remove
```
gauron99, line 106: should we add cleanup here as with my prior comment about leftover comments?

pkg/knative/labels_int_test.go
lkingland: Thanks for reviewing!  This is indeed redundant.  Looks like I left this test slightly incomplete when considering the commented out blocks below as well.   The PR's still a draft, so I have an excuse!

pkg/knative/remover_int_test.go
lkingland: I got this by adding an auto-registered cleanup to the namespace generation function used during setup

PR #3104, opened 2025-10-16
update go 1.25 with minor fixes

gauron99:
update as per https://cloud-native.slack.com/archives/C04LGHDR9K7/p1760551102514759

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module knative.dev/func
 
-go 1.24.0
+go 1.25.0
 
 // this is required bacause of bad dep in github.com/openshift-pipelines/pipelines-as-code
 replace github.com/imdario/mergo => dario.cat/mergo v1.0.1
--- a/hack/install-gitlab.sh
+++ b/hack/install-gitlab.sh
@@ -28,7 +28,16 @@ kind: Namespace
 apiVersion: v1
 metadata:
   name: gitlab
----
+EOF
+
+# Creation is nearly immediate but for some reason pod wouldnt get created at all
+# in some cases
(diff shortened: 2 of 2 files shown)


matejvasek:
Why would be the `GOTOOLCHAIN` envvar required?

@dprotaso do you know?

dprotaso:
It shouldn't. But to prevent bumps we use 'local' so that we always choose our CI version

https://github.com/knative-extensions/knobots/blob/853d8e021d1e78c31a126223abcbb0bc69b6694c/actions/update-deps/Dockerfile#L33

Makefile
```diff
@@ -69,7 +69,7 @@ $(BIN): generate/zz_filesystem_generated.go
 
 .PHONY: test
 test: generate/zz_filesystem_generated.go ## Run core unit tests
-	go test -race -cover -coverprofile=coverage.txt ./...
+	GOTOOLCHAIN=go1.25.0+auto go test -race -cover -coverprofile=coverage.txt ./...
```
dprotaso: You shouldn't need to set the GOTOOLCHAIN variable

gauron99: Without it I get `go: no such tool "covdata"`

gauron99: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_func/3095/unit-tests_func_main/1978703828245024768

dprotaso: We probably need this in first - https://github.com/knative/infra/pull/726

dprotaso: and https://github.com/knative-extensions/knobots/pull/444

gauron99:
the additional kubectl wait for SA is due to https://github.com/kubernetes/kubernetes/issues/66689

PR #3113, opened 2025-10-17
Feat/create path flag

intojhanurag:
This PR fixes the --path flag in the func create command. Previously, cfg.Path was empty when using --path because the path computation happened after the confirm check, causing initialization failure

## Before
image

## After
Screenshot 2025-10-18 014515

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -71,7 +71,7 @@ EXAMPLES
 	  $ {{.Name}} create -l go -t cloudevents myfunc
 		`,
 		SuggestFor: []string{"vreate", "creaet", "craete", "new"},
-		PreRunE:    bindEnv("language", "template", "repository", "confirm", "verbose"),
+		PreRunE:    bindEnv("language", "template", "repository", "confirm", "verbose", "path"),
 		Aliases:    []string{"init"},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runCreate(cmd, args, newClient)
@@ -93,11 +93,6 @@ EXAMPLES
 	// Add --path flag (default ".") for consistency with other commands.
 	// Retain positional [path] for backward compatibility (warned later).
 	cmd.Flags().StringP("path", "p", ".", "Path to the function project directory ($FUNC_PATH)")
-
-	// Bind the --path flag to viper so it can be read in newCreateConfig
-	if err := viper.BindPFlag("path", cmd.Flags().Lookup("path")); err != nil {
-		fmt.Fprintf(os.Stderr, "unable to bind path flag: %v\n", err)
-	}
 
 	addVerboseFlag(cmd, cfg.Verbose)
 


intojhanurag:

cmd/create.go
```diff
@@ -37,7 +37,7 @@ NAME
 
 SYNOPSIS
 	{{.Name}} create [-l|--language] [-t|--template] [-r|--repository]
-	            [-c|--confirm]  [-v|--verbose]  [path]
+	            [--path <dir>] [-c|--confirm]  [-v|--verbose]
```
gauron99: lets keep this consistent with the other flags. `-l` `-t` and `-r` all take additional followup arguments but dont explicitly show that in the synopsis text here.

cmd/create.go
```diff
@@ -198,6 +203,19 @@ func newCreateConfig(cmd *cobra.Command, args []string, newClient ClientFactory)
...
+	if pathFlag != "" && pathFlag != "." {
+		finalPath = pathFlag
+	} else if len(args) >= 1 && args[0] != "" {
+		finalPath = args[0]
+	}
+
+	dirName, absolutePath = deriveNameAndAbsolutePathFromPath(finalPath)
+
+	cfg.Name = dirName
+	cfg.Path = absolutePath
```
gauron99: can we move all of this new code above the cfg initialization? that way we dont have to assign it twice. Just like the `deriveNameAndAbsolute..` function call was before, just adding this new code

cmd/create.go
```diff
@@ -90,7 +90,15 @@ EXAMPLES
 	cmd.Flags().StringP("repository", "r", "", "URI to a Git repository containing the specified template ($FUNC_REPOSITORY)")
 
 	addConfirmFlag(cmd, cfg.Confirm)
-	// TODO: refactor to use --path like all the other commands
+	// Add --path flag (default ".") for consistency with other commands.
+	// Retain positional [path] for backward compatibility (warned later).
+	cmd.Flags().StringP("path", "p", ".", "Path to the function project directory (default: current directory) ($FUNC_PATH)")
+
+	// Bind the --path flag to viper so it can be read in newCreateConfig
```
gauron99: why do we need to bind this to read later? we are not binding other flags and we still read them in `createConfig`.

gauron99 (commented):
thanks for the contribution! I added a few questions/suggestions

cmd/create.go
intojhanurag: Done

cmd/create.go
intojhanurag: Done

cmd/create.go
intojhanurag: I initially tried reading --path without binding it to viper, but viper.GetString("path") returned empty, And with binding it is working🙂

docs/reference/func_create.md
```diff
@@ -70,6 +70,7 @@ func create
   -c, --confirm             Prompt to confirm options interactively ($FUNC_CONFIRM)
   -h, --help                help for create
   -l, --language string     Language Runtime (see help text for list) ($FUNC_LANGUAGE)
+  -p, --path string         Path to the function project directory (default: current directory) ($FUNC_PATH) (default ".")
```
gauron99: Lets please cleanup this "defaulting" string. As you can see its printed 2. We can keep the variable at the end as with the other commands. Cobra uses the default value you present in the `cmd.Flags()...` when doing `func create -h` and should display the default you added at the end

gauron99: ah this is the reference 🤔 looks like it works with the other commands hmm

intojhanurag:
Hey @gauron99,  let me know if any changes are needed. I'm not sure why one of the tests is failing.

cmd/create.go
gauron99: OK I see it. There is a `bindEnv` function thats being called. You should be able to add path as one of the arguments there and that binds all those flags i mentioned.

cmd/create.go
gauron99: we should be able to add `path` here

gauron99:
> Hey @gauron99, let me know if any changes are needed. I'm not sure why one of the tests is failing.

@intojhanurag looks like a flake, I will re-run if it fails again

cmd/create.go
intojhanurag: Make sense , thanks :)

gauron99:
Thanks!

/lgtm
/approve

PR #3115, opened 2025-10-19
add test cases to defaultNamespace testing

intojhanurag:
Update defaultNamespace testing to include latest deployed and user-forced namespace (`f.Deploy.Namespace` and `f.Namespace` respectively)

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -301,7 +301,7 @@ func Test_defaultNamespace(t *testing.T) {
 					Namespace: "deployns",
 				},
 			},
-		},{
+		}, {
 			name:     "function already deployed override",
 			context:  true,
 			global:   true,
@@ -311,7 +311,7 @@ func Test_defaultNamespace(t *testing.T) {
 					Namespace: "deployns",
 				},
 			},
-		},{
+		}, {
 			name:     "static default",
 			context:  false,            // no active kube context
 			global:   false,            // no global


intojhanurag:

gauron99:
@intojhanurag the title and description of the message makes it sound like you changed some actual logic but only tests were changed/updated. Were there some code changes supposed to be there as well?

gauron99:
I would also order the tests in the precedence order. Since the tests run in order from top to bottom I would put the highest precedence at the bottom to keep the nice sequence. The forced namespace case can also have the deployed namespace in the function structure `fn.Function` to make sure it overrides it

intojhanurag:
> I would also order the tests in the precedence order. Since the tests run in order from top to bottom I would put the highest precedence at the bottom to keep the nice sequence. The forced namespace case can also have the deployed namespace in the function structure `fn.Function` to make sure it overrides it

Haha, yesterday while working on this PR, I accidentally changed the order of the tests. This morning, I realized something was off and corrected the order 😅. I initially thought I was making a different change, but I was actually just fixing my own mistake 😂.

intojhanurag:
Hey @gauron99, any updates or changes you’d like me to make here?

cmd/root_test.go
```diff
@@ -287,10 +287,28 @@ func Test_defaultNamespace(t *testing.T) {
...
 	}{
-		// TODO cases for function state f.Namespace and f.Deploy.Namespace
 		{
+			name:     "function namespace overrides",
+			context:  true,
+			global:   true,
+			expected: "functionns",
+			f: fn.Function{
+				Namespace: "functionns",
+			},
```
gauron99: Here I would also add a deploy namespace which is expected to be overriden. This way we can have all the cases covered from static default to the forced namespace

intojhanurag: Done! , Can you review ?

cmd/root_test.go
```diff
@@ -287,10 +287,39 @@ func Test_defaultNamespace(t *testing.T) {
...
+			context:  true,
+			global:   true,
+			expected: "deployns",
+			f: fn.Function{
+				Namespace: "functionns",
+				Deploy: fn.DeploySpec{
+					Namespace: "deployns",
+				},
+			},
+		}, {
```
gauron99: we can make the change simpler with just 2 cases. `fn.Namespace` overrider the `Deploy.Namespace`
```suggestion
			name:     "function namespace overrides",
			context:  true,
			global:   true,
			expected: "functionns",
			f: fn.Function{
				Namespace: "functionns",
				Deploy: fn.DeploySpec{
					Namespace: "deployns",
				},
			},
		}, {
			name:     "function already deployed override",
			context:  true,
			global:   true,
			expected: "deployns",
			f: fn.Function{
				Deploy: fn.DeploySpec{
					Namespace: "deployns",
				},
			},
		}, {
```

PR #3117, opened 2025-10-19
Add early cluster validation to prevent wasted build time in func deploy

### LINKED ISSUE #3116: func deploy: Add early cluster validation to prevent wasted build time
## Problem

When running `func deploy` with invalid cluster configuration or connectivity issues, the command wastes time building the container image before failing with misleading or technical error messages.

### Scenario 1a: Invalid KUBECONFIG path
**Command:**
```bash
KUBECONFIG=/nonexistent/config func deploy --registry ghcr.io/user
```

**Current Output:**
```txt
Building function image
Still building
^CError: executing lifecycle: context canceled 
```
**Problems:**

-  Starts building before validating cluster connection
-  Wastes time (minutes) building an image that can't be deployed
-  Shows confusing "invalid run-image" error instead of cluster connection issue
-  User must manually cancel (Ctrl+C) the build

### Scenario 1b: Invalid KUBECONFIG path with build and push false 
**Command:**
```bash
KUBECONFIG=/nonexistent/config func deploy --registry ghcr.io/user --build=false  --push=false
```

**Current Output:**
```txt
Error: deploy error. failed to create new serving client: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```

**Problems:**
- Suggests outdated `KUBERNETES_MASTER` environment variable instead of `KUBECONFIG`
- Technical error message doesn't indicate the kubeconfig file path is invalid
- No guidance on how to fix the issue or where the config file should be located

### Scenario 2a: Empty kubeconfig or no cluster configuration
**Command:**
```bash
# Create empty kubeconfig
cat > /tmp/empty-kubeconfig.yaml << 'EOF'
apiVersion: v1
kind: Config
clusters: []
contexts: []
users: []
current-context: ""
EOF

KUBECONFIG=/tmp/empty-kubeconfig.yaml func deploy --registry ghcr.io/user
```

**Current Output:**
```txt
Building function image
Still building
^CError: executing lifecycle: context canceled 
```
**Problems:**

-  Starts building before validating cluster connection
-  Wastes time (minutes) building an image that can't be deployed
-  Shows confusing "invalid run-image" error instead of cluster connection issue
-  User must manually cancel (Ctrl+C) the build

### Scenario 2b: Empty kubeconfig or no cluster configuration with build and push false 
**Command:**
```bash
# Create empty kubeconfig
cat > /tmp/empty-kubeconfig.yaml << 'EOF'
apiVersion: v1
kind: Config
clusters: []
contexts: []
users: []
current-context: ""
EOF

KUBECONFIG=/tmp/empty-kubeconfig.yaml func deploy --registry ghcr.io/user  --build=false --push=false
```

**Current Output:**
```txt
Error: deploy error. failed to create new serving client: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```
**Problems:**
- Identical error message as Scenario 1b, making it impossible to distinguish between missing file and empty configuration
- Suggests deprecated `KUBERNETES_MASTER` variable instead of modern cluster setup commands (`minikube start`, `kind create cluster` )

### Scenario 3a: Valid kubeconfig but cluster is down
**Command:**
```bash
# Create kubeconfig with unreachable cluster
cat > /tmp/invalid-cluster-kubeconfig.yaml << 'EOF'
apiVersion: v1
kind: Config
clusters:
- cluster:
    server: https://127.0.0.1:6443
  name: invalid-cluster
contexts:
- context:
    cluster: invalid-cluster
    user: invalid-user
  name: invalid-context
current-context: invalid-context
users:
- name: invalid-user
  user: {}
EOF

KUBECONFIG=/tmp/invalid-cluster-kubeconfig.yaml func deploy --registry ghcr.io/user
```

**Current Output:**
```txt 
Building function image
Still building
^CError: executing lifecycle: context canceled
```
**Problems:**
-  Starts building before checking if cluster is accessible
- Wastes significant time and resources
-  Generic error doesn't indicate cluster connectivity issue

### Scenario 3b: Valid kubeconfig but cluster is down with build and push false 
**Command:**
```bash
# Create kubeconfig with unreachable cluster
cat > /tmp/invalid-cluster-kubeconfig.yaml << 'EOF'
apiVersion: v1
kind: Config
clusters:
- cluster:
    server: https://127.0.0.1:6443
  name: invalid-cluster
contexts:
- context:
    cluster: invalid-cluster
    user: invalid-user
  name: invalid-context
current-context: invalid-context
users:
- name: invalid-user
  user: {}
EOF

KUBECONFIG=/tmp/invalid-cluster-kubeconfig.yaml func deploy --registry ghcr.io/user
```

**Current Output:**
```txt 
Error: deploy error. knative deployer failed to get the Knative Service: Get "https://127.0.0.1:6443/apis/serving.knative.dev/v1/namespaces/default/services/myfunction": tls: failed to verify certificate: x509: certificate signed by unknown authority
```
**Problems:**
- Exposes technical error with full URL and certificate details that aren't helpful to most users
- Error message focuses on TLS certificate verification rather than the actual issue (cluster is down/unreachable)
- No guidance on checking cluster status or starting the cluster

## Proposed Improvement

Implement **early cluster validation** with a **2-layer error system** that checks cluster connectivity **before** starting the build process.

### Scenario 1: Invalid KUBECONFIG path
**Improved Output:**
```txt
Error: invalid kubeconfig

The kubeconfig file at '/nonexistent/config' does not exist or is not accessible.        

Try this:
  export KUBECONFIG=~/.kube/config           Use default kubeconfig
  kubectl config view                        Verify current config
  ls -la ~/.kube/config                      Check if config file exists

```

### Scenario 2: Empty kubeconfig or no cluster configuration
**Improved Output:**
```txt
Error: cluster not accessible

Cannot connect to Kubernetes cluster. No valid cluster configuration found.

Try this:
  minikube start                             Start Minikube cluster
  kind create cluster                        Start Kind cluster
  kubectl cluster-info                       Verify cluster is running
  kubectl config get-contexts                List available contexts

```

### Scenario 3: Valid kubeconfig but cluster is down
**Improved Output:**
```txt 
Error: cluster not accessible

Cannot connect to Kubernetes cluster. No valid cluster configuration found.

Try this:
  minikube start                             Start Minikube cluster
  kind create cluster                        Start Kind cluster
  kubectl cluster-info                       Verify cluster is running
  kubectl config get-contexts                List available contexts

```

RayyanSeliya:
# Changes

- :bug: Add early cluster validation to prevent wasted build time in `func deploy`
- :broom: Improve error messages for cluster connection failures with actionable guidance

## What Changed

`func deploy` now validates Kubernetes cluster connectivity **before** starting the container build process, preventing wasted time when the cluster is inaccessible.

**Before:**
- Built container image first (2-5 minutes)
- Then failed with confusing errors like "invalid run-image" or "context canceled"
- Different errors depending on whether `--build=false` was used

**After:**
- Validates cluster connection immediately (< 5 seconds)
- Fails fast with clear, specific error messages
- Consistent behavior regardless of build flags
- Provides actionable guidance for each error type

## Implementation

Added 2-layer error handling:
- **Layer 1** (`pkg/functions/errors.go`): Technical errors (`ErrInvalidKubeconfig`, `ErrClusterNotAccessible`)
- **Layer 2** (`cmd/deploy.go`): User-friendly CLI messages with examples

Detects three distinct error scenarios:
1. Invalid kubeconfig file path
2. Empty/no cluster configuration
3. Cluster unreachable (network, auth, down, etc.)

## Testing

Tested all combinations:
- ✅ Invalid KUBECONFIG path (with/without build flags)
- ✅ Empty kubeconfig (with/without build flags)  
- ✅ Unreachable cluster (with/without build flags)
- ✅ Cluster stopped after configuration (network error)
- ✅ Valid cluster with kind (success path)
- ✅ Unit tests pass (`TestDeploy_ConfigPrecedence`, etc.)

/kind bug

Fixes #3116 

**Release Note**

```release-note
`func deploy` now validates cluster connectivity before building, providing immediate feedback with clear error messages instead of wasting time on builds that will fail deployment.
```

### DIFF
--- a/pkg/functions/errors.go
+++ b/pkg/functions/errors.go
@@ -37,7 +37,7 @@ var (
 
 	// ErrInvalidDomain is returned when a domain name doesn't meet DNS subdomain requirements
 	ErrInvalidDomain = errors.New("invalid domain")
-	
+
 	// ErrInvalidKubeconfig is returned when the kubeconfig file path is invalid or inaccessible
 	ErrInvalidKubeconfig = errors.New("invalid kubeconfig")
 


cmd/deploy.go
```diff
@@ -315,6 +356,61 @@ For more options, run 'func deploy --help'`, err)
...
+			return fmt.Errorf(`%w
+
+Cannot connect to Kubernetes cluster.
+
+Try this:
+  kubectl cluster-info                       Verify cluster is accessible
+  minikube status                            Check Minikube cluster status
+  kubectl get nodes                          Test cluster connection
+
+For more options, run 'func deploy --help'`, fn.ErrClusterNotAccessible)
```
gauron99: i would be in favor of making these into functions so they dont take up so much space in the main-flow function

cmd/deploy.go
```diff
@@ -236,6 +237,46 @@ EXAMPLES
 	return cmd
 }
 
+// validateClusterConnection checks if the Kubernetes cluster is accessible before starting build
+func validateClusterConnection() error {
+	// Skip for test environments (check if using test kubeconfig)
+	kubeconfigPath := os.Getenv("KUBECONFIG")
+	if kubeconfigPath != "" && (strings.Contains(kubeconfigPath, "/testdata/") || strings.HasSuffix(kubeconfigPath, "default_kubeconfig")) {
+		return nil
```
gauron99: Im not really sure about these paths hardcoded here, wdyt @lkingland @matejvasek ?

cmd/deploy.go
```diff
@@ -236,6 +237,46 @@ EXAMPLES
 	return cmd
 }
 
+// validateClusterConnection checks if the Kubernetes cluster is accessible before starting build
+func validateClusterConnection() error {
+	// Skip for test environments (check if using test kubeconfig)
```
matejvasek: I would prefer if we called our `k8s.NewKubernetesClientset()` and checked the err return value.
Also I do not like making the code "test aware". Could we think of something better?

RayyanSeliya: Thxs for the feedback @matejvasek  ! I've updated it to use `k8s.NewKubernetesClientset()` and check the error as you suggested. The `/testdata/` check follows Go's convention for test fixtures  , it catches cases where tests use stale` kubeconfig` paths. The `.example.com` check skips API calls for `test/example` clusters (RFC 2606 reserved domain).

gauron99:
Im not much of a fan of this implementation... its still "test aware" as Matej said - with those literal strings and constants. Also there has to be some better way to implement this :thinking: 
Maybe as a method of client `Instance` (in pkg/functions/client.go) ? Instance has its method "CurrentState" or "State" or "Available" or something like this? 
@lkingland @matejvasek

RayyanSeliya:
> Im not much of a fan of this implementation... its still "test aware" as Matej said - with those literal strings and constants. Also there has to be some better way to implement this 🤔 Maybe as a method of client `Instance` (in pkg/functions/client.go) ? Instance has its method "CurrentState" or "State" or "Available" or something like this? @lkingland @matejvasek

@gauron99 Good point about the client architecture! I checked `pkg/functions/client.go` and saw it already has methods like `Deploy()`, `Build()`, etc.

I'm thinking we could add:
```go
func (c *Client) ClusterAvailable(ctx context.Context) error
```

Then just call it before building. Tests would naturally skip it since they use `NewTestClient()` with mocks.

The challenge is timing - we currently validate before creating the client (line 423 vs line 462 in `cmd/deploy.go`'s `runDeploy()`). We'd need to either:
1. Create the client earlier, or
2. Make it a standalone function like `functions.ValidateCluster()`

Which direction makes more sense for the existing code? Don't want to refactor the way which is not feasible ..! 

@lkingland @matejvasek

lkingland:
This one is a little tricky...
I understand the desire to validate in the CLI to "fail fast".  But too much pre-validation creates a hard dependency between the CLI and the deployer.

I think it might be worth looking into placing this validation in the deployer implementation itself, returing a typed error on failure, and then capturing this error in the CLI and adding "CLI specific" help text as necessary.

Remember that it's ok if the system builds and then fails on deployment, because it should not repeat the build on a subsequent deploy (it detects the build is "fresh").

RayyanSeliya:
> This one is a little tricky... I understand the desire to validate in the CLI to "fail fast". But too much pre-validation creates a hard dependency between the CLI and the deployer.
> 
> I think it might be worth looking into placing this validation in the deployer implementation itself, returing a typed error on failure, and then capturing this error in the CLI and adding "CLI specific" help text as necessary.
> 
> Remember that it's ok if the system builds and then fails on deployment, because it should not repeat the build on a subsequent deploy (it detects the build is "fresh").

thx for the feedback @lkingland   that makes  sense and sounds good to have the validation into the deployer itself with typed errors and CLI just catches those and provide a user-friendly errors ! can have a look now and ping me if any more changes needed !

RayyanSeliya:
hey @lkingland   dont know why the test are failing !!

gauron99:
/override "On Cluster RT Test (ubuntu-latest, pack)" "E2E Test (ubuntu-latest, springboot)"

gauron99:
These fails are unrelated. Its an issue with our custom pack builder currently

RayyanSeliya:
> These fails are unrelated. Its an issue with our custom pack builder currently

Yeah ! 👍Thx ..

PR #3126, opened 2025-10-22
feat(cmd): handle ErrEnvNotExist with typed error checking

intojhanurag:
## Changes
- Introduced typed error `ErrEnvNotExist` to clearly indicate missing environment variables.
- Added error check in tests using `errors.As` to assert that the correct typed error is returned.
- Validates that the `Name` field in `ErrEnvNotExist` matches the expected missing environment variable.
- Improves test reliability and clarity for environment variable management.

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -421,14 +421,6 @@ For more options, run 'func deploy --help'`, err)
 	return f.Stamp()
 }
 
-type ErrEnvNotExist struct {
-	Name string
-}
-
-func (e ErrEnvNotExist) Error() string {
-	return fmt.Sprintf("environment variable %q does not exist", e.Name)
-}
-
 // build when flag == 'auto' and the function is out-of-date, or when the
 // flag value is explicitly truthy such as 'true' or '1'.  Error if flag
 // is neither 'auto' nor parseable as a boolean.  Return CLI-specific error
--- a/cmd/deploy_test.go
+++ b/cmd/deploy_test.go
@@ -454,7 +454,7 @@ func TestDeploy_Envs(t *testing.T) {
 	cmd.SetArgs([]string{"--env=DOES_NOT_EXIST-"})
 	err = cmd.Execute()
 
-	var e *ErrEnvNotExist
+	var e *fn.ErrEnvNotExist
 	if !errors.As(err, &e) {
 		t.Fatalf("expected ErrEnvNotExist, got '%v'", err)
 	}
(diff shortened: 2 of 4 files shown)


intojhanurag:

cmd/deploy_test.go
```diff
@@ -448,7 +448,19 @@ func TestDeploy_Envs(t *testing.T) {
 		t.Fatalf("Expected no envs to remain, got '%v'", f.Run.Envs)
 	}
 
-	// TODO: create and test typed errors for ErrEnvNotExist etc.
```
gauron99, line 451: there is a "etc" so I assume we need to do this for more errors

intojhanurag: I can add , but how will I know which ones are missing? Otherwise, I can keep this comment so that we can add more in the future.

gauron99: yeah lets add that comment back just in case, thanks

intojhanurag: Done !

intojhanurag:
Hey @gauron99 , All tests passed in this pr .

cmd/deploy_test.go
```diff
@@ -449,6 +449,20 @@ func TestDeploy_Envs(t *testing.T) {
 	}
 
 	// TODO: create and test typed errors for ErrEnvNotExist etc.
+
+	cmd = NewDeployCmd(NewTestClient())
+	cmd.SetArgs([]string{"--env=DOES_NOT_EXIST-"})
```
gauron99, line 454: you are assigning `DOES_NOT_EXIST-` (with the hyphen at the end) thats intentional? Shouldnt you then test for the exact same name?

intojhanurag: Hey @gauron99 , It is intentional that is the sign of that "env" does not exist.

cmd/deploy.go
```diff
@@ -421,6 +421,14 @@ For more options, run 'func deploy --help'`, err)
 	return f.Stamp()
 }
 
+type ErrEnvNotExist struct {
+	Name string
+}
+
+func (e ErrEnvNotExist) Error() string {
+	return fmt.Sprintf("environment variable %q does not exist", e.Name)
+}
```
gauron99: can we put this into `pkg/functions/errors.go` please to keep consistent with the place for error definitions.
Otherwise looks good to me

cmd/deploy_test.go
gauron99: You are actually removing it prior (even though it didnt exist anyways). I think its fine

cmd/deploy.go
intojhanurag: Hey @gauron99 , Thanks for this suggestion ! Now Is it good to go ?

gauron99 (approved):
/lgtm

PR #3148, opened 2025-10-25
Fix InsecureSSL default for GitHub webhooks

intojhanurag:
Changed `InsecureSSL` from "1" (insecure) to "0" (secure) to ensure SSL certificates
are properly verified by default.


gauron99:
@matejvasek @lkingland can we do this? not sure if it used to break something beforehand or why havent we done this prior

PR #3149, opened 2025-10-26
feat: add interactive prompts for builder, push, and builder image options

intojhanurag:
### Changes
Adds interactive prompts to `func build --confirm` for additional build options, including:

- Builder selection (`--builder`)
- Push option (`--push`)
- Builder image (`--builder-image`)
- Base image (`--base-image`)

## Before
Screenshot 2025-10-26 124916

## After
Screenshot 2025-10-26 124222

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -406,7 +406,7 @@ func (c buildConfig) Prompt() (buildConfig, error) {
 		{
 			Name: "push",
 			Prompt: &survey.Confirm{
-				Message: "Push image after build?",
+				Message: "Push image to your registry after build?",
 				Default: c.Push,
 			},
 		},
@@ -422,7 +422,7 @@ func (c buildConfig) Prompt() (buildConfig, error) {
 			{
 				Name: "baseImage",
 				Prompt: &survey.Input{
-					Message: "Optional base image for your function (host builder only):",
+					Message: "Optional base image for your function (empty for default):",
 					Default: c.BaseImage,
 				},
 			},


gauron99:
Hmm Im thinking to include only the necessary options. Keep the optional ones out of the prompts. (remove the builderImage and base image since its only host builder -- or do a conditional check for baseImage only if host builder is used.

cmd/build.go
```diff
@@ -395,12 +395,45 @@ func (c buildConfig) Prompt() (buildConfig, error) {
...
+			Prompt: &survey.Select{
+				Message: "Select builder:",
+				Options: []string{"pack", "s2i", "host"},
+				Default: c.Builder,
+			},
+		},
+		{
+			Name: "push",
+			Prompt: &survey.Confirm{
+				Message: "Push image after build?",
```
gauron99:
```suggestion
				Message: "Push image to registry after build?",
```

third_party/VENDOR-LICENSE/github.com/hashicorp/go-cleanhttp/doc.go
gauron99, line 1: why is there a change to third party?

cmd/build.go
gauron99:
```suggestion
				Message: "Push image to your registry after build?",
```

cmd/build.go
```diff
@@ -395,12 +395,45 @@ func (c buildConfig) Prompt() (buildConfig, error) {
...
+	if err != nil {
+		return c, err
+	}
+
+	if c.Builder == "host" {
+		hostQs := []*survey.Question{
+			{
+				Name: "baseImage",
+				Prompt: &survey.Input{
+					Message: "Optional base image for your function (host builder only):",
```
gauron99:
```suggestion
					Message: "Optional base image for your function (empty for default):",
```
from the user perspective, is it left empty if they dont provide a value? In other words is the default shown in the prompt? if not I would suggest this.

gauron99 (commented):
Is there a prompt test already in use where we could add test cases for these new prompts?

third_party/VENDOR-LICENSE/github.com/hashicorp/go-cleanhttp/doc.go
intojhanurag: I just applied the goimports -w . to the whole repo. becasue test was failing

intojhanurag:
Hey @gauron99 , I added the all suggestion. Please take a look on it .

cmd/build.go
intojhanurag: Done

third_party/VENDOR-LICENSE/github.com/hashicorp/go-cleanhttp/doc.go
lkingland: That's interesting.  The check might fail asking to "update codegen" if this is a real problem (probably an IDE autocleanup or other)

Issue #3085, opened 2025-10-13
different func type args take different values for the same thing


dprotaso:
I noticed that when using `func init` we pass in `-t cloudevents` but when using `func invoke` the `-f=cloudevent` 

I understand the difference but it would be nice to use `cloudevents`/`cloudevent` interchangeably. This reduces the cognitive load of the user.

dprotaso:
I would suggest we advertise a single casing but accept both casing for backwards compatability

PR #3164, opened 2025-11-01
Fix/type args diff

intojhanurag:
## Fix: #3085

### DIFF
--- a/cmd/invoke.go
+++ b/cmd/invoke.go
@@ -261,7 +261,7 @@ func newInvokeConfig() (cfg invokeConfig, err error) {
 
 	switch strings.ToLower(cfg.Format) {
 	case "cloudevent", "cloudevents":
-		cfg.Format = "cloudevents"
+		cfg.Format = "cloudevent"
 	}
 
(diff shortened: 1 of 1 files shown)


intojhanurag:
Hey @gauron99 , Can you take a look on this one , if you have a moment ?

gauron99:
@lkingland I think we should actually promote the single case in both format (`--format`) and templates (`--template`) instead? (internally we support both with singular AND plural and we just overwrite the plural in cli)
Basically the other way around than this PR. 

WDYT?

lkingland:
Yes.  I agree.  Plurality is not easily understood for folks with english as a second language, so I would suggest we accept both in the CLI, but be strict in the "core", accepting only the singular.

It will be so nice to have this standardized!

intojhanurag:
Hey  @gauron99 , what I need to do ? Please lmk if anything .

gauron99:
It would be good to implement the reverse as we discussed with Luke @intojhanurag

intojhanurag:
Hey @gauron99 , I reversed the logic .

PR #3208, opened 2025-11-19
Add --registry-authfile to specify a specific auth file for the registry push

creydr:
# Changes

- :gift: Adding the `--registry-authfile string` build parameter to allow to specify a custom registry auth file location

/kind enhancement

**Release Note**

```release-note
Add the --registry-authfile build parameter to allow to specify a custom registry auth file location
```

### DIFF
--- a/pkg/creds/credentials_test.go
+++ b/pkg/creds/credentials_test.go
@@ -890,7 +890,7 @@ func TestCredentialsFromAuthfile(t *testing.T) {
 // ********************** helper functions below **************************** \\
 
 func resetHomeDir(t *testing.T) {
-	t.TempDir()
+	t.Helper()
 	if err := os.RemoveAll(homeTempDir); err != nil {
 		t.Fatal(err)
 	}
@@ -901,6 +901,7 @@ func resetHomeDir(t *testing.T) {
 
 // resetHomePermissions resets the HOME perms to 0700 (same as resetHomeDir(t))
 func resetHomePermissions(t *testing.T) {
+	t.Helper()
 	if err := os.Chmod(homeTempDir, 0700); err != nil {
 		t.Fatal(err)
 	}


creydr:
/hold
need to verify with remote deploy

lkingland (commented):
👍🏻 Nice

Can we make this change to the `oci.Pusher` as well, so that it works withe the `Host` builder (and for users with no docker/podman available)?  They're all using the `creds` package, so I assume this could work for both pushers.

Also, if you rebase on `main` you'll get some improved E2E tests.

creydr:
> 👍🏻 Nice
> 
> Can we make this change to the `oci.Pusher` as well, so that it works withe the `Host` builder (and for users with no docker/podman available)? They're all using the `creds` package, so I assume this could work for both pushers.
> 
> Also, if you rebase on `main` you'll get some improved E2E tests.

Hey @lkingland,
I added / reinitialized the pusher in the build.go clientOptions() with the credentials provider for host,pack,s2i builders (see https://github.com/creydr/knative-func/blob/2f563bfd250c78850f7b1fc61b232880cc1f3178/cmd/build.go#L498), as we there had the registry-authfile flag parsed. Or do you mean something different?

creydr:
Rebased. Tests seem to pass now again

creydr:
@lkingland @gauron99 can you recheck?

pkg/creds/credentials_test.go
```diff
@@ -815,6 +815,91 @@ func TestCredentialsHomePermissions(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			resetHomeDir(t)
+
+			authFile, err := os.CreateTemp("", "test-auth-*.txt")
+			if err != nil {
+				t.Fatalf("failed to create temp auth file: %s", err)
+			}
+			t.Cleanup(func() {
+				os.Remove(authFile.Name())
```
twoGiants: You could use `t.TempDir` and create the file there. Then you would not need to cleanup as the temp dir is removed automatically.

pkg/creds/credentials_test.go
```diff
@@ -815,6 +815,91 @@ func TestCredentialsHomePermissions(t *testing.T) {
...
+				creds.WithAuthFilePath(authFile.Name()),
+			)
+
+			got, err := credentialsProvider(context.Background(), tt.image)
+
+			// ASSERT
+			if err != nil {
+				t.Errorf("%v", err)
+				return
+			}
```
twoGiants:
```suggestion
			if err != nil {
				t.Fatalf("unexpected error: %v", err)
			}
```

pkg/creds/credentials_test.go
```diff
@@ -815,6 +815,91 @@ func TestCredentialsHomePermissions(t *testing.T) {
...
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("got: %v, want: %v", got, tt.want)
+			}
+		})
+	}
+}
+
 // ********************** helper functions below **************************** \\
 
 func resetHomeDir(t *testing.T) {
```
twoGiants, line 892: Right below that line in 906 is `t.TempDir()` which probably should be `t.Helper()`. And the function right below `resetHomePermissions` is missing the `t.Helper()` too.

pkg/creds/credentials_test.go
twoGiants: This can probably updated in few more tests in this file.

pkg/creds/credentials_test.go
creydr: Good point. Updated it

pkg/creds/credentials_test.go
creydr: :+1: Updated it

pkg/creds/credentials_test.go
creydr: :+1:

twoGiants (approved):
Looks great! :smile_cat: :+1:

Issue #3197, opened 2025-11-18
Execute Test_validatedLinkTarget from temp dir


lkingland:
The test `Test_validatedLinkTarget` creates absoluteLink and absoluteLinkWindows in `testdata` during execution.  This test should be executed from within a temp directory as with other tests, and the ignore directives removed from .gitignore

lkingland:
Also, let's please take the time to check if any _other_ tests are creating those two links, and if so, give them the same update.  That way we can remove them from .gitignore.

Issue #3158, opened 2025-10-29
Make /pkg/creds/auth.json go away


lkingland:
Track down which test is creating /pkg/creds/auth.json and either change the effective path to a temp directory (ideal) or remove it in a defer (suboptimal).

- track down `pkg/oci/testdata/test-links/absoluteLink*` as well (should be 2 files, second ending with 'Windows')

PR #3223, opened 2025-11-22
Fix/test validated link target tempdir

intojhanurag:
## Fixed: #3197

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,11 +49,6 @@ CLAUDE.md
 # https://github.com/knative/func/issues/3196
 /pkg/builders/testdata/go-fn-with-private-deps/.s2i
 
-# TODO: Run this test from a temp directory instead:
-# https://github.com/knative/func/issues/3197
-pkg/oci/testdata/test-links/absoluteLink
-pkg/oci/testdata/test-links/absoluteLinkWindows
-
 # TODO: update the test which creates this to run in a temp directory:
 # https://github.com/knative/func/issues/3158
 /pkg/creds/auth.json


intojhanurag:
Hey @lkingland , Now it is ready for review . Please take a look , if you have a moment.

gauron99:
This PR (and also I have seen a similar one recently perhaps from you @intojhanurag as well) removes the test leftover files, we should remove those .gitignore paths now that they are no longer necessary as part of cleanup

intojhanurag:
Hey @gauron99 , I think there was two-three issues of same type but not exactly same .

lkingland:
Looking good!

I made a little comment on the issue itself.  Can we either 1) open a new issue to track hunting down any other tests which create those links or 2) fix them as part of this issue #3197.  That way we can reach the final goal of being able to remove those lines from .gitignore 👍🏻

intojhanurag:
Hey @lkingland, I first deleted both files from the repo where they were being created, then I ran "make test". I checked, and they were not created again.
So I’m assuming the problem is solved, and I also removed both file locations from .gitignore.
Let me know if my approach is wrong.

gauron99:
Ideally I would like to make sure this is FOR SURE removed so we dont have to go back and forth. Ideally running `make test-full`. Im not sure what tests exactly interact with those files from the top of my head, perhaps not needed. 

If you're not sure lets keep the issue open so we know its not completely finished/investigated and please edit the issue to track what has been done and what hasnt, thanks!

intojhanurag:
Hey @gauron99 , Idk why "make test-full" command stuck in my laptop . but i checked on "make test" there is no file creating.  btw we can keep open till then.

gauron99:
`test-full` requires you to setup the cluster with some extra scripts along the way. Its all in `hack` directory. You should be able to run `hack/binaries.sh && hack/cluster.sh && hack/registry.sh && hack/gitlab.sh` The names might be slightly off but in general

gauron99:
This PR partially fixes https://github.com/knative/func/issues/3158; taking care of the `absoluteLink*` files

PR #3220, opened 2025-11-21
:bug: Populate image field in `func describe`

creydr:
# Changes

Currently the image field in `func describe` is not set:
```
{
  "Route": "http://func-go-hello-world.default.svc",
  "routes": [
    "http://func-go-hello-world.default.svc"
  ],
  "name": "func-go-hello-world",
  "image": "",
  "namespace": "default",
  "deployer": "raw",
  "subscriptions": null,
  "labels": {
    "boson.dev/function": "true",
    "function.knative.dev/name": "func-go-hello-world",
    "function.knative.dev/runtime": "go"
  }
}
```

This PR addressed it and populates the field:
```
{
  "Route": "http://func-go-hello-world.default.svc",
  "routes": [
    "http://func-go-hello-world.default.svc"
  ],
  "name": "func-go-hello-world",
  "image": "localhost:5001/func-go-hello-world@sha256:805b29c3ff6a379ebab4ddc27e974734d2e23baee3454c38abea9fccfc9b061c",
  "namespace": "default",
  "deployer": "raw",
  "subscriptions": null,
  "labels": {
    "boson.dev/function": "true",
    "function.knative.dev/name": "func-go-hello-world",
    "function.knative.dev/runtime": "go"
  }
}
```

/kind bug

**Release Note**
```release-note
fix: populate image field in func describe
```

PR #3239, opened 2025-11-24
fix: codecov reporting for e2es

gauron99:
add e2e reports back so it can be collected via codecov while taking the new FUNC_E2E* variable into consideration
(see comment below for more info)


gauron99:
This PR should fix recent codecov failures reporting about 7% drops (because less files are being reported) I noticed here for example: https://github.com/knative/func/pull/3220
This issue has been introduced in the e2e refactor here https://github.com/knative/func/pull/2938
The main CI has had some failures therefore the codecov can pick later main CI to compare to, its comparing to older coverages therefore these issues occur. I dont know why some PRs pick older ones while others dont. Weve had multiple (simple) PRs updating random stuff AFTER the e2e merge but this codecov failure has appeared only now

Issue #3256, opened 2025-11-29
Implement Github Workflow generation for building and deploying a function to a k8 cluster.


twoGiants:
## Summary

Extend the `func` cli with the capability to generate a GitHub Actions workflow setup (with yaml file and directory) in the given function repo. 

The workflow should contain:
- a checkout of the code
- get/setup of the `func` cli using the `knative-func-action`
- and running `func deploy|func deploy --remote`

The user should have the option to create the workflow with different configurations for a local build in the Github action runner or a remote on-cluster build.

Subtask of #2024.

## Resources
- sample workflow: https://github.com/functions-dev/templates/blob/main/.github/workflows/invoke-all.yaml
- func github action: https://github.com/functions-dev/action

## Pull Requests

1. #3257 
1. #3262 
1. #3266 
1. #3295 
1. #3297 
1. #3128
2. #3371 
3. #3379 
4. #3407 
5. #3457
6. _tba_
7. ...
8. ...

PR #3262, opened 2025-12-01
Refactor: extract common loader/saver interfaces (Github Workflow generation prep).

twoGiants:
# Changes

- :broom: Extract FunctionLoader, FunctionSaver, and FunctionLoaderSaver interfaces into new `cmd/common` package
- :broom: Move standardLoaderSaver implementation to `cmd/common` as DefaultLoaderSaver
- :white_check_mark: Add tests for the common loader/saver functionality
- :wrench: Add common testing factory function `CreateFuncInTempDir` for reuse in upcoming tests

/kind cleanup

Subtask of #3256.

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/cmd/common/common.go
+++ b/cmd/common/common.go
@@ -1,0 +1,47 @@
+package common
+
+import (
+	"fmt"
+
+	fn "knative.dev/func/pkg/functions"
+)
+
+// DefaultLoaderSaver implements FunctionLoaderSaver composite interface
+var DefaultLoaderSaver FunctionLoaderSaver = standardLoaderSaver{}
+
+// FunctionLoader loads a function from a filesystem path.
+type FunctionLoader interface {
+	Load(path string) (fn.Function, error)
+}
+
+// FunctionSaver persists a function to storage.
+type FunctionSaver interface {
+	Save(f fn.Function) error
+}
+
+// FunctionLoaderSaver combines loading and saving capabilities for functions.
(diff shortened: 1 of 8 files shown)


cmd/common/common.go
creydr, line 1: I am wondering, if we should have them in the `pkg/functions` package instead, like we have it for some function related interfaces too 🤔

twoGiants: I kept in in the cmd layer because it just a thin wrapper around very little functionality of the Function package which is not used outside the cmd layer.

i.e. it is not a real extension of the `Functions`  API...just a facade.

cmd/common/common.go
```diff
@@ -0,0 +1,47 @@
+package common
+
+import (
+	"fmt"
+
+	fn "knative.dev/func/pkg/functions"
+)
+
+// DefaultLoaderSaver implements FunctionLoaderSaver composite interface
+var DefaultLoaderSaver standardLoaderSaver
```
matejvasek: Probably should be of the interface type not the concrete type?

matejvasek:
```
var DefaultLoaderSaver FunctionLoaderSaver = standardLoaderSaver{}

matejvasek:
Ah it looks like you are not project reviewer/approver so you PRs are not auto approved.

PR #3270, opened 2025-12-04
Add label of scaffolded middleware version to function image

creydr:
# Changes

- :gift: Add the middleware version which was used during scaffolding as a label (`middleware-version`) to the function image
  - as we use scaffolding ATM only for s2i and host builder, this is only supported/added when one of those two builders is used

For example on a Golang build:
```
$ ~/development/knative-func/func build --registry quay.io/creydr/func-go-hello-world  --push --builder host        
Building function image
   f.linux.amd64
   f.linux.arm64
   f.linux.arm.v7
🙌 Function built: quay.io/creydr/func-go-hello-world/func-go-hello-world:latest

$ skopeo inspect --no-tags --tls-verify=false docker://quay.io/creydr/func-go-hello-world/func-go-hello-world:latest
{
    "Name": "quay.io/creydr/func-go-hello-world/func-go-hello-world",
    ...
    "Labels": {
        "middleware-version": "v0.21.3"
    },
   ...
}
```

/kind enhancement

**TODOs / Follow-ups:**
* we don't have a middlewareVersionDetector for Rust yet
* Provide this information as part of `func describe` (maybe only in `-o json|yaml`)
* Think about adding information about the middleware package (e.g. func-python, ...) as a label too

**Release Note**

```release-note
Provide used middleware version as a function label
```

### DIFF
--- a/pkg/scaffolding/middleware_version.go
+++ b/pkg/scaffolding/middleware_version.go
@@ -51,7 +51,7 @@ func getVersionDetector(runtime string) (middlewareVersionDetector, error) {
 	case "java":
 		return &springMiddlewareVersionDetector{}, nil
 	case "rust":
-		fallthrough //TODO: implement
+		return &rustMiddlewareVersionDetector{}, nil
 	default:
 		return nil, fmt.Errorf("unsupported runtime: %s", runtime)
 	}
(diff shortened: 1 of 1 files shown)


gauron99:
thanks Christoph!

PR #3275, opened 2025-12-05
feat: allow customizing host for local function runner

intojhanurag:
This PR introduces support for configuring the host used by `func run`.
Previously the local runner always bound to 127.0.0.1, which prevented use
cases such as exposing the function to other devices, local clusters, or
reverse proxies.

Improvement:
- Added a new CLI flag `--host` to allow overriding the default host.
- Updated the runner to resolve host/port from existing `--address` and the
  new `--host` option without breaking existing behavior.
- Ensured backward compatibility by preserving the previous default
  (`127.0.0.1:8080`) when no host is specified.

This enables workflows such as:
```bash
func run --host 0.0.0.0
```
and improves local development flexibility.

Screenshot 2025-12-05 220713

### DIFF
--- a/pkg/functions/client.go
+++ b/pkg/functions/client.go
@@ -484,7 +484,7 @@ func (c *Client) Update(ctx context.Context, f Function) (string, Function, erro
 		return "", f, err
 	}
 
-	// TODO: change this later when push does'nt return built image.
+	// TODO: change this later when push doesn't return built image.
 	// Assign this as c.Push is going to produce the built image (for now) to
 	// .Deploy.Image for the deployer -- figure out where to assign .Deploy.Image
 	// first, might be just moved above push
@@ -502,7 +502,7 @@ func (c *Client) Update(ctx context.Context, f Function) (string, Function, erro
 // Function. Used by Apply when the path is not yet an initialized function.
 // Errors if the path is already an initialized function.
 //
-// Use Apply for higher level control.  Use Init, Build, Push, Deploy
+// Use Apply for higher level control. Use Init, Build, Push, Deploy
 // independently for lower level control.
 //
 // Returns the primary route to the function or error.


intojhanurag:
Hey @gauron99, is it cool 😀? Please take a look whenever you have a moment.

intojhanurag:
Hey @lkingland , If you have a moment , Pls take a look on it :)

matejvasek:
Isn't `--host` redundant given `--address` exists?  If you mind that the address flags expects `host:port` we can change that and allow address without port (where 8080 would be implied port).

gauron99:
Yeah on the first glance it seems that we should be able to use `address` flag for this. We can just parse it for interface, port and both together. We have default for each and fill in what is missing or overriding the whole default via given address

intojhanurag:
Hey @gauron99 @matejvasek , thanks for the feedback ! I didn't notice :) 
Now i am going to make flexible the existing command .

matejvasek:
```go
func ParseAddressFlag(val string) (string, string) {
	if h, p, err := net.SplitHostPort(val); err == nil {
		return h, p
	}
	if val == "" {
		val = "localhost"
	}
	// maybe check that `val` is in fact valid ipv4,ipv6 or valid hostname here
	return val, "8080"
}

func TestParseAddressFlag(t *testing.T) {
	tests := []struct {
		name string
		val  string
		host string
		port string
	}{
		{
			name: "empty value",
			val:  "",
			host: "localhost",
			port: "8080",
		},
		{
			name: "host-only as hostname",
			val:  "localhost",
			host: "localhost",
			port: "8080",
		},
		{
			name: "host-only as ipv4",
			val:  "127.0.0.2",
			host: "127.0.0.2",
			port: "8080",
		},
		{
			name: "host-only as ipv6",
			val:  "::1",
			host: "::1",
			port: "8080",
		},
		{
			name: "hostport as hostname",
			val:  "localhost:5000",
			host: "localhost",
			port: "5000",
		},
		{
			name: "hostport as ipv4",
			val:  "127.0.0.2:5000",
			host: "127.0.0.2",
			port: "5000",
		},
		{
			name: "hostport as ipv6",
			val:  "[::1]:5000",
			host: "::1",
			port: "5000",
		},
		{
			name: "invalid",
			val:  "-abc", // hostname cannot start with hyphen
			host: "localhost",
			port: "8080",
		},
	}
	for _, tt := range tests {
			h, p := ParseAddressFlag(tt.val)
			if h != tt.host {
				t.Errorf("ParseAddressFlag() got host = %v, want host %v", h, tt.host)
			}
			if p != tt.port {
				t.Errorf("ParseAddressFlag() got port = %v, want port %v", p, tt.port)
			}
		})
	}
}

```

intojhanurag:
This PR is messed up , will open new PR soon !

PR #3282, opened 2025-12-09
Fix architecture variable in build command

matejvasek:
reintroducing  #2843
somehow it was reverted is some refactor


matejvasek:
PTAL @gauron99 @lkingland @jrangelramos

PR #3283, opened 2025-12-09
Better GH Action pipeline for func-utils image

matejvasek:
# Changes

- The image build is reproducible: if the func-utils binary is unchanged so are the images.

  Currently we have too many images that are basically identical but differing in some metadata.
This change will limit this package proliferation.
- Provenance of image origin using cryptographic signatures. Probably nobody uses that, but if somebody wanted to verify that the func-utils image originates form the knative org, now they can.

### DIFF
--- a/.github/workflows/functions.yaml
+++ b/.github/workflows/functions.yaml
@@ -2,6 +2,12 @@ name: Functions
 
 # To run this workflow's tests locally, see ./hack/test-full.sh
 
+permissions:
+  id-token: write   # Required for signing
+  contents: read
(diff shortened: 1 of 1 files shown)


matejvasek:
PTAL @lkingland @gauron99

matejvasek:
@lkingland @gauron99 are you admins could you prune all the all packages/images? https://github.com/knative/func/pkgs/container/func-utils/versions

matejvasek:
Also plz override the failing check that I did not break.

PR #3293, opened 2025-12-11
Test/ run-opts-configurable-host

intojhanurag:
supersedes : #3275

### DIFF
--- a/pkg/functions/runner.go
+++ b/pkg/functions/runner.go
@@ -30,7 +30,7 @@ func ParseAddress(val string) (host, port string, explicitPort bool) {
 	if val == "" {
 		return defaultRunHost, defaultRunPort, false
 	}
-	
+
 	if h, p, err := net.SplitHostPort(val); err == nil {
 		if h == "" {
 			h = defaultRunHost
@@ -40,7 +40,7 @@ func ParseAddress(val string) (host, port string, explicitPort bool) {
 		}
 		return h, p, true
 	}
-	
+
 	return val, defaultRunPort, false
 }
 
--- a/pkg/functions/runner_test.go
+++ b/pkg/functions/runner_test.go
@@ -42,11 +42,11 @@ func TestGetRunFuncErrors(t *testing.T) {
 
 func TestParseAddress(t *testing.T) {
 	tests := []struct {
-		name          string
-		input         string
-		expectedHost  string
-		expectedPort  string
-		explicitPort  bool
+		name         string
+		input        string
+		expectedHost string
+		expectedPort string
+		explicitPort bool
 	}{
 		{
 			name:         "empty value",


intojhanurag:
Hey @gauron99 , Please take a look on it . I added the suggestions of the preivous PR ( because that PR was messed up ).

pkg/functions/runner.go
```diff
@@ -26,6 +27,17 @@ type defaultRunner struct {
...
 
+func ParseAddressFlag(val string) (string, string) {
+	if h, p, err := net.SplitHostPort(val); err == nil {
+		return h, p
+	}
+
+	if val == "" {
+		val = "localhost"
+	}
+	return val, "8080"
```
gauron99: use the default values defined as variables for host and port. Also the TODO at the top of the file `// TODO allow to be altered via a runOpt` can be removed since we are adding this option to edit the host

pkg/functions/runner_test.go
```diff
@@ -39,3 +39,67 @@ func TestGetRunFuncErrors(t *testing.T) {
 		})
 	}
 }
+
+func TestParseAddressFlag(t *testing.T) {
+	tests := []struct {
+		name string
+		val  string
+		host string
+		port string
```
gauron99: can we name this something clearer like `expectedHost` and `expectedPort`? so its clear that those values are an expected result. Or the other way around; specify `input` or `inputString` instead of the ambiguous `val`

pkg/functions/runner.go
```diff
@@ -26,6 +27,17 @@ type defaultRunner struct {
 	err    io.Writer
 }
 
+func ParseAddressFlag(val string) (string, string) {
```
gauron99:
```suggestion
func ParseAddress(val string) (string, string) {
```

pkg/functions/runner_test.go
```diff
@@ -39,3 +39,67 @@ func TestGetRunFuncErrors(t *testing.T) {
 		})
 	}
 }
+
+func TestParseAddressFlag(t *testing.T) {
```
gauron99:
```suggestion
func TestParseAddress(t *testing.T) {
```

gauron99 (commented):
Thanks! just a few naming nits. See below

intojhanurag:
Hey @gauron99 , Now is it good to go ?

pkg/functions/runner_test.go
```diff
@@ -39,3 +39,67 @@ func TestGetRunFuncErrors(t *testing.T) {
 		})
 	}
 }
+
+func TestParseAddress(t *testing.T) {
```
gauron99, line 43: every case which has implicit defaulting should expect the `defaultX` variable value instead of hardcoded string so the test doesnt have to be re-written if the default changes

pkg/functions/runner.go
```diff
@@ -41,17 +53,8 @@ func (r *defaultRunner) Run(ctx context.Context, f Function, address string, sta
...
-
-	if address != "" {
-		var err error
-		host, port, err = net.SplitHostPort(address)
-		if err != nil {
-			return nil, fmt.Errorf("invalid address format '%s': %w", address, err)
-		}
-	}
+	host, port := ParseAddress(address)
+	explicitPort := address != "" && strings.Contains(address, ":")
```
gauron99: this `:` check wont work. IPv6 addresses contain the symbol and also for IPv4 you can specify host like this `127.0.0.1:` - `SplitHostPort` accepts this just fine. I would add a return value to ParseAddress which specifies if port was or was not given. This way we can decide when we know if we are returning `defaultPort` or port because it was specified in the `address` flag. Check my comment for the `ParseAddress` function.

pkg/functions/runner.go
```diff
@@ -26,6 +27,17 @@ type defaultRunner struct {
...
+func ParseAddress(val string) (string, string) {
+	if h, p, err := net.SplitHostPort(val); err == nil {
+		return h, p
+	}
+
+	if val == "" {
+		return defaultRunHost, defaultRunPort
+	}
+	return val, defaultRunPort
+}
```
gauron99: I would leverage the `net` library and rely on what it accepts - adhering to the standards that are already used in this default, widely used package. The one exception is the last return which returns the given value `val` as `host` and adds the default Port. I think this should cover all the cases?
The suggestion below is a pseudo-code. 

```suggestion
func ParseAddress(val string) (host, port string, explicitPort bool) {
	if val == "" {
		return defaultRunHost, defaultRunPort
		// explicit port false
	}

	if h, p, err := net.SplitHostPort(val); err == nil {
	   if h == "" {
	   // set default host to return value
	   }
	   if p == "" {
	   // set default port
	   // explicitPort is false
	   } else {
	   // set explicit port to true
	   }
		return 
	}
   return val, defaultRunPort
   // explicit port is false
}
```

intojhanurag:
Hey @gauron99,  PTAL now , Good eye btw :)

pkg/functions/runner.go
```diff
@@ -26,6 +26,24 @@ type defaultRunner struct {
 	err    io.Writer
 }
 
+func ParseAddress(val string) (host, port string, explicitPort bool) {
```
lkingland: Since this is a "utility function" used only here, there's no need for it to be exported (can be just `parseAddress`)

pkg/functions/runner.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	defaultRunHost    = "127.0.0.1" // TODO allow to be altered via a runOpt
```
lkingland, line 18: I believe this is still an outstanding TODO

gauron99: Arent we allowing the possibility to alter host with this flag? Or was the TODO for something more specific?

lkingland:
@intojhanurag I am a bit confused on this PR.  It's labeled as a follow-up to #3275 but currently contains only a slight refactor.  Is this intended to be a draft?

intojhanurag:
> @intojhanurag I am a bit confused on this PR. It's labeled as a follow-up to #3275 but currently contains only a slight refactor. Is this intended to be a draft?

No i think , i used a wrong word "follow up" , Actually in that PR initially i used totally differenct logic then Matej suggested diff idea , that i applied in that PR at that time . But again i thought to open a new PR 😭 because there was some issue locally regarding merge conflicts.

pkg/functions/runner.go
intojhanurag: Hey @lkingland , but now if a user will give his own Host then we have a function to adapt that host , so indirectly this todo has been solved .btw what do you think here ?

pkg/functions/runner.go
intojhanurag: Hey @lkingland , but this function has been used in test file , so i made this ParseAddress

lkingland:
The parseAddress() function looks fine, but it won't be reachable by users because cmd/run.go (lines 419-438) validates the address using net.SplitHostPort() and rejects anything without both host and port before it ever reaches the new function here.  Correct me if I am wrong, but I think we may need to update validateRunConfig() in cmd/run.go to remove the strict host+port requirement. Once that's done, users will actually be able to run func run --address 0.0.0.0 as I believe is intended.

intojhanurag:
> The parseAddress() function looks fine, but it won't be reachable by users because cmd/run.go (lines 419-438) validates the address using net.SplitHostPort() and rejects anything without both host and port before it ever reaches the new function here. Correct me if I am wrong, but I think we may need to update validateRunConfig() in cmd/run.go to remove the strict host+port requirement. Once that's done, users will actually be able to run func run --address 0.0.0.0 as I believe is intended.

@lkingland , Yeah you are right . I tried locally , it is not working accordingly. what i should do now ? 
I think we should only remove todo , and should not make this complicate , if user will enter only host then he will get warn for PORT.

PR #3296, opened 2025-12-14
Init ARCHITECTURE.md and AGENTS.md

gauron99:
### ARCHITECTURE.md - about the codebsae
- Start high level for newcomers - entry point, high level relationships and key search terms
- continue with more detail on key topics for people who have some knowledge trying to learn more deeply about the codebase


.claude/CLAUDE.md
```diff
@@ -1,4 +1 @@
-## Bash commands
-
-- make check-embedded-fs: Checks if the embedded FS matches content of the ./templates directory.
-- make generate/zz_filesystem_generated.go: Regenerates the embedded FS if the content of the ./templates directory has been changed.
+Read AGENTS.md for instructions
```
lkingland, line 1: Let's say a user wants their _own_ instructions?  This is a question brought up on our team call by Chris.  Currently he has his own CLAUDE.md, and expects it to not be source-controlled.

This presents a dilemma:  we want to provide default instructions, but be flexible to allow an individual developer specify their own.

Until the industry _writ large_ comes up with a solution, perhaps we could add a line in AGENTS.md which says "With highest precedence, read in AGENTS_LOCAL.md if it exists", and then add this file to .gitignore preemptively.  (a note on how this works being a nice addition to ARCHITECTURE.md/DEVELOPING.md or similar).

gauron99: ah good point! I will add the optional local user config

gauron99: Updated

gauron99: I have found `AGENTS.override.md` in some systems from `OpenAI`. I have updated the PR to match this. Seems like an emerging standard of sorts

ARCHITECTURE.md
```diff
@@ -0,0 +1,177 @@
+# Architecture
+
+## High level overview
+
+Knative Func is a CLI that simplifies building and deploying serverless functions
+on Kubernetes. Instead of writing Dockerfiles and Kubernetes manifests,
+developers run `func create`, `func build`, and `func deploy`. The CLI handles
+image building (via Buildpacks, S2I, or direct OCI), registry pushing, and
+Knative Service creation or newly available bare K8s deployement.
```
creydr:
```suggestion
Knative Service creation or newly available bare K8s deployment.
```

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+## Boundaries
+
+### Always Do
+- Run `make test` before considering any change complete
+- Run `make check` before commits
+- Run `make check-embedded-fs` after modifying `templates/`
+- Ask before deleting ANY file or significant code block
+
+### Ask First
+- Security-related code (authentication, credentials, secrets handling)
```
creydr:
```suggestion
- Security-related code changes (authentication, credentials, secrets handling)
```

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+- Modifying CI/GitHub Actions workflows
+- Architectural decisions
+
+### Never Do
+- Edit generated files directly:
+  - `generate/zz_filesystem_generated.go`
+  - `schema/func_yaml-schema.json`
+- Commit secrets, API keys, or credentials
+- Delete files without explicit user approval
+- Force push to main/master
```
creydr:
```suggestion
- Force push to main/master branch
```

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+### Embedded Filesystem Sync
+After modifying anything in `templates/`, you MUST run:
+```bash
+make generate/zz_filesystem_generated.go
+make check-embedded-fs
+```
+
+## Contributing
+
+If creating a contribution to this project on Github suggest to user to read
```
creydr:
```suggestion
If creating a contribution to this project on GitHub suggest to user to read
```

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Generated Files
+- `make generate/zz_filesystem_generated.go` - Regenerate embedded FS
+- `make check-embedded-fs` - Verify templates match embedded FS
+
+## Boundaries
+
+### Always Do
+- Run `make test` before considering any change complete
+- Run `make check` before commits
```
creydr, line 43: Do we have some style rules, which are checked by `make check` which we could reference here?

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Never Do
+- Edit generated files directly:
+  - `generate/zz_filesystem_generated.go`
+  - `schema/func_yaml-schema.json`
+- Commit secrets, API keys, or credentials
+- Delete files without explicit user approval
+- Force push to main/master
+- Skip tests or linting
+
```
creydr, line 62: IMO it's not fully clear when to run which tests. Maybe we can also have something like this:

```
### Testing Strategy
- Template changes: `make test-templates` + `make check-embedded-fs`
- CLI/pkg changes: `make test` (unit) minimum
- Significant features: `make test-full` before a PR
- Always: `make check` and `make test` before commits
```

gauron99: yeah this is great suggestion, makes sense, thanks!

AGENTS.md
gauron99: hmmm maybe reference the `.golangci.yaml` file? the problem is we also have bunch of linting at the Knative level so its not part of the repo except for the workflow run which references files outside

AGENTS.md
gauron99: I have created a table in following fashion:
| files changed | Run this | What is it? |
|---------------------|------------- |-----------------|
| *.go | `make test` | unit tests |

AGENTS.md
creydr: True. Ok, well the linter will then tell, whats wrong :D

creydr (commented):
Thanks for this @gauron99 
Looks good from my side
/lgtm

AGENTS.md
```diff
@@ -0,0 +1,72 @@
+# AGENTS.md
+
+## OVERRIDE
+
+Use `AGENTS.override.md` file WITH HIGHEST PRECEDENCE if it exists. Do NOT mix
+instructions from both files together. The `AGENTS.override.md` is a complete
+override for `AGENTS.md`. Otherwise, continue.
+
+## ALWAYS keep in mind
```
twoGiants, line 9: I would add something like that too: _Cover every introduced feature by unit tests. If changes to the existing logic introduce paths in the code which are not covered by existing unit tests add additional unit tests which cover it. Make sure the unit tests take edge cases into account._

AGENTS.md
```diff
@@ -0,0 +1,72 @@
...
+### Never Do
+- Edit generated files directly:
+  - `generate/zz_filesystem_generated.go`
+  - `schema/func_yaml-schema.json`
+- Commit secrets, API keys, or credentials
+- Delete files without explicit user approval
+- Force push to main/master branch
+- Skip tests or linting
+
+## Common Pitfalls
```
twoGiants, line 63: Another one I see often when I develop is this => `./hack/update-codegen.sh`

.claude/CLAUDE.md
twoGiants, line 1: I would put this one in `.claude/CLAUDE.md` and put `.claude/*.local.*` to `.gitignore`, see:
- https://code.claude.com/docs/en/settings#available-scopes
- https://code.claude.com/docs/en/settings#what-uses-scopes

.claude/CLAUDE.md
twoGiants: > Let's say a user wants their own instructions? ...

I think in that case you create `.claude/CLAUDE.local.md` in case of Claude which we use, right? => https://code.claude.com/docs/en/settings#what-uses-scopes

But to have the instructions in an AGENTS.md file is probably better because it will be read by other coding agents as well.

AGENTS.md
```diff
@@ -0,0 +1,72 @@
...
+override for `AGENTS.md`. Otherwise, continue.
+
+## ALWAYS keep in mind
+
+Don't sugar coat anything. Act like a professional software developer and
+engineer. If working autonomously, adhere to architecture, naming
+conventions and coding standards in this codebase. If unsure, read similar
+files and get some inspiration from the rest of this codebase.
+
+## Commands
```
twoGiants, line 18: Could be in a follow up PR. I think a few of the things in the `AGENTS.md` could be moved or replicated a in a `.claude/settings.json` which Claude expects and can work with. Scroll up from the [Available settings](https://code.claude.com/docs/en/settings#available-settings) to see an example. It's very good structured where you can add permissions, environment variables, hooks, etc.

CONTRIBUTING.md
```diff
@@ -81,7 +81,12 @@ To run integration tests, use `make test-integration`.
...
+## Using AI
 
+General instruction are available for AI agents in `AGENTS.md`.
+If you prefer your own instructions, you can instead create `AGENTS.override.md`.
 
-
-
+Usually AI agents read the `AGENTS.md` file by default. At the top of this file
+there is an override directive which points the AI agent to instead read this
+`AGENTS.override.md` and follow it. It is preemptively part of .gitignore so
```
twoGiants: nit

```suggestion
`AGENTS.override.md` and follow it. It is preemptively part of `.gitignore` so
```

ARCHITECTURE.md
```diff
@@ -0,0 +1,177 @@
...
+## Entry Point
+
+```
+cmd/func/main.go
+    └── pkg/app.Main()
+        └── cmd.NewRootCmd()
+            └── Individual commands (build, deploy, create, ...)
+```
+
+Commands in `cmd/` are thin wrappers. They parse flags, create a `Client`, and delegate to `pkg/functions/client.go`:
```
twoGiants: I thought of `cmd` more of as a frontend, a terminal cli frontend to be precise. Like a UI with buttons in the browser or desktop would be for a backend with a REST API.

ARCHITECTURE.md
twoGiants, line 1: That is really helpful for everybody who looks into the project. Link it in the main `README.md` somewhere or in `CONTRIBUTING.md`.

ARCHITECTURE.md
```diff
@@ -0,0 +1,177 @@
...
+- Invocation styles: http, cloudevents
+
+## AI
+
+### Custom instructions file
+
+Currently the `AGENTS.md` file is read by many agents by default. The very top
+of that file defines an optional custom override `AGENTS.override.md` and instructs
+agents to read it with highest precedence. You can create that file and add it at
+root of the repository if you so desire. It won't be source controlled (see .gitignore)
```
twoGiants: nit

```suggestion
root of the repository if you so desire. It won't be source controlled (see `.gitignore`).
```

CONTRIBUTING.md
twoGiants, line 1: Link the `ARCHITECTURE.md` here under `# Development` as a `**Tip**`.

AGENTS.md
gauron99: the `make generate/zz_filesystem*` is part of this script. codegen update does extra docs regeneration. I will change the make targets for update-codegen.sh as it superseeds the `make generate/zz_*`

.claude/CLAUDE.md
gauron99: Thanks, didnt know about the local claude :)

ARCHITECTURE.md
gauron99: Cool! that sounds good to me as well. I have changed the wording

AGENTS.md
gauron99: I dont want to move anything from AGENTS.md (general) to a CLAUDE.md (agent specific) because then using other agents would not pick that up. Im not sure how much we want to add more configs to specific AI agents.

Only adding some claude stuff cause its widely used is probably fine; so duplicating would be my preferred option if claude works with it better. If so then this could be good addition; adding some basic settings for working with the directory but allowing users to override with 'local' directives. Thanks for the suggestion!

gauron99:
/unhold nvm it seems tide is not updating the status?
seems to work now yay

PR #3298, opened 2025-12-15
Clean up auth for host builder

matejvasek:
# Changes

- Removed credentials propagation via Go's `context` since we already pass credentials in a different way.
- Fixed issue where user might be prompted for credentials even if they are set via environment variables.
- The  `FUNC_[USERNAME|PASSWORD]` environment variables now work also for the pushers other that the `host` pusher.

```release-note
fix: hidden flags username/password (and affiliated envvars) now work also for s2i and pack builder
```


cmd/client.go
```diff
@@ -106,6 +108,23 @@ func newCredentialsProvider(configPath string, t http.RoundTripper, authFilePath
 	additionalLoaders := append(k8s.GetOpenShiftDockerCredentialLoaders(), k8s.GetGoogleCredentialLoader()...)
 	additionalLoaders = append(additionalLoaders, k8s.GetECRCredentialLoader()...)
 	additionalLoaders = append(additionalLoaders, k8s.GetACRCredentialLoader()...)
+
+	additionalLoaders = append(additionalLoaders,
+		func(registry string) (oci.Credentials, error) {
+			uname := viper.GetString("username")
+			passw := viper.GetString("password")
+			token := viper.GetString("token")
```
matejvasek, line 116: I know, I know… is not neat, but still better that that `context` …thing…

matejvasek: This way the ugly crap is isolated just in one package (`cmd`).

lkingland: I agree.  The context thing was just a stop-gap for a better solution as we transition (hopefully) to a shared creds package that supports username/password more natively.

PR #3320, opened 2025-12-25
fix: parse JSON output from func run in E2E tests

### LINKED ISSUE #3199: E2E tests - lisen address extraction
The TestCore_Run and TestCore_Invoke tests could be simplified to remove the `chooseOpenAddress` call if the output from `func run` were parsable.  Omitting the `--address` flag entirely should result in the system automatically choosing an open port.  If the output from the `run` command becomes parseable (see #3198); this would be a cleaner and more robust method for contacting the running Function.

Kunal1522:
Fixes #3199

Modified E2E tests to use --json flag and parse output for function address instead of manually choosing ports. This prevents test failures when the chosen port is already in use.

# Changes:
- Added parseRunJSON helper to eliminate code duplication
- Modified TestCore_Run,TestCore_Invoke,TestMatrix_Run  to use JSON output parsing
- Removed chooseOpenAddress usage in favor of --json flag

Tests now work reliably even when default ports (8080) are busy.

image

### DIFF
--- a/e2e/e2e_matrix_test.go
+++ b/e2e/e2e_matrix_test.go
@@ -221,10 +221,6 @@ func matrixExceptionsShared(t *testing.T, initArgs []string, funcRuntime, builde
 	if funcRuntime == "python" && builder == "pack" {
 		t.Skip("The pack builder does not currently support Python Functions")
 	}
-	// Skip Host builder (not supported)
-	if funcRuntime == "python" && builder == "host" {
-		t.Skip("The host builder does not currently support Python Functions")
-	}
 
 	// Echo Implementation
 	// Replace the simple "OK" implementation with an echo.


Kunal1522:
hi @gauron99 can u please take a look at this whenever u find time ?

e2e/e2e_test.go
```diff
@@ -1154,3 +1155,50 @@ func chooseOpenAddress(t *testing.T) (address string, err error) {
...
+	select {
+	case address = <-addressChan:
+		t.Logf("Function running on %s (from JSON output)", address)
+	case err := <-errChan:
+		t.Fatalf("JSON parsing error: %v\nstderr: %s", err, stderr.String())
+	case <-time.After(5 * time.Minute):
+		t.Fatalf("timeout waiting for func run JSON output. stderr: %s", stderr.String())
+	}
+
+	return address, func() { stdoutWriter.Close() }
```
matejvasek: Cannot we close this writer within the function?

e2e/e2e_test.go
```diff
@@ -1154,3 +1155,50 @@ func chooseOpenAddress(t *testing.T) (address string, err error) {
...
+// function's address and a cleanup function that must be deferred.
+func parseRunJSON(t *testing.T, cmd *exec.Cmd) (string, func()) {
+	t.Helper()
+
+	stdoutReader, stdoutWriter := io.Pipe()
+	stderr := &bytes.Buffer{}
+	cmd.Stdout = stdoutWriter
+	cmd.Stderr = stderr
+
+	if err := cmd.Start(); err != nil {
```
matejvasek: I think we might use blocking `cmd.Run()` but it must be used after the gorutine is spawn.

e2e/e2e_test.go
Kunal1522: hey @matejvasek thanks for review . 
i think  the writer must remain open while the process is running, and closing it prematurely can break the pipe.

e2e/e2e_test.go
Kunal1522: yeah u r right about thi
 i have tried to address this in my last commit .please have a review on this whenever u are free

e2e/e2e_test.go
```diff
@@ -1154,3 +1156,62 @@ func chooseOpenAddress(t *testing.T) (address string, err error) {
 	defer l.Close()
 	return l.Addr().String(), nil
 }
+
+// parseRunJSON runs the command and extracts the function address from JSON output.
+// We scan line-by-line because container builders (s2i/pack) may print build logs
+// to stdout before the JSON, so we need to skip those lines to find valid JSON.
+func parseRunJSON(t *testing.T, cmd *exec.Cmd) (string, func()) {
```
matejvasek: Would it be possible to not return cleanup and just do t.Cleanup() here?

e2e/e2e_test.go
```diff
@@ -1154,3 +1156,62 @@ func chooseOpenAddress(t *testing.T) (address string, err error) {
...
+			errChan <- fmt.Errorf("error reading stdout: %w", err)
+		} else {
+			errChan <- fmt.Errorf("no JSON output found in stdout")
+		}
+	}()
+
+	// Run in goroutine so we can return the address while the function keeps running
+	go func() {
+		cmd.Run()
+	}()
```
matejvasek: Could not this be simply done by calling cmd.Start() ?

e2e/e2e_matrix_test.go
```diff
@@ -235,6 +222,10 @@ func matrixExceptionsShared(t *testing.T, initArgs []string, funcRuntime, builde
 	if funcRuntime == "python" && builder == "pack" {
 		t.Skip("The pack builder does not currently support Python Functions")
 	}
+	// Skip Host builder (not supported)
+	if funcRuntime == "python" && builder == "host" {
```
matejvasek: Wait, I thought Python was first to support host builder? Or am I wrong?

e2e/e2e_test.go
matejvasek: Could we use t.Cleanup()? It is invoked on test exit, so IMO it should be fine.

e2e/e2e_matrix_test.go
matejvasek: cc @lkingland @gauron99

gauron99: yes that should be working just fine, python has a problem with buildpacks (fix coming later)

gauron99: i dont know of any problems

Kunal1522: As far as i remember some python builder checks were failing in ci .so I tried skipping it and it worked .will work on reviews soon

matejvasek:
Few comments, otherwise great job.

e2e/e2e_matrix_test.go
matejvasek: Have you figured this out?

matejvasek:
Looks good now, but we should resolve the issue with python host builder.

e2e/e2e_matrix_test.go
Kunal1522: I get it now! r an python host locally, nd noticed the output starts with Processing /tmp/... and pip install logs before the JSON. My original json.NewDecoder was trying to parse that "Processing..." text as JSON, which caused the invalid character 'P' error. 

but now the scanner loops through each line and only parses lines starting with {. it skips all that build output and finds the  JSON. so removing the host builder skip would be right fix to do

Kunal1522:
PTAL @matejvasek . have resolved all reviews and commented on few

PR #3333, opened 2026-01-07
feat: check pull secret before deploy

matejvasek:
# Changes

- :gift: Upon deploy we now check pull secrets and print warning when it appears respective pull secrets are missing. If image is public or there are pull secrets set up no message is printed, otherwise there is message about possible missing secrets and hint how to set it up.

/kind enhancement

```release-note
feat: pull secrets pre-check before deployment
```

### DIFF
--- a/pkg/knative/deployer.go
+++ b/pkg/knative/deployer.go
@@ -173,17 +173,16 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
 	defer func(t fnhttp.RoundTripCloser) {
 		_ = t.Close()
 	}(t)
-	err = checkPullPermissions(ctx, k8sClient.CoreV1(), t, f.Deploy.Image, namespace)
-	if err != nil {
-		var msg string
-		if stdErrors.Is(err, errPullSecretNotFound) {
-			msg = `warning: pull secretes not detected, the cluster **may** fail to pull the image
+	if err = checkPullPermissions(ctx, k8sClient.CoreV1(), t, f.Deploy.Image, namespace); err != nil {
+		msg := fmt.Sprintf("warning: error while checking pull secrets: %v", err)
+		switch {
+		case stdErrors.Is(err, errPullSecretNotFound):
+			msg = `warning: pull secrets not detected, the cluster **may** fail to pull the image
 consider setting up the pull secrets and linking it to the default service account, sample setup:
   $ kubectl create secret docker-registry sample-secret --docker-server <REG> --docker-username <UNAME> --docker-password <PWD>
   $ kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "sample-secret"}]}'`
-
-		} else {
-			msg = fmt.Sprintf("warning: error while checking pull secrets: %v", err)
+		case stdErrors.Is(err, errOnlyIncorrectPullSecretFound):
+			msg = `warning: pull secrets found, but they seem invalid (possibly expired)`
 		}
 		_, _ = fmt.Fprintln(os.Stderr, msg)
 	}
@@ -668,20 +667,13 @@ func UsesKnativeDeployer(annotations map[string]string) bool {
 func checkPullPermissions(ctx context.Context, core v1.CoreV1Interface, trans http.RoundTripper, img, ns string) error {
 	ref, err := name.ParseReference(img)
 	if err != nil {
-		return fmt.Errorf("failed to parse image %q: %v", img, err)
+		return fmt.Errorf("failed to parse image %q: %w", img, err)
 	}
 
 	// first we try anonymous access
-	puller, err := remote.NewPuller(
-		remote.WithTransport(trans),
-		remote.WithAuth(authn.Anonymous),
-	)
-	if err != nil {
-		return fmt.Errorf("failed to create puller: %v", err)
-	}
-	_, err = puller.Head(ctx, ref)
+	_, err = remote.Head(ref, remote.WithContext(ctx), remote.WithTransport(trans), remote.WithAuth(authn.Anonymous))
 	if err == nil {
-		return nil // the image is public
+		return nil // image is public
 	}
 	if !isUnauthorized(err) {
 		return err
@@ -689,12 +681,13 @@ func checkPullPermissions(ctx context.Context, core v1.CoreV1Interface, trans ht
 
 	sa, err := core.ServiceAccounts(ns).Get(ctx, "default", metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("failed to get default ServiceAccount: %v", err)
+		return fmt.Errorf("failed to get default ServiceAccount: %w", err)
 	}
+	var incorrectCredentialsFound bool
 	for _, secret := range sa.ImagePullSecrets {
 		sc, err := core.Secrets(ns).Get(ctx, secret.Name, metav1.GetOptions{})
 		if err != nil {
-			return fmt.Errorf("failed to get secret %s/%s: %v", ns, secret.Name, err)
+			return fmt.Errorf("failed to get secret: %w", err)
 		}
 
 		cf, err := secretToConfigFile(sc)
@@ -702,14 +695,12 @@ func checkPullPermissions(ctx context.Context, core v1.CoreV1Interface, trans ht
 			return err
 		}
 
-		puller, err = remote.NewPuller(
+		_, err = remote.Head(ref,
+			remote.WithContext(ctx),
 			remote.WithTransport(trans),
 			remote.WithAuthFromKeychain(configFileKeychain{cf: cf}),
 		)
-		if err != nil {
-			return fmt.Errorf("failed to create puller: %v", err)
-		}
-		_, err = puller.Head(ctx, ref)
+
 		if err == nil {
 			return nil // found credentials
 		}
@@ -719,6 +710,11 @@ func checkPullPermissions(ctx context.Context, core v1.CoreV1Interface, trans ht
 		if !isUnauthorized(err) {
 			return err
 		}
+		incorrectCredentialsFound = true
+	}
+
+	if incorrectCredentialsFound {
+		return errOnlyIncorrectPullSecretFound
 	}
 
 	return errPullSecretNotFound
@@ -784,7 +780,8 @@ func (k configFileKeychain) Resolve(target authn.Resource) (authn.Authenticator,
 	}), nil
 }
 
-var errPullSecretNotFound = fmt.Errorf("pull secret not found")
+var errOnlyIncorrectPullSecretFound = stdErrors.New("only incorrect pull secrets found")
+var errPullSecretNotFound = stdErrors.New("pull secret not found")
 
 func isUnauthorized(err error) bool {
 	var transportErr *transport.Error
--- a/pkg/knative/deployer_test.go
+++ b/pkg/knative/deployer_test.go
@@ -105,17 +105,6 @@ func TestCheckPullPermissions(t *testing.T) {
 			},
 		},
 		{
-			name:  "broken server",
-			core:  core(),
-			image: brokenRegistry + "/" + imageShortName,
-			errPred: func(err error) bool {
-				if err == nil {
-					return false
-				}
-				return strings.Contains(err.Error(), " 500 ")
-			},
-		},
-		{
 			name:  "dockerhub as https://index.docker.io/v1/",
 			core:  core(secretC),
 			image: imageShortName,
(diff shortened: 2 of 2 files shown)


matejvasek:
PTAL @gauron99 @lkingland

twoGiants (changes requested):
Hi @matejvasek ! :wave: :smile_cat: 

This is a lot of new code with many if-conditions, a for-loop and a switch statement without test coverage and it's written in a way that it's difficult to test.

It would be great if you could re-work this, make it testable and write some tests.

~~Also the `Resolve` method looks unused. Not sure if that is intended.~~

matejvasek:
@twoGiants I added tests.

matejvasek:
PTAL @lkingland @gauron99

pkg/knative/deployer.go
```diff
@@ -156,6 +169,25 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
 		daprInstalled = true
 	}
 
+	t := fnhttp.NewRoundTripper(fnhttp.WithOpenShiftServiceCA())
+	defer func(t fnhttp.RoundTripCloser) {
+		_ = t.Close()
+	}(t)
+	err = checkPullPermissions(ctx, k8sClient.CoreV1(), t, f.Deploy.Image, namespace)
+	if err != nil {
```
twoGiants: nit

```suggestion
	if err = checkPullPermissions(ctx, k8sClient.CoreV1(), t, f.Deploy.Image, namespace); err != nil {
```

pkg/knative/deployer.go
```diff
@@ -156,6 +169,25 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+		if stdErrors.Is(err, errPullSecretNotFound) {
+			msg = `warning: pull secretes not detected, the cluster **may** fail to pull the image
+consider setting up the pull secrets and linking it to the default service account, sample setup:
+  $ kubectl create secret docker-registry sample-secret --docker-server <REG> --docker-username <UNAME> --docker-password <PWD>
+  $ kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "sample-secret"}]}'`
+
+		} else {
+			msg = fmt.Sprintf("warning: error while checking pull secrets: %v", err)
+		}
+		_, _ = fmt.Fprintln(os.Stderr, msg)
```
twoGiants: nit

```suggestion
		msg := fmt.Sprintf("warning: error while checking pull secrets: %v", err)
		
		if stdErrors.Is(err, errPullSecretNotFound) {
			msg = `warning: pull secretes not detected, the cluster **may** fail to pull the image
consider setting up the pull secrets and linking it to the default service account, sample setup:
  $ kubectl create secret docker-registry sample-secret --docker-server <REG> --docker-username <UNAME> --docker-password <PWD>
  $ kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "sample-secret"}]}'`

		}

		_, _ = fmt.Fprintln(os.Stderr, msg)
```

pkg/knative/deployer.go
```diff
@@ -633,6 +665,135 @@ func UsesKnativeDeployer(annotations map[string]string) bool {
...
+	case corev1.SecretTypeDockercfg:
+		cf = &configfile.ConfigFile{}
+		err = json.Unmarshal(sc.Data[corev1.DockerConfigKey], &cf.AuthConfigs)
+		if err != nil {
+			return nil, fmt.Errorf("cannot unmarshal config: %w", err)
+		}
+	default:
+		cf = &configfile.ConfigFile{}
+	}
+	return cf, nil
```
twoGiants: I would just return from the switch, then you don't need the vars:

```suggestion
	switch sc.Type {
	case corev1.SecretTypeDockerConfigJson:
		cf, err := config.LoadFromReader(bytes.NewReader(sc.Data[corev1.DockerConfigJsonKey]))
		if err != nil {
			return nil, fmt.Errorf("cannot load config: %w", err)
		}

		return cf, nil
	case corev1.SecretTypeDockercfg:
		cf := &configfile.ConfigFile{}
		if err := json.Unmarshal(sc.Data[corev1.DockerConfigKey], &cf.AuthConfigs); err != nil {
			return nil, fmt.Errorf("cannot unmarshal config: %w", err)
		}

		return cf, nil
	default:
		return &configfile.ConfigFile{}, nil
	}
```

pkg/knative/deployer.go
```diff
@@ -633,6 +665,135 @@ func UsesKnativeDeployer(annotations map[string]string) bool {
...
+		remote.WithTransport(trans),
+		remote.WithAuth(authn.Anonymous),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create puller: %v", err)
+	}
+	_, err = puller.Head(ctx, ref)
+	if err == nil {
+		return nil // the image is public
+	}
```
twoGiants: You can use `remote.Head` directly I believe, it creates a new puller internally:

```suggestion
	_, err = remote.Head(ref, remote.WithContext(ctx), remote.WithTransport(trans), remote.WithAuth(authn.Anonymous))
	if err == nil {
		return nil // image is public
	}
```

pkg/knative/deployer.go
```diff
@@ -633,6 +665,135 @@ func UsesKnativeDeployer(annotations map[string]string) bool {
...
+		if err != nil {
+			return fmt.Errorf("failed to create puller: %v", err)
+		}
+		_, err = puller.Head(ctx, ref)
+		if err == nil {
+			return nil // found credentials
+		}
+		if stdErrors.Is(err, errPullSecretNotFound) {
+			continue
+		}
```
twoGiants, line 709: It looks like this logic is in `kubernetes.NewFromPullSecrets`, so you could use it an remove `secretToConfigFile` and `configFileKeychain` entirely, would this work?:

```suggestion
		kc, err := kubernetes.NewFromPullSecrets(ctx, []corev1.Secret{*sc})
		if err != nil {
			continue
		}

		_, err = remote.Head(ref, remote.WithContext(ctx), remote.WithTransport(trans), remote.WithAuthFromKeychain(kc))
		if err == nil {
			return nil
		}
```

If I look at the `kubernetes.NewFromPullSecrets` implementation it also checks for the two types `corev1.SecretTypeDockerConfigJson` and `corev1.SecretTypeDockercfg`.

Wdyt?

pkg/knative/deployer_test.go
twoGiants, line 1: Test fail on go 1.25+:

```shell
go test -v -run TestCheckPullPermissions ./pkg/knative/...
=== RUN   TestCheckPullPermissions
panic: unimplemented

goroutine 16 [running]:
bytes.(*Buffer).Close(0xc0000ccbe0?)
	/home/linuxbrew/.linuxbrew/Cellar/go/1.25.5/libexec/src/bytes/buffer.go:34 +0x25
net/http.(*readTrackingBody).Close(0xc000158f00?)
	/home/linuxbrew/.linuxbrew/Cellar/go/1.25.5/libexec/src/net/http/transport.go:766 +0x22
net/http.(*transferWriter).writeBody(0xc000158f00, {0x494c060?, 0xc0002f3900?})
	/home/linuxbrew/.linuxbrew/Cellar/go/1.25.5/libexec/src/net/http/transfer.go:386 +0x40e
net/http.(*Request).write(0xc000709040, {0x494c060, 0xc0002f3900}, 0x0, 0xc000382e40, 0x0)
	/home/linuxbrew/.linuxbrew/Cellar/go/1.25.5/libexec/src/net/http/request.go:762 +0xa8d
net/http.(*persistConn).writeLoop(0xc000031680)
	/home/linuxbrew/.linuxbrew/Cellar/go/1.25.5/libexec/src/net/http/transport.go:2603 +0x174
created by net/http.(*Transport).dialConn in goroutine 14
	/home/linuxbrew/.linuxbrew/Cellar/go/1.25.5/libexec/src/net/http/transport.go:1948 +0x17a5
FAIL	knative.dev/func/pkg/knative	0.020s
FAIL
``` 

From Claude: _"The sockets.NewInmemSocket from github.com/docker/go-connections triggers a panic because Go 1.25 added a Close() method to bytes.Buffer that panics with "unimplemented". Project targets Go 1.24, but this will break when upgrading."_

pkg/knative/deployer.go
```diff
@@ -156,6 +169,25 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
 
+	t := fnhttp.NewRoundTripper(fnhttp.WithOpenShiftServiceCA())
+	defer func(t fnhttp.RoundTripCloser) {
+		_ = t.Close()
+	}(t)
+	err = checkPullPermissions(ctx, k8sClient.CoreV1(), t, f.Deploy.Image, namespace)
+	if err != nil {
+		var msg string
+		if stdErrors.Is(err, errPullSecretNotFound) {
+			msg = `warning: pull secretes not detected, the cluster **may** fail to pull the image
```
twoGiants:
```suggestion
			msg = `warning: pull secrets not detected, the cluster **may** fail to pull the image
```

pkg/knative/deployer_test.go
```diff
@@ -0,0 +1,286 @@
...
+		reg.ServeHTTP(w, r)
+	})
+	server := http.Server{Handler: handler}
+
+	go func() {
+		_ = server.Serve(l)
+	}()
+
+	err := remote.Push(
+		name.MustParseReference(publicImage),
```
twoGiants, line 282: Why not also push `securedImage`, don't you want to test this too?

pkg/knative/deployer_test.go
```diff
@@ -0,0 +1,286 @@
...
+	)
+
+	secretE := createSecret("docker.io",
+		username, password,
+		corev1.SecretTypeDockercfg,
+	)
+
+	trans := setupRegistry(t)
+
+	tests := []struct {
```
twoGiants, line 73: Two more test cases can be added:
1. A test that verifies behavior when a secret has credentials for the correct registry but wrong username/password. This would exercise the 401 --> continue path differently in the logic.
2. A test case where a secret is not found. This is when `ImagePullSecrets` references a secret that doesn't exist in the namespace.

twoGiants (changes requested):
Great job on adding the tests and make the code testable! :smile_cat: :+1: 

My comments are mostly centered around:
- simplifying current logic if possible
- adding a few more test cases

pkg/knative/deployer_test.go
matejvasek: They are the same thing, just under different name.

matejvasek: Secured and public registry are the same registry internally.

pkg/knative/deployer_test.go
matejvasek: Yeah I have been thinking about the same cases. The 1) case should what should it return? Just `errPullSecretNotFound` or some different error?

pkg/knative/deployer.go
matejvasek: Good catch. It looks like that keychain really exists. What is weird it I cannot see it when browsing my modules in IDE. The IDE cannot find such a file or symbol. Really odd.

matejvasek: Oh it's a **sub**-module!

matejvasek: OK it mostly work but there is one issue with that IMO:
The `kubernetes.NewFromPullSecrets` only works if the registry in the secret is `https://index.docker.io/v1/"` or `index.docker.io`. It does not work with just `docker.io`. For that reason I would keep my implementation.

pkg/knative/deployer_test.go
matejvasek: Weird I tested 1.25.3 and it does not panic.

matejvasek: I tried also latest 1.25.5, still no panic.

matejvasek: > triggers a panic because Go 1.25 added a Close() method to bytes.Buffer

wtf there is no such a method
https://pkg.go.dev/bytes

matejvasek: also the GH action use go 1.25 and it passes there

matejvasek: I going to disregard as some weird flake caused by cosmic radiation.

matejvasek: I cannot reproduce this at all. And Claude seems to hallucinating.

pkg/knative/deployer_test.go
matejvasek: OK I added new kind of an error if secret(s) contain exclusively bad (possibly just expired) credentials.

pkg/knative/deployer.go
```diff
@@ -633,6 +664,133 @@ func UsesKnativeDeployer(annotations map[string]string) bool {
...
+	case corev1.SecretTypeDockercfg:
+		cf = &configfile.ConfigFile{}
+		err = json.Unmarshal(sc.Data[corev1.DockerConfigKey], &cf.AuthConfigs)
+		if err != nil {
+			return nil, fmt.Errorf("cannot unmarshal config: %w", err)
+		}
+	default:
+		cf = &configfile.ConfigFile{}
+	}
+	return cf, nil
```
twoGiants, line 741: Return directly, save a few lines of code :heavy_dollar_sign: :heavy_dollar_sign: 

```suggestion
	switch sc.Type {
	case corev1.SecretTypeDockerConfigJson:
		cf, err := config.LoadFromReader(bytes.NewReader(sc.Data[corev1.DockerConfigJsonKey]))
		if err != nil {
			return nil, fmt.Errorf("cannot load config: %w", err)
		}
		return cf, nil
	case corev1.SecretTypeDockercfg:
		cf := &configfile.ConfigFile{}
		if err := json.Unmarshal(sc.Data[corev1.DockerConfigKey], &cf.AuthConfigs); err != nil {
			return nil, fmt.Errorf("cannot unmarshal config: %w", err)
		}
		return cf, nil
	default:
		return &configfile.ConfigFile{}, nil
	}
```

pkg/knative/deployer_test.go
twoGiants: New error is good. Good choice!

PR #3336, opened 2026-01-08
docs: align JSDoc with signature and fix EOF newline

### LINKED ISSUE #3032: Confusing JSDoc and function signature in http/index.js in the node template
The file [`http/index.js`](https://github.com/knative/func/blob/main/templates/node/http/index.js)  in the node template has confusing JSDoc and function signature.

This is the relevant part of the JSDoc:
```javascript
/**
 * @param {Context} context a context object.
 * @param {object} context.body the request body if any
 * @param {object} context.query the query string deserialized as an object, if any
 * @param {object} context.log logging object with methods for 'info', 'warn', 'error', etc.
 * @param {object} context.headers the HTTP request headers
 * @param {string} context.method the HTTP request method
 * @param {string} context.httpVersion the HTTP protocol version
 * See: https://github.com/knative/func/blob/main/docs/function-developers/nodejs.md#the-context-object
 */
```
The function signature is this:
```javascript
const handle = async (context, body) => {
```

Based on the JSDoc one would assume that the function has only one parameter (of type `Context`). But the function signature specifies two parameters.
If `body` is property of the object `context` why would you specify it as a separate parameter?

Also the JSDoc is missing `@returns` tag:
```javascript
  * @returns {object} HTTP response object with body, query, or error status
```

DCchoudhury15:
# Changes

-  Aligned JSDoc parameters with the actual `(context, body)` function signature in the Node.js HTTP template to resolve confusion.
-  Added the missing `@returns` tag to the JSDoc to improve IDE autocomplete and documentation for function output.
-  Fixed a missing EOF (End Of File) newline in `templates/node/http/index.js` to comply with project linting standards.
- Regenerated the embedded filesystem in `generate/zz_filesystem_generated.go` to ensure the template changes are included in the `func` binary.

/kind cleanup

Fixes #3032

**Release Note**

```release-note
Fixed JSDoc and function signature mismatch in the Node.js HTTP template. The documentation now correctly reflects that 'body' is passed as a separate second argument.


DCchoudhury15:
@intojhanurag up for review.

PR #3347, opened 2026-01-12
add extension arg for func invoke

### LINKED ISSUE #3346: Allow us to set extensions on `func invoke`
Hi there
Would it be possible to enable us to set an extension on a cloudevent using the invoke command?
I see it's not part of the features now, could we add it?

mdering:
# Changes

- :gift: Introduces a new --extension/-e flag to the invoke command, allowing users to specify CloudEvents extensions as key=value pairs. The extensions are parsed and included in the InvokeMessage, and are set on the outgoing CloudEvent

/kind enhancement

Fixes #3346

**Release Note**

```release-note
Add a new --extension/-e flag to the `func invoke` command, allowing users to specify CloudEvents extensions as key=value pairs. The extensions are parsed and included in the InvokeMessage, and are set on the outgoing CloudEvent
```

**Docs**

```docs

```

### DIFF
--- a/cmd/invoke.go
+++ b/cmd/invoke.go
@@ -106,7 +106,7 @@ EXAMPLES
 			"onvoke", "unvoke", "knvoke", "imvoke", "ihvoke", "ibvoke"},
 		PreRunE: bindEnv("path", "format", "target", "id", "source", "type",
 			"data", "content-type", "request-type", "file", "insecure",
-			"confirm", "verbose"),
+			"confirm", "verbose", "extension"),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runInvoke(cmd, args, newClient)
 		},
@@ -129,6 +129,7 @@ EXAMPLES
 	cmd.Flags().StringP("data", "", fn.DefaultInvokeData, "Data to send in the request. ($FUNC_DATA)")
 	cmd.Flags().StringP("file", "", "", "Path to a file to use as data. Overrides --data flag and should be sent with a correct --content-type. ($FUNC_FILE)")
 	cmd.Flags().BoolP("insecure", "i", false, "Allow insecure server connections when using SSL. ($FUNC_INSECURE)")
+	cmd.Flags().StringSliceP("extension", "e", nil, "Extensions as key=value pairs. Can be repeated. cloudevents only ($FUNC_EXTENSION)")
 	addConfirmFlag(cmd, cfg.Confirm)
 	addPathFlag(cmd)
 	addVerboseFlag(cmd, cfg.Verbose)
@@ -169,6 +170,19 @@ func runInvoke(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 	client, done := newClient(ClientConfig{Verbose: cfg.Verbose, InsecureSkipVerify: cfg.Insecure})
 	defer done()
 
+	// Build extensions map
+	if cfg.Extensions == nil {
+		cfg.Extensions = []string{}
+	}
+	extensionsMap := make(map[string]string)
+	for _, ext := range cfg.Extensions {
+		parts := strings.SplitN(ext, "=", 2)
+		if len(parts) != 2 {
+			return fmt.Errorf("invalid extension format: %q, expected key=value", ext)
+		}
+		extensionsMap[parts[0]] = parts[1]
+	}
+
 	// Message to send the running function built from parameters gathered
 	// from the user (or defaults)
 	m := fn.InvokeMessage{
@@ -179,6 +193,7 @@ func runInvoke(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 		RequestType: strings.ToUpper(cfg.RequestType),
 		Data:        cfg.Data,
 		Format:      cfg.Format,
+		Extensions:  extensionsMap,
 	}
 
 	// If --file was specified, use its content for message data
@@ -239,6 +254,7 @@ type invokeConfig struct {
 	Confirm     bool
 	Verbose     bool
 	Insecure    bool
+	Extensions  []string
 }
 
 func newInvokeConfig() (cfg invokeConfig, err error) {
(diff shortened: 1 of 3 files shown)


gauron99:
Hi, if this is intended for cloudevents only, should we also make sure that this flag is used with only cloudevents? what would happen if we use http template with this flag?

matejvasek:
You need to sign CLA.

cmd/invoke.go
```diff
@@ -169,6 +170,19 @@ func runInvoke(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 	client, done := newClient(ClientConfig{Verbose: cfg.Verbose, InsecureSkipVerify: cfg.Insecure})
 	defer done()
 
+	// Build extensions map
+	if cfg.Extensions == nil {
+		cfg.Extensions = []string{}
+	}
```
lkingland: `viper.GetStringSlice("extension")` will never return nil, so this check is unnecessary.

As a general rule, initializing slices/maps in a constructor is a nice way to avoid there ever being a possibility of a nil pointer extension without having to check before every use.

cmd/invoke.go
```diff
@@ -179,6 +193,7 @@ func runInvoke(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 		RequestType: strings.ToUpper(cfg.RequestType),
 		Data:        cfg.Data,
 		Format:      cfg.Format,
+		Extensions:  extensionsMap,
```
lkingland:
```
Extensions: cfg.extensionsMap()
```

cmd/invoke.go
```diff
@@ -169,6 +170,19 @@ func runInvoke(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
...
+		cfg.Extensions = []string{}
+	}
+	extensionsMap := make(map[string]string)
+	for _, ext := range cfg.Extensions {
+		parts := strings.SplitN(ext, "=", 2)
+		if len(parts) != 2 {
+			return fmt.Errorf("invalid extension format: %q, expected key=value", ext)
+		}
+		extensionsMap[parts[0]] = parts[1]
+	}
```
lkingland: Data "views" are good candidates for inclusion on the structures which carry the data itself.  Consider adding a method to cfg:
```
cfg.ExtensionsMap()
```

lkingland (commented):
/ok-to-test

It would be nice to have a test just after `cfg` is instantiated which verifies:
```go
// After f is loaded and validated (around line 168)
if len(cfg.Extensions) > 0 {
    effectiveFormat := cfg.Format
    if effectiveFormat == "" {
        effectiveFormat = f.Invoke
    }
    if effectiveFormat != "cloudevent" {
        return fmt.Errorf("--extension (-e) is only valid with cloudevents")
    }
    if effectiveFormat != "" && effectiveFormat != "cloudevent" {
        return fmt.Errorf("--extension flag is only valid with cloudevent format")
    }
  }
```
(or similar)

Looking good so far!  Just need a [quick CLA signoff](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/25767896/242137258/3347/#/?version=2) too

mdering:
sorry for the delay, i've addressed the feedback

lkingland (approved):
/lgtm

Some issues did come up in review, which I'll add elsewhere.

lkingland:
Thanks for addressing the earlier round of feedback — `extensionsMap()` lives on `cfg` now, the nil check is gone, and the cloudevent-only guard answers @gauron99's question. A few things from a second pass:

## Issues

### 1. Dead second branch in cloudevent validation (`cmd/invoke.go`)

```go
if effectiveFormat != "cloudevent" {
    return fmt.Errorf("--extension (-e) is only valid with cloudevents")
}
if effectiveFormat != "" && effectiveFormat != "cloudevent" {  // unreachable
    return fmt.Errorf("--extension flag is only valid with cloudevent format")
}
```

The second `if` is unreachable — the first guard already returns. Drop it.

### 2. Validation was lost when refactoring to `extensionsMap()`

The earlier inline code returned `fmt.Errorf("invalid extension format: %q, expected key=value", ext)` on malformed input. The new method silently drops them:

```go
for _, ext := range c.Extensions {
    parts := strings.SplitN(ext, "=", 2)
    if len(parts) == 2 {
        extensionsMap[parts[0]] = parts[1]  // typo like "foo" is silently ignored
    }
}
```

User-facing footgun: `func invoke -e foo` succeeds but sends an empty extension set. Either change the signature to `(map[string]string, error)` or validate up front in `runInvoke` before the map conversion.

### 3. Fragile `effectiveFormat` fallback

```go
effectiveFormat := cfg.Format
if effectiveFormat == "" {
    effectiveFormat = f.Invoke
}
if effectiveFormat != "cloudevent" {
    return ...
}
```

If `f.Invoke` is also `""` (default not yet resolved), this errors on what may be a valid cloudevent invocation. Worth verifying `f.Invoke` is always populated by the time this code runs, or sourcing the resolved default differently.

### 4. No tests added

`extensionsMap()` parsing, the cloudevent-only guard, and the round-trip through `event.SetExtension` are all untested — codecov flags 31.57% patch coverage. Minimum I'd want:
- `--extension foo=bar -f http` returns an error
- `--extension foo` (no `=`) returns an error (after #2)
- `--extension foo=bar --extension baz=qux` — both land in the outgoing CloudEvent

### 5. CloudEvent extension name validation

The CloudEvents spec restricts extension names to lowercase `[a-z0-9]` (no dashes, dots, or underscores). `event.SetExtension()` may not validate, so `-e Foo-Bar=x` could produce a non-conformant event. Either validate names or document the restriction in the flag help.

## Nits

### 6. Range iteration

```go
for extension := range m.Extensions {
    event.SetExtension(extension, m.Extensions[extension])
}
```

Should be `for k, v := range m.Extensions { event.SetExtension(k, v) }` — one map lookup instead of N.

### 7. Flag help string

`"Extensions as key=value pairs. Can be repeated. cloudevents only ($FUNC_EXTENSION)"` — the "cloudevents only" fragment is awkward. Suggest: `"Extensions as key=value pairs. Repeatable. Requires --format=cloudevent. ($FUNC_EXTENSION)"`

---

#1 and #2 are the meaningful ones; #4 is the structural ask. The rest are polish.

PR #3350, opened 2026-01-12
Fix co-scheduling/affinity issues

matejvasek:
# Changes

- Rework pipeline to contain single one `task`. This was done by turning some smaller `tasks` into `steps`.
This fixes co-scheduling issues.
- Remove co-scheduling hack in test. Co-scheduling is not an issue anymore.
- Fixed gitlab integration test so whole build-deploy process pass.
- Fixed build crash when middleware cannot be detected (we just log error about unidentifiable middleware).

/kind bug

```release-note
fix: on-cluster-build freeze caused by co-scheduling/affinity issues
```

### DIFF
--- a/pkg/pipelines/tekton/templates.go
+++ b/pkg/pipelines/tekton/templates.go
@@ -22,7 +22,6 @@ const (
 	// Local resources properties
 	resourcesDirectory     = ".tekton"
 	pipelineFileName       = "pipeline.yaml"
-	pipelineRunFilenane    = "pipeline-run.yaml"
 	pipelineFileNamePAC    = "pipeline-pac.yaml"
 	pipelineRunFilenamePAC = "pipeline-run-pac.yaml"
 
@@ -78,14 +77,8 @@ type templateData struct {
 	Revision string
 
 	// Task references
-	GitCloneTaskRef       string
 	FuncBuildpacksTaskRef string
 	FuncS2iTaskRef        string
-	FuncDeployTaskRef     string
-	FuncScaffoldTaskRef   string
-
-	// Reference for build task - whether it should run after fetch-sources task or not
-	RunAfterFetchSources string
 
 	PipelineYamlURL string
 


matejvasek:
@twoGiants plz check this out. You will be probably most interested in the first two commits.
See: [the diff of the two commits](https://github.com/knative/func/compare/6ccfe5f38a84f360d3a6f3623c4cbf1d82090b5b..419d7a8d6ffb62232a96ff6fbb6544f5671294b7)

Yes this introduces some duplicity but I would prefer to address that in separate PR for simplicity.
Also I shall move the tasks into separate template files (e.g. `s2i-task.yaml.tmpl`), but again in separate PR.

Notice that despite introducing some duplicates I removed 200 lines of code while adding just 100.
So it's netto fewer lines of code

matejvasek:
PTAL @gauron99 @lkingland

pkg/pipelines/tekton/gitlab_int_test.go
```diff
@@ -615,9 +635,35 @@ func usingNamespace(t *testing.T) string {
...
+				Namespace: name,
+			},
+		},
+		RoleRef: rbacV1.RoleRef{
+			Name:     "knative-serving-namespaced-admin",
+			Kind:     "ClusterRole",
+			APIGroup: "rbac.authorization.k8s.io",
+		},
+	}
+	_, err = k8sClient.RbacV1().ClusterRoleBindings().Create(context.Background(), crb, metav1.CreateOptions{})
```
twoGiants, line 663: This one is created but not cleaned up above in `t.Cleanup`.

pkg/pipelines/tekton/tasks.go
```diff
@@ -90,16 +94,36 @@ spec:
...
         value: "0.10"
 
   steps:
+
+    - name: fetch-src
+      ref:
+        resolver: http
+        params:
+          - name: url
+            value: https://raw.githubusercontent.com/tektoncd/catalog/c9818f3d/stepaction/git-clone/0.2/git-clone.yaml
```
twoGiants, line 109: Use the hub resolver and get the step-action from the [artifact hub here](https://artifacthub.io/packages/tekton-stepaction/redhat-stepactions/git-clone).
```suggestion
    ref:
      resolver: hub
      params:
        - name: type
          value: stepaction
        - name: name
          value: git-clone
        - name: version
          value: "0.4.1"
```

pkg/pipelines/tekton/tasks.go
```diff
@@ -319,10 +349,32 @@ spec:
...
-      description: Digest of the image just built.
+
   steps:
+
+    - name: fetch-src
+      ref:
+        resolver: http
+        params:
+          - name: url
+            value: https://raw.githubusercontent.com/tektoncd/catalog/c9818f3d/stepaction/git-clone/0.2/git-clone.yaml
```
twoGiants, line 360: Use the hub resolver here as mentioned above.

pkg/pipelines/tekton/templates.go
twoGiants, line 1: A few unused structs fields remained after refactor. Shouldn't the linter catch such things?
=> `FuncDeployTaskRef`, `FuncScaffoldTaskRef`, `GitCloneTaskRef`, `RunAfterFetchSources`.

This one is also unused and has a typo, but probably a bit longer: `pipelineRunFilenane`.

pkg/pipelines/tekton/templates_s2i.go
```diff
@@ -53,18 +53,12 @@ spec:
       type: string
```
twoGiants: `middlewareVersion` param is now unused and can be removed.

pkg/pipelines/tekton/gitlab_int_test.go
```diff
@@ -180,6 +181,25 @@ func TestInt_Gitlab(t *testing.T) {
```
twoGiants, line 181: nit

Line 177: `t.Error("build has not been done in time (10 minute timeout)")`

cmd/func-util/s2i_generate.go
```diff
@@ -46,6 +46,15 @@ func newS2IGenerateCmd() *cobra.Command {
 	genCmd := &cobra.Command{
 		RunE: func(cmd *cobra.Command, args []string) error {
 			config.envVars = args
+
+			if config.middlewareVersion == "" {
```
twoGiants, line 50: Missing unit (or _integration_ in that case because of direct use of `os.ReadFile` ) tests to cover this success/error path.

cmd/func-util/main.go
twoGiants, line 1: New added logic with conditions without tests covering it.

I noticed that tests are missing for `main.go` and `s2i_generate.go`. Please add them in a follow up PR. Or is there a reason not to add them?

cmd/func-util/main.go
```diff
@@ -73,6 +73,8 @@ func socat(ctx context.Context) error {
 	return cmd.Execute()
 }
 
+const middlewareFileName = "middleware-version"
```
twoGiants: nit

Traditionally constants which are used throughout the package are declared at the top below the imports.

cmd/func-util/main.go
```diff
@@ -141,6 +148,8 @@ func s2iCmd(ctx context.Context) error {
 	return cmd.Execute()
 }
 
+const imageDigestFileName = "image-digest"
```
twoGiants: Not used outside of `deploy`, better to declare it there.

pkg/pipelines/tekton/tasks.go
```diff
@@ -386,80 +437,25 @@ spec:
...
-      command: ["scaffold", "$(params.path)"]
-`, ScaffoldImage)
+`, ScaffoldImage, S2IImage, DeployerImage)
 }
 
 // GetClusterTasks returns multi-document yaml containing tekton tasks used by func.
 func GetClusterTasks() string {
-	tasks := getBuildpackTask() + "\n---\n" + getS2ITask() + "\n---\n" + getDeployTask() + "\n---\n" + getScaffoldTask()
+	tasks := getBuildpackTask() + "\n---\n" + getS2ITask()
 	tasks = strings.ReplaceAll(tasks, "kind: Task", "kind: ClusterTask")
```
twoGiants, line 459: `ClusterTasks` are deprecated in favor of remote resolution and namespaced tasks. This should be removed entirely.

pkg/pipelines/tekton/templates_pack.go
twoGiants, line 1: Use `apiVersion: tekton.dev/v1` for all pipelines.

pkg/pipelines/tekton/gitlab_int_test.go
matejvasek: I think it should be purged together with the NS.

matejvasek: Oh maybe not.

pkg/pipelines/tekton/gitlab_int_test.go
matejvasek: I see, this was dome somebody else in some other PR.

cmd/func-util/s2i_generate.go
matejvasek: I'll can add more test later. Integration tests for on cluster build should partly cover this already.

pkg/pipelines/tekton/gitlab_int_test.go
matejvasek: It's not NS scoped resource.

pkg/pipelines/tekton/tasks.go
matejvasek: Will this work by default on most clusters? I think this tries to use the defunct teknon hub by default unless admin set up artifact hub.

matejvasek:
```
0s          Warning   Failed                    taskrun/func-e865f837-pack-git-runhvzbf-build      error requesting remote resource: invalid resource request "gitlab-test-zswkg/hub-e417db434c251131b43423119114e5ca": failed to populate default params: failed to resolve catalog name with type: stepaction
```

matejvasek: It definitely does not work on my KinD cluster. Maybe it works on OCP, in which case I would do this as custom patch for our midstream.

matejvasek: But for midstream I might also use `cluster resolver` since we have these tasks installed by PipelinesOperator in OCP.

pkg/pipelines/tekton/templates.go
matejvasek: I do not think linter can be really sure about unused public fields.

pkg/pipelines/tekton/tasks.go
matejvasek: OK. I shall do that in separate PR.

matejvasek:
@twoGiants [I applied most of your suggestions](https://github.com/knative/func/compare/8c4e4072d8a649857c4b23114d2612e6d21143d1..c7def230cadf078b8f5cf6f916a0fbfb0ebdb233)

pkg/pipelines/tekton/tasks.go
matejvasek: hmm maybe it should be ?
```yaml
      ref:
        resolver: hub
        params:
          - name: type
            value: artifact
          - name: kind
            value: stepaction
          - name: name
            value: git-clone
          - name: version
            value: "0.4.1"
          - name: catalog
            value: redhat-stepactions
```

matejvasek: Ah ok with that I am getting:
```
0s          Warning   Failed                              taskrun/func-f4e4198d-pack-git-run-kr7t6-build                    the step "fetch-src" in TaskRun "func-f4e4198d-pack-git-run-kr7t6-build" failed to pull the image "". The pod errored with the message: "Back-off pulling image "registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:c4b2183f7c7997bd401d86b33eefb637b3ef2fa90618e875106292cd69a15c14": ErrImagePull: failed to pull and unpack image "registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:c4b2183f7c7997bd401d86b33eefb637b3ef2fa90618e875106292cd69a15c14": failed to resolve reference "registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:c4b2183f7c7997bd401d86b33eefb637b3ef2fa90618e875106292cd69a15c14": failed to authorize: failed to fetch anonymous token: unexpected status from GET request to https://registry.redhat.io/auth/realms/rhcc/protocol/redhat-docker-v2/auth?scope=repository%3Aopenshift-pipelines%2Fpipelines-git-init-rhel8%3Apull&service=docker-registry: 401 Unauthorized."
```

matejvasek: So it's probably only applicable for mids/down-stream.

matejvasek: I think I should leave this as it is using http resolver.

pkg/pipelines/tekton/tasks.go
matejvasek: won't fix, see above

pkg/pipelines/tekton/tasks.go
twoGiants: Yeah, keep the http then.

pkg/pipelines/tekton/gitlab_int_test.go
twoGiants: > It's not NS scoped resource.

Correct, so it stays in the cluster and "polutes" it. I would remove it.

pkg/pipelines/tekton/gitlab_int_test.go
twoGiants, line 1: `func usingNamespace(t *testing.T) string {` is missing `t.Helper()`

pkg/pipelines/tekton/templates.go
twoGiants: Yeah, good point. Could be used by a client of the package. In VScode I always check the "show references" and if it's 0 and not a desired public API I remove it.

PR #3354, opened 2026-01-13
Contour listens on IPv6

matejvasek:
# Changes

- Contour ingress listens on IPv6.


matejvasek:
PTAL @gauron99 @lkingland

Issue #3178, opened 2025-11-10
[go] func build installs go twice?


dprotaso:
I updated the go.mod file to set the go version to `1.25`

I'm seeing go toolchain being downloaded twice now which seems unnecessary

```
Running build command
Paketo Buildpack for Go Distribution 2.7.14
  Resolving Go version
    Candidate version sources (in priority order):
      <unknown> -> ""

    Selected Go version (using <unknown>): 1.23.11

  Executing build process
    Installing Go 1.23.11
      Completed in 56.384s

  Generating SBOM for /layers/paketo-buildpacks_go-dist/go
      Completed in 0s

Processing layers
Updating environment
Reading output files
Updating buildpack processes
Updating process list
Finished running build for buildpack paketo-buildpacks/go-dist@2.7.14
Running build for buildpack dev.knative-extensions.go@0.0.6
Looking up buildpack
Finding plan
Creating plan directory
Preparing paths
Running build command
> Knative Functions Go Buildpack
  Fetching framework dependencies
go: downloading go1.25.0 (linux/arm64)
```

gauron99:
We are in the middle of some buildpacks changes. I will check up on this after that change and revisit to see if its still an issue. Waiting to merge https://github.com/knative/func/pull/3067

dprotaso:
@gauron99 what's the plan to get that merged? 1.21 release?

gauron99:
yes, if urgent i can take a look at it tomorrow and backport

gauron99:
the PR mentioned is ready, will be up for review soon so

PR #3237, opened 2025-11-22
refactor: Extract scaffolding as per-builder, new buildpacks default builder and sync Go version

gauron99:
## Scaffolding unification
- Introducing new `client.Scaffold` step in build,deploy,run commands
- Extract scaffolding logic into dedicated per-builder components (`Scaffolder` structs)
New scaffolders
```
- `pkg/buildpacks/scaffolder.go`
- `pkg/oci/scaffolder.go`
- `pkg/s2i/scaffolder.go`
```
- Change default pack builders
```diff
- var DefaultBaseBuilder = "ghcr.io/knative/builder-jammy-base:latest"
- var DefaultTinyBuilder = "ghcr.io/knative/builder-jammy-tiny:latest"
+ var DefaultBaseBuilder = "ghcr.io/knative/builder-jammy-base:v2"
+ var DefaultTinyBuilder = "ghcr.io/knative/builder-jammy-tiny:v2"
```
### HOST builder changes
- Removing the "build history" - build hashes & pid links
- Removing the concurrency building protection (It will be part of a followup PR for separation)

## scaffolding file parsing
- Replace regex-based go.mod parsing in scaffolding with `golang.org/x/mod/modfile` 
- Dynamically sync Go version from user's function to scaffolding go.mod

// issue: remove custom buildpack for go, unify scaffolding
/fixes https://github.com/knative/func/issues/3005

// issue: double go bin download for pack
/fixes https://github.com/knative/func/issues/3178

## Follow-ups
- build concurrency protection (locking)
- python buildpack fix (remove injector)
- python remote deployment fix, if still issue

PR #3360, opened 2026-01-14
fix: double go download buildpacks

gauron99:
- ensure explicit go version is passed into buildpacks (from function's go.mod file)
/fixes https://github.com/knative/func/issues/3178


gauron99:
The problem here is that we need to identify "lowest supported version by bp"

gauron99:
I think I have a fix

gauron99:
fix now part of scaffolding https://github.com/knative/func/pull/3237

PR #3368, opened 2026-01-16
test: fewer domain levels for knative services domain

matejvasek:
# Changes

- Fewer domain levels for knative services url. Now test ksvc will be direct sub-domain of the cluster domain. This is useful when using wild-card certificate (one can have a cert like `*.localtest.me` but not like `*.*.localtest.me`).
- Refactor: move logic for determining ksvc url into a function instead or having code littered with `fmt.Sprintf(…)`.

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 	"strings"
 	"testing"
+	"text/template"
 	"time"
 
 	fn "knative.dev/func/pkg/functions"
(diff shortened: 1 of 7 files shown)


e2e/e2e_test.go
```diff
@@ -1216,3 +1217,22 @@ func parseRunJSON(t *testing.T, cmd *exec.Cmd) string {
 	t.Cleanup(func() { stdoutWriter.Close() })
 	return address
 }
+
+func ksvcUrl(name string) string {
+	// TODO get `default-external-scheme` from the config in cluster
+	const ksvcDefaultExternalScheme = `http`
+	// TODO get `domain-template` from the config in cluster
```
gauron99, line 1224: Are we planning these TODOs in a subsequent PR?

matejvasek: TBH I do not plan to do it soon. Even with this hardcoded value this is a lot better what what we had before.

matejvasek: We can mark it as good first issue.

gauron99: yep that sounds good

matejvasek:
cc @jrangelramos

PR #3371, opened 2026-01-19
Feature: Default branch and path flags intelligently.

twoGiants:
# Changes

- :gift: Default `--path` flag to current working directory when empty or "."
- :gift: Default `--branch` flag to current git branch instead of hardcoded "main"
- :broom: Consolidate duplicate `getCurrentGitBranch` from `cmd/deploy.go` into shared `gitCliWrapper`
- :broom: Reorganize tests: integration tests to `cmd/config_ci_int_test.go`, unit tests use stubs

/kind enhancement

Relates to #3256

**Release Note**

```release-note
The `func config ci` command now resolves --path and --branch flags intelligently:
- --path defaults to current working directory when empty or "."
- --branch defaults to current git branch when not specified (previously "main")
```

**Docs**

```docs

```

### DIFF
--- a/cmd/common/common.go
+++ b/cmd/common/common.go
@@ -1,7 +1,11 @@
 package common
 
 import (
+	"errors"
 	"fmt"
+	"os"
+	"os/exec"
+	"strings"
 
 	fn "knative.dev/func/pkg/functions"
 )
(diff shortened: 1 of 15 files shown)


twoGiants:
/cc @gauron99 @matejvasek @lkingland 

The PR is ready for review. It looks bigger than it is, because I separated the integration tests from the unit test in a file but there were no new integration tests added. I added just some fixes because they started to fail after I added the new logic.

cmd/common/common.go
```diff
@@ -75,3 +79,78 @@ func (m MockLoaderSaver) Load(path string) (fn.Function, error) {
 func (m MockLoaderSaver) Save(f fn.Function) error {
 	return m.SaveFn(f)
 }
+
+// GetCgbFunc is a function type that retrieves the current git branch for a given path.
+type GetCgbFunc func(path string) (string, error)
```
gauron99: I would probably go for a name like `GitBranchFunc` or `CurrentBranchFunc` because the `Cgb` sounds a bit mysterious to me and I need to think about what it is. Contrary the `GetCwdFunc` is ok-ish since its abbreviation makes more sense to me but in the same spirit could be simply `WorkDirFunc` (`Current...`) But I guess I can get used to it.

Following this the `DefaultGet...` would be `DefaultCurrentBranch` and `DefaultWorkDir` or similar.

cmd/config_ci_test.go
```diff
@@ -13,61 +12,57 @@ import (
...
 func TestNewConfigCICmd_CISubcommandExist(t *testing.T) {
 	// leave 'ci' to make this test explicitly use this subcommand
-	opts := opts{
-		enableFeature:       true,
-		withMockLoaderSaver: true,
-		withBufferWriter:    true,
-		args:                []string{"ci"},
-	}
+	opts := defaultOpts()
+	opts.args = []string{"ci"}
```
gauron99, line 35: arg "ci" is already part of the defaultOpts?

cmd/common/common_test.go
```diff
@@ -52,3 +54,55 @@ func TestDefaultLoaderSaver_ForwardsSaveError(t *testing.T) {
...
+
+func TestGitCliWrapper_CurrentBranch_ReturnsBranchName(t *testing.T) {
+	tempDir := fnTest.FromTempDirectory(t)
+	_, initErr := common.NewGitCliWrapper().Init(tempDir)
+
+	branch, err := common.NewGitCliWrapper().CurrentBranch(tempDir)
+
+	assert.NilError(t, initErr)
+	assert.NilError(t, err)
+	assert.Assert(t, branch != "")
```
gauron99: should we expect a Branch name here as the test name suggests?

cmd/common/common.go
twoGiants: Agree! Was also unsure about the naming. Thank you!

cmd/common/common_test.go
twoGiants: Updated!

cmd/common/common.go
```diff
@@ -75,3 +79,82 @@ func (m MockLoaderSaver) Load(path string) (fn.Function, error) {
...
+	var exitErr *exec.ExitError
+	argsJoined := strings.Join(args, " ")
+	if errors.As(err, &exitErr) {
+		return "", fmt.Errorf("git %s failed: %w\nstderr: %s", argsJoined, err, string(exitErr.Stderr))
+	}
+
+	return "", fmt.Errorf("git %s failed: %w", argsJoined, err)
+}
+
+// CurrentBranchStub creates a stub GetCgbFunc that returns the provided output or error.
```
gauron99: found an outdated comment here

cmd/common/common.go
```diff
@@ -75,3 +79,82 @@ func (m MockLoaderSaver) Load(path string) (fn.Function, error) {
...
+	return &gitCliWrapper{gitCmd}
+}
+
+// CurrentBranch returns the current git branch name for the repository at the given path.
+func (g *gitCliWrapper) CurrentBranch(path string) (string, error) {
+	if _, err := os.Stat(path); err != nil {
+		return "", err
+	}
+
+	return g.execGitCmdWith("-C", path, "symbolic-ref", "--short", "HEAD")
```
lkingland: If I understand this correctly, we'll correctly get errors bubbled up when there's no git repo, but we might want to help out a bit when the user tries to use this without having their Function in source-control.  Perhaps slightly more vebose error messaging would be enough?

```go
    branch, err := g.execGitCmdWith("-C", path, "symbolic-ref", "--short", "HEAD")
    if err != nil {
            return "", fmt.Errorf("could not detect git branch for '%s'. "+
                    "Has git been initialized for this Function? %w", path, err)
    }
    return branch, nil

```

lkingland (commented):
Clean!

I especially like the comprehensive tests, and that the unit tests and integration tests are properly split based on the use of a real FS.

Just one small suggestion about helpful error messages when there's no git repo (see below).

Other than that, it's just the comment David mentioned, and a misbehaving integration test.

cmd/common/common.go
twoGiants: fixed

cmd/common/common.go
twoGiants: Yes, much better! Will update :smile_cat: :+1:

Issue #3051, opened 2025-09-24
Make a 2 layer error typed system


gauron99:
Create a 2 layer system for handling errors.

layer 1. Underlying functions and frameworks used in the CLI (like client, deployer, builder etc.) will return typed errors.
layer 2. CLI will catch these errors and based on the type provided will return a **technical error** from within those functions and layer another **CLI specific error** on top for the user.

This way we can systematically have both a valid technical error and user friendly error returned to the user depending on what situation/edge case etc. is hit.

This will require creating more error types in the core of codebase. This should be accessible via all the interfaces used in the CLI so that it is not specific to them. Example: We have multiple builders and we want to have these errors consistent across all. We might have multiple deployers, same deal.

gauron99:
I think a good suggestion to make this work would be something along the lines of the following:

### Have the layer 1 (inner) error types.

This requires creating more error types alongside the already existing ones in `pkg/functions/errors.go`. These would be returned in the interfaces and functions used within the CLI -- meaning any deployer, builder, client functions etc. will return this type. Create as many error types necessary. These are intended to be technical, simple and straight to the point errors. Nothing CLI specific here. See already existing comment `pkg/functions/errors.go` lines 27-29 for a possible hint. 

### Have the layer 2 (outer) error types.

This layer is intended to be a CLI layer for the errors. CLI functions `runXXX` (in `cmd/deploy.go` etc.) will catch these inner errors in all its used methods, functions and interfaces and CHECK what has been returned. ( something like `if err == ErrTYPE` should suffice) . 

These CLI functions will catch inner errors and add a CLI specific error ON TOP. This means both straight-to-the-point technical error is returned from within as well as CLI specific user-friendly error.

I think there could even be something new like `cmd/errors` which would contain some CLI errors if they were generic or similar enough and could be used across multiple commands.

Actionable items:
- [x] Create new core technical error types to be used across all interfaces and functions
- [x] Use these errors in functions, interfaces and methods like deployers,builders etc.
- [x] Catch these errors in the CLI functions and layer on top a CLI-specific error for the user returning both

PR #3375, opened 2026-01-20
Improve error system

gauron99:
Introduced 2-layer error system:
  - Layer 1 (pkg/functions/errors.go): Technical/ library errors (short to the point)
  - Layer 2 (cmd/errors.go): CLI-friendly wrappers with examples and troubleshooting (user oriented)

- Added new error types which now use `Error()` and `Unwrap()` for using `errors.Is()/As()`
- Removed the verbosity from CMD package

/fix https://github.com/knative/func/issues/3051

### DIFF
--- a/cmd/errors.go
+++ b/cmd/errors.go
@@ -1,15 +1,294 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
+	"os"
 	"strings"
+
+	fn "knative.dev/func/pkg/functions"
 )
 
-// wrapNotInitializedError wraps an ErrNotInitialized error with CLI-specific guidance
-func wrapNotInitializedError(err error, command string) error {
-	switch command {
+// ----------------------------- META WRAPPERS ----------------------------- //
+
+// wrapValidateError is a meta wrapper for Validate function
+func wrapValidateError(err error, cmd string) error {
+	if cmd != "build" && cmd != "deploy" {
+		msg := fmt.Sprintf(`
+Internal error during error-wrapping: specified cmd '%s' not supported`, cmd)
+		return fmt.Errorf("%w\n%s", err, msg)
+	}
+
+	if errors.Is(err, fn.ErrInvalidDomain) {
+		return NewErrInvalidDomain(err, cmd)
+	}
+	if errors.Is(err, fn.ErrInvalidNamespace) {
+		return NewErrInvalidNamespace(err, cmd)
+	}
+	if errors.Is(err, fn.ErrConflictingImageAndRegistry) {
+		return NewErrConflictImageRegistry(err, cmd)
+	}
+	if errors.Is(err, fn.ErrPlatformNotSupported) {
+		return NewErrPlatformNotSupported(err, cmd)
+	}
+	return err
+}
+
+// wrapRunError wraps errors from client.Run with CLI-specific guidance
+func wrapRunError(err error, address string) error {
+	var portErr *fn.ErrPortUnavailableError
+	if errors.As(err, &portErr) {
+		if portErr.IsPermissionDenied() {
+			return NewErrPortPermissionDenied(portErr.Port, address)
+		}
+		return NewErrPortUnavailable(err, portErr.Port)
+	}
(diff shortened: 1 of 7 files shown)


cmd/errors.go
```diff
@@ -1,15 +1,260 @@
...
+}
+
+// ---------------------------- TYPES AND METHODS --------------------------- //
+
+type ErrPlatformNotSupported struct {
+	Err error
+	Cmd string
+}
+
+func NewPlatformNotSupported(err error, cmd string) error {
```
lkingland: Should probably be `NewErrPlatformNotSupported`

cmd/errors.go
```diff
@@ -1,15 +1,260 @@
...
+type ErrClusterNotAccessible struct {
+	Err error
+}
+
+func NewErrClusterNotAccessible(err error) error {
+	return &ErrClusterNotAccessible{Err: err}
+}
+
+func (e *ErrClusterNotAccessible) Error() string {
+	// fn.ErrorClusterNotAccessible was here used, anything better to do here?
```
lkingland: Is this comment still applicable?  Looks good to me.

cmd/errors.go
```diff
@@ -1,15 +1,260 @@
...
-	switch command {
+// ----------------------------- META WRAPPERS ----------------------------- //
+
+// wrapValidateError is a meta wrapper for Validate function
+func wrapValidateError(err error, cmd string) error {
+	if cmd != "build" && cmd != "deploy" {
+		msg := fmt.Sprintf(`
+Internal error during error-wrapping: specified cmd '%s' not supported`, cmd)
+		return fmt.Errorf("%w\n%s", err, msg)
+	}
```
lkingland, line 20: This returns an "internal error" for any other commandh. Maybe just pass through the original error for unsupported commands?

cmd/errors.go
```diff
@@ -91,37 +387,163 @@ Common registries:
...
+  127.0.0.1:8080    Localhost only
+  0.0.0.0:8080      All interfaces (IPv4)
+  [::]:8080         All interfaces (IPv6)
+
+For more options, run 'func run --help'`, e.Address)
+}
+
+func (e *ErrInvalidAddress) Unwrap() error {
+	return e.Err
+}
```
lkingland, line 519: Is there expected to always be a wrapped error for this?  If so, Maybe we should add the wrapped error to the constructor?  If not, we could remove Unwrap.

cmd/errors.go
```diff
@@ -91,37 +387,163 @@ Common registries:
...
+Examples:
+  func run --address 127.0.0.1:8080
+  func run --address 0.0.0.0:9090
+
+For more options, run 'func run --help'`, e.Port)
+}
+
+func (e *ErrInvalidPort) Unwrap() error {
+	return e.Err
+}
```
lkingland, line 545: Maybe we should add the wrapped error to the constructor? WDYT?

cmd/errors.go
gauron99: aha! the method was correct so linter wasnt screaming :facepalm:

cmd/errors.go
gauron99: The error is being passed through there using `%w` as well as at the end of function when the internal error couldnt be matched -- as a fallback to simply return what was passed in.

Did you mean to get rid of this `Sprintf` wrapping and simply return that internal error as is?

cmd/errors.go
gauron99: thanks for catching it! this one is a bit nuanced -- used only in `run` cmd and we "ignore the internal err" https://github.com/knative/func/blob/c2e4cf1ebe2c2e2f67b5293af03845f39c14514d/cmd/run.go#L430-L435

gauron99: I agree that keeping the error there looks a bit better, will add

gauron99:
new commit to main, had to rebase

lkingland:
/lgtm
/approve
(🤞🏻 the E2Es succeed)

PR #3378, opened 2026-01-21
Update templates (instanced signature)

gauron99:
- instanced go functions as default
- as instanced signature defaults to "OK" response, change this accordingly in tests (instead of echo response)
- edit tests to now use instanced function instead of the static one
- echo templates part of [templates](github.com/functions-dev/templates)
- add specific tests for Static go-signature
- some tests preserve the echo testing & use specific func template instead

http templates now return: "OK" string
cloudevents templates now return: json {"message":"OK"}


e2e/testdata/templates/python/http/function/func.py
```diff
@@ -44,7 +31,7 @@ async def handle(self, scope, receive, send):
         })
         await send({
             'type': 'http.response.body',
-            'body': url.encode(),
```
matejvasek, line 47: Does encode make sense with a string that is not an url?

matejvasek: Maybe it does I don't know.

e2e/e2e_core_test.go
```diff
@@ -260,7 +262,9 @@ func TestCore_Invoke(t *testing.T) {
 	name := "func-e2e-test-core-invoke"
 	_ = fromCleanEnv(t, name)
 
-	if err := newCmd(t, "init", "-l=go").Run(); err != nil {
+	if err := newCmd(t, "init", "-l=go",
```
matejvasek, line 265: Why now there is a git repo?

matejvasek: Ah that's a repo for templates, not func src code?

gauron99: yes im keeping the "echo" functionality and using the template to do it

e2e/testdata/templates/python/http/function/func.py
gauron99: This is string -> bytes conversion. thats whats required in the struct. I think this is equivalent to `b"OK"`. You can specify what encoding you want "utf-8" "ascii" but we just use basic (no argument) which is utf-8

gauron99:
rebased for zz file

lkingland (approved):
So excellent to see these now ready to go!
Simplified core handlers and instanced templates by default are a huge step forward
/lgtm
/approve

gauron99:
@lkingland could you re-lgtm pls, had to rebase because of zz file

lkingland (approved):

PR #3379, opened 2026-01-21
Feature: Add --platform flag for CI/CD platform selection.

twoGiants:
# Changes

- :gift: Add `--platform` flag to `config ci` command for CI/CD platform selection
- :gift: Default platform to "github" with case-insensitive matching
- :gift: Return descriptive error for unsupported platforms, preparing for future platform additions

/kind enhancement

Relates to #3256

**Release Note**

```release-note
Add --platform flag to `func config ci` command allowing users to specify which CI/CD platform to generate manifests for. Currently supports "github" (default).
```

**Docs**

```docs

```

### DIFF
--- a/cmd/ci/config.go
+++ b/cmd/ci/config.go
@@ -14,8 +14,8 @@ const (
 
 	PathFlag = "path"
 
-	CICDPlatformFlag    = "platform"
-	DefaultCICDPlatform = "github"
+	PlatformFlag    = "platform"
+	DefaultPlatform = "github"
 
 	DefaultGitHubWorkflowDir      = ".github/workflows"
 	DefaultGitHubWorkflowFilename = "func-deploy.yaml"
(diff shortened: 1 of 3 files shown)


creydr:
So IIUC correctly this is at the moment (without having other platforms supported) mostly to make the users aware, that only GH is supported at the moment. Correct?

cmd/ci/config.go
```diff
@@ -12,6 +14,9 @@ const (
 
 	PathFlag = "path"
 
+	CICDPlatformFlag    = "platform"
```
gauron99: can this be just `PlatformFlag` ? do we expect another platform to be included in `func config ci` subcommand?

twoGiants: @gauron99 sure! I'll update it. Yes, gitlab will also come, maybe Tekton PAC too and others.

@creydr Because `github` is the default anyway, they don't even have to select it. Just in case they try they will see the only one supported for now is Github.

gauron99: yes great. I meant "different platform" as to that new thing would conflict with the naming so that we have to specify the `CICD` as well. Thanks for simplicifation

PR #3388, opened 2026-01-22
refactor: big string literals to separate files

matejvasek:
# Changes

- Added basic tests for the tasks: we validate that the yaml contains valid task.
- Moved big yaml string literals (tekton tasks) into separate files and then use `go:embed` on these.
- Use Go templates instead of `fmt.Sprintf()`.
- Re-formatted the tekton tasks using `yq`.


matejvasek:
PTAL @gauron99 @lkingland

PR #3398, opened 2026-01-26
ACR Login: Use Azure SDK library to get the credential and token

### LINKED ISSUE #3288: Azure Container Repository Login Fails
## Context

Starting with `az` CLI v2.30, token persistence to accessTokens.json was [deprecated](https://learn.microsoft.com/en-us/cli/azure/msal-based-azure-cli?view=azure-cli-latest). However, `func` [still relies on this file](https://github.com/knative/func/blob/07092aad7c0072a349187cdd2184420809a3640f/pkg/k8s/keychains.go#L58-L61) to authenticate against ACR. As a result, we are hitting errors such as:

```
open Azure access tokens: no such file or directory
```
This breaks ACR authentication flows that previously worked without any user-side changes.

## Possible Approaches

1. **Call `az acr login --expose-token` or `az account get-access-token`** to get a fresh token from az CLI whenever we want to access an Azure Container Repository
2. **Use the [Azure SDK for Go](https://github.com/Azure/azure-sdk-for-go)** to get default credentials and a fresh token (this would delegate authentication to supported mechanisms (including, but not limited to `az` CLI)

## Related Work

I implemented a fix to validate the **second** approach here: https://github.com/stackbuilders/func/pull/2

The PR avoids relying on `accessTokens.json` and instead fetches a fresh token using the SDK.
Happy to open a PR here and adjust it to better align with the project’s preferred direction.

dbalseiro:
# Changes

- Add [Azure SDK for Go](https://github.com/Azure/azure-sdk-for-go) as a dependency
- Use Azure SDK for GO to get credentials and token to log in to [Azure Container Registry[(https://learn.microsoft.com/en-us/cli/azure/msal-based-azure-cli?view=azure-cli-latest)

/kind bug

Fixes #3288 

**Release Note**

```release-note
`func` no longer relies on Azure CLI’s deprecated `accessToken.json` file to authenticate with Azure Container Registry. Authentication now uses Azure’s supported credential mechanisms (e.g. Azure CLI, environment credentials, managed identity). No action is required for most users.
```

### DIFF
--- a/pkg/k8s/keychains.go
+++ b/pkg/k8s/keychains.go
@@ -2,16 +2,15 @@ package k8s
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
-	"os"
-	"path"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	ecr "github.com/awslabs/amazon-ecr-credential-helper/ecr-login"
 	dockercreds "github.com/docker/docker-credential-helpers/credentials"
 	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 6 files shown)


lkingland:
/ok-to-test

Thanks for the PR!

pkg/k8s/keychains.go
```diff
@@ -54,32 +54,22 @@ func GetACRCredentialLoader() []creds.CredentialsCallback {
...
-			target := "https://" + registry
-			for _, t := range tokens {
-				if t.Resource == target {
-					return oci.Credentials{
-						Username: "00000000-0000-0000-0000-000000000000",
-						Password: t.AccessToken,
-					}, nil
-				}
+			scope := "https://containerregistry.azure.net/.default"
+			token, err := azCredentials.GetToken(context.Background(), policy.TokenRequestOptions{
```
lkingland: We might want to plumb through a context object to GetACRCredentialLoader in case this thing can lock up or need canceling.

pkg/k8s/keychains.go
```diff
@@ -54,32 +54,22 @@ func GetACRCredentialLoader() []creds.CredentialsCallback {
...
+				Scopes: []string{scope},
+			})
+			if err != nil {
+				return oci.Credentials{}, fmt.Errorf("failed to get azure access token: %w", err)
 			}
-			return oci.Credentials{}, creds.ErrCredentialsNotFound
```
lkingland, line 204: I notice that `creds.ErrCredentialsNotFound` is no longer returned.  This sentinel value may be depended upon by other systems to fall-through (despite using error types as flow-control being an antipattern; it's life).  It might be a good idea to double-check that there's no other code depending on exactly that error type to ensure we're not breaking any "error fallback" systems if one of the other generic errors is returned instead.

lkingland (approved):
Looks good!  I have one minor improvement worth at least exploring the feasability of (plumbing through a context object) and one concern worth double-checking (that there's no other code depending on creds.ErrCredentialsNotFound being returned).

/approve
/hold pending possible changes per the above

pkg/k8s/keychains.go
dbalseiro: Yes, we are still returning `ErrCredentialsNotFound` when the registry's domain is not `azure.io`. And if I'm not mistaken, returning `ErrCredentialsNotFound` in this line made sense at the time, because it was trying to fetch the token from a JSON file. But now we try to _generate_ the token using the SDK. I think if the SDK fails, we should return a plain error. What do you think?

pkg/k8s/keychains.go
dbalseiro: giving it a try :)

dbalseiro (commented):
@lkingland Thanks for the review! Sorry it took me so long to do it, but I addressed your comments. Let me know what you think.

lkingland:
Where's all our tests I wonder?

dbalseiro:
Thanks for the review!
Does the PR also need a `lgtm` label? There is a pending check waiting for it.

lkingland:
@dbalseiro sorry for the late reply.  If you can do a rebase we can finally land this one.

dbalseiro:
@lkingland rebased and retested :)
Thank you!

dbalseiro:
Hello everyone! any news on this PR?

dbalseiro:
Hey @lkingland, I rebased this again, but I closed the PR by mistake and then reopened it 😓. Could you please take a look again and see if this is ready to merge -- sorry about the troubles

gauron99:
@dbalseiro the branch is only a test name tweak (`ErrCredentialsNotFound` → `… error`). The Azure SDK ACR change from the description is gone — likely dropped on rebase.

I assume that was accidental — are you still working on this?

dbalseiro:
@lkingland hey! yes this is still on my radar :) I can take care of this over the weekend

dbalseiro:
@gauron99 you were right I had force-pushed a bad rebase :( Now everything is how it's supposed to be

We are using a version of this in our project and I'd greatly appreciate if it can go upstream. Thanks for your time!

PR #3407, opened 2026-01-30
Feature: Protect existing workflow files from accidental overwrites.

twoGiants:
# Changes

- :gift: Add workflow file existence check to prevent accidental overwrites
- :gift: Add `--force` flag to explicitly allow overwriting existing workflow files
- :gift: Add warning message when `--force` is used to overwrite existing workflows
- :gift: Add distinct default workflow name for remote builds ("Remote Func Deploy") to avoid conflicts

/kind enhancement

Relates to #3256

**Release Note**

```release-note
The `func config ci` command now protects existing workflow files from accidental overwrites. Use the `--force` flag to explicitly overwrite existing files.
```

**Docs**

```docs

```

### DIFF
--- a/cmd/ci/config.go
+++ b/cmd/ci/config.go
@@ -23,8 +23,9 @@ const (
 	BranchFlag    = "branch"
 	DefaultBranch = "main"
 
-	WorkflowNameFlag    = "workflow-name"
-	DefaultWorkflowName = "Func Deploy"
+	WorkflowNameFlag               = "workflow-name"
+	DefaultWorkflowName            = "Func Deploy"
+	DefaultRemoteBuildWorkflowName = "Remote " + DefaultWorkflowName
 
 	KubeconfigSecretNameFlag    = "kubeconfig-secret-name"
 	DefaultKubeconfigSecretName = "KUBECONFIG"
@@ -52,6 +53,9 @@ const (
 
 	UseSelfHostedRunnerFlag    = "self-hosted-runner"
 	DefaultUseSelfHostedRunner = false
+
+	ForceFlag    = "force"
+	DefaultForce = false
 )
 
 // CIConfig readonly configuration
@@ -69,42 +73,37 @@ type CIConfig struct {
 	useRegistryLogin,
 	useSelfHostedRunner,
 	useRemoteBuild,
-	useWorkflowDispatch bool
+	useWorkflowDispatch,
+	force bool
 }
 
 func NewCIConfig(
 	currentBranch common.CurrentBranchFunc,
 	workingDir common.WorkDirFunc,
+	workflowNameExplicit bool,
 ) (CIConfig, error) {
-	platform := viper.GetString(PlatformFlag)
-	if strings.ToLower(platform) != DefaultPlatform {
-		return CIConfig{}, fmt.Errorf("%s support is not implemented", platform)
+	if err := resolvePlatform(); err != nil {
+		return CIConfig{}, err
 	}
 
-	path := viper.GetString(PathFlag)
-	if path == "" || path == "." {
-		cwd, err := workingDir()
-		if err != nil {
-			return CIConfig{}, err
-		}
-		path = cwd
+	path, err := resolvePath(workingDir)
+	if err != nil {
+		return CIConfig{}, err
 	}
 
-	branch := viper.GetString(BranchFlag)
-	if branch == "" {
-		var err error
-		branch, err = currentBranch(path)
-		if err != nil {
-			return CIConfig{}, err
-		}
+	branch, err := resolveBranch(path, currentBranch)
+	if err != nil {
+		return CIConfig{}, err
 	}
+
+	workflowName := resolveWorkflowName(workflowNameExplicit)
 
 	return CIConfig{
 		githubWorkflowDir:      DefaultGitHubWorkflowDir,
 		githubWorkflowFilename: DefaultGitHubWorkflowFilename,
 		path:                   path,
 		branch:                 branch,
-		workflowName:           viper.GetString(WorkflowNameFlag),
+		workflowName:           workflowName,
 		kubeconfigSecret:       viper.GetString(KubeconfigSecretNameFlag),
 		registryLoginUrlVar:    viper.GetString(RegistryLoginUrlVariableNameFlag),
 		registryUserVar:        viper.GetString(RegistryUserVariableNameFlag),
@@ -114,7 +113,58 @@ func NewCIConfig(
 		useSelfHostedRunner:    viper.GetBool(UseSelfHostedRunnerFlag),
 		useRemoteBuild:         viper.GetBool(UseRemoteBuildFlag),
 		useWorkflowDispatch:    viper.GetBool(WorkflowDispatchFlag),
+		force:                  viper.GetBool(ForceFlag),
 	}, nil
+}
+
+func resolvePlatform() error {
+	platform := viper.GetString(PlatformFlag)
+	if strings.ToLower(platform) != DefaultPlatform {
+		return fmt.Errorf("%s support is not implemented", platform)
+	}
+
+	return nil
+}
+
+func resolvePath(workingDir common.WorkDirFunc) (string, error) {
+	path := viper.GetString(PathFlag)
+	if path != "" && path != "." {
+		return path, nil
+	}
+
+	cwd, err := workingDir()
+	if err != nil {
+		return "", err
+	}
+
+	return cwd, nil
+}
+
+func resolveBranch(path string, currentBranch common.CurrentBranchFunc) (string, error) {
+	branch := viper.GetString(BranchFlag)
+	if branch != "" {
+		return branch, nil
+	}
+
+	branch, err := currentBranch(path)
+	if err != nil {
+		return "", err
+	}
+
+	return branch, nil
+}
+
+func resolveWorkflowName(explicit bool) string {
+	workflowName := viper.GetString(WorkflowNameFlag)
+	if explicit {
+		return workflowName
+	}
+
+	if viper.GetBool(UseRemoteBuildFlag) {
+		return DefaultRemoteBuildWorkflowName
+	}
+
+	return DefaultWorkflowName
 }
 
 func (cc *CIConfig) FnGitHubWorkflowDir(fnRoot string) string {
@@ -171,4 +221,8 @@ func (cc *CIConfig) RegistryPassSecret() string {
 
 func (cc *CIConfig) RegistryUrlVar() string {
 	return cc.registryUrlVar
+}
+
+func (cc *CIConfig) Force() bool {
+	return cc.force
 }
--- a/cmd/ci/workflow.go
+++ b/cmd/ci/workflow.go
@@ -2,10 +2,17 @@ package ci
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
+	"io"
 
 	"gopkg.in/yaml.v3"
 )
+
+const defaultFuncCliVersion = "knative-v1.21.0"
+
+// ErrWorkflowExists is returned when a GitHub workflow file already exists and --force is not specified.
+var ErrWorkflowExists = errors.New("existing GitHub workflow detected, overwrite using the --force option")
 
 type githubWorkflow struct {
 	Name string           `yaml:"name"`
@@ -35,7 +42,7 @@ type step struct {
 }
 
 func NewGitHubWorkflow(conf CIConfig) *githubWorkflow {
-	name := createWorkflowName(conf)
+	name := conf.WorkflowName()
 	runsOn := createRunsOn(conf)
 	pushTrigger := createPushTrigger(conf)
 
@@ -56,15 +63,6 @@ func NewGitHubWorkflow(conf CIConfig) *githubWorkflow {
 			},
 		},
 	}
-}
-
-func createWorkflowName(conf CIConfig) string {
-	result := conf.WorkflowName()
-	if conf.UseRemoteBuild() {
-		result = "Remote Func Deploy"
-	}
-
-	return result
 }
 
 func createRunsOn(conf CIConfig) string {
@@ -120,7 +118,7 @@ func createRegistryLoginStep(conf CIConfig, steps []step) []step {
 func createFuncCLIInstallStep(steps []step) []step {
 	installFuncCli := newStep("Install func cli").
 		withUses("functions-dev/action@main").
-		withActionConfig("version", "knative-v1.20.1").
+		withActionConfig("version", defaultFuncCliVersion).
 		withActionConfig("name", "func")
 
 	return append(steps, *installFuncCli)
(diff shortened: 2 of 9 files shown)


cmd/ci/workflow.go
```diff
@@ -2,11 +2,17 @@ package ci
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 
 	"gopkg.in/yaml.v3"
 )
 
+const defaultFuncCliVersion = "knative-v1.20.1"
```
gauron99: should we change this to "latest" or do we strictly want a static version here. I like using pinned for consistency -- in that case this can be already updated to 1.21

its a bit bitter-sweet: We can create _another_ script which updates this version in the `hack/cmd` files and creates bot PR :see_no_evil: 

PS script: I plan to bring that hack dir "together" so maybe some unification/simplification will be possible

twoGiants: Pinned version is safer. I will update to 1.21.0.

Ideally we would set it via some config file or "compiler flag" or env var at build time so that it always reflects the version which was actually build.

twoGiants:
/cc @gauron99 @matejvasek @lkingland

The PR is ready for review. Please read the description about the logging.

twoGiants:
> ## [Codecov](https://app.codecov.io/gh/knative/func/pull/3407?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=knative) Report
> 
> ❌ Patch coverage is `37.25490%` with `64 lines` in your changes missing coverage. Please review. ✅ Project coverage is 54.52%. Comparing base ([`c12ec96`](https://app.codecov.io/gh/knative/func/commit/c12ec9661d2ca1d5d9c52f1c952fe561217569bd?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=knative)) to head ([`8109c29`](https://app.codecov.io/gh/knative/func/commit/8109c29984be3a1cb52e8957e4bd9e9e4790f05e?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=knative)). ⚠️ Report is 1 commits behind head on main.
> [Files with missing lines](https://app.codecov.io/gh/knative/func/pull/3407?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=knative) 	
> ...
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/knative/func/pull/3407?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=knative). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=knative).
> 🚀 New features to boost your workflow:

The codecov tool is poorly configured or am I missing something? It's flagging the [`resolveBranch` function as uncovered](https://app.codecov.io/gh/knative/func/pull/3407#8572e2bc51e9f2b959ae59ba75ad094f-R142) and much more but I have explicit tests for this paths `TestNewConfigCICmd_BranchFlagResolution`, `TestNewConfigCICmd_BranchFlagResolutionError` etc.

@gauron99

cmd/ci/config.go
```diff
@@ -114,9 +112,61 @@ func NewCIConfig(
...
+		return workflowName
+	}
+
+	useRemoteBuild := viper.GetBool(UseRemoteBuildFlag)
+	if useRemoteBuild {
+		return DefaultRemoteBuildWorkflowName
+	}
+
+	return DefaultWorkflowName
+}
```
lkingland: So if I understand this right, if a user explicitly sets workflow-name="Func Deploy" (the default value), the function treats it as if no custom name was provided because of the workflowName != DefaultWorkflowName check. This means with --remote --workflow-name="Func Deploy", we'd get "Remote Func Deploy" instead of their explicitly requested "Func Deploy"?

Something like this might help:
```
workflowNameChanged := cmd.Flags().Changed(WorkflowNameFlag)
```

twoGiants: Yes, good catch! I was also unsure about this one. Thank you. Will correct it :smile_cat: :+1:

twoGiants: I am setting the default during resolution, the I don't need to pass down `cobra.Command`. Is that ok too?

twoGiants: After talking to David I got a small primer about how default values are set and it's probably best to evaluate the `cmd.Flags().Changed(WorkflowNameFlag)` and do what you proposed Luke :smiley_cat:

cmd/config_ci_int_test.go
```diff
@@ -77,14 +78,52 @@ func TestNewConfigCICmd_WritesWorkflowFileToFSWithCorrectYAMLStructure(t *testin
...
+	t.Run("overwrite without force flag fails", func(t *testing.T) {
+		opts := optsIntegration{
+			withFunc: baseOpts.withFunc,
+			args:     append(slices.Clone(baseOpts.args), "--workflow-name="+changedWorkflowName),
+		}
+
+		err := runConfigCiCmdIntegration(t, opts)
+		content := readWorkflowFile(t, opts.withFunc.Root)
+
+		assert.Error(t, err, ci.ErrWorkflowExists.Error())
```
matejvasek: Maybe `assert.ErrorIs()` would be better.

cmd/config_ci_test.go
```diff
@@ -276,15 +326,59 @@ func TestNewConfigCICmd_UnsupportedPlatformError(t *testing.T) {
...
+	})
+
+	t.Run("overwrite without force flag fails", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.withWriter = sharedWriter
+		opts.args = append(opts.args, "--workflow-name="+changedWorkflowName)
+
+		result := runConfigCiCmd(t, opts)
+
+		assert.Error(t, result.executeErr, ci.ErrWorkflowExists.Error())
```
matejvasek: Maybe `assert.ErrorIs()` would be better.

cmd/ci/writer.go
```diff
@@ -34,19 +35,32 @@ func (fw *fileWriter) Write(path string, raw []byte) error {
...
+func (bw *BufferWriter) Write(path string, raw []byte) error {
 	bw.Path = path
+	bw.Buffer.Reset()
 	_, err := bw.Buffer.Write(raw)
 	return err
 }
+
+// Exist is a fake implementation that returns true if the buffer has content.
+func (bw *BufferWriter) Exist(_ string) bool {
+	return bw.Buffer != nil && bw.Buffer.Len() > 0
```
matejvasek, line 65: nit: Should not this also check that data were written to appropriate file?

I guess this is fine for tests.

If we call `bw.Write("flow-a.yaml", flowYaml)` and then `bw.exists("flow-b.yaml")` this will true but I don't think that should be the case.

cmd/ci/writer.go
```diff
@@ -34,19 +35,32 @@ func (fw *fileWriter) Write(path string, raw []byte) error {
 	return nil
 }
 
-type bufferWriter struct {
+func (fw *fileWriter) Exist(path string) bool {
+	_, err := os.Stat(path)
```
matejvasek, line 39: I wonder whether we should rather call `os.Lstat()`.

But maybe not, I guess `func-deploy.yaml` should not ever be a symlink.

matejvasek:
Just thinking out loud.

Just matter of taste:
I would prefer no adding the `Exist()` function.  Rather I would make the `Write()` function to take an additional parameter `override` and let it handle possible override error. See patch below.

To me it feels more "atomic" as it does not do separate `check & create`, it simply just tries to create a file and possibly fails. Obviously this is not really important distinction, if we run two concurrent func at the same time we are in a trouble anyway. The fact that one instance of func may created the wf file while the other is in between `exits` and `write` is not really a problem.

```patch
diff --git a/cmd/ci/workflow.go b/cmd/ci/workflow.go
index c69c64b45..5382829fa 100644
--- a/cmd/ci/workflow.go
+++ b/cmd/ci/workflow.go
@@ -173,21 +173,22 @@ func newVariable(key string) string {
 }
 
 func (gw *githubWorkflow) Export(path string, w WorkflowWriter, force bool, m io.Writer) error {
-	if !force && w.Exist(path) {
-		return ErrWorkflowExists
-	}
-
-	if w.Exist(path) {
-		// best-effort user message; errors are non-critical
-		_, _ = fmt.Fprintf(m, "WARNING: --force flag is set, overwriting existing GitHub Workflow file\n")
+	raw, err := gw.toYaml()
+	if err != nil {
+		return err
 	}
 
-	raw, err := gw.toYaml()
+	err = w.Write(path, raw, force)
 	if err != nil {
 		return err
 	}
 
-	return w.Write(path, raw)
+	if force {
+		// best-effort user message; errors are non-critical
+		_, _ = fmt.Fprintf(m, "WARNING: --force flag is set, overwriting existing GitHub Workflow file\n")
+	}
+
+	return nil
 }
 
 func (gw *githubWorkflow) toYaml() ([]byte, error) {
diff --git a/cmd/ci/writer.go b/cmd/ci/writer.go
index f15be50b1..4c7a159a4 100644
--- a/cmd/ci/writer.go
+++ b/cmd/ci/writer.go
@@ -16,28 +16,39 @@ var DefaultWorkflowWriter = &fileWriter{}
 
 // WorkflowWriter defines the interface for writing workflow files.
 type WorkflowWriter interface {
-	Exist(path string) bool
-	Write(path string, raw []byte) error
+	Write(path string, raw []byte, override bool) error
 }
 
 type fileWriter struct{}
 
 // Write writes raw bytes to the specified path, creating directories as needed.
-func (fw *fileWriter) Write(path string, raw []byte) error {
+// If override is false and the file already exists, ErrWorkflowExists is returned.
+func (fw *fileWriter) Write(path string, raw []byte, override bool) error {
 	if err := os.MkdirAll(filepath.Dir(path), dirPerm); err != nil {
 		return err
 	}
 
-	if err := os.WriteFile(path, raw, filePerm); err != nil {
+	flags := os.O_WRONLY | os.O_CREATE | os.O_TRUNC | os.O_EXCL
+
+	if override {
+		// clear exclusive flag
+		flags &= ^os.O_EXCL
+	}
+
+	f, err := os.OpenFile(path, flags, filePerm)
+	if err != nil {
+		if os.IsExist(err) {
+			return ErrWorkflowExists
+		}
 		return err
 	}
+	defer f.Close()
 
-	return nil
-}
+	if _, err := f.Write(raw); err != nil {
+		return err
+	}
 
-func (fw *fileWriter) Exist(path string) bool {
-	_, err := os.Stat(path)
-	return err == nil
+	return nil
 }
 
 // BufferWriter is a test double (fake) that implements WorkflowWriter
@@ -53,14 +64,14 @@ func NewBufferWriter() *BufferWriter {
 }
 
 // Write is a fake implementation that stores content in the buffer.
-func (bw *BufferWriter) Write(path string, raw []byte) error {
+// If override is false and the buffer has content, ErrWorkflowExists is returned.
+func (bw *BufferWriter) Write(path string, raw []byte, override bool) error {
+	if !override && bw.Buffer != nil && bw.Buffer.Len() > 0 {
+		return ErrWorkflowExists
+	}
+
 	bw.Path = path
 	bw.Buffer.Reset()
 	_, err := bw.Buffer.Write(raw)
 	return err
 }
-
-// Exist is a fake implementation that returns true if the buffer has content.
-func (bw *BufferWriter) Exist(_ string) bool {
-	return bw.Buffer != nil && bw.Buffer.Len() > 0
-}

```

cmd/ci/writer.go
twoGiants: > nit: Should not this also check that data were written to appropriate file?

Yes, but it's a test double, a fake which does not have the exact same implementation / behavior. Its just good enough for the tests and the comment says exactly what it does.

> If we call bw.Write("flow-a.yaml", flowYaml) and then bw.exists("flow-b.yaml") this will true but I don't think that should be the case.

Yes, you're right and I agree. It was just not needed for the current test cases and I opted for keeping it simple and add a clarifying comment.

I would have to add a map for the paths and update the Exist and Write methods and a few unit tests.

Wdyt should I do it? It's no problem I'd do it in the next feature as a refactoring then.

If you and Claude point this out, it's probably good to fix :smiley_cat:

cmd/ci/writer.go
twoGiants: Should I check if it's a symbolic link and return an error?

I also think it should not ever be a symlink. The user would have to specifically create a symlink and name it the same as the yaml file and put into that directory replacing the original. :thinking: Do you think that would happen? Never trust the user :rofl:

twoGiants:
> Just thinking out loud.
> 
> Just matter of taste: I would prefer no adding the `Exist()` function. Rather I would make the `Write()` function to take an additional parameter `override` and let it handle possible override error. See patch below.
> 
> To me it feels more "atomic" as it does not do separate `check & create`, it simply just tries to create a file and possibly fails. Obviously this is not really important distinction, if we run two concurrent func at the same time we are in a trouble anyway. The fact that one instance of func may created the wf file while the other is in between `exits` and `write` is not really a problem.

Great comment! :smiley_cat: :+1: 

I like that you bring that up. I had the same idea when I started coding but then changed my mind. Here is my path of thoughts I was the following:
Writing a file/dir, reading a file/dir, checking file/dir existence, etc are atomic low level operations. It feels right to me keeping them separate in a kind of library and that is what Go does. Go keeps them separate in it's std lib. 
The GitHub Workflow generator is a higher level abstraction which happens to be working with files/dirs but has it's own notion how and in what order they are supposed to be used to achieve its goals. First thing that the GitHub Workflow generator decided is that the file always(!) goes into a specific directory and that's why `Write` creates the dir and the file in one method. The second thing GitHub Workflow generator decided is that it checks for an existing file IF and only IF there is an additional function parameter set. So now we could've put the check into `Write` but there was a fourth decision to print a message to the terminal IF and only IF there is an existing file already and the additional function parameter is set. And now to have everything in `Write` we would also have to forward the message writer `m io.Writer` to the `Write` method and putting logic into an infrastructure wrapper which actually is the core business logic of the GitHub Workflow generator.

To finish up and sum up my heuristic => I would prefer to keep business logic and infrastructure logic as separate as possible (not always possible I know). So if I would take my heuristic to the "extreme" I'd decompose `Write` and put even the `MkdirAll` in a separate method and use in the `Export` which is where the actual business logic lives I implement.

Now something else. In your patch this check should only happen when force is set and there is a file. So an accidental setting of force could create a bit of confusion and print a warning where none is needed:

```diff
+	if force {
+		// best-effort user message; errors are non-critical
+		_, _ = fmt.Fprintf(m, "WARNING: --force flag is set, overwriting existing GitHub Workflow file\n")
+	}
``` 

> ...if we run two concurrent func at the same time we are in a trouble anyway. 

Do you think this could be an issue? They would also have to be writing in the same directory.

lkingland:
/lgtm
Nice refinements

cmd/ci/writer.go
matejvasek: no need to fix it now

cmd/ci/writer.go
matejvasek: Let assume it's always our regular file.

Issue #3406, opened 2026-01-29
Add middleware-version image label to all builders


gauron99:
Set image labels of middleware into the function image so that its available to be fetched on cluster via the function operator. All three builders should add the middleware-version label to the function image, allowing users and
 tooling to inspect which middleware version was used during the build.

- [x] s2i
- [x] buildpack
- [x] host

/assign

PR #3408, opened 2026-01-30
add middleware version as image label

gauron99:
Add the middleware-version image label to function images built by S2I and Pack builders to preserve complete the implementation. OCI already implements this (only added unit test since we can inspect the actual blobs/files created)
- label is only set when non-empty

1. `func build --builder=s2i --push`
2. `Docker inspect...`
3.  Labels contain: `"middleware-version": "v0.21.3",`

PS:
some pack tests needed proper directory root because middleware fetch uses actual path detection for those templates (so it needs proper function root)

PSS: the s2i was added prior by Christoph but removed by me in the XXL scaffolding PR. Re-adding it here.

/kind enhancement
/fixes https://github.com/knative/func/issues/3406

```release-note
All function images now include the `middleware-version` label build by standard deployment
```

### DIFF
--- a/pkg/builders/middleware_labels_int_test.go
+++ b/pkg/builders/middleware_labels_int_test.go
@@ -1,0 +1,105 @@
+//go:build integration
+
+package builders_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/docker/docker/client"
+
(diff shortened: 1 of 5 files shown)


creydr:
@gauron99 can we backport this fix to 1.21?

PR #3426, opened 2026-02-09
Fix OpenShift detection logic

creydr:
Currently the OpenShift detection logic fails / returns IsOpenShift() == true, when the user does not have permissions to list namespaces.
This PR addresses it and queries the more low level api-resources API which _usually_ requires less permissions and checks if the `route.openshift.io` group exists.


gauron99 (commented):
/approve
/lgtm

PR #3427, opened 2026-02-09
Update docs with RBAC for raw & keda remote deploys

creydr:
Updating the docs with the require roles & rolebindings to make the remote deployments with the raw & keda deployer work

### DIFF
--- a/docs/building-functions/on_cluster_build.md
+++ b/docs/building-functions/on_cluster_build.md
@@ -13,37 +13,13 @@ kubectl apply -f https://storage.googleapis.com/tekton-releases/pipeline/previou
 ```
 
 ## Enabling a namespace to run Function related Tekton Pipelines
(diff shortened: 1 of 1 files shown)


gauron99:
If I understand correctly - on changing deployer this would require removing the old permissions to set new permissions. Should we also include a "one fits all" setup which could set all permissions for all deployers? 
Or do you think its too permission-less this way ?

creydr:
> If I understand correctly - on changing deployer this would require removing the old permissions to set new permissions. Should we also include a "one fits all" setup which could set all permissions for all deployers? Or do you think its too permission-less this way ?

TBH: I was wondering the same, but somehow didn't follow up on this. Let me add a solution for all deployers too

gauron99:
thanks! awesome
/lgtm
/approve

PR #3435, opened 2026-02-13
fix: install podman 5.x from Ubuntu Plucky in CI & fix its runs-on value

gauron99:
This PR introduces 2 things:
1) install podman 5.X from Ubuntu Plucky to fix API version mismatch of Docker engine
#### why?
- [GH Action runner updated to docker v29 on Feb 9](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) -> this raises minimum API to 1.44
-  podman 4.X (what we install currently on main) only provides docker engine API 1.41 -> with this PR we raise that to podman 5.X which has the minimal required 1.44

[original failed podman test in CI](https://github.com/knative/func/actions/runs/21971697863/job/63474860698?pr=3433)

2) fix `runs-on: ${{ matrix.os }}` so that this runs as intended on both OSs

### DIFF
--- a/.github/workflows/functions.yaml
+++ b/.github/workflows/functions.yaml
@@ -248,12 +248,12 @@ jobs:
   test-e2e-podman:
     name: E2E - Podman
     needs: precheck
-    runs-on: ubuntu-latest
     strategy:
       matrix:
         os:
           - "ubuntu-latest"      # x86_64
           - "ubuntu-24.04-arm"   # ARM64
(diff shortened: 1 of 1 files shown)


gauron99:
This seems to have been fixed by itself. Either ubuntu podman bump to match api versions or buildpack update (downgrade to support under 1.44 docker engine api)
/close

PR #3436, opened 2026-02-13
Move S2I scaffolding to .func/build/ to match other builders

gauron99:
- Move the generated `.s2i/*` files during s2i build with scaffold-enabled runtimes into `.func/build` to align it with the other builders. All scaffolders now have the same destination :tada: 
- Update Tekton pipelines to use custom assemble scripts for Go/Python
### side effect of this change
- pre-existing bug on integration test for private git repo `TestInt_PrivateGitRepository`:

This test on main is actually a false-positive -- go function is built via `pack` builder with Java buildpack! (because s2i runs first, pre-existing `root/.s2i/assemble` script which gets detected via pack ([current main latest integration job](https://github.com/knative/func/actions/runs/22227573049/job/64298869985)).
```
=== RUN   TestInt_PrivateGitRepository/pack
...
===> DETECTING
Paketo Buildpack for Quarkus 0.8.0
...
===> BUILDING
Paketo Buildpack for BellSoft Liberica 11.5.4
Paketo Buildpack for Syft 2.29.1
Paketo Buildpack for DistZip 5.11.1
  Process types:
    dist-zip: /workspace/.s2i/bin/assemble
    task:     /workspace/.s2i/bin/assemble
    web:      /workspace/.s2i/bin/assemble
```
This `DistZip` buildpack sees pre-existing .s2i/bin/assemble executable and is detected as "java thing"
 
Its wrong but builds successfuly, tests succeeds --> This test revealed the issue - therefore this PR adds go.mod/go.sum merging for Paketo compatibility

### DIFF
--- a/pkg/pipelines/tekton/templates.go
+++ b/pkg/pipelines/tekton/templates.go
@@ -29,9 +29,9 @@ const (
 	// S2I related properties
 	defaultS2iImageScriptsUrl = "image:///usr/libexec/s2i"
 	quarkusS2iImageScriptsUrl = "image:///usr/local/s2i"
-	// URL (not a filesystem path), so forward slashes are used instead of filepath.Join.
+	// Note: function calls (like path.Join()) can't be used in const declaration
+	// therefore this URL is built via concatenation.
 	scaffoldedS2iImageScriptsUrl = "file://" + fn.RunDataDir + "/" + fn.BuildDir + "/bin"
-
 	// The branch or tag we are targeting with Pipelines (ie: main, refs/tags/*)
 	defaultPipelinesTargetBranch = "main"
 )
--- a/pkg/s2i/builder.go
+++ b/pkg/s2i/builder.go
@@ -169,16 +169,19 @@ func (b *Builder) Build(ctx context.Context, f fn.Function, platforms []fn.Platf
 		// Excludes
 		// Do not include .git, .env or any language-specific cache directories
 		// (node_modules, etc) in the tar file sent to the builder, as this both
-		// bloats the build process and can cause unexpected errors in the resultant
-		// function. Note: .func is NOT excluded as it contains scaffolding for go/python.
-		ExcludeRegExp: "(^|/)\\.git|\\.env|node_modules(/|$)",
+		// bloats the build process and can cause unexpected errors in the
+		// resultant function.
+		// Note: .func/build is allowed through (scaffolding for go/python), but
+		// other .func subdirs (blob-cache, built-hash, etc.) are excluded to
+		// avoid bloating the tar and leaking local metadata.
+		ExcludeRegExp: "(^|/)\\.git|\\.env|\\.func/(blob-cache|built-hash|built-image|built\\.log|local\\.yaml)|node_modules(/|$)",
 	}
 
 	// Tell S2I where to find the assemble script written by Scaffolder.
-	// Use an absolute path because s2i's FileURLReader resolves file:// URLs
-	// relative to the process CWD, not relative to the source directory.
+	// Use forward slashes (not filepath.Join) because this is a file:// URL,
+	// and filepath.Join would use OS-specific separators.
 	if f.HasScaffolding() {
-		cfg.ScriptsURL = "file://" + filepath.Join(f.Root, fn.RunDataDir, fn.BuildDir, "bin")
+		cfg.ScriptsURL = "file://" + f.Root + "/" + fn.RunDataDir + "/" + fn.BuildDir + "/bin"
 	}
 
 	// Set middleware version label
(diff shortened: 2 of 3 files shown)


pkg/oci/scaffolder.go
```diff
@@ -26,9 +26,9 @@ func NewScaffolder(verbose bool) *Scaffolder {
 // Scaffold the function so that it can be built via oci builder.
 // 'path' is an optional override. Assign "" (empty string) most of the time
 func (s Scaffolder) Scaffold(ctx context.Context, f fn.Function, path string) error {
-	if f.Runtime != "go" && f.Runtime != "python" {
```
lkingland, line 29: Nitpick:  Perhaps `HasScaffolding`

pkg/s2i/builder.go
```diff
@@ -158,31 +158,43 @@ func (b *Builder) Build(ctx context.Context, f fn.Function, platforms []fn.Platf
...
-		return fmt.Errorf("cannot get middleware version: %w", err)
-	}
-	if middlewareVersion != "" {
-		cfg.Labels = map[string]string{fn.MiddlewareVersionLabelKey: middlewareVersion}
+		// Add middleware version label
+		embeddedRepo, repoErr := fn.NewRepository("", "")
+		if repoErr != nil {
+			return fmt.Errorf("unable to load the embedded scaffolding: %w", repoErr)
+		}
+		middlewareVersion, mwErr := scaffolding.MiddlewareVersion(f.Root, f.Runtime, f.Invoke, embeddedRepo.FS())
```
lkingland: We might want to put a note here about _why_ ony languages which have scaffolding get their middlewareVersion populated.

I suspect it's because, if it's out of date, the operator will try rebuilding it repeatedly.

pkg/s2i/builder_test.go
```diff
@@ -356,3 +357,48 @@ func (n notFoundErr) NotFound() {
...
+		t.Fatal(err)
+	}
+
+	// Call Scaffold
+	if err := scaffolder.Scaffold(context.Background(), f, ""); err != nil {
+		t.Fatal(err)
+	}
+
+	// Assert: scaffolding should be in .func/build/
+	expectedPath := root + "/" + fn.RunDataDir + "/build"
```
lkingland: Might want to use filepath.Join for consistency

matejvasek:
note: podman failure is related to this PR, override it

pkg/s2i/builder.go
gauron99: actually even non-scaffolded languages get middleware popuplated... its taken straight from the template! thanks for making me notice that!

gauron99:
@lkingland @matejvasek PTAL :pray:

pkg/pipelines/tekton/templates.go
```diff
@@ -28,6 +29,8 @@ const (
 	// S2I related properties
 	defaultS2iImageScriptsUrl = "image:///usr/libexec/s2i"
 	quarkusS2iImageScriptsUrl = "image:///usr/local/s2i"
+	// URL (not a filesystem path), so forward slashes are used instead of filepath.Join.
+	scaffoldedS2iImageScriptsUrl = "file://" + fn.RunDataDir + "/" + fn.BuildDir + "/bin"
```
matejvasek, line 34: btw `path.Join()` uses forward slashes.

matejvasek:
Just: what about older functions, should we not clean .s2i up?

gauron99:
yes this is good point. Maybe I could add a warning when using s2i builder that the scaffolding has been moved and its advised to delete the `f.Root/.s2i` directory? 

I dont want to just start deleting things from users function... although in this case I guess its ok since its all generated.

Im for either or both -- warning / delete.

pkg/pipelines/tekton/templates.go
gauron99: the `scaffoldedS2iImageScriptsUrl` is a constant so when using `.Join` i get err `(value of type string) is not constant [InvalidConstInit]`

also i liked that since its a URL im not using OS-specific separators (i thought that's what it would use) and wanted to enforce the `/`

gauron99: although i guess the comment was the confusing part? i can change that too...

lkingland: "[This is a URL] not a filesystem path, so forward slashes..." perhaps

lkingland:
> yes this is good point. Maybe I could add a warning when using s2i builder that the scaffolding has been moved and its advised to delete the `f.Root/.s2i` directory?
> 
> I dont want to just start deleting things from users function... although in this case I guess its ok since its all generated.
> 
> Im for either or both -- warning / delete.

That would be a nice one to add as a "good first issue" separately.

Issue #2172, opened 2024-02-20
when redeploying function in openshift dont force its internal registry 


gauron99:
This is a followup for https://github.com/knative/func/pull/1962

when usign openshift cluster its useful to use its internal registry when deploying images. But you might want to use your own image in different registry like (docker.io etc). One can deploy with `f deploy --registry=docker.io/XXX` and the correct registry will be used. BUT when redeploying after, openshift registry is used overriding the existing one.

Are you willing to work on this yourself?
yes

either simple logic refactor for redeploy or a complete registry derivation change is needed here

PR #3445, opened 2026-02-19
bug-fix: preserve registry on OpenShift namespace change

gauron99:
OS registry was being forced when working on OS cluster but wanted to have deployed a function which has image in different registry eg. `docker.io`. On namespace change given `--namespace` this would instead force the OS internal registry to be used ignoring the previously given one. This PR ensures that the registry is preserved in this scenario and the resolved `f.Registry` is taken into consideration.

Introducing
```
func SetOpenShiftForTest(val bool) func() {
	checkOpenShiftOnce.Do(func() {}) // ensure real detection won't run
	prev := isOpenShift
	isOpenShift = val
	return func() { isOpenShift = prev }
}
```
to test openshift without running the real detection
- regression tests included

/kind bug
/fix https://github.com/knative/func/issues/2172

**Release Note**

```release-note
bug-fix: dont force openshift registry when changing namespace
```

### DIFF
--- a/pkg/k8s/openshift.go
+++ b/pkg/k8s/openshift.go
@@ -5,6 +5,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"errors"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


gauron99:
@matejvasek @lkingland PTAL

pkg/k8s/openshift.go
```diff
@@ -126,6 +133,15 @@ func GetOpenShiftDockerCredentialLoaders() []creds.CredentialsCallback {
 var isOpenShift bool
 var checkOpenShiftOnce sync.Once
 
+// SetOpenShiftForTest overrides OpenShift detection for testing.
+// Returns a cleanup function that restores the previous state.
+func SetOpenShiftForTest(val bool) func() {
```
matejvasek, line 138: Can't we get rid of the cleanup return? Just use t.Cleanup() ?

gauron99: cant, it is in non test file (testing pkg would need to be imported) and it needs to be there because of the global variables it accesses there like `isOpenshift` and `checkOpenShiftOnce`

PR #3459, opened 2026-02-27
Feature: Add test step to CI workflow.

twoGiants:
# Changes

- :gift: Add `--test-step` flag to `func config ci`, enabled by default, disable with `--test-step=false`
- :gift: Generate a "Run tests" step in the GitHub Actions workflow that executes before deployment
- :gift: Select the test command based on function runtime: `go test ./...` for Go, `npm test` for Node.js, `python -m pytest` for Python
- :gift: Skip the test step with a warning for unsupported runtimes

/kind enhancement

Relates to #3256

**Release Note**

```release-note
New `--test-step` flag for `func config ci` adds a runtime-appropriate test step to the CI workflow before deployment.
```

**Docs**

```docs

```

### DIFF
--- a/cmd/ci/workflow.go
+++ b/cmd/ci/workflow.go
@@ -41,9 +41,10 @@ type step struct {
 	With map[string]string `yaml:"with,omitempty"`
 }
 
-func NewGitHubWorkflow(conf CIConfig) *githubWorkflow {
+func NewGitHubWorkflow(conf CIConfig, runtime string, messageWriter io.Writer) *githubWorkflow {
 	var steps []step
 	steps = createCheckoutStep(steps)
+	steps = createRuntimeTestStep(conf, steps, runtime, messageWriter)
 	steps = createK8ContextStep(conf, steps)
 	steps = createRegistryLoginStep(conf, steps)
 	steps = createFuncCLIInstallStep(steps)
@@ -61,23 +62,36 @@ func NewGitHubWorkflow(conf CIConfig) *githubWorkflow {
 	}
 }
 
-func createPushTrigger(conf CIConfig) workflowTriggers {
-	result := workflowTriggers{
-		Push: &pushTrigger{Branches: []string{conf.Branch()}},
-	}
-
-	if conf.UseWorkflowDispatch() {
-		result.WorkflowDispatch = &struct{}{}
-	}
-
-	return result
-}
-
 func createCheckoutStep(steps []step) []step {
 	checkoutCode := newStep("Checkout code").
 		withUses("actions/checkout@v4")
 
 	return append(steps, *checkoutCode)
+}
+
+func createRuntimeTestStep(conf CIConfig, steps []step, runtime string, messageWriter io.Writer) []step {
+	if !conf.TestStep() {
+		return steps
+	}
+
+	testStep := newStep("Run tests")
+
+	switch runtime {
+	case "go":
+		testStep.withRun("go test ./...")
+	case "node", "typescript":
+		testStep.withRun("npm ci && npm test")
+	case "python":
+		testStep.withRun("pip install . && python -m pytest")
+	case "quarkus":
+		testStep.withRun("./mvnw test")
+	default:
+		// best-effort user message; errors are non-critical
+		_, _ = fmt.Fprintf(messageWriter, "WARNING: test step not supported for runtime %s\n", runtime)
+		return steps
+	}
+
+	return append(steps, *testStep)
 }
 
 func createK8ContextStep(conf CIConfig, steps []step) []step {
@@ -90,7 +104,7 @@ func createK8ContextStep(conf CIConfig, steps []step) []step {
 }
 
 func createRegistryLoginStep(conf CIConfig, steps []step) []step {
-	if !conf.UseRegistryLogin() {
+	if !conf.RegistryLogin() {
 		return steps
 	}
 
@@ -114,18 +128,30 @@ func createFuncCLIInstallStep(steps []step) []step {
 
 func createFuncDeployStep(conf CIConfig, steps []step) []step {
 	runFuncDeploy := "func deploy"
-	if conf.UseRemoteBuild() {
+	if conf.RemoteBuild() {
 		runFuncDeploy += " --remote"
 	}
 
 	registryUrl := newVariable(conf.RegistryUrlVar())
-	if conf.UseRegistryLogin() {
+	if conf.RegistryLogin() {
 		registryUrl = newVariable(conf.RegistryLoginUrlVar()) + "/" + newVariable(conf.RegistryUserVar())
 	}
 	deployFunc := newStep("Deploy function").
 		withRun(runFuncDeploy + " --registry=" + registryUrl + " -v")
 
 	return append(steps, *deployFunc)
+}
+
+func createPushTrigger(conf CIConfig) workflowTriggers {
+	result := workflowTriggers{
+		Push: &pushTrigger{Branches: []string{conf.Branch()}},
+	}
+
+	if conf.WorkflowDispatch() {
+		result.WorkflowDispatch = &struct{}{}
+	}
+
+	return result
 }
 
 func newStep(name string) *step {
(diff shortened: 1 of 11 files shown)


twoGiants:
/cc @matejvasek @lkingland @gauron99 

Next one is ready for review! :smile_cat: :+1:

cmd/config_ci_test.go
```diff
@@ -525,6 +529,71 @@ func TestNewConfigCICmd_VerboseAndPostExportMessageAreMutuallyExclusive(t *testi
 	})
 }
 
+func TestNewConfigCICmd_TestStepPerRuntime(t *testing.T) {
```
gauron99, line 532: should we also add go runtime here for completeness?

cmd/ci/workflow.go
```diff
@@ -41,9 +41,10 @@ type step struct {
 	With map[string]string `yaml:"with,omitempty"`
 }
 
-func NewGitHubWorkflow(conf CIConfig) *githubWorkflow {
+func NewGitHubWorkflow(conf CIConfig, runtime string, messageWriter io.Writer) *githubWorkflow {
```
gauron99, line 44: would it make sense here to have runtime as part of the `CIConfig` as well? not sure about the `messageWriter`

cmd/ci/workflow.go
```diff
@@ -61,25 +62,36 @@ func NewGitHubWorkflow(conf CIConfig) *githubWorkflow {
...
+	testStep := newStep("Run tests")
+
+	switch runtime {
+	case "go":
+		testStep.withRun("go test ./...")
+	case "node":
+		testStep.withRun("npm test")
+	case "python":
+		testStep.withRun("python -m pytest")
+	default:
```
matejvasek, line 88: Please add `quarkus` -- `./mvnw test`.

cmd/ci/workflow.go
```diff
@@ -61,25 +62,36 @@ func NewGitHubWorkflow(conf CIConfig) *githubWorkflow {
...
+		return steps
+	}
+
+	testStep := newStep("Run tests")
+
+	switch runtime {
+	case "go":
+		testStep.withRun("go test ./...")
+	case "node":
+		testStep.withRun("npm test")
```
matejvasek: Don't we need to call `npm install` first?

cmd/ci/workflow.go
```diff
@@ -61,25 +62,36 @@ func NewGitHubWorkflow(conf CIConfig) *githubWorkflow {
...
+	case "python":
+		testStep.withRun("python -m pytest")
```
matejvasek: Don't we need to call `pip instal…` first?

cmd/config_ci_test.go
twoGiants: It is tested implicitly with the default options and I should've added a comment saying that. But you're right mentioning it explicitly. I'll add it, it's better to have it explicitly.

cmd/ci/workflow.go
twoGiants: `messageWriter` probably not, but possible, it's probably not go style but normal OOP style. `runtime` maybe. I didn't do it because the `CIConfig` is created before loading the function from the fs. We need the `path` from the `CIConfig` to load the function and only when we loaded the function we have the `runtime`. I don't want to add a setter to the config and make it mutable and pass `runtime` to it after it's created. One way of doing it would be having the `CIConfig` factory also take care of loading the function because it needs the `Root` and `Runtime` anyway. 

Wdys?

cmd/ci/workflow.go
twoGiants: We do, thx!

cmd/ci/workflow.go
twoGiants: We do, thx!

cmd/ci/workflow.go
gauron99: I see. since we now need 2 items here (runtime+root), my thinking is that adding a `CIConfig` factory is a little overkill probably and it can remain as is. Maybe if we were to add more of these arguments they could all be resolved at start via this factory.

I would keep it as is and propose a followup PR or make this part of next revamp?

twoGiants: > ...adding a CIConfig factory is a little overkill...

It's already there. The [`NewCIConfig`](https://github.com/twoGiants/knative-func/blob/93278b3b2d64dbe95b437f64214b6982169a78e9/cmd/ci/config.go#L88) function is the `CIConfig` factory function I was talking about.

We would've need to provide the `loader` to the factory so that it can load the function from the fs and get the needed Root and Runtime. Can do that in a follow up PR.

gauron99: Aha! shouldve checked the unchanged code :D ok thanks :)

cmd/config_ci_test.go
```diff
@@ -525,6 +529,86 @@ func TestNewConfigCICmd_VerboseAndPostExportMessageAreMutuallyExclusive(t *testi
...
+	}{
+		{
+			name:        "go runtime adds go test step",
+			runtime:     "go",
+			expectedRun: "go test ./...",
+		},
+		{
+			name:        "nodejs runtime adds npm test step",
+			runtime:     "node",
+			expectedRun: "npm i && npm test",
```
lkingland: I seem to remember something about an `npm ci` command specific for this use-case?

twoGiants: Oh yes of course! :see_no_evil: 

Updated it!

matejvasek:
Doc `docs/reference/func_config_ci.md` is not regenerated.

twoGiants:
> Doc `docs/reference/func_config_ci.md` is not regenerated.

Running `./hack/update-codegen.sh` did not produce any changes. But the CI should've caught this anyway, or? @matejvasek

matejvasek:
> > Doc `docs/reference/func_config_ci.md` is not regenerated.
> 
> Running `./hack/update-codegen.sh` did not produce any changes. But the CI should've caught this anyway, or? @matejvasek

The command (and therefore docs gen) is behind feature flag `export FUNC_ENABLE_CI_CONFIG=true`.

`````patch

diff --git a/docs/reference/func_config_ci.md b/docs/reference/func_config_ci.md
index 1c87a5fbd..74cfd1748 100644
--- a/docs/reference/func_config_ci.md
+++ b/docs/reference/func_config_ci.md
@@ -9,17 +9,21 @@ func config ci
 ### Options
 
 ```
-      --branch string                             Use a custom branch name in the workflow (default "main")
+      --branch string                             Use a custom branch name in the workflow
+      --force                                     Use to overwrite an existing GitHub workflow
   -h, --help                                      help for ci
       --kubeconfig-secret-name string             Use a custom secret name in the workflow, e.g. secret.YOUR_CUSTOM_KUBECONFIG (default "KUBECONFIG")
   -p, --path string                               Path to the function.  Default is current directory ($FUNC_PATH)
+      --platform string                           Pick a CI/CD platform for which a manifest will be generated. Currently only GitHub is supported. (default "github")
+      --registry-login                            Add a registry login step in the github workflow (default true)
       --registry-login-url-variable-name string   Use a custom registry login url variable name in the workflow, e.g. vars.YOUR_REGISTRY_LOGIN_URL (default "REGISTRY_LOGIN_URL")
       --registry-pass-secret-name string          Use a custom registry pass secret name in the workflow, e.g. secret.YOUR_REGISTRY_PASSWORD (default "REGISTRY_PASSWORD")
       --registry-url-variable-name string         Use a custom registry url variable name in the workflow, e.g. vars.YOUR_REGISTRY_URL (default "REGISTRY_URL")
       --registry-user-variable-name string        Use a custom registry user variable name in the workflow, e.g. vars.YOUR_REGISTRY_USER (default "REGISTRY_USERNAME")
       --remote                                    Build the function on a Tekton-enabled cluster
       --self-hosted-runner                        Use a 'self-hosted' runner instead of the default 'ubuntu-latest' for local runner execution
-      --use-registry-login                        Add a registry login step in the github workflow (default true)
+      --test-step                                 Add a language-specific test step (supported: go, node, python) (default true)
+  -v, --verbose                                   Print verbose logs ($FUNC_VERBOSE)
       --workflow-name string                      Use a custom workflow name (default "Func Deploy")
 ```
 

`````

PR #3470, opened 2026-03-04
use python3 instead of hardcoded python version in assembler

gauron99:
Dont hardcode the python minor version in s2i assemble script. This way the version can be resolved based on the builder-image


gauron99:
@matejvasek @lkingland PTAL

PR #3473, opened 2026-03-05
fix: e2e test naming to match category conventions + add verbose flag to 'func subscribe'

gauron99:
Some e2e test never actually ran!... fix this for e2e scripts accept a naming convention
Had to create a verbose flag for `func subscribe` since it was lacking the implementation as well


jrangelramos:
apparently it doesn't like `-v` on the `func subscribe` command (verbose flag is not implemented there). Other than that it looks good to me.

PR #3477, opened 2026-03-09
fix: workaround podman bug wrt error reporting

matejvasek:
# Changes

In case that the image registry hostname is not resolvable podman now return error in bad JSON shape so it's not decodable by client.

For now we assume that `500.*but provided no error-message` means hostname resolution error.

/kind bug

```release-note
fix: cannot push to cluster internal registry with podman (error API returned a 500 (Internal Server Error) but provided no error-message)
```


gauron99 (commented):
/lgtm
/approve

PR #3479, opened 2026-03-09
Feature+Test: Add runtime-aware builder config and Config CI e2e tests.

twoGiants:
# Changes

- :gift: Add runtime-aware builder selection to generated GitHub Actions workflows — `func config ci` now sets `FUNC_BUILDER` based on the function runtime (host for Go/Python local, pack for Node/TypeScript/Rust/Quarkus/Springboot, s2i for Python remote) and uses environment variables (`FUNC_BUILDER`, `FUNC_REGISTRY`, `FUNC_REMOTE`, `FUNC_VERBOSE`) instead of CLI flags
- :gift: Add e2e tests using nektos/act to deploy functions via generated workflows against a real cluster for Go, Node, TypeScript, Python, and Quarkus runtimes
- :gift: Add `test-e2e-config-ci` GitHub Actions job, Makefile target, and `hack/test-full.sh` entry
- :gift: Install `act` binary via `hack/binaries.sh`
- :broom: Add missing `.PHONY` declarations for `test-e2e-matrix` and `test-full-logged`
- :broom: Refactor CIConfig to load the function upfront, encapsulating runtime and root path instead of deferring to callers
- :broom: Improve platform validation: reject empty values and include supported platforms in error messages
- :bug: Fix typos in CI workflow comments (`Cluser` → `Cluster`, `runtim` → `runtime`)

/kind enhancement

Relates to #3256

**Release Note**

```release-note
`func config ci` now generates workflows with runtime-aware builder selection (FUNC_BUILDER) and uses environment variables instead of CLI flags for deploy configuration.
```

**Docs**

```docs

```


twoGiants:
/kind enhancement

/cc @gauron99 @creydr @lkingland 

The e2e test are done with a few other features and refactorings. Check out the PR description for details.

Thank you for the review :smile_cat: :+1: .

cmd/ci/common.go
```diff
@@ -2,8 +2,30 @@ package ci
 
 import "fmt"
 
-func runner(conf CIConfig) string {
-	if conf.SelfHostedRunner() {
+func determineBuilder(runtime string, remote bool) (string, error) {
```
gauron99: Could we determine the builder in the `CIConfig`? make it resolve once instead of calling it in both during the DeployStep creation and Printing the config.

The builder needs runtime & deployment type combo (local/remote) which are both also part of the `CIConfig`. This should be resolvable at `CIConfig` struct creation, if Im not missing anything?

This would then simply add `fnBuilder` value to `CIConfig` struct

templates/typescript/cloudevents/test/integration.ts
```diff
@@ -3,7 +3,7 @@ import { CloudEvent, HTTP } from 'cloudevents';
 import { start, InvokerOptions } from 'faas-js-runtime';
 import request from 'supertest';
 
-import * as func from '../build';
+import * as func from '../src';
```
gauron99: adding this template change to this PR is perhaps a bit unfortunate (unless its needed for the config changes?) because the major change of `func config ci` needs a release note but I guess it should not be an issue.

twoGiants: It's a fix, same as for http. I wasn't able to run the typescript tests without this change, got a ts-node error.

Do you have a test executing the tests in the typescript function?

cmd/ci/common.go
twoGiants: Yes, good catch. It makes sense to put it there. I'll update it quick!

templates/typescript/cloudevents/test/integration.ts
twoGiants: I can add the fix to the release note too.

twoGiants: Will revert. Lets see if the CI passes.

cmd/ci/common.go
twoGiants: This one is here => https://github.com/knative/func/pull/3479/commits/da0f72fb18d9946d68bf3192ed70f7fbfd93d584

templates/typescript/cloudevents/test/integration.ts
twoGiants: This one is here => https://github.com/knative/func/pull/3479/commits/266830b60a082adb4ab65f104ca0e2d614ea087d

If the CI passes I'll rebase & squash.

gauron99:
All looks good to me! feel free to unhold whenever you please. @twoGiants

PR #3490, opened 2026-03-12
Remember `--registry-insecure` setting on recurring runs

### LINKED ISSUE #3489: `--registry-insecure` not remembered on recurring runs of `func deploy`
When I deploy a function with an insecure registry via `--registry-insecure` it works on the initial run. But on consecutive runs it is not "remembered" anymore:

Initial run works:
```
$ func deploy --registry localhost:5001 --registry-insecure
Building function image
🙌 Function built: localhost:5001/func-go-hello-world:latest
Pushing function image to the registry "localhost:5001" using the "" user credentials
warning: error while checking pull secrets: Get "https://localhost:5001/v2/": tls: failed to verify certificate: x509: certificate signed by unknown authority; GET http://localhost:5001/v2/: unexpected status code 400 Bad Request: Client sent an HTTP request to an HTTPS server.

🎯 Creating Triggers on the cluster
✅ Function deployed in namespace "default" and exposed at URL: 
   http://func-go-hello-world.default.svc.cluster.local
```

Afterwards when only `func deploy` is used, it is not "remembered":
```
$  func deploy                                              
function up-to-date. Force rebuild with --build
Error: failed to get credentials: creating push check transport for localhost:5001 failed: Get "https://localhost:5001/v2/": tls: failed to verify certificate: x509: certificate signed by unknown authority; GET http://localhost:5001/v2/: unexpected status code 400 Bad Request: Client sent an HTTP request to an HTTPS server.
```

I think this is, because the option is not stored in the `func.yaml`:
```
specVersion: 0.36.0
name: func-go-hello-world
runtime: go
registry: localhost:5001
namespace: default
created: 2025-11-17T11:20:29.114748843+01:00
build:
  builder: pack
deploy:
  namespace: default
  image: localhost:5001/func-go-hello-world@sha256:cb0b6d0ffb1d025f90da6a1d46c908f8f3d4bf10ec8e9608f228474c2470c21d
```

Also not sure about the `unexpected status code 400 Bad Request: Client sent an HTTP request to an HTTPS server`, while I'd expect `--registry-insecure` should only skip the TLS verification according to its docs. But that's probably a different issue

creydr:
Fixes #3489 

# Changes

- :bug: Remember `--registry-insecure` setting on recurring runs

Hint: This also reverts da21fa29868238a34d301f5e439a6dfe726a4c27, because when we have the `RegistryInsecure` in the function, this can be simplified.

### DIFF
--- a/cmd/build_registry_insecure_test.go
+++ b/cmd/build_registry_insecure_test.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"bytes"
+	"strings"
 	"testing"
 
 	"github.com/spf13/cobra"
@@ -185,4 +187,82 @@ func testRegistryInsecurePersists(cmdFn func(factory ClientFactory) *cobra.Comma
 			}
 		}
 	})
+}
+
+// TestWarnRegistryInsecureChange ensures that the warning is printed when
+// the registry changes but registryInsecure is still true.
+func TestWarnRegistryInsecureChange(t *testing.T) {
+	tests := []struct {
+		name            string
+		cfgRegistry     string
+		funcRegistry    string
+		funcInsecure    bool
+		expectWarning   bool
+		expectedMessage string
+	}{
+		{
+			name:          "no warning - registry not changed",
+			cfgRegistry:   "example.com/alice",
+			funcRegistry:  "example.com/alice",
+			funcInsecure:  true,
+			expectWarning: false,
+		},
+		{
+			name:          "no warning - registryInsecure is false",
+			cfgRegistry:   "example.com/bob",
+			funcRegistry:  "example.com/alice",
+			funcInsecure:  false,
+			expectWarning: false,
+		},
+		{
+			name:          "no warning - func registry is empty",
+			cfgRegistry:   "example.com/bob",
+			funcRegistry:  "",
+			funcInsecure:  true,
+			expectWarning: false,
+		},
+		{
+			name:          "no warning - cfg registry is empty",
+			cfgRegistry:   "",
+			funcRegistry:  "example.com/alice",
+			funcInsecure:  true,
+			expectWarning: false,
+		},
+		{
+			name:            "warning - registry changed and insecure is true",
+			cfgRegistry:     "example.com/bob",
+			funcRegistry:    "example.com/alice",
+			funcInsecure:    true,
+			expectWarning:   true,
+			expectedMessage: "Warning: Registry changed from 'example.com/alice' to 'example.com/bob', but registryInsecure is still true.",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			f := fn.Function{
+				Registry:         tt.funcRegistry,
+				RegistryInsecure: tt.funcInsecure,
+			}
+
+			// Capture output
+			var buf bytes.Buffer
+			warnRegistryInsecureChange(&buf, tt.cfgRegistry, f)
+
+			output := buf.String()
+
+			if tt.expectWarning {
+				if output == "" {
+					t.Fatal("expected warning but got none")
+				}
+				if tt.expectedMessage != "" && !strings.Contains(output, tt.expectedMessage) {
+					t.Fatalf("expected message to contain '%s', got '%s'", tt.expectedMessage, output)
+				}
+			} else {
+				if output != "" {
+					t.Fatalf("expected no warning but got: %s", output)
+				}
+			}
+		})
+	}
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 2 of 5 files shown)


cmd/build_registry_insecure_test.go
```diff
@@ -0,0 +1,183 @@
...
+		t.Fatal(err)
+	}
+
+	var (
+		builder = mock.NewBuilder()
+		pusher  = mock.NewPusher()
+	)
+
+	// Test 1: Initial state - registryInsecure should be false
+	t.Run("initial_state_is_false", func(t *testing.T) {
```
twoGiants: nit

No need for underscore in `t.Run` name argument. Go will add those underscores by itself when the tests are executed.

Same for all the others below.

```suggestion
	t.Run("initial state is false", func(t *testing.T) {
```

pkg/functions/function.go
```diff
@@ -81,6 +81,10 @@ type Function struct {
 	// [registry]/[user].
 	Registry string `yaml:"registry,omitempty"`
 
+	// RegistryInsecure defines it the TLS verification of the registry
```
twoGiants: nit

```suggestion
	// RegistryInsecure defines if the TLS verification of the registry
```

schema/func_yaml-schema.json
```diff
@@ -182,6 +182,10 @@
 					"type": "string",
 					"description": "Registry at which to store interstitial containers, in the form\n[registry]/[user]."
 				},
+				"registryInsecure": {
+					"type": "boolean",
+					"description": "RegistryInsecure defines it the TLS verification of the registry\nshould be skipped."
```
twoGiants: nit

```suggestion
					"description": "RegistryInsecure defines if the TLS verification of the registry\nshould be skipped."
```

twoGiants (approved):
This is really useful and it all looks good! 😸 👍 

I have only a few nits.

[edit]: Did you run `./hack/update-codegen.sh`?

/approve
/lgtm
/hold for other reviewers

pkg/config/config.go
```diff
@@ -70,6 +71,15 @@ func (c Global) RegistryDefault() string {
 	}
 }
 
+// WarnRegistryInsecureChange checks if the registry has changed but
+// registryInsecure is still set to true, and prints a warning if so.
+// This helps users avoid accidentally skipping TLS verification on a new registry.
+func (c Global) WarnRegistryInsecureChange(w io.Writer, f fn.Function) {
```
twoGiants: Unit tests missing.

pkg/functions/client.go
```diff
@@ -372,6 +373,14 @@ func WithRegistry(registry string) Option {
 	}
 }
 
+// WithRegistryInsecure sets if the TLS verification of the registry should
+// be skipped.
```
gauron99, line 377:
```suggestion
// WithRegistryInsecure sets if the TLS verification of the registry should
// be skipped.
//
// NOTE: Unlike WithRegistry and other Options, which act as a fallback (function's value
// takes precedence when set), WithRegistryInsecure(true) WILL override the 
// function's value unconditionally -> bool has no "unset" (unlike string's ""),
// so we cannot distinguish "not configured" from "explicitly set to false".
// As a result:
// - WithRegistryInsecure(true)  → always sets f.RegistryInsecure = true
// - WithRegistryInsecure(false) → preserves the function's existing value
//
// NOTE-2: gauron99 - If we ever need to make this consistent with the rest of Options
// RegistryInsecure needs to become *bool in Function's struct & global.Config struct
// (nil == unset == empty value) and it would most likely require special use of
// cmd.Flags.Changed() and viper.IsSet() (non-standard for rest of commands) 
```

pkg/functions/client.go
```diff
@@ -372,6 +373,14 @@ func WithRegistry(registry string) Option {
...
+func WithRegistryInsecure(insecure bool) Option {
```
twoGiants, line 391: Should we add a comment here, that  a `WithRegistryInsecure(false)` does not override function's existing `true`? Because one could think that this setting would set `registryInsecure` definitely to `false`.

pkg/functions/client.go
```diff
@@ -690,6 +699,11 @@ func (c *Client) Build(ctx context.Context, f Function, options ...BuildOption)
 		f.Registry = c.registry
 	}
 
+	if c.registryInsecure {
+		// only override it, when given via the client
```
gauron99:
```suggestion
        // Unlike other With* functions (like WithRegistry) this takes bool.
        // Bool has no "empty value" (like "" for string) therefore we always
        // override if client's value is true. 
        // See WithRegistryInsecure definition comment for more info.
```

pkg/config/config.go
```diff
@@ -138,6 +148,8 @@ func (c Global) Apply(f fn.Function) Global {
 	if f.Registry != "" {
 		c.Registry = f.Registry
 	}
+	c.RegistryInsecure = f.RegistryInsecure
```
gauron99, line 143:
```suggestion
    // Unconditional because bool has no "empty value". Works because
    // viper resolves the correct precedence via our defaulting.
	c.RegistryInsecure = f.RegistryInsecure
```

pkg/config/config.go
```diff
@@ -156,6 +168,8 @@ func (c Global) Configure(f fn.Function) fn.Function {
 	if c.Registry != "" {
 		f.Registry = c.Registry
 	}
+	f.RegistryInsecure = c.RegistryInsecure
```
gauron99, line 165:
```suggestion
    // Unconditional because bool has no "empty value". Works because
    // viper resolves the correct precedence via our defaulting.
	f.RegistryInsecure = c.RegistryInsecure
```

cmd/build_registry_insecure_test.go
creydr: updated

pkg/functions/function.go
creydr: good catch. updated

schema/func_yaml-schema.json
creydr: updated

pkg/config/config.go
creydr: Added one

pkg/functions/client.go
creydr: I accepted the suggestion from @gauron99 in https://github.com/knative/func/pull/3490#discussion_r2930302974

PR #3522, opened 2026-03-25
Podman next containerized

matejvasek:
# Changes

Podman next test runs in Fedora container.

### DIFF
--- a/Dockerfile.podman-next
+++ b/Dockerfile.podman-next
@@ -1,0 +1,27 @@
+FROM fedora:43
+
+RUN dnf -y copr enable rhcontainerbot/podman-next && \
+    dnf -y install podman socat go make git && \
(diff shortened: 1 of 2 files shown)


Dockerfile.podman-next
```diff
@@ -0,0 +1,27 @@
...
+
+RUN mkdir -p /etc/containers && \
+cat > /etc/containers/registries.conf <<EOF
+unqualified-search-registries = ["docker.io", "quay.io", "registry.fedoraproject.org", "registry.access.redhat.com"]
+short-name-mode="permissive"
+
+[[registry]]
+location="localhost:50000"
+insecure=true
+EOF
```
matejvasek, line 15: This part might be redundant as the test script sets this.

matejvasek: but for now leave it

PR #3538, opened 2026-03-27
ci: fix test-e2e-podman to use matrix.os instead of hardcoded runner

### LINKED ISSUE #3537: ci: test-e2e-podman ignores ARM matrix entry due to hardcoded runs-on
## Problem

In `.github/workflows/functions.yaml`, the `test-e2e-podman` job defines a strategy matrix with two OS entries:

```yaml
strategy:
  matrix:
    os:
      - "ubuntu-latest"    # x86_64
      - "ubuntu-24.04-arm" # ARM64
```

However, `runs-on` is hardcoded to `ubuntu-latest` instead of referencing the matrix:

```yaml
runs-on: ubuntu-latest
```

This means both matrix entries run on the same x86_64 runner, so the ARM64 variant is never actually tested on ARM hardware. CI time is wasted on a duplicate x86_64 run.

## Fix

Change `runs-on: ubuntu-latest` to `runs-on: ${{ matrix.os }}` so each matrix entry runs on the correct runner.

Ankitsinghsisodya:
## Summary

- Fixes `test-e2e-podman` job which defines a matrix with `ubuntu-latest` (x86_64) and `ubuntu-24.04-arm` (ARM64) but had `runs-on: ubuntu-latest` hardcoded
- Both matrix entries were running on x86_64, so ARM64 was never actually tested on ARM hardware
- Changed to `runs-on: ${{ matrix.os }}` so each matrix entry runs on the correct runner

## Test plan

- [ ] Verify the `E2E - Podman` job spawns two matrix runs: one on `ubuntu-latest` and one on `ubuntu-24.04-arm`
- [ ] Confirm both matrix entries complete successfully

Fixes #3537

### DIFF
--- a/.github/workflows/functions.yaml
+++ b/.github/workflows/functions.yaml
@@ -247,12 +247,12 @@ jobs:
   test-e2e-podman:
     name: E2E - Podman
     needs: precheck
-    runs-on: ubuntu-latest
     strategy:
       matrix:
         os:
           - "ubuntu-latest" # x86_64
           - "ubuntu-24.04-arm" # ARM64
+    runs-on: ${{ matrix.os }}
     env:
       FUNC_CLUSTER_RETRIES: 5
       FUNC_E2E_PODMAN: true


Ankitsinghsisodya:
@matejvasek Can you look into this PRs whenever you are free.

Ankitsinghsisodya:
@matejvasek sir, I think we can go for the merge since it passed all the test cases.

PR #3540, opened 2026-03-27
ci: add top-level permissions for least-privilege security

### LINKED ISSUE #3539: ci: add top-level permissions for least-privilege security in test-podman-next workflow
## Description

The `test-podman-next.yaml` workflow does not declare a top-level `permissions` block, which means the default `GITHUB_TOKEN` has broad read/write access. This is unnecessary — the workflow only needs `contents: read` for checkout and artifact upload.

## Proposed Fix

Add `permissions: contents: read` at the top level of `test-podman-next.yaml` to restrict the token to read-only access, following the principle of least privilege.

Ankitsinghsisodya:
## Summary

- Add `permissions: contents: read` to `test-podman-next.yaml` to restrict the default GITHUB_TOKEN to read-only access, following the principle of least privilege.

Fixes #3539

## Test plan

- [ ] Verify CI passes — no functional change, workflow only needs read access

### DIFF
--- a/.github/workflows/test-podman-next.yaml
+++ b/.github/workflows/test-podman-next.yaml
@@ -1,5 +1,8 @@
 name: Func Podman Next Test
 
+permissions:
+  contents: read
+
 on:
   schedule:
     - cron: '0 2 * * *'


matejvasek:
Decreasing privilege makes sense, however I don't see how the change in [.github/workflows/functions.yaml](https://github.com/knative/func/pull/3540/files#diff-4fea8201320c4fc7e7398b42336051af35414173237947560d2454a47da035e8) is related.

Ankitsinghsisodya:
@matejvasek I've removed the functions.yaml change from this PR. It's already covered in #3538 as a separate fix.

PR #3572, opened 2026-04-02
Refactor: Move CI workflow generation from `cmd/ci` to `pkg/ci/github`.

### LINKED ISSUE #3256: Implement Github Workflow generation for building and deploying a function to a k8 cluster.

twoGiants:
# Changes

- :gift: Move CI workflow generation from `cmd/ci/` to `pkg/ci/github/` as a self-contained, reusable package that can be imported by library consumers
- :gift: Introduce `CIGenerator` interface and `GenerateCIWorkflow` method on the client, routing `func config ci` through the client instead of calling the generator directly from the command layer
- :gift: `NewWorkflowGenerator` constructor with functional options (`WithWorkflowConfig`, `WithWorkflowWriter`, `WithMessageWriter`, `WithVerbose`) and sensible defaults -- works out of the box with zero configuration
- :broom: Introduce `ciGeneratorFactory` in the CLI layer to decouple CLI tests from generator internals -- CLI tests verify flag-to-config mapping via a mock, pkg tests verify workflow generation
- :broom: Migrate workflow structure tests from `cmd/config_ci_test.go` to `pkg/ci/github/generator_test.go`
- :broom: `WorkflowConfig` uses exported struct fields instead of getter methods, with `FnRuntime`/`FnRoot` sourced from the `Function` passed to `Generate`
- :bug: Fix typos: `--patform` -> `--platform`, `functionl` -> `function`, `proivdes` -> `provides`, `precldes` -> `precedes`, `sevices` -> `services`, `suppot` -> `support`, `Desribe` -> `Describe`, `pipilines` -> `pipelines`, `intance` -> `instance`

/kind enhancement

Closes #3256

**Release Note**

```release-note
CI workflow generation (`func config ci`) is now available as a reusable library package at `pkg/ci/github`. External consumers can import `github.NewWorkflowGenerator()` to programmatically generate GitHub Actions deployment workflows for Knative functions.
```

**Docs**

```docs

```


cmd/config_ci.go
```diff
@@ -1,175 +1,304 @@
...
 	if err != nil {
 		return err
 	}
 
-	githubWorkflow := ci.NewGitHubWorkflow(cfg, messageWriter)
-	if err := githubWorkflow.Export(cfg.FnGitHubWorkflowFilepath(), writer, cfg.Force(), messageWriter); err != nil {
+	client, done := newClient(
+		ClientConfig{Verbose: viper.GetBool(verboseFlag)},
+		fn.WithPathWriter(pathWriter),
+		fn.WithStdout(cmd.OutOrStdout()),
```
lkingland: These could be constructor arguments on the CI generator implementation.

pkg/ci/github/config.go
```diff
@@ -0,0 +1,54 @@
...
+	DefaultRegistryUrlVariableName      = "REGISTRY_URL"
+	DefaultRegistryLogin                = true
+	DefaultWorkflowDispatch             = false
+	DefaultRemoteBuild                  = false
+	DefaultSelfHostedRunner             = false
+	DefaultTestStep                     = true
+	DefaultForce                        = false
+)
+
+type Config struct {
```
lkingland: Perhaps turn these all (except FnRuntime and FnRoot) into members of the Generator, set as part of construction, and add `verbose`.  Then, for example, if this struct is renamed `WorkflowGeneratorOptions`, then the constructor becomes simply: `NewWorkflowGenerator(opts WorkflowGeneratorOptions) *WorkflowGenerator` and the `CI` interface in client becomes:
`Generate(ctx context.Context, f Function)`

The implementation then pulls the FnRuntime and FnRoot from `f`, and everything else from it's WorkflowGeneratorOptions struct.

pkg/functions/client.go
```diff
@@ -221,6 +224,16 @@ type MCPServer interface {
 	Start(context.Context, bool) error
 }
 
+type CI interface {
+	Generate(context.Context, any, PathWriter, io.Writer) error
```
lkingland: This is definitely the right pattern; keeping everything neatly gated behind an interface, and dependencies held at bay.
Simply moving most of the config to the concrete implementation's constructor (as well as the writers), and we can keep only the universal settings here.  That could yield an actualy gramatically correct interface:
"Generate a CI Workflow for function X" ==  `Generate(context.Context, Function) error`

pkg/functions/client.go
```diff
@@ -221,6 +224,16 @@ type MCPServer interface {
 	Start(context.Context, bool) error
 }
 
+type CI interface {
```
lkingland: While I know it's not nearly as pretty, the Go idiom is to use something like "CIGenerator" here.  That way it flows rather nicely:
```
generator := github.CIGenerator(...)
generator.Generate(ctx, f)
```

cmd/config_ci_test.go
lkingland: The CLI's job is to translate a request into an API call, so if you use a mock CIGenerator here to ensure flags show up in that struct's constructor as-expected, and that it's `Generate` method is invoked when expected, then the testing here in the CLI is complete!

Then, you can rely on the tests in the `github` package to ensure that the implementation is correct (writes a workflow file,  and writes it with the correct structure).  That might simplify things here.

cmd/config_ci_test.go
```diff
@@ -362,7 +362,7 @@ func TestNewConfigCICmd_PlatformFlagErrors(t *testing.T) {
 func TestNewConfigCICmd_ForceFlagOverwritesExistingWorkflow(t *testing.T) {
```
lkingland: Here's a concrete example!  Notice how the behavior is : the GitHub CI Generator overwrites an existing workflow file when `force` is enabled.

That's a test for the implementation itself.  The CLI needs to make sure that that `force` is set to `true` if `--force` is provided, and leave the rest up to the implementation.

This tests actually looks a lot like an E2E.  Where you'd actually call the cli command twice and confirm it overwrote interstital changes.

pkg/ci/github/printer.go
```diff
@@ -12,7 +12,7 @@ GitHub Workflow Configuration
   Workflow name:      %s
   Branch:             %s
   Builder:            %s
-  Remote build:       %s
+	Remote build        %s
```
lkingland: This looks like an inadvertent change

cmd/config_ci.go
twoGiants: Agree, done!

pkg/ci/github/config.go
twoGiants: Great! Thx, did mostly like that.

```go
type workflowGenerator struct {
	verbose        bool
	workflowWriter WorkflowWriter
	messageWriter  io.Writer
	cfg           WorkflowConfig
}

type WorkflowConfig struct {
	GithubWorkflowDir,
	GithubWorkflowFilename,
	// ...
}

type Option func(*workflowGenerator)

func WithWorkflowConfig(wc WorkflowConfig) Option {
	return func(wg *workflowGenerator) {
		wg.cfg = wc
	}
}

// ...

func NewWorkflowGenerator(options ...Option) *workflowGenerator {
	wg := &workflowGenerator{}
	for _, o := range options {
		o(wg)
	}

	return wg
}
```

pkg/functions/client.go
twoGiants: Thx, updated!

cmd/config_ci_test.go
twoGiants: Agree, understood. Working on it.

pkg/ci/github/printer.go
twoGiants: Fixed.

cmd/config_ci_test.go
twoGiants: All done.

cmd/config_ci_test.go
twoGiants: Understood and addressed.

> This tests actually looks a lot like an E2E. Where you'd actually call the cli command twice and confirm it overwrote interstital changes.

I actually have an integration test for that here => https://github.com/twoGiants/knative-func/blob/c49b404ba0b7656706b1b9e0eb728bb423b1f34a/cmd/config_ci_int_test.go#L89

:)

twoGiants:
Hey @lkingland. I am done with the re-work. All comments are addressed. Here is a little guide on where to start best.

## Review Guide

Start with the two interface changes. They define the contract everything else follows:

1. **`pkg/functions/client.go`**: `CIGenerator` interface, `WithCIGenerator` option, and `GenerateCIWorkflow` method. Addressed your comment.

2. **`pkg/ci/github/generator.go`**: The concrete implementation. Constructor with functional options and sensible defaults. `Generate(ctx, Function)` is the entry point.

3. **`pkg/ci/github/workflow.go`**: `WorkflowConfig` struct and defaults live here now. Workflow YAML construction was moved as is, no changes.

4. **`cmd/config_ci.go`**: CLI wiring. The key addition is `ciGeneratorFactory` which decouples the CLI from the generator implementation, enabling mock-based testing.

5. **`pkg/ci/github/generator_test.go`**: All workflow behavior tests live here now. `cmd/config_ci_test.go` only tests flag-to-config mapping via a mock.

pkg/ci/github/workflow.go
```diff
@@ -0,0 +1,334 @@
...
+	"path/filepath"
+
+	"gopkg.in/yaml.v3"
+)
+
+// ErrWorkflowExists is returned when a GitHub workflow file already exists and --force is not specified.
+var ErrWorkflowExists = errors.New("existing GitHub workflow detected, overwrite using the --force option")
+
+const (
+	defaultFuncCliVersion               = "knative-v1.22.0"
```
lkingland, line 17: We might want to see if we can source this from the version static defaults.  Future PR is fine.

lkingland (approved):
/approve
/lgtm

What a great addition. Glad to see it landing.  Thanks @twoGiants!

Issue #3514, opened 2026-03-24
Environment variables passed via -e flag not applied on deploy


lkingland:
Environment variables passed with `func deploy -e KEY=VALUE` are not
included in the deployed service spec. They work correctly with
`func run -e`, but on deploy the env vars are missing from the
container spec.

Expected: `-e MYVAR=value` on deploy sets the env var in the Knative
service container spec, the same way it does for local runs.

Related: [SRVOCF-650](https://redhat.atlassian.net/browse/SRVOCF-650)

lkingland:
This can be closed by adding an E2E **matrix test** that:

1. Creates a new Function which returns (in its response) whether a given env variable was set in its environment.
2. Runs `func deploy -e KEY=VALUE` against it, then invokes the deployed function and asserts the env reached the container.

Because it's a matrix test, the assertion runs for **every builder × OS × language combination** we support — so we're verifying that each builder's distinct env-handling path is wired up correctly, and any future regression in any one builder's env-applying logic has not regressed.

Checked existing coverage: `TestMetadata_Envs_Add` in `e2e/e2e_metadata_test.go:37` covers the `func config envs add` path end-to-end (deploy → invoke → `os.LookupEnv`), but only for Go. The matrix deploy test (`e2e/e2e_matrix_test.go:82`) covers builders/runtimes but doesn't assert envs. Unit coverage at `cmd/deploy_test.go:386` only asserts the env round-trips to `func.yaml`. Nothing covers the `-e` flag → runtime path across builders. That's the gap to close.

PR #3579, opened 2026-04-03
fix: pass referenced resource sets to generateNewService for proper validation

Ankitsinghsisodya:
# Changes

`generateNewService` in `pkg/knative/deployer.go` was allocating its own local
`referencedSecrets`, `referencedConfigMaps`, and `referencedPVCs` sets
internally. Those sets were never visible to the caller, so
`CheckResourcesArePresent` on the first-deploy path was always invoked with
empty sets, silently passing validation regardless of whether referenced
Secrets, ConfigMaps, or PVCs actually existed on the cluster.

The fix adds the three sets as parameters to `generateNewService` and removes
the internal allocations. `ProcessEnvs` and `ProcessVolumes` now populate the
caller-owned sets, so `CheckResourcesArePresent` receives the correct data and
can reject a deployment that references missing resources.

On the test side, Two new tests are added:
`TestUpdateService_EnvsPropagated` calls the `updateService` closure directly
against a service with a stale env var and asserts that the container env list
is replaced, not merged; `TestGenerateNewService_ResourceSetsPopulated` is a
regression test for the create path that asserts the caller-supplied
`referencedSecrets` and `referencedConfigMaps` sets are populated after
`generateNewService` returns — a future regression that reintroduces internal
allocation would cause this test to fail.

relates #3514

### DIFF
--- a/cmd/deploy_test.go
+++ b/cmd/deploy_test.go
@@ -465,6 +465,49 @@ func TestDeploy_Envs(t *testing.T) {
 
 }
 
+// TestDeploy_EnvsPassedToDeployer ensures that environment variables provided
+// via the -e flag are included in the function passed to the deployer.
+// This is a regression test for issue #3514 where env vars were persisted to
+// func.yaml but not included in the deployed service spec.
+func TestDeploy_EnvsPassedToDeployer(t *testing.T) {
+	root := FromTempDirectory(t)
+	ptr := func(s string) *string { return &s }
+
+	_, err := fn.New().Init(fn.Function{Runtime: "go", Root: root, Registry: TestRegistry})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	deployer := mock.NewDeployer()
+	var deployedEnvs fn.Envs
+	deployer.DeployFn = func(_ context.Context, f fn.Function) (fn.DeploymentResult, error) {
+		deployedEnvs = f.Run.Envs
+		return fn.DeploymentResult{
+			Status:    fn.Deployed,
+			URL:       "http://example.com",
+			Namespace: "default",
+		}, nil
+	}
+
+	cmd := NewDeployCmd(NewTestClient(fn.WithDeployer(deployer)))
+	cmd.SetArgs([]string{"--env=MYVAR=myvalue", "--env=OTHER=othervalue"})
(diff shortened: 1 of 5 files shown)


Ankitsinghsisodya:
cc @gauron99  @matejvasek

Ankitsinghsisodya:
cc @Ikingland @matejvasek

matejvasek:
Good catch!

Could we have some integration test for this too?
We have some common scenarios in: https://github.com/knative/func/blob/3c78bf04288717e7aecbb4789e96ff68ffe869c3/pkg/deployer/testing/integration_test_helper.go
Then we invoke these in respective `Deployer` interface implementation, e.g.:  https://github.com/knative/func/blob/2e8f9af5bda15db79d7d6dac252b4008789d9f9a/pkg/knative/deployer_int_test.go

Ankitsinghsisodya:
> Good catch!
> 
> Could we have some integration test for this too? We have some common scenarios in: https://github.com/knative/func/blob/3c78bf04288717e7aecbb4789e96ff68ffe869c3/pkg/deployer/testing/integration_test_helper.go Then we invoke these in respective `Deployer` interface implementation, e.g.: https://github.com/knative/func/blob/2e8f9af5bda15db79d7d6dac252b4008789d9f9a/pkg/knative/deployer_int_test.go

I have added an integration test as requested. TestInt_ResourceValidationOnFirstDeploy is now in pkg/deployer/testing/integration_test_helper.go and wired up in pkg/knative/deployer_int_test.go following the same pattern as the existing tests.

The test deploys a function that references a non-existent Secret and asserts the deploy is rejected with an error that names the missing Secret (the old code would have silently succeeded here). It then creates the Secret on the cluster and redeploys, asserting that it now succeeds. This directly exercises the generateNewService path that the fix targets.

Ankitsinghsisodya:
cc : @matejvasek

matejvasek:
sry forgot to finish review

matejvasek:
/approve
/lgtm

Thanks for your patience @Ankitsinghsisodya 

If you have time you could try running `TestInt_ResourceValidationOnFirstDeploy` with the other deployers (raw,keda) too.

matejvasek:
https://github.com/knative/func/pull/3670

PR #3597, opened 2026-04-09
Replace generated byte array with go:embed for templates

cardil:
## Summary

Replaces the codegen-based template embedding with direct `//go:embed templates/`, eliminating the 14K-line generated file `generate/zz_filesystem_generated.go` and the codegen tool that produced it.

## Problem

The old approach used `generate/templates/main.go` to walk the `templates/` directory, zip its contents in memory, and write the result as a Go byte array literal (`var TemplatesZip`) into `generate/zz_filesystem_generated.go` (~14K lines of hex). This required running codegen before any build or test, produced a large committed generated file, and made it harder to inspect embedded template contents or review template diffs.

## Solution

Use `//go:embed` directly on the `templates/` directory tree, working around two hard `go:embed` limitations:

1. **Directories with `go.mod` are silently skipped** → rename `go.mod`/`go.sum` to `go.mod.embd`/`go.sum.embd`
2. **Symlinks are silently dropped** → replace symlinks with `.symlink` plain-text marker files containing the target path

A new `manglingFS` wrapper in `pkg/filesystem` transparently reverses this mangling at runtime:
- `*.embd` files are exposed with their real names
- `*.symlink` files are presented as symlink directory entries
- `//go:build embd` constraints are stripped from Go source file content
- File permissions (lost by `embed.FS` which returns 0444 for everything) are restored from a committed `templates/.permissions` manifest

## New Tools

- **`hack/cmd/embd`** — `embd`/`unembd` subcommands to temporarily unmangle a template directory so normal Go tooling (`go mod tidy`, `go get`, `go test`) can operate on it, then re-mangle it. Uses the same constants/logic as `manglingFS` to keep mangling rules in sync.
- **`hack/cmd/permissiongen`** — codegen tool that scans `templates/` and generates `templates/.permissions` listing non-standard file permissions (only executable files differ from the default 0644). Run via `go run ./hack/cmd/permissiongen`.

## Build tag approach

Go template source files are tagged with `//go:build embd` so the Go toolchain does not attempt to compile them as part of the main module. The `go/build/constraint` package is used for correct AST-based manipulation of build tags (preserving any existing tags on round-trips through `embd`/`unembd`).

## Changes

- `templates/embed.go` — `//go:embed` directive
- `pkg/filesystem/mangling.go` — `manglingFS` runtime FS wrapper
- `pkg/functions/templates_embedded.go` — wires up `manglingFS`
- `hack/cmd/embd/main.go` — embd/unembd CLI tool
- `hack/cmd/permissiongen/main.go` — permissions manifest codegen
- `templates/.permissions` — committed generated manifest
- `Makefile` — updated `check-go`, `test-go` and `update-runtime-go` targets
- `hack/update-codegen.sh` / `hack/verify-codegen.sh` — updated
- `.gotagsignore` — excludes `embd` from CI build tag matrix
- Removed: `generate/zz_filesystem_generated.go`, `generate/templates/main.go`

Assisted-by: 🤖 Claude Opus/Sonnet 4.6


matejvasek:
Would this work even if I imported function as a library (We doing that now in Functions OCP dynamic plugin)?

lkingland:
Sorry, but we tried `go:embed` as soon as it was available, but it doesn't work for our use-case for multiple reasons.

EDIT: Ah, I see we have a Slack thread about this.

> **The generated Go source file changed in almost every PR that touched templates, causing constant merge conflicts. A binary zip + //go:embed eliminates this entirely since templates.zip is gitignored.

tl;dr, This could perhaps be minimized by splitting the monolithic generated file into a per-language, per-template set of generated files.

cardil:
I think I have a solution. Will keep you posted...

/hold

cardil:
E2E Runtime failures (go, python, quarkus, rust, typescript) are infra flakes: Docker Hub rate limiting (TOOMANYREQUESTS) and job timeouts. Go E2E TestMatrix_Run and TestMatrix_Deploy all passed — only TestMatrix_Remote was cancelled mid-run.

/retest

cardil:
Thanks for the feedback @matejvasek @lkingland. Your comments were about the earlier version of this PR, which took a different approach. The PR has been fully reworked since then.

@matejvasek good point about library import. The current approach uses `//go:embed templates/` directly in the `templates/` package. Go resolves embed directives at build time relative to the source package, so the embedded templates travel with the module when imported as a dependency. The public API (`Filesystem` interface from `pkg/functions`) hasn't changed. The `manglingFS` wrapper that reverses the file mangling is transparent to callers.

@lkingland you're right that raw `go:embed` doesn't work for our templates because dirs with `go.mod` get silently skipped and symlinks are dropped. This version works around both:

1. `go.mod`/`go.sum` are renamed to `go.mod.embd`/`go.sum.embd` at rest
2. Symlinks are replaced with `.symlink` marker files containing the target path
3. Go source files in templates get a `//go:build embd` tag so the toolchain won't try to compile them

A `manglingFS` wrapper reverses all of this at runtime. Consumers see real `go.mod` files, real symlinks, and untagged Go source.

The main motivation was merge conflicts. The old `zz_filesystem_generated.go` (14K lines of hex byte array) changed in almost every template PR. With direct `go:embed`, the template files are committed as-is (with mangling), so diffs are human-readable and conflicts are trivially resolvable. No generated file to maintain.

The tradeoff is the mangling tooling (`hack/cmd/embd` for embd/unembd, `hack/cmd/permissiongen` for the permissions manifest), but these are small, well-tested tools that keep the mangling rules in sync with the runtime `manglingFS`.

cardil:
/retest

E2E Runtime failures are infra flakes: Red Hat container registry returned 502 Bad Gateway when pulling `ubi8/go-toolset:latest` during s2i remote build (`TestMatrix_Remote/func-e2e-matrix-remote-go-s2i-cloudevents`). Python, quarkus, rust, typescript were cancelled (not failed). All unit tests, template tests, lint, and codegen checks pass.

hack/cmd/embd/main.go
```diff
@@ -0,0 +1,189 @@
+// Package main implements the embd/unembd tool for reversibly mangling
```
matejvasek, line 1: More fitting than embd/unembd would be mangele/unmangle (or demangle?) IMO.

But more importantly we should just need unmangle (unembd).
We could unmangle (unembd) templates to some temp directory and run smoke test on that temp directory.
IMO that would be more robust -- imagine test gets killed/interrupted in the middle, then templates re not re-mangled.

Furthermore the unmangling seems to be re-implemented here. The re-implementation may potentially differ/diverge from actual demangling as defined in the filesystem module. We should rather call filesystem here and do demangling.
 Something like:
* ```go
    fsToCopy := functions.EmbeddedTemplatesFS
    filesystem.CopyFromFS("go", tempDir, fsToCopy)
   ```
   or
* ```go
	fsToCopy := filesystem.NewOsFilesystem(dir)
	filesystem.CopyFromFS(".", tempDir, fsToCopy)
   ```

matejvasek: Reimplementing unmangling may not be hard, still it best just call the unmangling code that we aready have, and more importantly the one that is actually called during `func create`.

matejvasek: Also looking at `Makefile` it appears that this is called only for `Go` since only that has some serious manglin going on. However `Java` templates also have some subtle transformation it it -- the permissions restoration. I would prefer running the `Java` test on demangled version too, so we are sure the permissions are correctly passed trought embedding process.

Makefile
```diff
@@ -207,16 +203,26 @@ presubmit-unit-tests: ## Run prow presubmit unit tests locally
 check-embedded-fs: ## Check the embedded templates FS
```
matejvasek, line 203: IMO We might to delete this test. It purpose is just to test that the generated file is in sync with templates content.

hack/cmd/embd/main.go
cardil: Let me take a look. I intended to reuse that package...

To the mangle/unmangle -- we need to un-mangle to upgrade deps, and then re-mangle after changes, with `make update-runtime-go`

matejvasek: >I would prefer running the Java test on demangled version too, so we are sure the permissions are correctly passed trought embedding process.

but might be overkill if we reliably test perms somewhere else.

pkg/filesystem/mangling.go
```diff
@@ -0,0 +1,527 @@
...
+		}
+	}
+	return perms
+}
+
+// parsePermLine parses a line like "0755 quarkus/cloudevents/mvnw".
+func parsePermLine(line string, mode *uint32, path *string) (int, error) {
+	var n int
+	var err error
+	_, err = func() (int, error) {
```
matejvasek, line 98: Why the immediately called closure here. Sometimes closure that is immediately called makes sense, e.g.: goroutine, complex defer, defer in loop. But it does not seems to apply here?

pkg/filesystem/mangling.go
```diff
@@ -0,0 +1,527 @@
...
+	fs.FileInfo
+	name string
+}
+
+func (s *symlinkFileInfo) Name() string       { return s.name }
+func (s *symlinkFileInfo) Mode() fs.FileMode  { return fs.ModeSymlink | 0777 }
+func (s *symlinkFileInfo) Size() int64        { return 0 }
+func (s *symlinkFileInfo) ModTime() time.Time { return s.FileInfo.ModTime() }
+func (s *symlinkFileInfo) IsDir() bool        { return false }
+func (s *symlinkFileInfo) Sys() interface{}   { return nil }
```
matejvasek, line 480: nit: return `any` instead of `interface{}`

pkg/filesystem/mangling.go
```diff
@@ -0,0 +1,527 @@
...
+
+	// EmbdBuildConstraint is the build-constraint tag name used to mark Go
+	// source files inside templates/ so they are not compiled by go build.
+	EmbdBuildConstraint = "embd"
+
+	// PermissionsManifest is the name of the file listing non-default
+	// permissions in the templates/ tree, generated by hack/cmd/permissiongen.
+	PermissionsManifest = ".permissions"
+
+	permissionsManifest = PermissionsManifest
```
matejvasek, line 33: Why this alias?

pkg/filesystem/mangling.go
```diff
@@ -0,0 +1,527 @@
...
+	}
+	return 0644
+}
+
+// Open opens the named file. If the file is not found, it also tries
+// name + ".embd" (for go.mod / go.sum files).
+func (m manglingFS) Open(name string) (fs.File, error) {
+	f, err := m.inner.Open(name)
+	if err != nil {
+		// Try with .embd suffix (handles go.mod.embd, go.sum.embd)
```
matejvasek, line 132: IMO this should be done only if `errors.Is(err, fs.ErrNotExist)`, okay in practice it always will be `fs.ErrNotExist` but still...

pkg/filesystem/mangling.go
```diff
@@ -0,0 +1,527 @@
...
+			return 0, err
+		}
+		raw = stripEmbdBuildTag(raw)
+		s.buf = bytes.NewReader(raw)
+		s.stripped = true
+	}
+	return s.buf.Read(p)
+}
+
+func (s *strippedFile) Stat() (fs.FileInfo, error) {
```
matejvasek, line 272: Is this correct implementation? Won't this return bigger `Size` that the actual file data,  since we are removing the `Go` build tags? Right now, it may not cause any troubles but if some implementation details changes this may lead to funny bugs.

templates/embed.go
```diff
@@ -0,0 +1,8 @@
+package templates
+
+import "embed"
+
+//go:embed all:go all:node all:python all:quarkus all:rust
+//go:embed all:springboot all:typescript all:certs
+//go:embed manifest.yaml README.md .permissions
+var Content embed.FS
```
matejvasek, line 8: Can't we just embed "."?

matejvasek: Is there rule against it?

matejvasek: `//go:embed all:*`

matejvasek: Okey, maybe it unnecessary embed the `embed.go` file but so what?

matejvasek:
@lkingland PTAL

PR #3600, opened 2026-04-10
Use correct describer in pipelines provider

creydr:
Currently the pipelines provider tries to get the function information from the Knative Service. As we have additional deploy methods (raw + keda), this does not work anymore in all cases.
This PR addresses it and gets the deployments information correctly based on the deployer.

### DIFF
--- a/pkg/functions/client.go
+++ b/pkg/functions/client.go
@@ -188,6 +188,7 @@ type Instance struct {
 	Subscriptions []Subscription    `json:"subscriptions" yaml:"subscriptions"`
 	Labels        map[string]string `json:"labels" yaml:"labels" xml:"-"`
 	Middleware    Middleware        `json:"middleware,omitempty" yaml:"middleware,omitempty"`
+	Generation    int64             `json:"generation,omitempty" yaml:"generation,omitempty"`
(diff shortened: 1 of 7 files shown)


pkg/functions/client.go
```diff
@@ -188,6 +188,7 @@ type Instance struct {
 	Subscriptions []Subscription    `json:"subscriptions" yaml:"subscriptions"`
 	Labels        map[string]string `json:"labels" yaml:"labels" xml:"-"`
 	Middleware    Middleware        `json:"middleware,omitempty" yaml:"middleware,omitempty"`
+	Generation    int64             `json:"generation,omitempty" yaml:"generation,omitempty"`
```
creydr, line 191: This is simply added, so we can say if some operations were an update or a create. See for example here: https://github.com/knative/func/blob/959228011c3d6b95ae2113dcdc1763e6f7c70c0e/pkg/pipelines/tekton/pipelines_provider.go#L256-L260

matejvasek:
Looks good but I am not sure if this would be all tested. Maybe modify `TestInt_Remote_Default` to run for each deployer?

PR #3601, opened 2026-04-10
Use correct namespace in KEAD and k8s listers

creydr:
Use correct namespace for Keda and k8s lister.

Before:

```
$ func list --all-namespaces                                    
Error: unable to get details about function: unable to get HTTPScaledObject: the server could not find the requested resource (get httpscaledobjects.http.keda.sh my-func)

...
```

With the fix of this PR:
```
$ ./func list --all-namespaces                                                                                                          1 ↵
NAME     NAMESPACE  RUNTIME  DEPLOYER  URL                                                 READY
my-func  default             keda      http://my-func-interceptor-bridge.default.svc:8080  True
```

### DIFF
--- a/pkg/k8s/lister.go
+++ b/pkg/k8s/lister.go
@@ -42,7 +42,7 @@ func (l *Lister) List(ctx context.Context, namespace string) ([]fn.ListItem, err
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


gauron99 (approved):
/lgtm

PR #3603, opened 2026-04-12
feat: add platform support to pack builder

matejvasek:
# Changes

* feat: add platform support to pack builder

  The pack builder previously rejected any --platform flag with a hard error. Since the pack library already supports a Platform field in BuildOptions, this enables single-platform builds following the same pattern as the s2i builder: resolve the builder image via docker.GetPlatformImage and set opts.Platform so pack also pulls the correct run image.

  Multiple platforms are still rejected with a clear error, and omitting --platform continues to work as before.

```release-note
feat: add platform support to pack builder
```


matejvasek:
ptal @lkingland @gauron99

gauron99 (approved):
/lgtm

PR #3611, opened 2026-04-15
warn on extraneous keys used in func.yaml

gauron99:
adding a warning for extraneous keys
```
❯ func run
Warning (unknown fields will be ignored):
 'func.yaml' is not valid:
  line 7: unknown field "junk"
  line 8: unknown field "builder"
  line 9: unknown field "tralala"
```

when we have the latest (migrated) spec present, we do a strict Unmarshal (which errors on extraneous keys) and we print a warning like above.
Note that Im adding the "once" because NewFunction is actually called few times throughout a lets say "func deploy" call.

### DIFF
--- a/pkg/functions/function_migrations.go
+++ b/pkg/functions/function_migrations.go
@@ -5,19 +5,34 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/coreos/go-semver/semver"
 	"gopkg.in/yaml.v2"
 )
 
+var unknownFieldsOnce sync.Once
+
 // Migrate applies any necessary migrations, returning a new migrated
 // version of the function.  It is the caller's responsibility to
-// .Write() the function to persist to disk.
+// .Write() the function to persist to disk. Additionally it will warn on
+// up-to-date spec but wrong func.yaml (eg. extraneous fields)
 func (f Function) Migrate() (migrated Function, err error) {
 	// Return immediately if the function indicates it has already been
 	// migrated.
 	if f.Migrated() {
+		// Already at the latest spec — check for unknown fields
+		if f.Root != "" {
+			if bb, readErr := os.ReadFile(filepath.Join(f.Root, FunctionFile)); readErr == nil {
+				unknownFieldsOnce.Do(func() {
+					var strict Function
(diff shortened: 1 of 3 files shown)


twoGiants (changes requested):
Looks good! Just add a few test cases covering the new conditional logic. Thank you 😸 👍

pkg/functions/function_migrations.go
```diff
@@ -5,19 +5,34 @@ import (
...
+		if f.Root != "" {
+			if bb, readErr := os.ReadFile(filepath.Join(f.Root, FunctionFile)); readErr == nil {
+				unknownFieldsOnce.Do(func() {
+					var strict Function
+					if strictErr := yaml.UnmarshalStrict(bb, &strict); strictErr != nil {
+						fmt.Fprintf(os.Stderr, "Warning (unknown fields will be ignored):\n %v\n.\n", formatUnmarshalError(strictErr))
+					}
+				})
+			}
+		}
```
twoGiants, line 35: nit: use early return


```suggestion
		if f.Root == "" {
				return f, nil
		}
		
		if bb, readErr := os.ReadFile(filepath.Join(f.Root, FunctionFile)); readErr == nil {
			unknownFieldsOnce.Do(func() {
				var strict Function
				if strictErr := yaml.UnmarshalStrict(bb, &strict); strictErr != nil {
					fmt.Fprintf(os.Stderr, "Warning (unknown fields will be ignored):\n %v\n.\n", formatUnmarshalError(strictErr))
				}
			})
		}
```

twoGiants (approved):
Looks good! Thank you for the tests.

I have just a minor nit => optional.

/approve
/lgtm
/hold for others

pkg/functions/function_migrations.go
```diff
@@ -5,19 +5,34 @@ import (
...
 	"github.com/coreos/go-semver/semver"
 	"gopkg.in/yaml.v2"
 )
 
+var unknownFieldsOnce sync.Once
+
 // Migrate applies any necessary migrations, returning a new migrated
 // version of the function.  It is the caller's responsibility to
-// .Write() the function to persist to disk.
+// .Write() the function to persist to disk. Aditionaly it will warn on
```
matejvasek: typo: Aditionaly => Additionally

PR #3612, opened 2026-04-15
Fix: raise soft ulimit for open files in Python runtime startup

### LINKED ISSUE #3513: Support low-ulimit platforms in Python runtime
The Python function runtime should raise its soft ulimit for open files
to match the hard limit at startup. Go and Java runtimes already do this
automatically. Without it, the Python runtime fails under load on
platforms with a low default soft limit (e.g. 1024).

The fix belongs in the middleware startup path.

Related: [SRVOCF-788](https://redhat.atlassian.net/browse/SRVOCF-788)

Ankitsinghsisodya:
# Changes

- :bug: Raise the soft open-file limit to match the hard limit at Python middleware startup, preventing failures under load on platforms with a low default soft limit (e.g. 1024)

/kind bug

Fixes #3513

**Release Note**

```release-note
Python functions now raise the soft open-file limit to match the hard limit at startup, matching the behaviour of the Go and Java runtimes. This prevents failures under load on platforms where the default soft limit is low (e.g. 1024).
```

**Docs**

```docs

```


Ankitsinghsisodya:
Closing in favour of the correct upstream fix.

The fix belongs in the middleware library, not in the scaffolding templates. Scaffolding is baked into container images at build time, so this PR would only help functions that are rebuilt after it merges — already-deployed functions would get nothing.

The correct fix is in knative-extensions/func-python#83, where it lives inside serve() and benefits all deployed functions on the next library version bump.

A follow-up PR to knative/func will be raised to bump the pinned func-python version in the scaffolding pyproject.toml once that PR merges and a release is cut.

PR #3614, opened 2026-04-16
fix: conform dialer and volume-uploader pods to restricted pod security profile

### LINKED ISSUE #3517: Dialer pod does not conform to restricted pod security profile
The in-cluster dialer pod used during deployment does not set the
security context fields required by the "restricted" pod security
profile: allowPrivilegeEscalation, capabilities drop, runAsNonRoot,
and seccompProfile.

This produces warnings on clusters enforcing pod security admission
and will fail on clusters that enforce (rather than warn) the
restricted profile.

Related: [SRVOCF-445](https://redhat.atlassian.net/browse/SRVOCF-445)

Ankitsinghsisodya:
# Changes

- `defaultPodSecurityContext` now sets `RunAsNonRoot: true` and `SeccompProfile: RuntimeDefault` unconditionally at pod level, so pods satisfy the `restricted` PSA profile on any cluster
- On OpenShift the function previously returned `nil`, leaving the pod with no security context at all; it now returns a minimal context without `RunAsUser`/`RunAsGroup`/`FSGroup` (which OpenShift SCCs manage) so PSA admission passes
- `defaultSecurityContext` restores `Privileged: false` explicitly for defence-in-depth, removes the version-gated seccomp check (`RuntimeDefault` is GA since k8s 1.25; this project tracks k8s 1.35), and drops the now-unused `client` parameter
- `SeccompProfile` is set at both pod and container level intentionally — pod-level covers all containers by default; container-level ensures compliance even if the pod-level context is overridden downstream
- `RunAsGroup: 0` is retained on non-OpenShift to preserve Tekton buildpack task compatibility; this does not violate the restricted profile (which checks UID, not GID)
- Unit tests cover both OpenShift and non-OpenShift paths and include a compliance test asserting all four restricted-profile requirements; end-to-end admission validation against a real restricted namespace is covered by `make test-full`

/kind bug

Fixes #3517

**Release Note**

```release-note
Pods created by func during deployment (in-cluster dialer, volume uploader) now
conform to the Kubernetes "restricted" pod security profile. Deployments no
longer fail or produce warnings on clusters or namespaces that enforce
pod-security.kubernetes.io/enforce: restricted.
```

### DIFF
--- a/pkg/k8s/dialer.go
+++ b/pkg/k8s/dialer.go
@@ -290,7 +290,7 @@ func (c *contextDialer) startDialerPod(ctx context.Context) (err error) {
 					Stdin:           true,
 					StdinOnce:       true,
 					Command:         []string{"socat", "-u", "-", "OPEN:/dev/null"},
-					SecurityContext: defaultSecurityContext(client),
+					SecurityContext: defaultSecurityContext(),
 				},
 			},
 			DNSPolicy:     coreV1.DNSClusterFirst,
--- a/pkg/k8s/persistent_volumes.go
+++ b/pkg/k8s/persistent_volumes.go
@@ -141,7 +141,7 @@ func runWithVolumeMounted(ctx context.Context, podImage string, podCommand []str
 							MountPath: volumeMntPoint,
 						},
 					},
-					SecurityContext: defaultSecurityContext(client),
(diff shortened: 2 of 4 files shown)


matejvasek:
Good job! Look all right, I just test it on OCP cluster just to be sure.

PR #3621, opened 2026-04-21
fix make targets exit code

lkingland:
Fixes a problem where some targets don't properly pass through non-zero exit codes.


gauron99 (approved):
/lgtm

PR #3623, opened 2026-04-21
Remove legacy build directives

lkingland:
The `// +build` syntax is only for compatibility with Go < 1.17. We now require Go 1.25.


gauron99 (approved):
/lgtm

PR #3624, opened 2026-04-21
feat: improved host builder .funcignore support

lkingland:
The host (OCI) builder walks the function root to build its data layer and applies a fixed defaultIgnored list (.git, .func, .funcignore, .gitignore). Workspace-specific directories outside that list — most commonly .jj/, which contains absolute symlinks to system sockets — trip validatedLinkTarget and make the build fail, with no way for the user to opt out.

Fix is to read .funcignore (which the s2i builder already honors via a .s2iignore symlink) and append its patterns to defaultIgnored. Supports two pattern forms:

  /foo/bar  root-relative; excludes that path and its descendants
  foo       base-name; excludes any entry named 'foo' at any depth

### DIFF
--- a/pkg/oci/builder.go
+++ b/pkg/oci/builder.go
@@ -41,6 +41,25 @@ var defaultIgnored = []string{
 	".gitignore",
 }
 
+// readFuncIgnore reads the .funcignore file from the given root directory
+// and returns its patterns with blank lines and '#' comments removed. A
+// missing or unreadable file returns nil; .funcignore is optional.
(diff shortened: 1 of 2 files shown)


gauron99 (approved):
/lgtm
/approve

as a followup, if im understanding this correctly, I think it might be nice to have a concrete way of ignoring directories with an explicit `"/"` at the end like `"dir/"`. Also I think we now support s2i and host but pack builder is still lacking this feature.

PR #3625, opened 2026-04-21
Remove stern dependency from subscription e2e tests

lkingland:
When running E2E tests on a new machine, I noticed that stern became a hard dependency in our test suite.  This is a quick fix, because the use case can be covered with this somewhat awkward kubectl logs command, however.

A more complete solution could be to use a second Function, deployed alongside the subscriber, which awaits a POST from the subscriber indicating it received the event, and the test just waits for that "verifier" to retuturn the expected event was received via its default endpoint.

### DIFF
--- a/e2e/e2e_metadata_test.go
+++ b/e2e/e2e_metadata_test.go
@@ -4,7 +4,6 @@
 package e2e
 
 import (
-	"bufio"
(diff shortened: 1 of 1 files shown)


gauron99 (approved):
/lgtm

PR #3628, opened 2026-04-21
Remove nested git repository from pipelines tests

lkingland:
Nested git repositories cause problems with a plethora of tools.


gauron99 (approved):
/lgtm

PR #3629, opened 2026-04-22
Add revision label to function images and show in describe

creydr:
# Changes

- :gift: Add a `org.opencontainers.image.revision` image label containing the short git commit SHA of the function source directory at build time, with a `-dirty` suffix when the working tree has uncommitted changes
- :gift: Show the `revision` label in `func describe` output (JSON/YAML)

```
$ func describe my-func -o yaml
...
revision: 0cf9be2-dirty
```

/kind enhancement

**Release Note**

```release-note
Function images now include a `commit` label with the short git commit SHA of the function source at build time. The commit is also shown in `func describe` JSON/YAML output.
```


pkg/functions/function_labels.go
```diff
@@ -10,6 +10,7 @@ import (
 
 const (
 	MiddlewareVersionLabelKey = "middleware-version"
+	CommitLabelKey            = "commit"
```
matejvasek: I would name this `revision` to be more VCS agnostic. Or even `org.opencontainers.image.revision`.

pkg/pipelines/tekton/templates.go
```diff
@@ -90,6 +90,9 @@ type templateData struct {
 
 	// TLS verification for registry operations
 	TlsVerify string
+
+	// Git commit SHA of the function source
+	Commit string
```
matejvasek, line 95: Why we need to parameterize? Cannot the task read current commit on its own?

matejvasek: I think we upload source as is including .git so the task could read it.

creydr: I had it initially in the task itself, but I wasn't able to get the commit hash there. This would also have involved to write it to a file and then read it during the prepare step. So I thought it'd be cleaner to have it simply as a parameter for the task.

pkg/functions/function_labels.go
creydr: Good point. Let me update it

pkg/functions/image_commit.go
```diff
@@ -0,0 +1,36 @@
+package functions
+
+import (
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+// ImageCommit gets the commit SHA label from a function image.
+// Returns an empty string and no error if the image was built without this label.
+func ImageCommit(image string) (string, error) {
```
matejvasek: You should user our custom transport otherwise internal registry will cause troubles.

matejvasek: e.g. https://github.com/knative/func/blob/main/pkg/oci/pusher.go#L185
https://github.com/knative/func/blob/main/cmd/build.go#L459

pkg/functions/image_inspector.go
```diff
@@ -0,0 +1,65 @@
+package functions
+
+import (
+	"net/http"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+type ImageInspector struct {
```
lkingland: This could use a comment.

pkg/functions/image_inspector.go
```diff
@@ -0,0 +1,65 @@
...
+	transport http.RoundTripper
+}
+
+func NewImageInspector(transport http.RoundTripper) *ImageInspector {
```
lkingland: This is just a nit, but I'm not seeing any reason this _needs_ to be a pointer, and `nil`s as args begs the question.

Is that just to enable the nil in method signatures, like `NewDeployer(x, nil)`?  This could be accomplished with an fn.ImageInspector interface (which is perhaps a good idea for other reasons too).  The simplest option to get rid of the ponters/nils would then be to have a singleton `fn.DefaultImageInspector` which is a noop and can be passed in place of the nil.

lkingland (approved):
/lgtm
/hold for the good suggestions from others

cmd/client.go
```diff
@@ -62,15 +62,16 @@ func NewClient(cfg ClientConfig, options ...fn.Option) (*fn.Client, func()) {
 		t  = newTransport(cfg.InsecureSkipVerify)        // may provide a custom impl which proxies
 		c  = newCredentialsProvider(config.Dir(), t, "") // for accessing registries
 		d  = newKnativeDeployer(cfg.Verbose)             // default deployer (can be overridden via options)
-		pp = newTektonPipelinesProvider(c, cfg.Verbose)
+		ii = fn.NewImageInspector(t)
```
lkingland: Is the transport the reason for having the ImageInspector exposed outside the core into the cli (or anyone using the core api)?  If so, it might be possible to just pass the transport to the describers/pipeline provider.  Then they can use that internally to inspect using a helper method.

pkg/functions/function_labels.go
creydr: I updated the label key to `org.opencontainers.image.revision` and the key for the "func describe" output to `revision`

pkg/functions/image_inspector.go
creydr: Removed the struct as part of the refactoring

pkg/functions/image_inspector.go
creydr: The struct got removed during the refactoring. I am using the http.transport directly now (passed via our `Options` pattern)

cmd/client.go
creydr: Good point. I updated it. I am passing the `http.Transport` now to the describers only (via a `WithDescriberTransport(t)` option func). I hope this makes sense :)

PR #3631, opened 2026-04-22
ci: skip functions workflow on doc-only changes

### LINKED ISSUE #3634: ci: skip functions workflow on doc-only changes
## Problem

The `Functions` CI workflow runs the full suite — unit tests across 5 OSes, template tests, integration tests, and a 7-runtime E2E matrix — even when a PR only touches documentation or non-code files (`*.md`, `docs/**`, `CODEOWNERS`, `LICENSE`). This wastes CI minutes and runner time.

## Proposed Solution

Add `paths-ignore` filters to the `push` and `pull_request` triggers so the workflow is skipped when only documentation or non-code files are modified.

Ankitsinghsisodya:
Closes #3634

## Summary

- Adds `paths-ignore` filters to both `push` and `pull_request` triggers in the `Functions` CI workflow
- Skips the full CI suite (unit, template, integration, e2e) when only documentation or non-code files are changed
- Ignored paths: `docs/**`, `**/*.md`, `CODEOWNERS`, `LICENSE`

## Motivation

The `Functions` workflow spawns many expensive jobs (unit tests across 5 OSes, integration tests, 7-runtime E2E matrix). Running all of these on a docs-only commit wastes CI minutes and runner time.

## Test plan

- [ ] Open a PR that only modifies a `.md` file — verify `Functions` workflow is skipped
- [ ] Open a PR that modifies a `.go` file — verify `Functions` workflow still runs

### DIFF
--- a/.github/workflows/ci-check.yaml
+++ b/.github/workflows/ci-check.yaml
@@ -1,0 +1,31 @@
+name: Check
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  precheck:
+    name: Precheck
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+      - uses: knative/actions/setup-go@main
+
+      - name: Check and Lint
+        run: make check
+      - name: Check Generated Schema File
+        run: make check-schema
+      - name: Check Templates
+        run: make check-templates
+      - name: Check Embedded FS
+        run: make check-embedded-fs
--- a/.github/workflows/functions.yaml
+++ b/.github/workflows/functions.yaml
@@ -12,9 +12,19 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - 'docs/**'
+      - '**/*.md'
+      - 'CODEOWNERS'
+      - 'LICENSE'
   pull_request:
     branches:
       - main
+    paths-ignore:
+      - 'docs/**'
+      - '**/*.md'
+      - 'CODEOWNERS'
+      - 'LICENSE'
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/functions.yaml
+++ b/.github/workflows/functions.yaml
@@ -13,7 +13,7 @@ on:
     branches:
       - main
     paths-ignore:
-      - 'docs/**'
+      - 'docs/**/*.md'
       - '*.md'
       - 'OWNERS'
       - 'OWNERS_ALIASES'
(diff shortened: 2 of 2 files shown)


matejvasek:
This would also exclude `docs/generator/main.go` (IMO it should not).
This also skips some spell checks that may be useful for documentation.

Ankitsinghsisodya:
> This would also exclude `docs/generator/main.go` (IMO it should not). This also skips some spell checks that may be useful for documentation.

For the `docs/**` issue, I can narrow that to `docs/**/*.md` so only markdown files under docs are excluded and `docs/generator/main.go` still triggers CI normally.

For the spell check concern — I'm thinking of adding a small separate `ci-check.yaml` workflow that runs the same precheck steps (`make check`, `make check-schema`, `make check-templates`, `make check-embedded-fs`) on every push/PR with no `paths-ignore`. That way misspell checking always runs regardless of what files changed. The `precheck` job would stay in `functions.yaml` too since the expensive jobs depend on it — there'd be a small duplication on non-doc PRs, but given how fast precheck is I think that's a reasonable trade-off.

Let me know if that approach works or if you'd prefer something different.

Ankitsinghsisodya:
> > This would also exclude `docs/generator/main.go` (IMO it should not). This also skips some spell checks that may be useful for documentation.
> 
> For the `docs/**` issue, I can narrow that to `docs/**/*.md` so only markdown files under docs are excluded and `docs/generator/main.go` still triggers CI normally.
> 
> For the spell check concern — I'm thinking of adding a small separate `ci-check.yaml` workflow that runs the same precheck steps (`make check`, `make check-schema`, `make check-templates`, `make check-embedded-fs`) on every push/PR with no `paths-ignore`. That way misspell checking always runs regardless of what files changed. The `precheck` job would stay in `functions.yaml` too since the expensive jobs depend on it — there'd be a small duplication on non-doc PRs, but given how fast precheck is I think that's a reasonable trade-off.
> 
> Let me know if that approach works or if you'd prefer something different.

cc @matejvasek

lkingland:
/ok-to-test

Three follow-ups before this lands:
1. Narrow `docs/**` → `docs/**/*.md` as agreed in the thread above, so `docs/generator/main.go` still triggers CI.
2. Rebase needed.
3.  Need that dedicated workflow to run non-code checks like spelling etc.

Ankitsinghsisodya:
@lkingland @gauron99

The actual E2E Podman tests passed on both matrix legs. The only failure is in the codecov upload step, which exits with code 1 because secrets.CODECOV_TOKEN is not exposed to fork PRs under the pull_request trigger. With fail_ci_if_error: true, that turns a missing-token warning into a hard job failure. This is a pre-existing issue, not introduced by this PR.

Two small fixes would address it:
1. Set fail_ci_if_error: false on the codecov step in test-e2e-podman so a missing token does not block CI.
2. Set fail-fast: false on the test-e2e-podman matrix so a codecov failure on ubuntu-latest does not cancel the ubuntu-24.04-arm leg.

Happy to fold these into this PR or open a separate one, whichever you prefer.

PR #3639, opened 2026-04-23
ci: pin remaining GitHub Actions to commit SHAs

### LINKED ISSUE #3638: ci: pin remaining GitHub Actions to commit SHAs
## Problem

Multiple workflows reference GitHub Actions using **mutable** refs: semver tags (e.g. `@v4`, `@v5`) and `@main` for `knative/actions`. Tags and branch heads can be moved, which complicates supply-chain review and is flagged by many security checklists (e.g. OpenSSF) in favor of **full commit SHAs** with a human-readable version in a comment.

## Expected

- Every `uses:` in `.github/workflows/` should pin to a 40-char commit SHA, with a trailing comment noting the previous tag (e.g. `# v4`).

## Scope (example)

- `actions/*`, `docker/*`, `codecov/*`, and other third-party actions currently using tags
- `knative/actions/setup-go@main` and `knative/actions/.github/workflows/reusable-*.yaml@main` — pin to a concrete SHA of the `knative/actions` repository

/area CI

/assign @Ankitsinghsisodya

Ankitsinghsisodya:
# Changes

- :broom: Pin every `uses:` in `.github/workflows/` to a full 40-character commit SHA, with a short comment (e.g. `# v4`, `# knative@abc1234`) for the previous tag or `main` ref.
- Covers `actions/*`, `codecov/*`, `docker/*`, `imjasonh/setup-ko`, `peter-evans/create-pull-request`, `endersonmenezes/free-disk-space`, `actions-rust-lang/setup-rust-toolchain`, and all `knative/actions` setup-go and reusable workflow references that previously used `@main` or semver tags.

/kind cleanup

Fixes #3638

## References

- GitHub docs: [Using third-party actions - Pinning to a commit SHA](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)
- StepSecurity: Why pin GitHub Actions to commit SHAs

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/workflows/functions.yaml
+++ b/.github/workflows/functions.yaml
@@ -36,7 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - uses: knative/actions/setup-go@main
 
       - name: Check and Lint
@@ -70,13 +70,13 @@ jobs:
         if: runner.os == 'Windows'
         run: git config --global core.autocrlf false
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - uses: knative/actions/setup-go@main
 
       - name: Run Unit Tests
         run: make test
 
-      - uses: codecov/codecov-action@v5
+      - uses: codecov/codecov-action@75cd11691c0faa626561e295848008c8a7dddffe # v5
         with:
           files: ./coverage.txt
           flags: unit ${{ matrix.os }}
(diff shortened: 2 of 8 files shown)


matejvasek:
@lkingland @dsimansk  PTAL

gauron99:
This does not look good to me. I would much prefer to keep this as it was. introduces a lot of overhead to bump all of these as well

gauron99:
+ i dont think this will work for the knative-wide workflows -- these are automatically generated and maintained elsewhere so i suspect some changes would not last, anyways.

Ankitsinghsisodya:
@gauron99 
I've updated the PR:

- Reverted all the `knative-*.yaml` files back to `@main` — makes no sense to pin those locally if knobots is just going to overwrite them anyway. Same for `knative/actions/setup-go`.
- Only the third-party actions in the repo-owned workflows are pinned now (`actions/*`, `codecov`, `docker`, etc.).
- Added a `dependabot.yml` to handle the weekly SHA bumps automatically, so there's no manual toil involved in keeping these up to date.

Let me know if you have any other concerns!

matejvasek:
Aren't `actions/*` GitHub's? Don't we trust these?

Ankitsinghsisodya:
> Aren't `actions/*` GitHub's? Don't we trust these?

Yeah we trust them, but that's not really the point. tags like `@v4` are mutable, gitHub can move them at any time without any notice. A commit SHA is immutable so you know exactly what's running.

Ankitsinghsisodya:
GitHub's own security hardening docs actually recommend this for all actions including first-party ones: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

matejvasek:
It also says:
>Pin actions to a tag only if you trust the creator:
    Although pinning to a commit SHA is the most secure option, specifying a tag is more convenient and is widely used. If you’d like to specify a tag, then be sure that you trust the action's creators.

Ankitsinghsisodya:
> Although pinning to a commit SHA is the most secure option, specifying a tag is more convenient and is widely used. If you’d like to specify a tag, then be sure that you trust the action's creators.

Fair point, you’re right about what the docs say. SHA is described as the more secure option, while tags are more of a convenience trade off.
Happy to go with whatever you and the other reviewers prefer. If tags are good enough, I can close it

matejvasek:
I would prefer tag/branch for `actions/*` and `knative/*` and keep sha for the rest.

PR #3642, opened 2026-04-25
bump: update TypeScript templates to TypeScript 6

Ankitsinghsisodya:
## What

Bumps TypeScript from v5 to v6 across both TypeScript function templates (`templates/typescript/http` and `templates/typescript/cloudevents`), including all required tooling adjustments.

This is the TypeScript follow-up to #3582 (Node template bumps).

### `package.json` (both templates)

| Package | Old | New | Notes |
|---|---|---|---|
| `typescript` | `^5.5.2` | `^6.0.3` | TypeScript 6 |
| `typescript-eslint` | `^8.0.0` | `^8.59.0` | first version with TS6 peer dep support |
| `ts-node` | `^10.9.2` | removed | ts-node 10 does not support TypeScript 6 |
| `tsx` | — | `^4.21.0` | drop-in replacement for ts-node |
| `tsd` | `^0.31.1` | `^0.33.0` | adds TypeScript 6 support |

### `tsconfig.json` (both templates)

| Option | Old | New | Why |
|---|---|---|---|
| `rootDir` | not set | `"./src"` | TS6 no longer infers rootDir; without this, output lands in `build/src/` instead of `build/` |
| `moduleResolution` | `"node"` | `"bundler"` | `"node"` (node10) is deprecated in TS6; bundler+commonjs is the recommended migration path |

### Scripts (both templates)

```diff
- "test:unit": "ts-node node_modules/tape/bin/tape test/unit.ts",
- "test:integration": "ts-node node_modules/tape/bin/tape test/integration.ts",
+ "test:unit": "tsx node_modules/tape/bin/tape test/unit.ts",
+ "test:integration": "tsx node_modules/tape/bin/tape test/integration.ts",
```

`package-lock.json` regenerated for both templates.

## Verification

- `tsc --noEmit` passes on both templates with TypeScript 6.0.3
- `tsc` build produces `build/index.js` at the expected path in both templates
- No source changes required in `src/index.ts` or test files


Ankitsinghsisodya:
@matejvasek, just looked into the Podman CI failure — it's not related to the TypeScript changes in this PR at all.

The issue is in `hack/cluster.sh`. The local registry is started with `-p "127.0.0.1:50000:5000"` which only binds to IPv4. On the Ubuntu runner, Go resolves `localhost` to `[::1]` (IPv6) first, so when `func deploy` tries to reach the registry it gets connection refused:

```
dial tcp [::1]:50000: connect: connection refused
```

The fix is straightforward — drop the `127.0.0.1:` prefix so the registry binds on all interfaces (both IPv4 and IPv6):

```diff
-    $CONTAINER_ENGINE run -d --restart=always -p "127.0.0.1:50000:5000" --name "func-registry" registry:2
+    $CONTAINER_ENGINE run -d --restart=always -p "50000:5000" --name "func-registry" registry:2
```

This applies to both the `docker` and `podman` branches inside the `registry()` function in `cluster.sh`.

What are your thoughts?

matejvasek:
> @matejvasek, just looked into the Podman CI failure — it's not related to the TypeScript changes in this PR at all.
> 
> The issue is in `hack/cluster.sh`. The local registry is started with `-p "127.0.0.1:50000:5000"` which only binds to IPv4. On the Ubuntu runner, Go resolves `localhost` to `[::1]` (IPv6) first, so when `func deploy` tries to reach the registry it gets connection refused:
> 
> ```
> dial tcp [::1]:50000: connect: connection refused
> ```

That's odd I would expect Go to use `Happy Eyeballs`, dial both 127.0.0.1 and ::1 and use the one that responds first.

> 
> The fix is straightforward — drop the `127.0.0.1:` prefix so the registry binds on all interfaces (both IPv4 and IPv6):
> 
> ```diff
> -    $CONTAINER_ENGINE run -d --restart=always -p "127.0.0.1:50000:5000" --name "func-registry" registry:2
> +    $CONTAINER_ENGINE run -d --restart=always -p "50000:5000" --name "func-registry" registry:2
> ```
> 
> This applies to both the `docker` and `podman` branches inside the `registry()` function in `cluster.sh`.
> 
> What are your thoughts?

There are various school of thought.
I myself would prefer listening on all interfaces `::` as you suggested. Some people prefer listening on localhost only for security reason, but IMO that folly and one should use firewall for that.

matejvasek:
Maybe the podman issue was just a flake?

matejvasek:
@twoGiants please check this PR, you know JS more than me

Claude reviewd this and does not like the change of bundler

>1. moduleResolution: "bundler" is incompatible with module: "commonjs" (likely broken)...

But maybe it's all right, I do not know.

templates/typescript/cloudevents/package.json
```diff
@@ -8,10 +8,10 @@
     "url": ""
   },
   "scripts": {
-    "build": "npx -p typescript tsc",
```
twoGiants, line 11: Good catch 😸 👍

templates/typescript/cloudevents/tsconfig.json
```diff
@@ -15,7 +16,7 @@
     "noUnusedLocals": true,                      /* Report errors on unused locals. */
     "noUnusedParameters": true,                  /* Report errors on unused parameters. */
     "noImplicitReturns": true,                   /* Report error when not all code paths in function return a value. */
-    "moduleResolution": "node",                  /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
+    "moduleResolution": "bundler",               /* Specify module resolution strategy. */
```
twoGiants: This Node.js function is not intended to be bundled by esbuild, webpack etc. and as @matejvasek and Claude pointed out `bundle` resolution does not go well with `commonjs` module.

I saw in your PR description that you tested it and it worked which is great and due to Microsofts relaxation of the TS5 rule where `bundle` resolution with `commonjs` modules was forbidden. Here is [blog post](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#combining---moduleresolution-bundler-with---module-commonjs) about it. But this is not supposed to be the final state.

If we change this at all I'd go with the recommendation from Microsoft and migrate towards `--module nodenext` which is the right choice for a Node.js backend.

twoGiants (changes requested):
Hi @Ankitsinghsisodya ! 👋 😸 

Thank you for your contribution. See my comments below.

@matejvasek I addressed your comment.

templates/typescript/cloudevents/tsconfig.json
Ankitsinghsisodya: @twoGiants sir done.

templates/typescript/cloudevents/package.json
Ankitsinghsisodya: @twoGiants sir done.

twoGiants (approved):
/approve
/lgtm

matejvasek:
I did rerun it, but it may be legit error:
```
node:internal/errors:484
    ErrorCaptureStackTrace(err);
    ^
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension "" for /home/prow/go/src/knative.dev/func/templates/typescript/cloudevents/node_modules/tape/bin/tape
    at __node_internal_captureLargerStackTrace (node:internal/errors:484:5)
    at new NodeError (node:internal/errors:393:5)
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:79:11)
    at defaultGetFormat (node:internal/modules/esm/get_format:121:38)
    at defaultLoad (node:internal/modules/esm/load:81:20)
    at nextLoad (node:internal/modules/esm/loader:163:28)
    at load (file:///home/prow/go/src/knative.dev/func/templates/typescript/cloudevents/node_modules/tsx/dist/esm/index.mjs:2:1777)
    at nextLoad (node:internal/modules/esm/loader:163:28)
    at ESMLoader.load (node:internal/modules/esm/loader:601:26)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:457:22)
    at new ModuleJob (node:internal/modules/esm/module_job:63:26)
    at #createModuleJob (node:internal/modules/esm/loader:476:17)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:434:34)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:526:24)
    at async loadESM (node:internal/process/esm_loader:91:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}
Node.js v18.10.0
make: *** [Makefile:265: test-typescript] Error 1
╔═════════════════════════════════════════════════╗
║                                                 ║
║   BUILT-IN TEMPLATE TESTS FAILED TESTS PASSED   ║
║                                                 ║
║           at 2026-04-28 12:49:31.389            ║
║                                                 ║
╚═════════════════════════════════════════════════╝
+ EXIT_VALUE=2
+ set +o xtrace
```

matejvasek:
The test failed again. Possibly because the knative test infra uses NodeJS 18?

matejvasek:
You probably also need to update test/presubmit-tests.sh too.
https://github.com/knative/func/blob/main/test/presubmit-tests.sh#L23

matejvasek:
Why you changed `tsx` => `node --require tsx/cjs` in the latest commit?
Wasn't updating node version in `test/presubmit-tests.sh` sufficient?

Ankitsinghsisodya:
> Why you changed `tsx` => `node --require tsx/cjs` in the latest commit? Wasn't updating node version in `test/presubmit-tests.sh` sufficient?

Locally some tests were failing. Changing tsx => node --require tsx/cjs worked because it bypasses esm entirely and uses  commonjs hooks, which has no problem with extensonless files

PR #3648, opened 2026-04-29
Upgrade go-containerregistry to v0.21.5 and fix moby upgrade

creydr:
go-containerregistry v0.21.3 migrated its `daemon.Client` interface from `docker/docker` to `moby/moby` types. Since `PusherDockerClient` embedded `daemon.Client`, the `closeGuardingClient` (wrapping `docker/docker/client.APIClient`) no longer satisfied the interface. Rather than switching the entire docker client stack to moby/moby, this PR decouples the two concerns:

- `PusherDockerClient` handles pushing via the Docker daemon API (`ImagePush` + `Close`)
- `DaemonClientFactory` provides a separate `daemon.Client` for loading images from the local daemon when direct registry push is needed (e.g., cluster-private registries)

Hint: This is some prework to get func CLI automatically creating a Function CR from func-operator (dependency mismatch - and this bump should be done at some point anyhow)


pkg/docker/pusher.go
```diff
@@ -29,26 +29,31 @@ import (
...
 type Opt func(*Pusher)
 
 // PusherDockerClient is sub-interface of client.APIClient required by pusher.
 type PusherDockerClient interface {
-	daemon.Client
 	ImagePush(ctx context.Context, ref string, options image.PushOptions) (io.ReadCloser, error)
 	Close() error
 }
 
 type PusherDockerClientFactory func() (PusherDockerClient, error)
```
matejvasek, line 44: Is this still used?

pkg/docker/pusher.go
```diff
@@ -29,26 +29,31 @@ import (
...
+// DaemonClientFactory creates a client that satisfies go-containerregistry's
+// daemon.Client interface for loading images from the local docker daemon.
+type DaemonClientFactory func() (daemon.Client, error)
+
 // Pusher of images from local to remote registry.
 type Pusher struct {
 	verbose             bool // verbose logging.
 	credentialsProvider oci.CredentialsProvider
 	transport           http.RoundTripper
 	dockerClientFactory PusherDockerClientFactory
```
matejvasek, line 55: Is this still used?

pkg/docker/pusher.go
creydr: The Pusher.dockerClientFactory uses this type. This get's set in NewPusher() and WithPusherDockerClientFactory() and is used in Pusher.daemonPush()

pkg/docker/pusher.go
creydr: seems to be set in NewPusher() and WithDaemonClientFactory() and used in Pusher.push()

matejvasek:
superseded by https://github.com/knative/func/pull/3649

PR #3653, opened 2026-05-01
fix nightly build

dprotaso:
Fix the nightly build - it runs build tests which does a go-licenses check and requires us to ignore a test library that it can't infer the license for


Cali0707 (approved):
/lgtm
/approve

PR #3655, opened 2026-05-04
fix(knative): handle multi-path KUBECONFIG entries

### LINKED ISSUE #3534: Fails when KUBECONFIG contains multiple colon-separated paths
# `kn func` fails when `KUBECONFIG` contains multiple colon-separated paths

## Describe the bug

`kn func` commands that require cluster access (e.g., `deploy`, `list`, `invoke`) fail with an `ErrInvalidKubeconfig` error when the `KUBECONFIG` environment variable is set to a colon-separated list of paths — which is standard Kubernetes behavior supported by `kubectl` and `client-go`.

## Expected behavior

`kn func` should accept a `KUBECONFIG` value with multiple colon-separated paths (e.g., `~/.kube/config:/other/kubeconfig`) the same way `kubectl` and other Kubernetes tooling does.

## Actual behavior

The command fails immediately with an error like:

```
Error: invalid kubeconfig: kubeconfig file does not exist at path: /home/user/.kube/config:/other/kubeconfig
```

## Possible root cause

`validateKubeconfigFile()` in `pkg/knative/client.go` reads the `KUBECONFIG` env var and passes the entire string directly to `os.Stat()`, treating it as a single file path:

```go
func validateKubeconfigFile() error {
    kubeconfigPath := os.Getenv("KUBECONFIG")
    if kubeconfigPath == "" {
        return nil
    }
    if _, err := os.Stat(kubeconfigPath); os.IsNotExist(err) {
        return fmt.Errorf("%w: kubeconfig file does not exist at path: %s", fn.ErrInvalidKubeconfig, kubeconfigPath)
    }
    return nil
}
```

When `KUBECONFIG=/path/a:/path/b`, `os.Stat("/path/a:/path/b")` fails because `/path/a:/path/b` is not a valid single path.

## Steps to reproduce

1. Set `KUBECONFIG` to multiple paths: `export KUBECONFIG=~/.kube/config:/path/to/second/config`
2. Run any cluster-bound command: `kn func list`
3. Observe the `ErrInvalidKubeconfig` error.

## Environment

- `kn func` version: v0.48.3
- OS: macOS 26.4
- `KUBECONFIG` value format: multiple colon-separated paths

xenonnn4w:
# Reproduction
Successfully reproduced the bug

# Changes

- Fix validation of KUBECONFIG when multiple paths are provided
- Use filepath.SplitList to correctly handle OS-specific path separators
- Treat KUBECONFIG as a list of paths instead of a single file path
- Validation now succeeds if at least one kubeconfig file exists
- Return ErrInvalidKubeconfig only when all paths are missing or unreadable
- Add unit tests covering:
  - empty KUBECONFIG
  - single valid path
  - multiple paths (valid + missing)
  - all paths missing

/kind bug

Fixes #3534

### Case 1: Multiple valid paths

```sh
export KUBECONFIG=~/.kube/config:/tmp/second-config
func list
```

**Expected:**

- Validation succeeds (no invalid path error)
- Command proceeds to cluster connection stage

### Case 2: One valid + one missing

```sh
export KUBECONFIG=~/.kube/config:/tmp/missing
func list
```

**Expected:**

- Validation succeeds

### Case 3: All paths missing

```sh
export KUBECONFIG=/tmp/miss1:/tmp/miss2
func list
```

**Expected:**

- invalid kubeconfig: kubeconfig file does not exist at path: /tmp/miss1:/tmp/miss2

### Case 4: Empty KUBECONFIG

```sh
unset KUBECONFIG
func list
```

**Expected:**

- No validation error

# Extra

make test ; passed
make check ;passed

**Release Note**

```release-note
func commands now correctly support multi-path KUBECONFIG values. Validation succeeds if at least one referenced kubeconfig file exists, aligning behavior with kubectl and client-go.```

### DIFF
--- a/pkg/knative/client.go
+++ b/pkg/knative/client.go
@@ -3,6 +3,7 @@ package knative
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	clienteventingv1 "knative.dev/client/pkg/eventing/v1"
 	clientservingv1 "knative.dev/client/pkg/serving/v1"
@@ -60,9 +61,18 @@ func validateKubeconfigFile() error {
 		return nil
 	}
 
-	if _, err := os.Stat(kubeconfigPath); os.IsNotExist(err) {
-		return fmt.Errorf("%w: kubeconfig file does not exist at path: %s", fn.ErrInvalidKubeconfig, kubeconfigPath)
+	for _, path := range filepath.SplitList(kubeconfigPath) {
+		if path == "" {
+			continue
+		}
+		_, statErr := os.Stat(path)
+		if statErr == nil {
+			return nil
+		}
+		if !os.IsNotExist(statErr) {
+			return fmt.Errorf("%w: kubeconfig file is not accessible at path: %s", fn.ErrInvalidKubeconfig, path)
+		}
 	}
 
-	return nil
+	return fmt.Errorf("%w: kubeconfig file does not exist at path: %s", fn.ErrInvalidKubeconfig, kubeconfigPath)
 }
--- a/pkg/knative/client_test.go
+++ b/pkg/knative/client_test.go
@@ -1,0 +1,79 @@
+package knative
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	fn "knative.dev/func/pkg/functions"
+)
+
+func TestValidateKubeconfigFile(t *testing.T) {
+	testDir := t.TempDir()
+	existingPath := filepath.Join(testDir, "kubeconfig")
+	if err := os.WriteFile(existingPath, []byte("apiVersion: v1\n"), 0o600); err != nil {
+		t.Fatalf("write kubeconfig: %v", err)
+	}
+	missingPath := filepath.Join(testDir, "missing")
+	listSep := string(filepath.ListSeparator)
+
+	cases := []struct {
+		name        string
+		kubeconfig  string
+		wantErr     bool
+		wantErrText string
+	}{
+		{
+			name:       "empty env",
+			kubeconfig: "",
+			wantErr:    false,
+		},
+		{
+			name:       "single existing",
+			kubeconfig: existingPath,
+			wantErr:    false,
+		},
+		{
+			name:        "single missing",
+			kubeconfig:  missingPath,
+			wantErr:     true,
+			wantErrText: "kubeconfig file does not exist at path",
+		},
+		{
+			name:       "multiple with existing",
+			kubeconfig: missingPath + listSep + existingPath,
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: pkg/knative/client.go [hunk 1]
```diff
@@ -60,7 +61,27 @@ func validateKubeconfigFile() error {
 		return nil
 	}
 
-	if _, err := os.Stat(kubeconfigPath); os.IsNotExist(err) {
+	paths := filepath.SplitList(kubeconfigPath)
+	foundExisting := false
+	missingCount := 0
+	for _, path := range paths {
```
gauron99: can we simplify this to simply return the first found instead?
something like
```
 for _, path := range paths {
       // if path == "" - skip
       if _, err := os.Stat(path); err == nil {
               return nil
       }
       // handle errors

```

pkg/knative/client_test.go
```diff
@@ -0,0 +1,74 @@
...
+		kubeconfig  string
+		wantErr     bool
+		wantErrText string
+	}{
+		{
+			name:       "empty env",
+			kubeconfig: "",
+			wantErr:    false,
+		},
+		{
```
gauron99, line 38: add the "single correct" case please

gauron99 (commented):
thanks for the PR, few comments below

pkg/knative/client_test.go
xenonnn4w: i see!

xenonnn4w:
thanks for the review!

pkg/knative/client.go
```diff
@@ -60,7 +61,28 @@ func validateKubeconfigFile() error {
 		return nil
 	}
 
-	if _, err := os.Stat(kubeconfigPath); os.IsNotExist(err) {
+	paths := filepath.SplitList(kubeconfigPath)
+	havePaths := false
+	missingCount := 0
```
gauron99: do we still need both of these variables?

pkg/knative/client.go
gauron99: I mightve not been clear enough in my explanation. I dont think we need to capture the number of  missing paths at all? If I understand it, we can simplify the loop to simply skip empty, return on first successful path to validate the kubeconfig or return error if no valid paths are present.
isnt `havePaths` unnecessary? we return good path immediately. And following logically if we ever get to missingCount, isnt it always true? If there is some error found in the loop and 0 successful paths (nothing was returned) then the `missingCount` is always true?

pkg/knative/client.go
xenonnn4w: ah i see, both were redundant. Simplified!

gauron99:
/lgtm
/approve
lets see if the integration tests were a flake

PR #3657, opened 2026-05-05
Add func-operator integration to deploy command 

creydr:
# Changes

- :gift: Create a Function CR on `func deploy` when func-operator is installed
- :gift: Add option to skip Function CR creation via `func deploy --management-disabled`
- :wrench: Bump Go to 1.26 (required by `controller-runtime` dependency)

/kind enhancement

**Release Note**
```release-note
Create Function CR on `func deploy`
```


go.mod
```diff
@@ -1,6 +1,6 @@
 module knative.dev/func
 
-go 1.25.9
+go 1.26
```
lkingland, line 3: I presume this was necessary due to `controller-runtime` dependency?  Maybe just add to the commit description this PR also bumps to Go 1.26

pkg/operator/sync.go
```diff
@@ -0,0 +1,173 @@
...
+	hasCRD, err := hasFunctionCRD(disc)
+	if err != nil {
+		return fmt.Errorf("checking for Function CRD: %w", err)
+	}
+	if !hasCRD {
+		return nil
+	}
+
+	if cfg.RepoURL == "" {
+		fmt.Fprintln(os.Stderr, "Function CR not created: no git remote found. Set --git-url to enable operator management.")
```
lkingland: This warning will fire on every deploy without a configured git remote, given `--manage=true` is the default.  I think that there are actually two preconditions for a Function to be "managed":  The operator is installed AND they're using git.

The base case:  A user just initializes a function and deployes it directly to a vanilla cluster, shouldn't really even print a warning (it's an expected base case).

pkg/operator/sync.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	// Look up existing CR
+	existing, err := findExistingCR(ctx, cl, cfg.FunctionName, cfg.Namespace)
+	if err != nil {
+		return fmt.Errorf("looking up existing Function CR: %w", err)
+	}
+
+	if existing != nil {
+		existing.Spec = desiredSpec
```
lkingland, line 112: Just double-checking:  without using a merge, this will always overwrite the spec, even if changed in some different way.  I think that's what we want (again encouraging declarative config), but just want to flag this to be sure.

pkg/operator/sync.go
```diff
@@ -0,0 +1,173 @@
...
+		return fmt.Errorf("creating Function CR: %w", err)
+	}
+	fmt.Fprintf(os.Stderr, "Function CR %q created in namespace %q\n", cfg.FunctionName, cfg.Namespace)
+	return nil
+}
+
+func hasFunctionCRD(disc discovery.DiscoveryInterface) (bool, error) {
+	resources, err := disc.ServerResourcesForGroupVersion("functions.dev/v1alpha1")
+	if err != nil {
+		return false, nil
```
lkingland: This swallows any error from `ServerResourcesForGroupVersion` and treats it identically to "CRD not present." If there's a legitimate reason, we probably want to raise an error.  Perhaps something like:

```go
resources, err := disc.ServerResourcesForGroupVersion("functions.dev/v1alpha1")
if err != nil {
    if apierrors.IsNotFound(err) || meta.IsNoMatchError(err) {
        return false, nil
    }
    return false, fmt.Errorf("checking for Function CRD: %w", err)
}
```

(I haven't fully traced the call path here; just an example)

pkg/operator/sync.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	for i := range list.Items {
+		if list.Items[i].Status.Name == funcName {
+			return &list.Items[i], nil
+		}
+	}
+
+	for i := range list.Items {
+		if list.Items[i].Name == funcName {
```
lkingland: Minor nit:  these two loops over list.Items made me double-check to see if it was an error.  maybe just put both `if` statements in one for-loop for clarity that this was intended.

cmd/deploy.go
```diff
@@ -808,6 +821,15 @@ func (c deployConfig) clientOptions() ([]fn.Option, error) {
...
 	}
 
+	if c.Manage {
+		o = append(o, fn.WithPostDeploy(func(ctx context.Context, f fn.Function) error {
+			if err := syncFunctionCR(ctx, f, c, creds); err != nil {
+				fmt.Fprintf(os.Stderr, "Warning: failed to sync Function CR: %v\n", err)
+			}
+			return nil
+		}))
+	}
```
lkingland: This should probably be part of core rather than in the CLI.  Adding a boolean to the Function structh which states this is a "managed" Function, and a CR should be created (if the operator is installed).

pkg/functions/client.go
```diff
@@ -82,6 +82,7 @@ type Client struct {
 	pipelinesProvider PipelinesProvider // CI/CD pipelines management
 	mcpServer         MCPServer         // MCP Server
 	startTimeout      time.Duration     // default start timeout for all runs
+	postDeploy        func(context.Context, Function) error
```
lkingland: No need for a post-deploy hook if we just extend the core's "deploy" logic to do the CR work if the given function is a "managed" function.

To make it even simpler, the flag could be "ManagementDisabled", such that the deafult will always be to create the CR if the system detects the operator is installed.

lkingland (commented):
Looks great!  My only substantive suggestion is that this be integrated more deeply into core instead of an ad-hoc post-deploy hook 👍🏻

creydr:
rebased and resolved merge conflicts

go.mod
creydr: Yes, required by controller-runtime. Added a note to the PR description.

pkg/operator/sync.go
creydr: Good point. Changed to skip silently when no git remote is configured — the two preconditions are now operator CRD installed AND git URL available.

pkg/operator/sync.go
creydr: Correct, this is intentional. The spec is always overwritten with the local state to encourage declarative config.

pkg/operator/sync.go
creydr: Fixed. Now only `IsNotFound` errors are treated as CRD absence; other errors are propagated.

pkg/operator/sync.go
creydr: Combined into a single loop. Status name match still takes priority (returns immediately); metadata name match is remembered as fallback.

cmd/deploy.go
creydr: Done. Moved into core `Client.Deploy()` via a `FunctionSyncer` interface, with `ManagementDisabled` field on `DeploySpec` (managed by default).

pkg/functions/client.go
creydr: Done. Replaced `postDeploy` hook with `FunctionSyncer` interface. `--manage` flag replaced by `--management-disabled`.

creydr:
Thanks for the review @lkingland 
I addressed your review (hopefully :D )

Only maybe one thing we could also address in this PR: WDYT about removing the CR when `--management-disabled` is set?

creydr:
rebased

creydr:
@lkingland can you recheck?

creydr:
rebased

lkingland:
Looks good @creydr!  Let's rebase and then land this one

creydr:
@lkingland rebased :)

lkingland (approved):

PR #3661, opened 2026-05-06
mcp: split config tools by action for accurate tool annotations

Ankitsinghsisodya:
## Problem

`config_envs`, `config_labels`, and `config_volumes` each accepted an `action`
parameter (`list`, `add`, `remove`) but declared `DestructiveHint: true` at the
tool level. MCP tool annotations are static metadata evaluated at discovery time,
before any invocation arguments exist. A client that respects hints treats the tool
as destructive on every call — including a read-only `list`.

## Changes

Split each combined tool into three single-action tools, each carrying annotations
that accurately describe that action's semantics:

| Tool | ReadOnlyHint | DestructiveHint |
|---|---|---|
| `config_{envs,labels,volumes}_list` | true | — (irrelevant per spec when ReadOnly) |
| `config_{envs,labels,volumes}_add` | false | false (additive only) |
| `config_{envs,labels,volumes}_remove` | false | true |

The `action` field is removed from all input structs. Each tool's `Args()` method
hardcodes the correct subcommand. `IdempotentHint` is set to `true` on list tools.

## Tests

Nine dedicated tests (list/add/remove per resource) replace the previous six combined
tests. All pass.

### DIFF
--- a/pkg/mcp/instructions.md
+++ b/pkg/mcp/instructions.md
@@ -33,7 +33,7 @@ This is essential because:
 - `create` tool: path = absolute path to directory where Function will be created
 - `deploy` tool: path = absolute path to Function directory (where func.yaml exists)
 - `build` tool: path = absolute path to Function directory (where func.yaml exists)
-- `config_*` tools: path = absolute path to Function directory (where func.yaml exists)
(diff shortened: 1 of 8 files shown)


Ankitsinghsisodya:
cc @gauron99

lkingland (approved):
/lgtm

Technically a breaking change, but LLM MCP clients do discovery, so should be fine.

PR #3662, opened 2026-05-06
mcp: use build-time version instead of hardcoded "0.1.0"

Ankitsinghsisodya:
## Summary

- The MCP server advertised a static `version = "0.1.0"` constant to every
  MCP client regardless of the actual binary version, causing stale version
  metadata in clients that display server info.
- Replace the constant with `version.Vers` from `pkg/version`, which is
  already injected at build time via ldflags
  (`-X knative.dev/func/pkg/version.Vers=$(VERS)`).
- Fall back to `"0.0.0+source"` when the variable is empty (source builds
  that bypass the Makefile), consistent with how `cmd/root.go` handles the
  same case.

## Changes

- `pkg/mcp/mcp.go`: remove `const version = "0.1.0"`, import
  `knative.dev/func/pkg/version`, and use `version.Vers` (with fallback)
  when constructing the `mcp.Implementation` passed to `mcp.NewServer`.

## Test plan

- [ ] `go test ./pkg/mcp/...` passes
- [ ] `make test` passes for mcp package
- [ ] `func version` and `func mcp start` report the same version string
- [ ] MCP client (e.g. Claude Code) shows the correct version after
      `func mcp start`

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -8,12 +8,12 @@ import (
 	"sync/atomic"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"knative.dev/func/pkg/version"
 )
 
 const (
-	name    = "func"
-	title   = "func"
-	version = "0.1.0"
+	name  = "func"
+	title = "func"
 )
 
 // NOTE: Invoking prompts in some interfaces (such as Claude Code) when all
@@ -87,11 +87,16 @@ func New(options ...Option) *Server {
 		o(s)
 	}
 
+	vers := version.Vers
+	if vers == "" {
+		vers = "0.0.0+source"
+	}
+
 	i := mcp.NewServer(
 		&mcp.Implementation{
 			Name:    name,
 			Title:   title,
-			Version: version},
+			Version: vers},
 		&mcp.ServerOptions{
 			Instructions:       instructions(s.readonly.Load()),
 			HasPrompts:         true,
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -87,16 +87,11 @@ func New(options ...Option) *Server {
 		o(s)
 	}
 
-	vers := version.Vers
-	if vers == "" {
-		vers = "0.0.0+source"
-	}
-
 	i := mcp.NewServer(
 		&mcp.Implementation{
 			Name:    name,
 			Title:   title,
-			Version: vers},
+			Version: version.Get().String()},
 		&mcp.ServerOptions{
 			Instructions:       instructions(s.readonly.Load()),
 			HasPrompts:         true,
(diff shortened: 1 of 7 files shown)


Ankitsinghsisodya:
cc @gauron99

gauron99:
I wonder if mcp version should be different to func cli version. Underneath we just run 'func' as a command so technically its possible to run `/func1 mcp start` and that invokes `func` as a different binary, resulting in 2 different versions being used for starting the mcp and running the commands.

I like more that the mcp has its own versioning to be fair but we can have it be the same for the sake of identifying what func is used to run the mcp server. or we can display both the mcp version and func version?
maybe its an over-optimized approach for a problem we yet dont have :smile: 
@lkingland wdyt?

pkg/mcp/mcp.go
```diff
@@ -74,11 +74,16 @@ func New(options ...Option) *Server {
 		o(s)
 	}
 
+	vers := version.Vers
+	if vers == "" {
+		vers = "0.0.0+source"
+	}
```
lkingland: This splits the logic of the default version into two places.

Let's consolidate into a single `Get() *semver.Version`.

Then it's all in one place, AND we can decide on the leading `v` depending on the situation:  human-readable output prefers leading v, whereas machine-readable (like the MCP server) perfer a proper semver format which does not.

Quick sketch:
```go
  // pkg/version/version.go
  package version

  import "github.com/Masterminds/semver/v3"

  var Vers, Kver, Hash string

  const DefaultVers = "v0.0.0+source"

  // Get returns the parsed semver, with a fallback when no
  // build-time version was injected.
  func Get() *semver.Version {
      s := Vers
      if s == "" {
          s = DefaultVers
      }
      v, err := semver.NewVersion(s)  // permissive: accepts leading 'v'
      if err != nil {
          v, _ = semver.NewVersion(DefaultVers)
      }
      return v
  }

  Callers:
  // cmd/version.go (human form):
  fmt.Printf("v%s\n", version.Get().String())  // String() returns "0.0.0+source"; we add the v
  // or use .Original() to round-trip whatever was injected, including the v.

  // pkg/mcp/mcp.go (machine form):
  Version: version.Get().String(),  // "0.0.0+source", clean semver

```

lkingland (changes requested):

lkingland:
> I wonder if mcp version should be different to func cli version. Underneath we just run 'func' as a command so technically its possible to run `/func1 mcp start` and that invokes `func` as a different binary, resulting in 2 different versions being used for starting the mcp and running the commands.
> 
> I like more that the mcp has its own versioning to be fair but we can have it be the same for the sake of identifying what func is used to run the mcp server. or we can display both the mcp version and func version? maybe its an over-optimized approach for a problem we yet dont have 😄 @lkingland wdyt?

That's a good point about perhaps the MCP server having a different version than func, but for simplicity let's just keep them the same for now so we don't forget to bump versions!

pkg/mcp/mcp.go
Ankitsinghsisodya: done sir!!

lkingland (approved):

lkingland:
I think that failing unit test would be fixed with a rebase.

PR #3663, opened 2026-05-06
fix: write `func.yaml` before remote deploy upload

creydr:
Previously, `func.yaml` was only written to disk after deployment completed. For remote deploys, the pipeline uploads sources from disk to a PVC before the on-cluster deploy step runs. This meant CLI flags like `--image-pull-secret`, `--service-account`, or `--deployer` were not reflected in the uploaded `func.yaml`.

Write `func.yaml` before the pipeline starts so the on-cluster deploy step sees the latest configuration.

# Changes

- :bug: Fix `--image-pull-secret` support on remote deploys

/kind bug

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -318,6 +318,12 @@ func runDeploy(cmd *cobra.Command, newClient ClientFactory) (err error) {
 
 	// Deploy
 	if cfg.Remote {
+		// Write func.yaml before the pipeline uploads sources to the PVC,
+		// so that the on-cluster deploy step sees the latest config
+		// (e.g. --image-pull-secret, --service-account, --deployer).
+		if err = f.Write(); err != nil {
+			return
+		}
 		var url string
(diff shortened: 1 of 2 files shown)


cmd/deploy.go
```diff
@@ -318,6 +318,12 @@ func runDeploy(cmd *cobra.Command, newClient ClientFactory) (err error) {
 
 	// Deploy
 	if cfg.Remote {
+		// Write func.yaml before the pipeline uploads sources to the PVC,
+		// so that the on-cluster deploy step sees the latest config
+		// (e.g. --image-pull-secret, --service-account, --deployer).
```
lkingland, line 323: Thank you for the explanatory comment.

Creating a follow-up for this becasue we don't write func.yaml until the end for purposes of keeping the source in a clean state, but this gets us over the current hurdle.

lkingland (approved):
/lgtm

lkingland:
Created a follow-up for this: #3679

PR #3667, opened 2026-05-06
fix(mcp): omit empty subcommand in defaultExecutor argument list

Ankitsinghsisodya:
## Problem

Every help resource handler calls `executor.Execute(ctx, "", cmd..., "--help")`, passing an empty string as the `subcommand` argument. `defaultExecutor.Execute` unconditionally appended that empty string to the command parts, producing invocations like:

```
func "" create --help
```

Cobra receives `""` as the first positional argument, fails to match it against any known subcommand, and returns an error. Every help resource was broken when used against a real `func` binary.

A second instance of the same class of bug existed on the root help resource registration, which passed `""` as a `cmd` variadic argument:

```go
// before
newHelpResource(s, "Help", "help for the command root", "")
```

This injected a second empty argument even with the executor fix in place.

The test suite did not catch either defect because every resource test uses a mock executor via `WithExecutor`, so `defaultExecutor` is never exercised.

## Fix

- Extract `buildArgs(prefix, subcommand string, args []string) []string` from `defaultExecutor.Execute`. The helper omits the subcommand when it is empty.
- Drop the spurious `""` argument from the root help resource registration.

## Tests

- `TestBuildArgs` — five table-driven cases directly testing `buildArgs`, including the critical empty-subcommand case and multi-word prefix handling.
- `TestResource_Help` — eleven table-driven cases covering every registered help URI end-to-end through the resource handler → mock executor path. Asserts that `subcommand` is always `""` and no element in `args` is ever an empty string.

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -114,7 +114,7 @@ func New(options ...Option) *Server {
 	// Help
 	// A resource for each command which returns its help
 	// eg. "config volumes add" -> "func://help/config/volumes/add")
-	i.AddResource(newHelpResource(s, "Help", "help for the command root", ""))
+	i.AddResource(newHelpResource(s, "Help", "help for the command root"))
 	i.AddResource(newHelpResource(s, "Create Help", "help for 'create'", "create"))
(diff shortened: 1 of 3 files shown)


lkingland (approved):
/lgtm
/approve

PR #3671, opened 2026-05-07
fix: apply resource validation to first-deploy path in k8s and keda deployers

Ankitsinghsisodya:
## Problem

`generateDeployment` in `pkg/k8s/deployer.go` allocated its own local
`referencedSecrets`, `referencedConfigMaps`, and `referencedPVCs` sets
internally. Those sets were never visible to the caller, so
`CheckResourcesArePresent` was never called on the k8s deploy path —
silently passing validation regardless of whether referenced Secrets,
ConfigMaps, or PVCs actually existed on the cluster.

The keda deployer is a thin wrapper around the k8s deployer and inherited
the same gap.

This is the same class of bug fixed for the knative deployer in #3579.

Pass the three sets as parameters into `generateDeployment` instead of
allocating them locally. `Deploy()` now:

1. Allocates the sets before calling `generateDeployment`
2. Calls `CheckResourcesArePresent` after `generateDeployment` on both the
   create path (first deploy) and the update path (redeployment), before
   touching the API

The keda deployer gets the fix for free since it delegates to the k8s
deployer.

## Tests

- `TestInt_ResourceValidationOnFirstDeploy` wired in `pkg/k8s/deployer_int_test.go`
  and `pkg/keda/deployer_int_test.go` — exercises the same shared scenario
  from `pkg/deployer/testing/integration_test_helper.go` used by the
  knative deployer since #3579
- Existing unit tests in `pkg/k8s/deployer_test.go` updated to match the
  new `generateDeployment` signature

### DIFF
--- a/pkg/k8s/deployer.go
+++ b/pkg/k8s/deployer.go
@@ -157,7 +157,7 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
 		}
 
 		if err = CheckResourcesArePresent(ctx, namespace, &referencedSecrets, &referencedConfigMaps, &referencedPVCs, f.Deploy.ServiceAccountName, f.Deploy.ImagePullSecret); err != nil {
-			return fn.DeploymentResult{}, fmt.Errorf("failed to update deployment: %w", err)
+			return fn.DeploymentResult{}, fmt.Errorf("failed to validate referenced resources: %w", err)
 		}
 
 		svc, err := d.generateService(f, namespace, daprInstalled, existingDeployment)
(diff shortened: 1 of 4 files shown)


Ankitsinghsisodya:
cc @matejvasek

lkingland (approved):
LGTM. The bug fix is correct, well-scoped, and well-tested. Validation now fires early on both the create and update paths in the k8s deployer, and keda inherits the fix cleanly via delegation. This is a real user-facing bug — failing fast on missing Secrets/ConfigMaps/PVCs beats failing late at mount time with cryptic errors — and the parallel to #3579 keeps the deployers consistent.

One design observation worth flagging, captured as follow-up at #3678 so it doesn't get lost. It's pre-existing and explicitly out of scope for this PR.

The reference-tracking pattern this PR extends — three `*sets.Set[string]` out-parameters threaded through `generateDeployment`, alongside the existing `ProcessEnvs`, `ProcessVolumes`, `createEnvFromSource`, `createEnvVarSource`, and `CheckResourcesArePresent` — has three layers of smell:

1. **Pointer-to-map is doubly indirect.** `sets.Set[T]` is `map[T]Empty`; the map is already a reference. The `*` doesn't unlock any capability the callers use (nobody reassigns or null-checks). Idiomatic Go would pass `sets.Set[T]` by value.
2. **A coupled triple, passed loose.** `referencedSecrets, referencedConfigMaps, referencedPVCs` always travel together. Naming them as a `References` struct collapses every signature.
3. **Accumulator-by-out-param.** Multi-return or a `Tracker` receiver would express the producer-plus-tracking semantics more directly than threading mutable parameters down the call chain. (And, notably, a `Tracker` receiver would have made the original bug — locally allocated sets in `generateDeployment` invisible to the caller — much harder to introduce.)

This PR is consistent with the surrounding code, so changing the convention here would be a refactor unrelated to the bug. Asking @Ankitsinghsisodya to absorb that scope would punish the wrong person and gate a real fix on cleanup work. Issue #3678 captures the cleanup with enough back-story for someone with cycles to pick it up — it's a reasonable good-first-issue candidate too.

/approve
/lgtm

PR #3676, opened 2026-05-07
Fix PVC cleanup between on-cluster builds

### LINKED ISSUE #3516: On-cluster build PersistentVolume fills up across builds
The PersistentVolume used for on-cluster builds (both standard and
Pipelines-as-Code) is reused across builds without cleanup. Over time
it fills up, causing subsequent builds to fail with "No space left on
device" during git-init.

The build pipeline should clean up or rotate workspace PVs between
builds to prevent accumulation.

Related: [SRVOCF-583](https://redhat.atlassian.net/browse/SRVOCF-583)

Itx-Psycho0:
## Problem

This PR addresses issue #3516, where the PersistentVolumeClaim used for on-cluster builds was being reused without cleanup, causing 'No space left on device' errors after multiple builds.

## Solution

The fix ensures that each build starts with a clean PVC by deleting and recreating it before each build. This approach is simpler and more reliable than trying to clean up the PVC contents while it's in use.

## Changes

- Added `DeletePersistentVolumeClaim()` function in `pkg/k8s/persistent_volumes.go` to delete a single PVC by name
- Added `WaitForPVCDeletion()` function in `pkg/k8s/persistent_volumes.go` to wait for complete PVC deletion (not just Terminating state)
- Modified `createPipelinePersistentVolumeClaim()` in `pkg/pipelines/tekton/pipelines_provider.go` to:
  1. Check if PVC exists
  2. Delete it if found (ignoring NotFound errors)
  3. Wait for complete deletion
  4. Create a fresh PVC
- Updated tests in `pkg/pipelines/tekton/pipelines_provider_test.go` to properly mock the new k8s functions

## Testing

- All unit tests pass
- Linting checks pass
- Added test cases for:
  - Successful deletion and recreation when PVC exists
  - Error handling when deletion fails
  - Error handling when waiting for deletion fails

Fixes #3516

### DIFF
--- a/pkg/k8s/persistent_volumes.go
+++ b/pkg/k8s/persistent_volumes.go
@@ -72,6 +72,47 @@ func DeletePersistentVolumeClaims(ctx context.Context, namespaceOverride string,
 	return client.CoreV1().PersistentVolumeClaims(namespace).DeleteCollection(ctx, metav1.DeleteOptions{}, listOptions)
 }
 
+// DeletePersistentVolumeClaim deletes a single PVC by name
+func DeletePersistentVolumeClaim(ctx context.Context, name, namespaceOverride string) error {
+	client, namespace, err := NewClientAndResolvedNamespace(namespaceOverride)
+	if err != nil {
+		return err
+	}
+
+	err = client.CoreV1().PersistentVolumeClaims(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return fmt.Errorf("failed to delete PVC %s: %w", name, err)
+	}
+	return nil
+}
+
+// WaitForPVCDeletion waits for a PVC to be fully deleted (not just in Terminating state)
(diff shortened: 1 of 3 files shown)


Itx-Psycho0:
Hi maintainers,

I've implemented a fix for issue #3516 that addresses the PVC space exhaustion problem during on-cluster builds. The solution deletes and recreates the PVC before each build to ensure a clean workspace.

All unit tests and linting checks pass. I'd appreciate your review when you have a chance.

@lkingland @matejvasek @jrangelramos @gauron99 

Thanks!

lkingland (approved):

matejvasek:
This seems to break build. Second build simply hangs on deletion.

Have you tested this at least locally by hand? @Itx-Psycho0 
Why are there no **integration** tests?

matejvasek:
Also why delete it? We can use the volume for more efficient upload. We don't do it now but we conceivably could upload only changed files for instance.

matejvasek:
The volume should be deleted only on `func delete`.

PR #3666, opened 2026-05-06
pkg/mcp: add tests for readonly enforcement, delete validation, and help resources

Ankitsinghsisodya:
## Summary

- `tools_deploy_test.go` and `tools_delete_test.go` each had a single happy-path test that manually set `server.readonly = false`, leaving the readonly guard in both handlers completely untested.
- The mutual-exclusion validation in `deleteHandler` (exactly one of `path` or `name` must be provided) had no test coverage for either branch.
- All 14 `func://help/*` resources registered in `mcp.go` had zero tests; the `newHelpResource` handler — URI construction, executor dispatch, args assembly, and response shape — was entirely unexercised.

## Changes

- `pkg/mcp/tools_deploy_test.go`: add `TestTool_Deploy_Readonly` — uses `newTestPairWithReadonly(t, true)` and asserts `result.IsError`.
- `pkg/mcp/tools_delete_test.go`: add `TestTool_Delete_Readonly`, `TestTool_Delete_BothPathAndName`, and `TestTool_Delete_NeitherPathNorName`.
- `pkg/mcp/resources_test.go`: add `TestResource_Help` — table-driven test over 6 `func://help/*` URIs that validates executor subcommand, args, response URI, MIME type, and body text.

## Test plan

- [ ] `go test ./pkg/mcp/...` passes (all 10 new tests green)
- [ ] `make test` passes
- [ ] `make check` passes

PR #3681, opened 2026-05-08
mcp: populate Image and URL structured output fields

Ankitsinghsisodya:
## What

The `build` and `deploy` MCP tool handlers declared structured output fields
(`image`, `url`) that were never populated — only the raw `message` string was
set. Agents inspecting the structured response to learn the deployed URL or
built image name always received empty strings and had to parse raw text instead.

## Why

Structured output fields exist so MCP clients can consume typed values without
text scraping. Empty fields make the schema misleading and break any agent that
relies on them.

## How

**`tools_build.go`** — after `func build` exits, reload the function from disk
and read `f.Build.Image`. The build CLI writes the final image reference (with
digest) to `.func/built-image`; `fn.NewFunction` picks it up via
`getLastBuiltImage()`.

**`tools_deploy.go`** — two changes:
- Added `parseDeployedURL` to extract the route URL from combined stdout. Handles
  both the local-deploy format (`exposed at URL:\n   <url>`) and the remote
  pipeline format (`Function Deployed at <url>`).
- After the subprocess exits, reload the function and read `f.Deploy.Image` from
  `func.yaml` (where the deploy step persists it).

## Tests

- `TestTool_Build_StructuredOutput` — verifies `Image` is read from
  `.func/built-image` after a successful build.
- `TestTool_Build_StructuredOutput_NoFuncYaml` — verifies that a missing
  `func.yaml` does not cause the handler to fail; `Image` is empty but no
  error is returned.
- `TestParseDeployedURL` — table-driven coverage of both output formats,
  surrounding noise, and the empty / no-URL cases.
- `TestTool_Deploy_StructuredOutput` — end-to-end: mock executor returns
  local-deploy output; asserts `URL` is correctly extracted.

### DIFF
--- a/pkg/mcp/tools_build.go
+++ b/pkg/mcp/tools_build.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
+	fn "knative.dev/func/pkg/functions"
 )
 
 var buildTool = &mcp.Tool{
@@ -27,6 +28,9 @@ func (s *Server) buildHandler(ctx context.Context, r *mcp.CallToolRequest, input
 	}
 	output = BuildOutput{
 		Message: string(out),
+	}
+	if f, ferr := fn.NewFunction(input.Path); ferr == nil {
+		output.Image = f.Build.Image
 	}
 	return
 }
--- a/pkg/mcp/tools_build.go
+++ b/pkg/mcp/tools_build.go
@@ -31,6 +31,8 @@ func (s *Server) buildHandler(ctx context.Context, r *mcp.CallToolRequest, input
 	}
 	if f, ferr := fn.NewFunction(input.Path); ferr == nil {
 		output.Image = f.Build.Image
+	} else {
+		output.Message += fmt.Sprintf("\n(warning: could not read built image from func.yaml: %v)", ferr)
 	}
 	return
 }
--- a/pkg/mcp/tools_build_test.go
+++ b/pkg/mcp/tools_build_test.go
@@ -2,6 +2,9 @@ package mcp
 
 import (
 	"context"
+	"encoding/json"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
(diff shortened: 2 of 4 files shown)


pkg/mcp/tools_deploy.go
```diff
@@ -31,10 +35,48 @@ func (s *Server) deployHandler(ctx context.Context, r *mcp.CallToolRequest, inpu
...
+			if u := strings.TrimSpace(line[idx+len(remotePrefix):]); u != "" {
+				return u
+			}
+		}
+	}
+	return ""
+}
+
 // DeployInput defines the input parameters for the deploy tool.
 type DeployInput struct {
```
lkingland, line 88: This works for the two current formats but is structurally brittle:

1. `"exposed at URL:"` and `"Function Deployed at "` are user-facing log strings. They've changed before and will likely change again (i18n, output redesign, etc.). When they do, `URL` will silently return `""` — no test fails, no error, agents just see empty fields.
2. `DeploySpec` doesn't currently have a `URL` field (I just checked `pkg/functions/function.go:195`). So the only options today are (a) parse output, (b) query the cluster post-deploy.

The better long-term shape — same conversation that came up on #3662 — is:

- Promote `URL` to a first-class field on `DeploySpec`.
- Have the deploy step write it to `func.yaml` alongside `Image`.
- MCP just reads it via `NewFunction`, same as you do for `Image`.

That eliminates this parser, makes URL discoverable to any downstream tool (not just MCP), and ends the prose-scraping pattern.

For this PR, the parsing approach is reasonable as an interim — but worth filing as a follow-up issue rather than letting it accrete more output formats over time.

pkg/mcp/tools_build.go
```diff
@@ -28,6 +29,9 @@ func (s *Server) buildHandler(ctx context.Context, r *mcp.CallToolRequest, input
 	output = BuildOutput{
 		Message: string(out),
 	}
+	if f, ferr := fn.NewFunction(input.Path); ferr == nil {
+		output.Image = f.Build.Image
```
lkingland, line 33: Silent failure: if `NewFunction` errors (corrupt `func.yaml`, permissions, etc.), `Image` is left empty with no signal to the caller. The build succeeded from the agent's POV, but it can't actually use the structured output.

At minimum, attach the error to the message:

```go
if f, ferr := fn.NewFunction(input.Path); ferr == nil {
    output.Image = f.Build.Image
} else {
    output.Message += fmt.Sprintf("\n(warning: could not read built image from func.yaml: %v)", ferr)
}
```

Same pattern applies to the deploy handler.

pkg/mcp/tools_deploy_test.go
```diff
@@ -79,3 +80,88 @@ func TestTool_Deploy_Args(t *testing.T) {
 		t.Fatal("executor was not invoked")
 	}
 }
+
+// TestParseDeployedURL verifies URL extraction from both local and remote deploy output.
+func TestParseDeployedURL(t *testing.T) {
+	tests := []struct {
+		name string
```
lkingland, line 87: Heads-up: #3666 is replacing the `server.readonly = false` direct-mutation pattern with `newTestPairWithReadonly(t, false)` (or by making the default `false`). When that lands, this test will want to use the new pattern. Easy rebase.

pkg/mcp/tools_deploy.go
```diff
@@ -31,10 +35,48 @@ func (s *Server) deployHandler(ctx context.Context, r *mcp.CallToolRequest, inpu
 	}
 	output = DeployOutput{
 		Message: string(out),
+		URL:     parseDeployedURL(out),
+	}
```
lkingland, line 43: Minor: `parseDeployedURL(out)` is called whether the deploy succeeded or not. If the executor returns an error, the message string may contain stderr that happens to contain `"Function Deployed at"` from a previous run cached somewhere — unlikely but possible. Worth gating: only parse on success.

lkingland (changes requested):

pkg/mcp/tools_deploy.go
Ankitsinghsisodya: Fixed. parseDeployedURL now returns (string, error) and propagates scanner.Err(). Also bumped the buffer to 1MB so large output does not hit the 64KB line limit.

pkg/mcp/tools_build.go
Ankitsinghsisodya: Fixed in both buildHandler and deployHandler. The error is appended to output.Message so the caller can see it.

pkg/mcp/tools_deploy_test.go
Ankitsinghsisodya: Got it, will update once #3666 lands.

pkg/mcp/tools_deploy.go
Ankitsinghsisodya: The early return on error already gates it. Split the call to a separate variable to make control flow clearer.

Ankitsinghsisodya:
@gauron @lkingland  sir can you please re run the failed test.

pkg/mcp/tools_build_test.go
```diff
@@ -67,3 +70,86 @@ func TestTool_Build_Args(t *testing.T) {
...
 }
+
+// TestTool_Build_StructuredOutput verifies that the Image field is populated
+// from the .func/built-image file written by the func CLI after a successful build.
+func TestTool_Build_StructuredOutput(t *testing.T) {
+	const wantImage = "ghcr.io/user/my-func:latest"
+
+	// Create a minimal function directory that fn.NewFunction can read.
+	root := t.TempDir()
+	if err := os.WriteFile(filepath.Join(root, "func.yaml"), []byte("name: my-func\nruntime: go\n"), 0644); err != nil {
```
lkingland, line 81: This couples the MCP implementation to that of the func internals.

Rather:  create a function in a temp directory (i.e. use the API)

pkg/mcp/tools_build_test.go
```diff
@@ -67,3 +70,86 @@ func TestTool_Build_Args(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	funcDir := filepath.Join(root, ".func")
+	if err := os.MkdirAll(funcDir, 0755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(funcDir, "built-image"), []byte(wantImage), 0644); err != nil {
```
lkingland, line 88: This also couples to the func internals.

Why not create a new image, then load it and ask it using the typesafe API?

pkg/mcp/tools_deploy.go
```diff
@@ -29,12 +33,57 @@ func (s *Server) deployHandler(ctx context.Context, r *mcp.CallToolRequest, inpu
...
 
+// parseDeployedURL extracts the deployed function URL from combined command
+// output. It handles two formats produced by the func CLI:
+//
+//   - Local deploy (written to stderr by the functions client):
+//     "✅ Function deployed/updated in namespace "ns" and exposed at URL: \n   <url>"
+//
+//   - Remote pipeline deploy (written to stdout by cmd/deploy.go):
+//     "Function Deployed at <url>"
+func parseDeployedURL(out []byte) (string, error) {
```
lkingland, line 60: parsing command output is brittle.  Prefer using structured output, or the API.

lkingland (commented):
I like that this gets the job done, but the implementation needing to parse output etc. is a good sign something deeper needs to change.

lkingland:
Opened #3770, which will supercede the need for parsing output or other gymnastics here.

If we ensure that every command has structured output, then the MCP can just use that directly; a clean passthrough.

PR #3682, opened 2026-05-08
feat: make cluster domain configurable in KEDA deployer

### LINKED ISSUE #3403: Clusterdomain is not configurable in Keda deployer
The external name service to the keda-add-ons-http-interceptor-proxy is currently hard coded with the clusterdomain set to `cluster.local`:

https://github.com/knative/func/blob/a48fb707067d670c19c1a7362e86dac9db57a235/pkg/keda/deployer.go#L223

`cluster.local` seems to be fine for most cases, but in some cases it doesn't. So we should find a way to make this configurable somehow.

Additional details: https://github.com/knative/func/pull/3386#discussion_r2728815144

Itx-Psycho0:
## Problem

The external name service to the keda-add-ons-http-interceptor-proxy was hardcoded with the cluster domain set to `cluster.local`. This works for most cases, but fails in clusters with custom domain configurations.

## Solution

Made the cluster domain configurable by adding a `ClusterDomain` field to the `DeploySpec` struct. The KEDA deployer now uses this field to build the interceptor proxy service URL, with `cluster.local` as the default fallback for backward compatibility.

## Changes

- Added `ClusterDomain` field to `DeploySpec` in `pkg/functions/function.go`
- Updated `interceptorBridgeService()` in `pkg/keda/deployer.go` to use the configurable cluster domain
- Added unit tests in `pkg/keda/deployer_test.go` to verify the functionality

## Usage

Users can now configure the cluster domain in their `func.yaml`:

```yaml
deploy:
  deployer: keda
  clusterDomain: "custom.domain"  # Optional, defaults to "cluster.local"
```

Testing
✅ All unit tests pass
✅ Linting checks pass
✅ Backward compatible (defaults to "cluster.local")
Fixes #3403

### DIFF
--- a/pkg/functions/function.go
+++ b/pkg/functions/function.go
@@ -227,6 +227,12 @@ type DeploySpec struct {
 	// Defaults to "knative" for backwards compatibility
 	Deployer string `yaml:"deployer,omitempty" jsonschema:"enum=knative,enum=raw,enum=keda"`
 
+	// ClusterDomain specifies the cluster domain used for building internal service URLs.
+	// Defaults to "cluster.local" if not specified.
+	// This is particularly useful for KEDA deployments where the interceptor proxy
+	// service URL needs to match the cluster's actual domain configuration.
+	ClusterDomain string `yaml:"clusterDomain,omitempty"`
+
 	Subscriptions []KnativeSubscription `yaml:"subscriptions,omitempty"`
 }
 
--- a/pkg/keda/deployer.go
+++ b/pkg/keda/deployer.go
@@ -204,6 +204,14 @@ func (d *Deployer) interceptorBridgeServiceName(f fn.Function) string {
 }
 
 func (d *Deployer) interceptorBridgeService(f fn.Function, namespace string, deployment *v1.Deployment) *corev1.Service {
+	// Use configured cluster domain, default to "cluster.local"
+	clusterDomain := f.Deploy.ClusterDomain
+	if clusterDomain == "" {
+		clusterDomain = "cluster.local"
+	}
+
+	externalName := fmt.Sprintf("keda-add-ons-http-interceptor-proxy.keda.svc.%s", clusterDomain)
+
 	return &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      d.interceptorBridgeServiceName(f),
@@ -220,7 +228,7 @@ func (d *Deployer) interceptorBridgeService(f fn.Function, namespace string, dep
 		},
 		Spec: corev1.ServiceSpec{
 			Type:         corev1.ServiceTypeExternalName,
-			ExternalName: "keda-add-ons-http-interceptor-proxy.keda.svc.cluster.local",
+			ExternalName: externalName,
 		},
 	}
 }
--- a/pkg/keda/deployer_test.go
+++ b/pkg/keda/deployer_test.go
@@ -1,0 +1,86 @@
+package keda
+
+import (
+	"testing"
+
+	v1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	fn "knative.dev/func/pkg/functions"
+)
+
+func TestDeployer_interceptorBridgeService(t *testing.T) {
+	tests := []struct {
+		name                 string
+		function             fn.Function
+		expectedExternalName string
+	}{
+		{
+			name: "uses default cluster.local when ClusterDomain is empty",
+			function: fn.Function{
+				Name: "test-func",
+				Deploy: fn.DeploySpec{
+					ClusterDomain: "",
+				},
+			},
+			expectedExternalName: "keda-add-ons-http-interceptor-proxy.keda.svc.cluster.local",
+		},
+		{
+			name: "uses custom cluster domain when specified",
+			function: fn.Function{
+				Name: "test-func",
+				Deploy: fn.DeploySpec{
+					ClusterDomain: "custom.domain",
+				},
+			},
+			expectedExternalName: "keda-add-ons-http-interceptor-proxy.keda.svc.custom.domain",
+		},
+		{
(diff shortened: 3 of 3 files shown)


Itx-Psycho0:
Hi maintainers,

I've implemented a fix for issue #3403 that makes the cluster domain configurable in the KEDA deployer.

Summary:
- Added a `ClusterDomain` field to `DeploySpec` that allows users to configure the cluster domain
- Updated the KEDA deployer to use this field when building the interceptor proxy service URL
- Defaults to "cluster.local" for backward compatibility
- Added comprehensive unit tests

The implementation is backward compatible and follows the existing patterns in the codebase. All tests pass and linting is clean.

I'd appreciate your review when you have a chance.

@lkingland @creydr @matejvasek

Thanks!

matejvasek:
Is this only keda specific?

matejvasek:
@creydr please look at this.

matejvasek:
Isn't cluster domain setting of a cluster, not a function?

gauron99:
> Is this only keda specific?

wouldnt this also touch on raw k8s deployer?

Itx-Psycho0:
You're absolutely right on all points. Let me address the concerns:

"Is this only KEDA specific?"
No, it's not. The cluster domain affects any deployer that builds internal service URLs. The raw K8s deployer likely has similar hardcoded references to cluster.local.

"Isn't cluster domain setting of a cluster, not a function?"
You're correct, the cluster domain is a cluster-level configuration, not a function-level one. Putting it in DeploySpec (per-function) is architecturally wrong.

Proposed Better Solution - **Auto-detect cluster domain** from the cluster's DNS configuration
**Make it a global/cluster-level config** (not per-function)
**Apply to all deployers** (KEDA, raw K8s, Knative if needed)

Implementation approach:
Add cluster domain detection in the deployer initialization
Fall back to "cluster.local" if detection fails
Allow override via global config (not per-function)
Apply consistently across all deployers

Should I rework the PR with this approach? I can:
Move cluster domain to a global config
Add auto-detection logic
Apply it to both KEDA and raw K8s deployers
Add tests for all deployers

Let me know if this direction makes sense!
@gauron99 @matejvasek

lkingland:
This one is... nontrivial.

However, if you're interested in tackling it (with @creydr's help I hope, *nudge nudge*) your other PRs have been quite nice, so I am sure we can figure out a good solution.

I've added comments to the parent issue #3403

lkingland:
Here, I had AI do some research and break it down into actionable steps.  Beware potential hallucinations may follow:

Shared contract (decide first — blocks everything else)

  - Pick the ConfigMap coordinates. Propose func-system/cluster-config; agree namespace + name as a stable client/server contract.
  - Define the schema. At minimum clusterDomain. Reserve keys for defaultRegistry, installedComponents, supportedRuntimes, defaultNamespace. Document forward-compat
  rules (unknown keys ignored, never repurpose a key).
  - Decide kubeconfig-context as the cluster identity key on the CLI side (used for cache directory and user-override scoping).

  Operator (depends on contract)

  - Add cluster-domain detection that parses /etc/resolv.conf search entries in the operator pod (port of Knative's GetClusterDomainName logic).
  - Reconcile the cluster-config ConfigMap on operator startup, creating/owning it via metadata.ownerReferences to the operator Deployment so it cleans up on
  uninstall.
  - Add a periodic reconciliation timer (e.g., 1h) so post-install kubelet reconfigs eventually propagate.
  - Emit a ConfigMap event on change so consumers can watch instead of poll if they want to.
  - Document the ConfigMap as the operator's public API surface in the operator README.

  CLI core helper

  - Introduce a single func ClusterDomain(...) helper that every deployer goes through. No deployer should reference cluster.local directly.
  - Implement the precedence chain inside the helper:
    a. --cluster-domain flag value (if set on the command)
    b. ~/.config/func user override for the active kubeconfig context
    c. func-system/cluster-config ConfigMap read from the cluster
    d. cluster.local
  - Add CLI cache at ~/.cache/func/<context>/cluster.json with a short TTL (default ~1h, env override). Invalidate when kubeconfig context changes.
  - Add --cluster-domain flag to func deploy (and any other commands that build cluster-internal URLs) as the explicit override.
  - Improve the fallback error UX: when a deploy fails on the cluster.local default with no operator and no user override, the error message should say "assumed
  cluster.local; set X to override" — not silent.

  Global config (~/.config/func)

  - Add a clusters: section keyed by kubeconfig context name, each with a clusterDomain field (and reserved space for the same future keys as the cluster ConfigMap).
  - Define a single source of truth for config schema so the in-cluster ConfigMap and the user-local file share field names.
  - Add a func config cluster set <context> clusterDomain <value> subcommand (or similar) so users don't have to hand-edit YAML.

  Deployers (consume — depends on CLI helper)

  - pkg/keda/deployer.go:231 — replace hardcoded ExternalName with ClusterDomain(...) lookup.
  - pkg/k8s/deployer.go:322 — replace hardcoded Trigger filter Host with same.
  - pkg/pipelines/tekton/templates.go — audit and replace any cluster.local literals (verify which need it; some Tekton templates may run inside the cluster and could
   use env-var injection instead).
  - Grep for any remaining cluster.local literals in pkg/ and cmd/ and either route through the helper or annotate why they're intentional (e.g., test fixtures).

  Docs

  - New reference page: "Cluster Configuration" — explains the operator ConfigMap, the user-local override, the flag, and the precedence order.
  - Troubleshooting entry for the symptom (KEDA interceptor unreachable / Trigger filter 404s) pointing at the override.
  - Note in operator README that the cluster-config ConfigMap is part of the operator's contract.

  Suggested PR sequence

  1. Contract + helper skeleton (no behavior change yet — helper just returns cluster.local)
  2. Deployers wired through the helper (still cluster.local, but centralized)
  3. Global config schema + user override path
  4. Operator side (detection + ConfigMap publish)
  5. CLI ConfigMap reader + cache
  6. Docs + troubleshooting

PR #3685, opened 2026-05-09
feat(mcp): add secret and ConfigMap support to config_envs_add tool

Ankitsinghsisodya:
## What

Add secret-backed and ConfigMap-backed environment variable support to the `config_envs_add` MCP tool. Previously, `ConfigEnvsAddInput` only exposed `Name` and `Value` fields, leaving agents unable to configure Kubernetes-sourced env vars through MCP.

## Changes

- **`pkg/mcp/tools_config_envs.go`**: Add four optional fields to `ConfigEnvsAddInput`: `SecretName`, `SecretKey`, `ConfigMapName`, and `ConfigMapKey`. Update `Args()` to construct the appropriate `{{ secret:… }}` / `{{ configMap:… }}` value template automatically. Explicit `Value` takes precedence when provided alongside source fields. Update the tool description to document all sourcing modes.

- **`pkg/mcp/tools_config_envs_test.go`**: Add five new test cases covering single-key secret, all-keys secret, single-key ConfigMap, all-keys ConfigMap, and precedence of explicit `Value` over `SecretName`.

## Sourcing modes

| Mode | Fields required | Generated value |
|---|---|---|
| Literal value | `name`, `value` | value passed through |
| Secret (one key) | `name`, `secretName`, `secretKey` | `{{ secret:name:key }}` |
| Secret (all keys) | `secretName` | `{{ secret:name }}` |
| ConfigMap (one key) | `name`, `configMapName`, `configMapKey` | `{{ configMap:name:key }}` |
| ConfigMap (all keys) | `configMapName` | `{{ configMap:name }}` |

## Testing

All 11 tests in `pkg/mcp` pass locally.

### DIFF
--- a/pkg/mcp/tools_config_envs.go
+++ b/pkg/mcp/tools_config_envs.go
@@ -82,7 +82,7 @@ The explicit Value field takes precedence if provided alongside source fields.`,
 type ConfigEnvsAddInput struct {
 	Path          string  `json:"path" jsonschema:"required,Path to the function project directory"`
 	Name          *string `json:"name,omitempty" jsonschema:"Name of the environment variable"`
-	Value         *string `json:"value,omitempty" jsonschema:"Literal value for the environment variable"`
+	Value         *string `json:"value,omitempty" jsonschema:"Literal value or template expression (e.g. '{{ env:MY_VAR }}') for the environment variable"`
 	SecretName    *string `json:"secretName,omitempty" jsonschema:"Name of the Kubernetes Secret to source the value from"`
 	SecretKey     *string `json:"secretKey,omitempty" jsonschema:"Key within the Secret; omit to import all keys as env vars"`
 	ConfigMapName *string `json:"configMapName,omitempty" jsonschema:"Name of the Kubernetes ConfigMap to source the value from"`
@@ -98,6 +98,21 @@ func (i ConfigEnvsAddInput) validate() error {
 	}
 	if i.ConfigMapKey != nil && i.ConfigMapName == nil {
 		return fmt.Errorf("configMapKey requires configMapName to be set")
+	}
+	if i.SecretName != nil && i.ConfigMapName != nil {
+		return fmt.Errorf("secretName and configMapName are mutually exclusive; provide only one source")
+	}
+	// All-keys import (no secretKey/configMapKey) is incompatible with --name: the
+	// underlying env validation only allows whole-secret/configMap templates when name is nil.
+	// The guard is skipped when an explicit Value is present because Value takes precedence
+	// and SecretName/ConfigMapName is effectively ignored in that path.
+	if i.Value == nil {
+		if i.SecretName != nil && i.SecretKey == nil && i.Name != nil {
+			return fmt.Errorf("name must not be set when importing all keys from a Secret (omit secretKey to import all keys)")
+		}
+		if i.ConfigMapName != nil && i.ConfigMapKey == nil && i.Name != nil {
+			return fmt.Errorf("name must not be set when importing all keys from a ConfigMap (omit configMapKey to import all keys)")
+		}
 	}
 	if i.SecretName != nil && !validResourceName.MatchString(*i.SecretName) {
 		return fmt.Errorf("invalid secretName %q: only word characters, hyphens and apostrophes are allowed", *i.SecretName)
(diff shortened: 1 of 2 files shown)


pkg/mcp/tools_config_envs.go
```diff
@@ -3,10 +3,18 @@ package mcp
...
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
 
+// validResourceName matches Kubernetes resource names as accepted by the env template syntax.
+// Mirrors the name capture group used in pkg/functions/function.go validation regexes.
+var validResourceName = regexp.MustCompile(`^((?:\w|['-]\w)+)$`)
+
+// validKeyName matches Secret/ConfigMap key names as accepted by the env template syntax.
+var validKeyName = regexp.MustCompile(`^[-._a-zA-Z0-9]+$`)
```
lkingland: These regexes mirror func's *template parser* validation, but they don't match real **K8s DNS-1123** resource name rules. For example, `validResourceName` accepts apostrophes (`'`) — those aren't valid in any K8s Secret/ConfigMap name, so the cluster will reject the manifest later.

Two options:
1. **Tighten** to DNS-1123: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$` (max 253 chars, lowercase, no underscores/apostrophes).
2. **Drop the regexes entirely** and rely on func's existing validation when the `--value` is parsed — the MCP layer becomes a thinner pass-through, with one less place to maintain validation logic.

Leaning toward (2) personally — the duplication here is brittle. If func updates its name rules, this regex won't follow, and any drift causes either false rejections or false acceptances. Pass-through with good error reporting from the CLI is probably enough.

pkg/mcp/tools_config_envs.go
```diff
@@ -60,16 +80,80 @@ var configEnvsAddTool = &mcp.Tool{
...
+	// underlying env validation only allows whole-secret/configMap templates when name is nil.
+	// The guard is skipped when an explicit Value is present because Value takes precedence
+	// and SecretName/ConfigMapName is effectively ignored in that path.
+	if i.Value == nil {
+		if i.SecretName != nil && i.SecretKey == nil && i.Name != nil {
+			return fmt.Errorf("name must not be set when importing all keys from a Secret (omit secretKey to import all keys)")
+		}
+		if i.ConfigMapName != nil && i.ConfigMapKey == nil && i.Name != nil {
+			return fmt.Errorf("name must not be set when importing all keys from a ConfigMap (omit configMapKey to import all keys)")
+		}
```
lkingland: Silent precedence is a footgun for the agent. If an LLM client passes both `Value` and `SecretName` (e.g., copy-pasted from a config and forgot to clear one field), the SecretName is dropped without a peep. Consider rejecting the combination explicitly:

```go
if i.Value != nil && (i.SecretName != nil || i.ConfigMapName != nil) {
    return fmt.Errorf("value is mutually exclusive with secretName/configMapName; provide one or the other")
}
```

Alternatively, at least make the precedence rule prominent in the tool description (it's currently a one-liner at the end of a long block).

pkg/mcp/tools_config_envs_test.go
```diff
@@ -72,6 +72,500 @@ func TestTool_ConfigEnvsAdd(t *testing.T) {
...
+		if len(args) < 2 || args[0] != "envs" || args[1] != "add" {
+			t.Fatalf("expected positional args ['envs','add'], got %v", args[:min(2, len(args))])
+		}
+		argsMap := argsToMap(args[2:])
+		wantValue := "{{ configMap:my-config:DB_HOST }}"
+		if got, ok := argsMap["--value"]; !ok || got != wantValue {
+			t.Fatalf("expected --value=%q, got %q", wantValue, got)
+		}
+		if got, ok := argsMap["--name"]; !ok || got != "DATABASE_HOST" {
+			t.Fatalf("expected --name='DATABASE_HOST', got %q", got)
```
lkingland, line 184: Heads-up: #3694 (path validation) rejects `"."` and other relative paths. If that lands first, all these test cases will need `t.TempDir()` instead of `"."`. Worth coordinating with @Shrey327 on merge order.

lkingland (changes requested):
Thanks for the PR.  A couple small changes.  Mostly around keeping the MCP server with as little logic as possible (especially regex).

pkg/mcp/tools_config_envs_test.go
Ankitsinghsisodya: Noted, I will coordinate with @Shrey327 on merge order. If #3694 lands first, will update these tests to use `t.TempDir()`.

lkingland (approved):

PR #3687, opened 2026-05-09
refactor: consolidate reference tracking with Tracker in deployer

### LINKED ISSUE #3678: Refactor: consolidate reference tracking in pkg/k8s/deployer.go
## Background

Several functions in `pkg/k8s/deployer.go` track references to cluster
resources (Secrets, ConfigMaps, PVCs) so that `CheckResourcesArePresent`
can fail fast if anything referenced by a function isn't on the
cluster. The current shape passes three independent sets as
out-parameters via pointer:

```go
func ProcessEnvs(envs []fn.Env, referencedSecrets, referencedConfigMaps *sets.Set[string]) (...)
func ProcessVolumes(volumes []fn.Volume, referencedSecrets, referencedConfigMaps, referencedPVCs *sets.Set[string]) (...)
func createEnvFromSource(value string, referencedSecrets, referencedConfigMaps *sets.Set[string]) (...)
func createEnvVarSource(slices []string, referencedSecrets, referencedConfigMaps *sets.Set[string]) (...)
func (d *Deployer) generateDeployment(f fn.Function, namespace string, daprInstalled bool, referencedSecrets, referencedConfigMaps, referencedPVCs *sets.Set[string]) (...)
func CheckResourcesArePresent(ctx context.Context, namespace string, referencedSecrets, referencedConfigMaps, referencedPVCs *sets.Set[string], referencedServiceAccount, imagePullSecret string) error
```

Most recently propagated to `generateDeployment` in #3671 to fix a
silent-validation gap on the first-deploy path. That PR is the right
fix for the bug; this issue captures the design cleanup that the PR
deliberately did not absorb.

## Why this is worth cleaning up

Three layers of smell sit on top of each other:

**1. Pointer-to-map is doubly indirect.**
`k8s.io/apimachinery/pkg/util/sets.Set[T]` is `map[T]Empty`. Maps in
Go are reference types — passing a `Set` by value already lets the
callee mutate the same backing storage. The `*` adds a level of
indirection that buys nothing the callers actually use: nobody
reassigns the set, nobody distinguishes a nil pointer from an empty
set, and the only operations the callees invoke are `Insert` and
`Has`. Idiomatic Go here is `sets.Set[T]` by value.

**2. The triple is structurally coupled but passed as three loose
parameters.** Every signature that touches reference tracking repeats
`referencedSecrets, referencedConfigMaps, referencedPVCs`. That's a
cohesive triple — it deserves a name:

```go
type References struct {
    Secrets    sets.Set[string]
    ConfigMaps sets.Set[string]
    PVCs       sets.Set[string]
}
```

Once the triple has a type, every call site collapses from three
arguments to one.

**3. Accumulator-by-out-param is C-style in a language with
multi-return.** Two cleaner shapes are available:

```go
// (a) multi-return — pure-function flavor:
func ProcessEnvs(envs []fn.Env) ([]corev1.EnvVar, []corev1.EnvFromSource, References, error)

// (b) receiver-as-tracker — OO flavor:
type Tracker struct { References }
func (t *Tracker) ProcessEnvs(envs []fn.Env) ([]corev1.EnvVar, []corev1.EnvFromSource, error)
```

Either reads better than threading three out-pointers through the call
chain. (b) also makes it harder to forget to wire one of the sets
through, which is exactly the bug class that #3671 fixed (the local
sets in `generateDeployment` were silently invisible to the caller).

## Suggested approach

Pick one shape — (a) or (b) above — and apply it consistently.
Either change is mechanical:

- Define `References` (and `Tracker`, if going with shape (b)) in a
  shared spot, likely near the top of `pkg/k8s/deployer.go`.
- Update the affected functions to use it:
  - `ProcessEnvs`
  - `ProcessVolumes`
  - `createEnvFromSource`
  - `createEnvVarSource`
  - `generateDeployment`
  - `CheckResourcesArePresent`
- Update call sites in `Deploy()` — both the create branch and the
  update branch in `pkg/k8s/deployer.go`. The keda deployer inherits
  via delegation, no separate work needed there.
- Update the unit tests in `pkg/k8s/deployer_test.go` for the new
  signatures. Integration tests should not need behavior changes.
- Confirm no behavior changes — this is purely a shape refactor.

## Out of scope

- The validation logic itself.
- Any change to the bug fix in #3671.
- Reference-tracking work in other deployers (knative, keda) — they
  use the k8s functions but don't define their own. Once the k8s side
  is done, the rest follows.

## Why this wasn't done in #3671

#3671 is a focused bug fix from a contributor. Asking them to absorb
a refactor of a pre-existing pattern would have widened scope unrelated
to the bug and gated a real user-facing fix on cleanup work. This issue
captures the cleanup as separable, picker-uppable work.

## Good first issue?

Reasonable candidate — the change is mechanical, the affected
functions are localized to one file (plus the test file), and the
desired end state is clearly specified. The judgment call between
shapes (a) and (b) is the only design decision; either is acceptable.

Ankitsinghsisodya:
## What

Fixes #3678
Fixes the three code smells called out in #3678 by introducing a `Tracker`
struct (shape b from the issue) that owns the three reference sets and
accumulates them via receiver methods.

## Changes

**`pkg/k8s/deployer.go`**
- Add `References` struct holding `Secrets`, `ConfigMaps`, `PVCs` as
  `sets.Set[string]` values (not pointers — maps are already reference types).
- Add `Tracker` struct embedding `References` and `NewTracker()` constructor
  that initialises all three sets.
- Convert `ProcessEnvs`, `ProcessVolumes`, `createEnvFromSource`,
  `createEnvVarSource` from free functions with `*sets.Set[string]`
  out-parameters to `*Tracker` methods that write directly into the receiver.
- Update `generateDeployment` to accept `*Tracker` instead of three pointer
  parameters.
- Update `CheckResourcesArePresent` to accept `References` by value instead of
  three `*sets.Set[string]` parameters.
- Update both call sites in `Deploy()` (create and update paths).

**`pkg/knative/deployer.go`**
- Update `generateNewService` signature to accept `*k8s.Tracker`.
- Update both call sites in `Deploy()` (create and update paths).
- Remove now-unused `sets` import.

**`pkg/k8s/deployer_test.go`** / **`pkg/knative/deployer_test.go`**
- Replace three-variable `sets.New[string]()` + pointer passing with
  `NewTracker()` in all affected tests.
- Update regression test assertions to use `tracker.Secrets` /
  `tracker.ConfigMaps`.
- Remove unused `sets` imports.

## Why shape (b) — Tracker receiver

- **Structurally prevents the #3671 bug class.** With out-parameters a caller
  can silently create a disconnected set and pass it alongside the one that
  gets validated; with a single Tracker receiver there is nothing to
  disconnect.
- **Eliminates the pointer-to-map double-indirection.** `sets.Set[string]` is
  `map[string]Empty`; passing it by value inside `References` already lets
  callees mutate the same backing store without the extra `*`.
- **One argument instead of three at every call site.**

## Behavior

No behavior changes. This is a pure shape refactor.

## Test plan

- [ ] `go test ./pkg/k8s/ ./pkg/knative/` — both pass
- [ ] `make check` — 0 lint issues
- [ ] `make test` — all pre-existing unit tests pass

### DIFF
--- a/pkg/k8s/deployer.go
+++ b/pkg/k8s/deployer.go
@@ -69,6 +69,34 @@ func WithDeployerDecorator(decorator deployer.DeployDecorator) DeployerOpt {
 	}
 }
 
+// References holds the names of cluster resources referenced by a function.
+// It is populated during deployment manifest generation and consumed by
+// CheckResourcesArePresent to validate that every referenced resource exists.
+type References struct {
+	Secrets    sets.Set[string]
+	ConfigMaps sets.Set[string]
+	PVCs       sets.Set[string]
+}
+
+// Tracker accumulates References across a chain of processing calls.
+// Using a receiver ensures all processing steps write into the same sets,
+// making it structurally impossible for any caller to silently bypass
+// validation by threading a disconnected set through the call chain.
+type Tracker struct {
+	References
+}
+
+// NewTracker returns a Tracker with all sets initialized and ready for use.
+func NewTracker() *Tracker {
+	return &Tracker{
+		References: References{
+			Secrets:    sets.New[string](),
+			ConfigMaps: sets.New[string](),
+			PVCs:       sets.New[string](),
+		},
+	}
+}
+
 func onClusterFix(f fn.Function) fn.Function {
 	// This only exists because of a bootstrapping problem with On-Cluster
 	// builds:  It appears that, when sending a function to be built on-cluster
@@ -147,16 +175,14 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
 	var status fn.Status
 	if err == nil {
 		// Update the existing function
-		referencedSecrets := sets.New[string]()
-		referencedConfigMaps := sets.New[string]()
-		referencedPVCs := sets.New[string]()
+		tracker := NewTracker()
 
-		deployment, err := d.generateDeployment(f, namespace, daprInstalled, &referencedSecrets, &referencedConfigMaps, &referencedPVCs)
+		deployment, err := d.generateDeployment(f, namespace, daprInstalled, tracker)
 		if err != nil {
 			return fn.DeploymentResult{}, fmt.Errorf("failed to generate deployment resources: %w", err)
 		}
 
-		if err = CheckResourcesArePresent(ctx, namespace, &referencedSecrets, &referencedConfigMaps, &referencedPVCs, f.Deploy.ServiceAccountName, f.Deploy.ImagePullSecret); err != nil {
+		if err = CheckResourcesArePresent(ctx, namespace, tracker.References, f.Deploy.ServiceAccountName, f.Deploy.ImagePullSecret); err != nil {
 			return fn.DeploymentResult{}, fmt.Errorf("failed to validate referenced resources: %w", err)
 		}
 
@@ -196,16 +222,14 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
 			return fn.DeploymentResult{}, fmt.Errorf("failed to check for existing deployment: %w", err)
 		}
 
-		referencedSecrets := sets.New[string]()
-		referencedConfigMaps := sets.New[string]()
-		referencedPVCs := sets.New[string]()
+		tracker := NewTracker()
 
-		deployment, err := d.generateDeployment(f, namespace, daprInstalled, &referencedSecrets, &referencedConfigMaps, &referencedPVCs)
+		deployment, err := d.generateDeployment(f, namespace, daprInstalled, tracker)
 		if err != nil {
 			return fn.DeploymentResult{}, fmt.Errorf("failed to generate deployment resources: %w", err)
 		}
 
-		if err = CheckResourcesArePresent(ctx, namespace, &referencedSecrets, &referencedConfigMaps, &referencedPVCs, f.Deploy.ServiceAccountName, f.Deploy.ImagePullSecret); err != nil {
+		if err = CheckResourcesArePresent(ctx, namespace, tracker.References, f.Deploy.ServiceAccountName, f.Deploy.ImagePullSecret); err != nil {
 			return fn.DeploymentResult{}, fmt.Errorf("failed to validate referenced resources: %w", err)
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/k8s/deployer.go
```diff
@@ -69,6 +69,92 @@ func WithDeployerDecorator(decorator deployer.DeployDecorator) DeployerOpt {
...
+			ConfigMaps: sets.New[string](),
+			PVCs:       sets.New[string](),
+		},
+	}
+}
+
+// ensureInit initializes any nil sets in the embedded References so that
+// methods on a zero-value or partially-constructed Tracker never panic on a
+// nil-map Insert.
+func (t *Tracker) ensureInit() {
```
lkingland: Why both NewTracker() (initializes the sets) and ensureInit() (lazy-inits them)? They serve the same purpose. Just the constructor should be sufficient.

pkg/k8s/deployer.go
```diff
@@ -69,6 +69,92 @@ func WithDeployerDecorator(decorator deployer.DeployDecorator) DeployerOpt {
...
+		t.PVCs = sets.New[string]()
+	}
+}
+
+// ProcessEnvs is a package-level wrapper kept for backwards compatibility.
+//
+// Deprecated: Create a Tracker via NewTracker() and call its ProcessEnvs
+// method instead. The out-parameter sets are populated from the tracker's
+// accumulated references after the call.
+func ProcessEnvs(envs []fn.Env, referencedSecrets, referencedConfigMaps *sets.Set[string]) ([]corev1.EnvVar, []corev1.EnvFromSource, error) {
```
lkingland: The free-function ProcessEnvs and ProcessVolumes wrappers state "for backwards compatibility", but no one uses them except for known packages.  The compiler will catch that.  Unnecessary at this stage.

pkg/k8s/deployer.go
```diff
@@ -69,6 +69,92 @@ func WithDeployerDecorator(decorator deployer.DeployDecorator) DeployerOpt {
 	}
 }
 
+// References holds the names of cluster resources referenced by a function.
+// It is populated during deployment manifest generation and consumed by
+// CheckResourcesArePresent to validate that every referenced resource exists.
+type References struct {
```
lkingland, line 75: I am not understanding why Tracker needs to be a pointer, and itself hold a strcuture which hold sets (which are themselves pointers; they're maps).

Just a single flat, non-pointer struct `References` with the sets would be sufficient I believe?

lkingland (changes requested):

PR #3690, opened 2026-05-09
mcp: set HasPrompts to false since no prompts are registered

### LINKED ISSUE #3689: mcp: HasPrompts capability advertised but no prompts registered
## Bug

The MCP server declares `HasPrompts: true` in its capabilities (`pkg/mcp/mcp.go`, line 84), but no prompts are registered anywhere in the codebase:

```go
&mcp.ServerOptions{
    Instructions:       instructions(s.readonly),
    HasPrompts:         true,   // no prompts are registered
    HasResources:       true,
    HasTools:           true,
    InitializedHandler: ...,
}
```

Per the [MCP specification](https://modelcontextprotocol.io/specification/2025-03-26/server/prompts), when a server declares prompts capability, clients may call `prompts/list` and expect results.

## Impact

- MCP clients that enumerate prompts will get an empty list, which is misleading
- Some clients may present empty prompt UI or make unnecessary protocol round-trips
- Incorrectly signals to MCP clients that the server supports a feature it does not

## Fix

Set `HasPrompts` to `false` until prompts are actually implemented:

```diff
 &mcp.ServerOptions{
     Instructions:       instructions(s.readonly),
-    HasPrompts:         true,
+    HasPrompts:         false,
     HasResources:       true,
     HasTools:           true,
     InitializedHandler: ...,
 }
```

Elvand-Lie:
The MCP server declares `HasPrompts: true` in its capabilities but no prompts are registered. Per the MCP specification, clients may call `prompts/list` when this capability is advertised and expect results.

Set `HasPrompts` to `false` to accurately reflect the server capabilities.

# Changes

- Set `HasPrompts` to `false` in MCP server options since no prompts are registered

Fixes #3689

**Release Note**

```release-note
NONE

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -81,7 +81,7 @@ func New(options ...Option) *Server {
 			Version: version},
 		&mcp.ServerOptions{
 			Instructions:       instructions(s.readonly),
-			HasPrompts:         true,
+			HasPrompts:         false,
 			HasResources:       true,
 			HasTools:           true,
 			InitializedHandler: func(ctx context.Context, _ *mcp.InitializedRequest) { s.OnInit(ctx) },


matejvasek:
PTAL @lkingland

lkingland:
Duplicate of https://github.com/knative/func/issues/3734. PR https://github.com/knative/func/pull/3737 there adds an actual prompt instead of flipping the flag — closing this in favor of that path.

PR #3692, opened 2026-05-10
fix: use net.Dial in socat for dual-stack support (Happy Eyeballs)

matejvasek:
# Changes

ResolveTCPAddr + DialTCP resolves the hostname to a single address and dials it directly.  When DNS returns both A and AAAA records (dual-stack Docker network), the IPv4 address is typically returned first.  On IPv6-only pods this fails with "network is unreachable" because the pod has no IPv4 connectivity.

Switch to net.Dial which implements Happy Eyeballs (RFC 6555): it resolves all addresses, tries them concurrently with IPv6 preferred, and uses whichever connects first.  The tcp4/tcp6 explicit cases are also handled correctly by net.Dial.

/kind bug

```release-note
fix: in cluster dialer (socat) uses Happy Eyeballs
```

### DIFF
--- a/cmd/func-util/socat.go
+++ b/cmd/func-util/socat.go
@@ -65,15 +65,9 @@ func createConnection(address string, stdio connection) (connection, error) {
 	switch typ {
 	case "tcp", "tcp4", "tcp6":
 		_, _ = fmt.Fprintln(os.Stderr, "opening connection")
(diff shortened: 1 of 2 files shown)


matejvasek:
PTAL @lkingland @gauron99

gauron99 (approved):
/lgtm

PR #3693, opened 2026-05-10
fix: improve IPv6 support in local Kind cluster setup

matejvasek:
## Summary

Fixes two IPv6-related issues in `hack/cluster.sh` that break the local Kind development cluster on IPv6-only and dual-stack Docker networks:

- **Contour IPv6 bindings:** Contour defaults its xDS server, Envoy listeners, and stats endpoint to `0.0.0.0` (IPv4-only). On IPv6-only clusters the control plane cannot reach Envoy and ingress traffic is blocked. The Contour deployment is now patched at install time to rebind `--xds-address` and `--stats-address` to `::` alongside the existing HTTP/HTTPS address overrides.
Note: `::` listens on both IPv4 and IPv6 on Linux by default. However if machines completely disables IPv6  or deliberately sets `net.ipv6.bindv6only = 1`  the it might not listen on IPv4, but people doing that had it coming anyway.

- **CoreDNS DNS forwarding via socat:** Docker's embedded DNS proxy only listens on IPv4 (`moby/moby#41651`), and Kind replaces the node's nameserver with the IPv4 bridge gateway (e.g. `172.18.0.1`). On IPv6-only clusters CoreDNS pods cannot reach this address, breaking all DNS resolution. A socat UDP/TCP proxy is now installed on the Kind node when IPv4-only nameservers are detected, forwarding DNS from the node's IPv6 address to the IPv4 nameserver. The CoreDNS Corefile is patched to use this proxy instead of `/etc/resolv.conf`. See also https://github.com/kubernetes-sigs/kind/issues/4152

## Test plan

- [ ] Run `hack/cluster.sh` on a Docker host with an IPv6-only Kind cluster and verify DNS resolution works inside pods
- [ ] Run `hack/cluster.sh` on a dual-stack Kind cluster and verify both IPv4 and IPv6 DNS resolution work
- [ ] Run `hack/cluster.sh` on a standard IPv4-only Kind cluster and verify no regressions (socat proxy should be skipped, Contour bindings are additive)
- [ ] Verify Contour/Envoy ingress serves traffic on IPv6-only clusters

### DIFF
--- a/hack/cluster.sh
+++ b/hack/cluster.sh
@@ -239,8 +239,9 @@ networking() {
 
   echo "Installing a configured Contour."
   curl -sSL "https://github.com/knative/net-contour/releases/download/knative-${contour_version}/contour.yaml" \
+    | $YQ '(select(.kind == "Deployment" and .metadata.name == "contour").spec.template.spec.containers[0].args[] | select(. == "--xds-address=0.0.0.0")) = "--xds-address=::"' \
     | $YQ '(select(.kind == "Deployment" and .metadata.name == "contour").spec.template.spec.containers[0].args)
-          += ["--envoy-service-http-address=::", "--envoy-service-https-address=::"]' \
+          += ["--envoy-service-http-address=::", "--envoy-service-https-address=::", "--stats-address=::"]' \
     | $KUBECTL apply -f -
 
   sleep 5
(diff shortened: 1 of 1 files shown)


matejvasek:
PTAL @gauron99 @lkingland

PR #3694, opened 2026-05-10
mcp: reject relative paths in tool handlers

Shrey327:
## Description
Adds `validatePath()` to the MCP server that rejects non-absolute path arguments before shelling out to the func CLI. Since `cmd.Dir` is never set in the executor, relative paths silently resolved against the MCP server's working directory instead of the user's project turning a wrong directory failure into a silent misbehavior. This change turns that into an actionable error, matching the guidance already documented in instructions.md.All existing tests updated to use `t.TempDir()` and a `TestValidatePath` unit test added.

### DIFF
--- a/pkg/mcp/tools.go
+++ b/pkg/mcp/tools.go
@@ -2,6 +2,7 @@ package mcp
 
 import (
 	"fmt"
+	"path/filepath"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
@@ -39,4 +40,13 @@ func appendBoolFlag(args []string, flag string, value *bool) []string {
 // Useful for setting optional annotation fields.
 func ptr[T any](v T) *T {
 	return &v
+}
+
+// validatePath rejects relative paths. Since cmd.Dir is not set in the executor,
+// relative paths resolve against the MCP server's CWD, not the user's project root.
+func validatePath(path string) error {
+	if !filepath.IsAbs(path) {
+		return fmt.Errorf("path must be absolute, got %q", path)
+	}
+	return nil
 }
--- a/pkg/mcp/tools.go
+++ b/pkg/mcp/tools.go
@@ -46,7 +46,7 @@ func ptr[T any](v T) *T {
 // relative paths resolve against the MCP server's CWD, not the user's project root.
 func validatePath(path string) error {
 	if !filepath.IsAbs(path) {
-		return fmt.Errorf("path must be absolute, got %q", path)
+		return fmt.Errorf("path must be absolute (the MCP server's working directory is not the user's project root), got %q", path)
 	}
 	return nil
 }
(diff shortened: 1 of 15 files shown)


pkg/mcp/tools.go
```diff
@@ -40,3 +41,12 @@ func appendBoolFlag(args []string, flag string, value *bool) []string {
 func ptr[T any](v T) *T {
 	return &v
 }
+
+// validatePath rejects relative paths. Since cmd.Dir is not set in the executor,
+// relative paths resolve against the MCP server's CWD, not the user's project root.
+func validatePath(path string) error {
+	if !filepath.IsAbs(path) {
+		return fmt.Errorf("path must be absolute, got %q", path)
+	}
```
lkingland, line 50: Consider a more agent-friendly error message — the client is usually an LLM. Something like:

```go
return fmt.Errorf("path must be absolute (the MCP server's working directory is not the user's project root), got %q", path)
```

The parenthetical hints at the *why*, which helps the agent fix its call instead of just retrying with a different relative path.

pkg/mcp/tools_test.go
```diff
@@ -62,6 +62,19 @@ func validateBoolFlags(t *testing.T, args []string, boolFlags map[string]string)
...
+func TestValidatePath(t *testing.T) {
+	if err := validatePath("/absolute/path"); err != nil {
+		t.Fatalf("expected no error for absolute path, got: %v", err)
+	}
+	if err := validatePath("relative/path"); err == nil {
+		t.Fatal("expected error for relative path, got nil")
+	}
+	if err := validatePath("."); err == nil {
+		t.Fatal("expected error for '.', got nil")
+	}
```
lkingland, line 118: Worth adding handler-level rejection tests so the wiring is covered, not just the helper. Something like:

```go
func TestTool_Deploy_RejectsRelativePath(t *testing.T) {
    client, _, err := newTestPair(t)
    if err != nil { t.Fatal(err) }
    result, err := client.CallTool(t.Context(), &mcp.CallToolParams{
        Name:      "deploy",
        Arguments: map[string]any{"path": "."},
    })
    if err != nil { t.Fatal(err) }
    if !result.IsError { t.Fatal("expected error for relative path") }
}
```

One per handler is verbose but mechanical. Alternatively, a table-driven test parameterized over tool names would cover all of them in one place.

lkingland (commented):
Thanks for the PR.  Just two small suggestions.

pkg/mcp/tools.go
Shrey327: Thanks for your review i have made the changes

pkg/mcp/tools_test.go
Shrey327: made the changes , i used the table-driven approach

lkingland (approved):

PR #3714, opened 2026-05-12
git/github: default GitHub webhooks to TLS verification

### LINKED ISSUE #3713: git/github: default repository webhook insecure_ssl to TLS verification ("0")
## Summary
When creating a GitHub repository webhook (`Client.CreateWebHook` in `pkg/git/github/github.go`), we currently set `HookConfig.InsecureSSL` to `"1"`, which tells GitHub **not** to verify TLS when delivering webhooks to an **HTTPS** payload URL. There is an inline TODO noting this should default to secure behavior.

## Problem
- GitHub’s `insecure_ssl` field: `"0"` = verify TLS to the webhook URL, `"1"` = skip verification ([Create a repository webhook](https://docs.github.com/en/rest/repos/webhooks#create-a-repository-webhook)).
- Defaulting to `"1"` is unsafe for normal HTTPS endpoints with valid certificates.

## Proposed change
- Set `InsecureSSL` to `"0"` so GitHub verifies TLS by default for HTTPS webhook targets.
- Keep behavior documented in code (and optionally add a small unit test on the hook config helper).

## Area
`pkg/git/github`

Vi-shub:
- Default GitHub repository webhooks created by `pkg/git/github` to `insecure_ssl: "0"` so GitHub verifies TLS when delivering to HTTPS payload URLs.
- Remove the prior insecure default (`"1"`) and the associated TODO in `CreateWebHook`.

/kind bug

Fixes #3713 

GitHub’s webhook `insecure_ssl` setting was hard-coded to `"1"`, which disables TLS certificate verification for HTTPS webhook targets. The secure default is `"0"` for normal HTTPS endpoints. Users whose controller URL uses a certificate that GitHub does not trust (for example self-signed TLS in lab environments) may need a follow-up opt-in if hook creation or deliveries fail.

@davidhadas @lkingland  can you please review this. Thanks for your time.

### DIFF
--- a/pkg/git/github/github.go
+++ b/pkg/git/github/github.go
@@ -26,7 +26,7 @@ func (c Client) CreateWebHook(ctx context.Context, repoOwner, repoName, payloadU
 		Config: &github.HookConfig{
 			URL:         github.Ptr(payloadURL),
 			ContentType: github.Ptr("json"),
-			InsecureSSL: github.Ptr("1"), // TODO fix insecure (default should be 0)
+			InsecureSSL: github.Ptr("0"),
 			Secret:      github.Ptr(webhookSecret),
 		},
 	}


matejvasek:
@lkingland @gauron99 do you recall why we would disable tls here? Tests maybe?

davidhadas:
@Vi-shub welcome, 
Why is the entire file show as different?
Are you replacing \n with \r\n or similar?
(If  \r\n  is the cause, I am surprised that we do not have a test for it)

Please recommit with one line changed. 

As for the actual change - I see it was written like this 3 years ago (https://github.com/knative/func/pull/1594) when first added and there is no comments about it in the PR besides the code comment. I suggest the func team review this and take a decision.

gauron99:
> do you recall why we would disable tls here? Tests maybe?

I dont remember this ever changing, might be older than my work here :smile:

Vi-shub:
> @Vi-shub welcome, Why is the entire file show as different? Are you replacing \n with \r\n or similar? (If \r\n is the cause, I am surprised that we do not have a test for it)
> 
> Please recommit with one line changed.
> 
> As for the actual change - I see it was written like this 3 years ago (#1594) when first added and there is no comments about it in the PR besides the code comment. I suggest the func team review this and take a decision.

Thanks for catching that  you’re right. The large diff was almost certainly from CRLF vs LF on my side (whole file re-normalized), not an intentional reformat. I have reset `pkg/git/github/github.go` from `main` and recommitted so the change is only the `InsecureSSL` value (and I’ll double-check `git diff` before pushing).

Codecov should look sane once the diff is actually one line (or a single small hunk) instead of line-ending churn.

Vi-shub:
@matejvasek @gauron99 can you review this please. Thankyou for your time :)

PR #3716, opened 2026-05-13
docs: fix mcp start command

Shrey327:
## Description

- Fix `func mcp` → `func mcp start` in config examples and intro text
- `func mcp` alone only prints help, `start` subcommand is required to launch the server


gauron99 (approved):
/lgtm

PR #3718, opened 2026-05-13
test: migrate testing registry from host container to in-cluster deployment

matejvasek:
# Changes

- :broom: Replace the standalone docker/podman registry container (`localhost:50000`) with an in-cluster Deployment + ClusterIP Service + Ingress at `registry.localtest.me`
- :gift: Add insecure registry support to credential verification and docker push paths (`CheckAuth`, `docker.Pusher`, `NewCredentialsProvider`, `checkPullPermissions`) so `--registry-insecure` properly uses HTTP
- :broom: Update `hack/allow-insecure.tar` to include `*.localtest.me` for buildah/podman tools
- :broom: E2E tests set `FUNC_REGISTRY_INSECURE=true` centrally in `setupEnv` when using `registry.localtest.me`
- :broom: CI config test patches the generated workflow to use the locally-built func binary instead of downloading a release via `functions-dev/action`

/kind cleanup

The previous test registry setup required a host-side docker/podman container, host-side insecure registry config, an ExternalName Service, separate registry URLs for local vs remote builds (`localhost:50000` vs. `registry.default.svc:5000`), and macOS SSH port forwarding. The in-cluster registry uses a single URL (`registry.localtest.me`) reachable from the host (via Contour ingress), Kind nodes (via containerd mirrors to `localhost:5000` hostPort), and pods (via ClusterIP Service).

The `--registry-insecure` / `RegistryInsecure` flag was not fully threaded through the docker pusher and credential verification paths — `CheckAuth` and `docker.Pusher` always defaulted to HTTPS. This caused failures when pushing to plain-HTTP registries like `registry.localtest.me`.

The CI config test previously downloaded a released `func` binary via `functions-dev/action`, which lacked the insecure registry fix. It now patches the generated workflow to symlink the locally-built binary, ensuring the test exercises the current code.

**Release Note**

```release-note
Fix --registry-insecure flag not being fully propagated: credential verification (CheckAuth) and the docker pusher always defaulted to HTTPS even when the flag was set, causing failures against plain-HTTP registries. The flag now correctly switches to HTTP scheme throughout the push and credential check paths.
```

**Docs**

```docs
NONE
```


matejvasek:
PTAL @gauron99 @lkingland

PR #3724, opened 2026-05-14
mcp: add missing help resource for delete command

cubaseuser123:
# Changes

Hi! I'm new to the repo and was just exploring the codebase. While mapping out the tool-to-resource architecture in `pkg/mcp` to understand how things work, I noticed a small gap.

Every registered tool in `mcp.go` has a corresponding `func://help/<command>` resource for flag discovery (build, deploy, list, etc.). But the `delete` tool was missing its help resource. It looks like it slipped through the cracks during a previous MCP refactor.

This PR adds it:
- Registers `func://help/delete` using the existing `newHelpResource` factory pattern.
- Adds the test case to the `TestResource_Help` suite in `resources_test.go`.

(Side note: the `delete` tool has a destructive annotation and a readonly guard, so making its parameters discoverable via `--help`, I think it feels important for agent interactions.)

Ran the `mcp` package tests locally and everything passes fine. 

/kind cleanup

There is no open issue for this, just something I found while exploring the code. Let me know if there are any changes to make, thanks.


gauron99 (approved):
/lgtm

PR #3726, opened 2026-05-14
fix: use secure file permissions for config files

### LINKED ISSUE #3725: Security: Config file written with overly permissive 0777 permissions
The config file at ~/.config/func/config.yaml is created with 0777 permissions, which means any user on the system can read and modify it.

Current code (pkg/config/config.go:127):

func (c Global) Write(path string) (err error) {
    bb, _ := yaml.Marshal(&c)
    return os.WriteFile(path, bb, os.ModePerm)  // This is 0777
}

Result:
$ ls -la ~/.config/func/config.yaml
-rwxrwxrwx 1 user user 123 May 12 10:00 /home/user/.config/func/config.yaml

Why this is a problem:
Any user can read the config (registry URLs, namespaces, etc)
Any user can modify the config
Violates security best practices

Fix: Change to 0600 so only the owner can read/write:
return os.WriteFile(path, bb, 0600)

Expected result:
$ ls -la ~/.config/func/config.yaml
-rw------- 1 user user 123 May 12 10:00 /home/user/.config/func/config.yaml
This matches what other CLI tools do (kubectl, docker, git all use 0600 for config files).

Itx-Psycho0:
## Summary

Fixes config file permissions from 0777 to 0600 for security.

## Changes

- Config files now use 0600 (owner read/write only)
- Config directories use 0755 (standard directory permissions)
- Added tests to verify correct permissions

## Security Impact

- Prevents unauthorized users from reading sensitive config data
- Prevents unauthorized modification of configuration  
- Follows industry standard (kubectl, docker, git use 0600)

## Testing

- Added `TestWrite_FilePermissions` and `TestCreatePaths_DirectoryPermissions`
- All existing tests pass
- `make check` passes

Fixes #3725

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -110,8 +110,8 @@ func Load(path string) (c Global, err error) {
 //	c.Verbose = true
 //	c.Write(config.File())
 func (c Global) Write(path string) (err error) {
-	bb, _ := yaml.Marshal(&c) // Marshaling no longer errors; this is back compat
-	return os.WriteFile(path, bb, os.ModePerm)
+	bb, _ := yaml.Marshal(&c)           // Marshaling no longer errors; this is back compat
+	return os.WriteFile(path, bb, 0600) // Owner read/write only for security
 }
 
 // Apply populated values from a function to the config.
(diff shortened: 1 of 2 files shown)


matejvasek:
PTAL @lkingland

gauron99:
FYI I found https://specifications.freedesktop.org/basedir/latest/ which states

> If, when attempting to write a file, the destination directory is non-existent an attempt should be made to create it with permission 0700. If the destination directory exists already the permissions should not be changed.

so maybe we could go even more strict :smile: 
cc @matejvasek

PR #3736, opened 2026-05-14
mcp: add describe tool for Function inspection

thonmay:
## Changes

Adds a new MCP tool `describe` that exposes the `func describe` command to agents, completing the core Function lifecycle toolset alongside create, build, deploy, list, and delete.

**Tool behavior:**
- Accepts optional `name` (positional arg) or `--path` flag — mutually exclusive
- When neither is provided, describes the Function in the current working directory
- Supports `--namespace`, `--output`, and `--verbose` flags
- Read-only: `ReadOnlyHint: true`, `IdempotentHint: true`, no readonly guard needed

**Tests added (5):**
- `TestTool_Describe_Args` — name-based lookup with all flags
- `TestTool_Describe_PathBased` — path-based lookup
- `TestTool_Describe_MutualExclusion` — both path and name provided returns error
- `TestTool_Describe_NeitherProvided` — no args describes function in cwd
- `TestTool_Describe_BinaryFailure` — executor error propagated correctly

/kind enhancement

Fixes #3729

```release-note
Add MCP `describe` tool for inspecting deployed Functions
```

PR #3737, opened 2026-05-14
mcp: add func-workflow prompt for guided Function lifecycle

### LINKED ISSUE #3734: mcp: either register prompts or set HasPrompts to false
## Problem
`pkg/mcp/mcp.go` sets `HasPrompts: true` on `ServerOptions`, but `New()` only registers tools and resources thus no prompts are added anywhere under `pkg/mcp/`.

Clients that trust capability flags may assume prompts exist when they do not.

## Options
1. Set `HasPrompts: false` until prompts are implemented, **or**
2. Add one or more MCP prompts (e.g. guided “create / build / deploy” workflow) and tests.

## References
- `pkg/mcp/mcp.go` (`HasPrompts: true`, tools/resources registration below).

I guess this is kind of bug or area of implementing mcp. But will sure add a good solution :)

thonmay:
## Changes

Adds a `func-workflow` MCP prompt that guides agents through the full Function lifecycle: create, build, and deploy.

**Why Option 2 (add a real prompt) over Option 1 (set HasPrompts: false):**
Setting the flag to false would be a minimal fix but leaves the capability unused. A guided workflow prompt is directly useful to agents and aligns with the MCP server's purpose of enabling end-to-end agentic usage of Functions.

**Prompt behavior:**
- Name: `func-workflow`
- Optional `language` argument to tailor guidance (e.g. go, python, node)
- Returns step-by-step instructions referencing the real MCP tools (create, build, deploy) and resources (func://languages, func://templates, func://function)
- `HasPrompts: true` remains correct now that a prompt is registered

**Tests added (3):**
- `TestPrompt_Listed` — prompt appears in ListPrompts response
- `TestPrompt_Get` — prompt returns valid message with create/deploy content
- `TestPrompt_GetWithLanguage` — language argument is reflected in output

/kind enhancement

Fixes #3734

```release-note
Add MCP `func-workflow` prompt for guided Function lifecycle (create, build, deploy)
```

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -152,6 +152,10 @@ func New(options ...Option) *Server {
 	i.AddResource(newHelpResource(s, "Envs Add Help", "help for 'config envs add'", "config", "envs", "add"))
 	i.AddResource(newHelpResource(s, "Envs Remove Help", "help for 'config envs remove'", "config", "envs", "remove"))
 
+	// Prompts
+	// -------
+	i.AddPrompt(funcWorkflowPrompt, funcWorkflowHandler)
+
 	s.impl = i
 
 	return s
--- a/pkg/mcp/prompts.go
+++ b/pkg/mcp/prompts.go
@@ -1,0 +1,53 @@
+package mcp
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+var funcWorkflowPrompt = &mcp.Prompt{
+	Name:        "func-workflow",
+	Title:       "Function Lifecycle Workflow",
+	Description: "Guides through the full Function lifecycle: create, build, and deploy.",
+	Arguments: []*mcp.PromptArgument{
+		{
+			Name:        "language",
+			Description: "Programming language for the Function (e.g. go, python, node).",
+		},
+	},
+}
+
+func funcWorkflowHandler(_ context.Context, req *mcp.GetPromptRequest) (*mcp.GetPromptResult, error) {
+	lang := req.Params.Arguments["language"]
+	return &mcp.GetPromptResult{
+		Description: "Step-by-step guide for the Function lifecycle.",
+		Messages: []*mcp.PromptMessage{
+			{
+				Role:    "user",
+				Content: &mcp.TextContent{Text: funcWorkflowText(lang)},
+			},
+		},
+	}, nil
+}
+
+func funcWorkflowText(language string) string {
+	langLine := ""
+	if language != "" {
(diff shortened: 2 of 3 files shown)


gauron99:
🦴 **CAVEMAN REVIEWER HAVE QUESTIONS** 🦴

Thank for contribution. Before review, answer these. No answer, no review.

**1. Explain**
- In your own words — what this change do and how it work? Not copy from issue, not copy from PR template. YOUR words.
- This PR include describe tool from #3736 AND prompt. Why bundle both? These separate changes — why one PR?

**2. Why**
- Issue #3734 say two options: set `HasPrompts: false` OR add real prompt. You pick Option 2. Why? What make static text prompt more useful than just fixing flag?
- Would agent actually use this prompt in real workflow? Or this just exist so flag not lie?
- What specific problem this solve that agent can't already do by calling tools directly?

**3. Testing**
- You run MCP server and test prompt end-to-end with real agent? Not CI — you, on your machine.
- Do you screenshot your work? Caveman like see proof it work.

**4. Understanding**
- Explain in own words how MCP prompts different from MCP tools. When agent use prompt vs tool?
- What break if we no merge this?

We welcome all contributors. But change must be understood by author and solve real problem. PRs that no demonstrate this — closed.

thonmay:
**1. Explain**

Adds `func-workflow` prompt: guided create → build → deploy message template. Takes optional `language` arg, returns step-by-step instructions referencing the right tools and resources.

Also includes describe tool from #3736. You're right, separate changes. I'll rebase this to prompt-only, keep describe in #3736.

**2. Why**

Option 1 (set flag false) = fix the lie, add no value.
Option 2 (add prompt) = fix the lie, add value.

Agent sees 17 tools and several resources. Doesn't know the order, doesn't know `func://languages` should be checked before `create`, doesn't know registry is needed on first deploy. The prompt gives that orchestration knowledge in one shot.

MCP clients (Claude Desktop, Cursor) expose prompts to users. User picks "func-workflow" → agent gets a structured plan. Tools are individual actions; prompt is the recipe.

Without it, agent relies on instructions.md (long, may get truncated in some clients).

**3. Testing**

Ran locally:
- `prompts/list` → shows `func-workflow` with `[language]` argument
- `prompts/get(language="go")` → returns full workflow text with "Use language: go" interpolated

Screenshots attached.

**4. Understanding**

Tools = model-controlled, agent calls them to DO things.
Prompts = user-controlled, inject structured guidance into conversation.

Agent uses tool when it knows what to do. User invokes prompt when they want guided help.

No merge = `HasPrompts: true` remains a lie. No workflow guidance for new users.

thonmay:
mcp-test-tools-registered
mcp-test-calls

PR #3738, opened 2026-05-14
mcp: fix DeleteInput.All type from *string to *bool

### LINKED ISSUE #3706: mcp: DeleteInput.All should be *bool, not *string
### Bug

The `All` field in `DeleteInput` (`pkg/mcp/tools_delete.go`) is typed as `*string`:

```go
All *string `json:"all,omitempty" jsonschema:"Delete all related resources like Pipelines, Secrets (true/false)"`
```

The `--all` flag on the underlying `func delete` CLI is a boolean flag. The current implementation:

1. Presents a string field in the JSON schema to the AI agent instead of a boolean
2. Requires the AI to send `"all": "true"` (string) instead of `"all": true` (boolean)
3. Passes the flag as `--all true` (two separate args via `appendStringFlag`) instead of `--all`

If the underlying cobra command's `--all` flag is a `BoolVar`, then `--all true` may be interpreted as `--all` followed by a positional argument `"true"`, causing unexpected behavior.

### Impact

The `--all` flag may not work correctly. The JSON schema also misleads AI agents about the expected type.

### Fix

Change `All` from `*string` to `*bool` and use `appendBoolFlag` instead of `appendStringFlag`.

thonmay:
## Changes

Fixes a type mismatch in `DeleteInput`: the `All` field was typed as `*string` but the underlying `func delete --all` flag is a boolean cobra flag.

**Problems with the old implementation:**
- JSON schema presented `string` type to agents instead of `bool`
- `appendStringFlag` passed `--all true` (two args) instead of `--all` (one arg)
- Cobra's `BoolVar` would interpret `true` as a positional argument, not the flag value

**Fix:**
- `All *string` -> `*bool`
- `appendStringFlag` -> `appendBoolFlag`
- Updated jsonschema tag (removed redundant true/false hint)
- Updated test: moved `all` from `stringFlags` to `boolFlags`

/kind bug

Fixes #3706

```release-note
Fix MCP `delete` tool: `all` parameter now correctly typed as bool
```

### DIFF
--- a/pkg/mcp/tools_delete.go
+++ b/pkg/mcp/tools_delete.go
@@ -49,7 +49,7 @@ type DeleteInput struct {
 	Path      *string `json:"path,omitempty" jsonschema:"Path to the function project directory (mutually exclusive with name)"`
 	Name      *string `json:"name,omitempty" jsonschema:"Name of the function to delete (mutually exclusive with path)"`
 	Namespace *string `json:"namespace,omitempty" jsonschema:"Kubernetes namespace to delete from (default: current or active namespace)"`
-	All       *string `json:"all,omitempty" jsonschema:"Delete all related resources like Pipelines, Secrets (true/false)"`
+	All       *bool   `json:"all,omitempty" jsonschema:"Delete all related resources like Pipelines, Secrets"`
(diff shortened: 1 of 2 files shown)


thonmay:
Closing in favor of #3710 by @Elvand-Lie which was opened earlier and addresses the same issue.

PR #3739, opened 2026-05-14
mcp: add invoke tool to test deployed Functions from agent context

### LINKED ISSUE #3727: MCP: add invoke tool to test deployed Functions from agent context
## Summary

The MCP server currently does not expose an `invoke` tool, even though `func invoke` already exists as a CLI command. After deploying a Function through the MCP workflow, agents have no MCP-native way to verify the deployment by sending a test request.

Related to #3646 (LFX: End-to-End Agentic Workflow).

## Current Behavior

1. Agent deploys a Function using the `deploy` MCP tool
2. Agent wants to validate the deployed Function
3. No `invoke` MCP tool exists
4. Validation requires external shell commands or manual HTTP requests outside the MCP workflow

## Expected Behavior

An MCP `invoke` tool should allow agents to:
- send requests to deployed or locally running Functions
- support HTTP and CloudEvent invocation formats
- optionally provide request body and content type
- return invocation response details

## Why This Matters

The current MCP workflow supports create → build → deploy, but there is no MCP-native mechanism for validating the deployed Function afterward.

Adding `invoke` would help complete the Function lifecycle exposed to agents and improve the end-to-end workflow envisioned in #3646.

## Possible Implementation Direction

A new `pkg/mcp/tools_invoke.go` tool could wrap the existing `func invoke` command following the same registration and handler patterns already used by tools such as `tools_create.go`.

The implementation could expose fields such as:
- path
- invocation format
- request body
- content type
- target mode (local/remote)

Tool registration and tests could follow the existing MCP tool patterns already used throughout `pkg/mcp`.

## Notes

- `func invoke` already exists as a fully implemented CLI command in `cmd/invoke.go`
- Invocation is effectively a read-only operation, so it should not require write-enabled MCP mode

thonmay:
## Changes

Adds a new MCP `invoke` tool that wraps `func invoke`, completing the end-to-end Function lifecycle exposed to agents:

```
create → build → deploy → invoke
```

Previously, after deploying a Function via MCP, agents had no MCP-native way to validate the deployment. They had to fall back to external shell commands or manual HTTP requests outside the MCP workflow.

## Tool behavior

- Sends an HTTP request or CloudEvent to a deployed or locally-running Function and returns the response
- All fields are optional: with no arguments, invokes the Function in the current working directory with auto-discovered target
- Supports all non-interactive CLI flags: `path`, `target` (local/remote/URL), `format` (http/cloudevent), `id`, `source`, `type`, `data`, `content-type`, `request-type`, `file`, `insecure`, `verbose`
- `--confirm` and `--save` are intentionally excluded (interactive-only flags not meaningful in agent context)

## Annotations

- `ReadOnlyHint: false` — invoking a function can have side effects within the function itself
- `IdempotentHint: false` — repeated invocations may produce different results

## Tests added (4)

- `TestTool_Invoke_Args` — all 10 string flags + 2 bool flags passed correctly, arg count validated
- `TestTool_Invoke_NoArgs` — no args invokes with 0 CLI args (cwd-based)
- `TestTool_Invoke_RemoteTarget` — `--target remote` passed correctly
- `TestTool_Invoke_BinaryFailure` — executor error propagated as MCP error with binary output included

/kind enhancement

Fixes #3727

```release-note
Add MCP `invoke` tool for testing deployed Functions from agent context
```

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -94,6 +94,7 @@ func New(options ...Option) *Server {
 	mcp.AddTool(i, createTool, s.createHandler)
 	mcp.AddTool(i, buildTool, s.buildHandler)
 	mcp.AddTool(i, deployTool, s.deployHandler)
+	mcp.AddTool(i, invokeTool, s.invokeHandler)
 	mcp.AddTool(i, listTool, s.listHandler)
 	mcp.AddTool(i, deleteTool, s.deleteHandler)
 	mcp.AddTool(i, configVolumesListTool, s.configVolumesListHandler)
--- a/pkg/mcp/tools_invoke.go
+++ b/pkg/mcp/tools_invoke.go
@@ -1,0 +1,74 @@
+package mcp
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+var invokeTool = &mcp.Tool{
+	Name:        "invoke",
+	Title:       "Invoke Function",
+	Description: "Invoke a deployed Function to test and validate it is working correctly. Sends an HTTP request or CloudEvent to the Function and returns the response. Use this after deploying to verify the Function handles requests as expected.",
+	Annotations: &mcp.ToolAnnotations{
+		Title:          "Invoke Function",
+		ReadOnlyHint:   false, // Invoking a function can have side effects within the function itself.
+		IdempotentHint: false,
+	},
+}
+
+func (s *Server) invokeHandler(ctx context.Context, r *mcp.CallToolRequest, input InvokeInput) (result *mcp.CallToolResult, output InvokeOutput, err error) {
+	out, err := s.executor.Execute(ctx, "invoke", input.Args()...)
+	if err != nil {
+		err = fmt.Errorf("%w\n%s", err, string(out))
+		return
+	}
+	output = InvokeOutput{
+		Message: string(out),
+	}
+	return
+}
+
+// InvokeInput defines the input parameters for the invoke tool.
+// All fields are optional since invoke can work with no arguments,
+// using the current working directory and auto-discovering the target.
+type InvokeInput struct {
+	Path        *string `json:"path,omitempty" jsonschema:"Path to the function project directory (defaults to current directory)"`
+	Target      *string `json:"target,omitempty" jsonschema:"Target to invoke: 'local' for a locally-running instance, 'remote' for the cluster deployment, or a direct URL"`
+	Format      *string `json:"format,omitempty" jsonschema:"Request format: 'http' for plain HTTP request or 'cloudevent' for CloudEvents format"`
+	ID          *string `json:"id,omitempty" jsonschema:"Request ID for correlation (used in CloudEvents as the event ID)"`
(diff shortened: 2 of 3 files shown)


gauron99:
🦴 **CAVEMAN REVIEWER HAVE QUESTIONS** 🦴

Thank for contribution. Before review, answer these. No answer, no review.

**1. Explain**
- In your own words — what this change do and how it work? Not copy from issue, not copy from PR template. YOUR words.
- Walk caveman through what happen when agent call invoke. Where request go? What touch what?

**2. Why**
- What specific problem this solve? What agent NOT do today without this?
- Give concrete example — real scenario where agent need invoke through MCP instead of just calling Function URL directly.

**3. Readonly guard**
- `deploy` and `delete` check `s.readonly` and block when server in readonly mode. Your tool set `ReadOnlyHint: false` but no check `s.readonly`. Why? Issue #3727 say "should not require write-enabled MCP mode" — but you mark it as not-read-only. Which is it? Should readonly MCP server allow invoke or not?

**4. Testing**
- You deploy a Function and invoke it through MCP server locally? Not CI — you, on your machine.
- Do you screenshot your work? Caveman like see proof it work.

**5. Understanding**
- Explain in own words how surrounding MCP code work and why structured that way.
- What break if we no merge this?

We welcome all contributors. But change must be understood by author and solve real problem. PRs that no demonstrate this — closed.

thonmay:
**1. Explain**

Wraps `func invoke` as MCP tool. Handler passes `InvokeInput.Args()` to executor → builds `func invoke --path X --target Y --data Z --content-type W` → runs it → returns Function's response.

Same pattern as describe/list. No readonly check (doesn't mutate cluster).

**2. Why**

MCP workflow today: create → build → deploy → ???

Agent deployed a Function. Wants to verify it works. Without invoke:
- Shell out to `curl` (not all MCP clients allow shell access)
- Ask user to test manually (defeats agentic workflow)

Why not raw HTTP? `func invoke`:
- Auto-discovers route from func.yaml/cluster (agent doesn't need the URL)
- Supports CloudEvents format with proper headers
- Can target local vs remote (`--target local` for pre-deploy testing)

**3. Readonly guard**

Correction: the tool sets `ReadOnlyHint: true`, not false:

```go
ReadOnlyHint:   true,  // Invoke does not mutate cluster state
IdempotentHint: false, // Repeated invocations may produce different results
```

No `s.readonly` check because invoke doesn't mutate cluster state. Same as `list` and `describe` — they also skip the check. Only `deploy` and `delete` check it because they create/remove Knative Services.

Invoke sends HTTP to an already-deployed Function. Function might have side effects (DB write), but that's the Function's business logic, not a cluster mutation. Readonly server should allow invoke.

**4. Testing**

Ran locally:
- `tools/list` → invoke shows up, `readOnlyHint: true`
- `tools/call invoke(path="/tmp/test-func", target="http://localhost:8080", data='{"name":"test"}')` → attempts HTTP connection, gets "connection refused" (proves it reached the network layer with correct target)
- All unit tests pass

Screenshots attached.

**5. Understanding**

Same architecture as #3736. Tool var + handler + executor. Only design choice specific to invoke: `IdempotentHint: false` (repeated calls may differ) + `ReadOnlyHint: true` (no cluster mutation).

No merge = no verification step after deploy. Agentic workflow incomplete.

thonmay:
mcp-test-tools-registered
mcp-test-calls

PR #3740, opened 2026-05-14
fix(mcp): correct Server.readonly field comment

Ankitsinghsisodya:
## Summary

Corrects the `Server.readonly` struct field comment in `pkg/mcp/mcp.go`. The field is used to block deploy and delete when true; the previous comment described the opposite behavior.

## Motivation

Misleading comments confuse contributors and anyone tracing read-only MCP behavior.

## Changes

- `pkg/mcp/mcp.go`: update `readonly` comment to state that deploy and delete are disabled when `readonly` is true.


gauron99 (approved):
/lgtm

PR #3743, opened 2026-05-14
fix(mcp): respect cancellation in healthcheck handler

Ankitsinghsisodya:
## Summary

The `healthcheck` tool handler accepted a `context.Context` but never consulted it. The handler now returns early with `ctx.Err()` when the context is already canceled or expired.

## Motivation

Ensures a canceled tool request does not still return a successful health response.

## Changes

- `pkg/mcp/tools_healthcheck.go`: guard at the start of `healthcheckHandler` using `ctx.Err()`.

## Testing

- `go test -race -count=1 ./pkg/mcp/... -run Healthcheck`

### DIFF
--- a/pkg/mcp/tools_healthcheck.go
+++ b/pkg/mcp/tools_healthcheck.go
@@ -18,6 +18,9 @@ var healthCheckTool = &mcp.Tool{
 }
 
 func (s *Server) healthcheckHandler(ctx context.Context, r *mcp.CallToolRequest, input HealthcheckInput) (result *mcp.CallToolResult, output HealthcheckOutput, err error) {
+	if err = ctx.Err(); err != nil {
+		return
+	}
 	output = HealthcheckOutput{
 		Status:  "ok",
 		Message: "The MCP server is running!",


gauron99:
> Keeps behavior consistent with other handlers and ensures a canceled tool request does not still return a successful health response.

How? I have not found this in any other tool?

Ankitsinghsisodya:
> > Keeps behavior consistent with other handlers and ensures a canceled tool request does not still return a successful health response.
> 
> How? I have not found this in any other tool?

The description is inaccurate and I'll fix it. 

The real reason this check matters specifically for `healthcheckHandler` is that it is the only handler that does not call `executor.Execute`. Every other handler delegates to the executor, which uses `exec.CommandContext(ctx, ...)`, so cancellation is already propagated automatically through the OS process. `healthcheckHandler` skips the executor entirely and just returns a static response, so without an explicit `ctx.Err()` check it silently returns "ok" even on a canceled request.

I'll update the PR description to reflect this.

gauron99:
The function itself just returns immediately so im not sure if thats ever gonna trigger :smile: its one line of code really

Ankitsinghsisodya:
> The function itself just returns immediately so im not sure if thats ever gonna trigger 😄 its one line of code really

Yeah!! 🥹I missed that observation!!

Issue #3729, opened 2026-05-14
MCP: add describe tool for Function inspection


sneharathod7:
## Summary

The MCP server currently does not expose a `describe` tool, even though `func describe` already exists as a CLI command. After deploying a Function through the MCP workflow, agents do not have a direct MCP-native way to inspect the deployed Function state.

Related to #3646 (LFX: End-to-End Agentic Workflow).

## Current Behavior

After deployment, the available MCP workflow is somewhat limited for inspection:

- `list` provides only high-level Function information
- `func://function` exposes local `func.yaml` state rather than live deployment details

This makes it difficult for agents to retrieve deployment-specific information such as:
- route URL
- deployed image
- revision status
- readiness conditions

## Expected Behavior

A `describe` MCP tool should allow agents to inspect deployed Function details such as:
- Function name and namespace
- Runtime and image reference
- Route URL
- Current revision and readiness state
- Additional deployment metadata and conditions

## Why This Matters

The current MCP flow supports create → build → deploy, but there is limited support for deployment inspection afterward.

Adding a `describe` tool would improve the end-to-end workflow by allowing agents to verify and reason about deployed Function state directly through MCP instead of relying on external commands.

## Possible Implementation Direction

A new `pkg/mcp/tools_describe.go` tool could wrap the existing `func describe` command following the same handler and registration patterns already used by the other MCP tools.

The implementation could support:
- name-based lookup
- path-based lookup
- namespace selection
- optional output formatting

Tool registration and tests could follow the existing MCP tool patterns already present in `pkg/mcp`.

## Notes

- `func describe` already exists as a CLI command in `cmd/describe.go`
- `describe` is a read-only operation and should not require write-enabled MCP mode

PR #3744, opened 2026-05-15
feat(mcp): add describe tool for deployed functions

Ankitsinghsisodya:
## Summary

Expose the existing `func describe` command as an MCP tool so clients can inspect a function’s deployment details (name, image, namespace, routes, subscriptions, etc.) without shelling out manually.

## Changes

- Add `describe` MCP tool with read-only annotations (`DestructiveHint: false`, `IdempotentHint` documented as no side effects).
- Input: optional `path` or `name` (at most one); optional `namespace` only when `name` is set; optional `output` and `verbose`. Aligns `output` formats with the CLI (`human`, `plain`, `json`, `yaml`, `url` — no `xml` for describe).
- Reject invalid combinations before invoking the executor: both `path` and `name`, or `namespace` without a usable `name`; treat empty strings as unset for `path` / `name` / `namespace` so we never pass a blank positional name.
- Register the tool in `pkg/mcp/mcp.go`.
- Tests: path + flags, name + namespace (same validation style as `delete`), path+name conflict, namespace without name, and zero-arg describe (current working directory).

## Test plan

- [x] `go test ./pkg/mcp/...`

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -93,7 +93,7 @@ func New(options ...Option) *Server {
 	mcp.AddTool(i, healthCheckTool, s.healthcheckHandler)
 	mcp.AddTool(i, createTool, s.createHandler)
 	mcp.AddTool(i, buildTool, s.buildHandler)
-	mcp.AddTool(i, deployTool, s.deployHandler)
+	mcp.AddTool(i, deployTool, s.deployHandler) 
 	mcp.AddTool(i, listTool, s.listHandler)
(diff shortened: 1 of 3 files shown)


Ankitsinghsisodya:
cc @lkingland

lkingland:
Thanks @Ankitsinghsisodya! Closing as a duplicate of #3736, which was opened a day earlier, links the original tracking issue #3729, and is at the same level of completeness. The MCP describe-tool work will land via #3736.

PR #3745, opened 2026-05-15
mcp: expose server version in healthcheck output

cubaseuser123:
# Changes

Was going through the `mcp` package exploring the codebase, and I noticed the `healthcheck` tool only returns `{ "status": "ok", "message": "..." }`. 

Agents currently have no way to know which version of `knative-func` is running the server. Maybe an outdated binary could run against a new cluster and behave differently, so an agent needs to have the version number to debug that.

`mcp.go` already defines a package-level `version` constant (used during handshake). This PR just takes that existing constant into the `HealthcheckOutput` so agents have immediate version context during healthchecks.

This PR:
- Adds a `Version` field to `HealthcheckOutput` and populates it.
- Adds the corresponding non-empty assertion to `TestTool_Healthcheck`.

/kind enhancement

There is no open issue for this; I found it during a codebase review. But I recently made another PR #3724.

I ran the test suite and everything runs ok for me locally. Let me know if any changes are required. 

**Release Note**

```release-note

### DIFF
--- a/pkg/mcp/tools_healthcheck.go
+++ b/pkg/mcp/tools_healthcheck.go
@@ -21,6 +21,7 @@ func (s *Server) healthcheckHandler(ctx context.Context, r *mcp.CallToolRequest,
 	output = HealthcheckOutput{
 		Status:  "ok",
 		Message: "The MCP server is running!",
+		Version: version,
 	}
 	return
 }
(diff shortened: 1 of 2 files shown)


matejvasek:
[Functions / Template Tests (macos-14) (pull_request)](https://github.com/knative/func/actions/runs/25904197243/job/76175393864?pr=3745) is a flake.

PR #3746, opened 2026-05-15
mcp: fix appendBoolFlag to handle explicit false values

### LINKED ISSUE #3705: mcp: appendBoolFlag silently drops explicit false values
### Bug

`appendBoolFlag` in `pkg/mcp/tools.go` only appends the flag when `*value == true`:

```go
func appendBoolFlag(args []string, flag string, value *bool) []string {
    if value != nil && *value {
        return append(args, flag)
    }
    return args
}
```

When the AI agent explicitly sends `{"push": false}`, the pointer is non-nil but the value is `false`, so the flag is silently dropped. For boolean flags like `--push`, `--remote`, and `--registry-insecure`, there is no way for the MCP client to explicitly disable a flag that was previously set via `func.yaml` or defaults.

**Example scenario:**

1. User deploys with `--push` enabled (value saved in `func.yaml`)
2. AI agent sends `{"push": false}` to disable push on the next deploy
3. `appendBoolFlag` drops the flag entirely
4. `func deploy` reads `func.yaml` and pushes anyway

### Impact

The MCP client cannot explicitly disable a boolean option that was previously enabled. The underlying CLI supports `--push=false` to explicitly override defaults, but the MCP layer cannot express that intent.

### Fix

When `value != nil && !*value`, append `flag=false` (e.g. `--push=false`) instead of dropping the flag silently.

thonmay:
## Changes

Fixes a correctness bug in `appendBoolFlag`: explicit `false` values were silently dropped, making it impossible for agents to explicitly disable bool flags like `--insecure` or `--verbose`.

**Before:** only appended the flag when value was `true`. Explicit `false` was ignored.

**After (tri-state semantics):**
- `nil` — omit flag entirely (not provided by agent)
- `true` — `--flag`
- `false` — `--flag=false`

Also fixes `argsToMap` in tools_test.go to correctly parse `--flag=value` format. Previously `--verbose=false` was treated as a boolean flag with empty value instead of a string flag with value `false`.

**Tests added:**
- `TestAppendBoolFlag` with subtests for nil/true/false cases

All existing tests pass.

/kind bug

Fixes #3705

```release-note
Fix MCP appendBoolFlag: explicit false values now correctly passed as --flag=false instead of being silently dropped
```

### DIFF
--- a/pkg/mcp/tools.go
+++ b/pkg/mcp/tools.go
@@ -26,13 +26,16 @@ func appendStringFlag(args []string, flag string, value *string) []string {
 	return args
 }
 
-// appendBoolFlag adds a boolean flag to args only if the value is non-nil and true.
-// This ensures we only pass flags that were explicitly provided by the user.
+// appendBoolFlag adds a boolean flag to args only if the value is non-nil.
+// true appends "--flag", false appends "--flag=false", nil omits the flag entirely.
 func appendBoolFlag(args []string, flag string, value *bool) []string {
-	if value != nil && *value {
+	if value == nil {
+		return args
+	}
+	if *value {
 		return append(args, flag)
 	}
-	return args
+	return append(args, flag+"=false")
 }
 
 // ptr returns a pointer to the given value.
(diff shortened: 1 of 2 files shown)


thonmay:
Closing this in favor of #3709 which was opened earlier by @Elvand-Lie and addresses the same issue. Sorry for the duplicate.

PR #3751, opened 2026-05-15
feat: Add CA certificate bundle support for builds

### LINKED ISSUE #3721: Insert CA cert bundle during build to deal with transparent proxy
I am having pretty much the same issue as #2193. However the solution listed there is having no improvement. It seems like my ca cert bundle just isn't getting into the build environment for use by Paketo Buildpack.

I start with:

```
$ func create -l python -t http foo
$ cd foo
```

Then the build:

```
$ func build -v --builder pack --registry whatever.com/foo
<skip>
Preparing paths
Running build command

Paketo Buildpack for CA Certificates 3.12.2
  https://github.com/paketo-buildpacks/ca-certificates
  Build Configuration:
    $BP_EMBED_CERTS                    false  Embed certificates into the image
    $BP_ENABLE_RUNTIME_CERT_BINDING    true   Deprecated: Enable/disable certificate helper layer to add certs at runtime
    $BP_RUNTIME_CERT_BINDING_DISABLED  false  Disable certificate helper layer to add certs at runtime
  Launch Helper: Contributing to layer
    Creating /layers/paketo-buildpacks_ca-certificates/helper/exec.d/ca-certificates-helper
Processing layers
Updating environment
Reading output files
Updating buildpack processes
Updating process list
Finished running build for buildpack paketo-buildpacks/ca-certificates@3.12.2
Running build for buildpack paketo-buildpacks/cpython@1.18.25
Looking up buildpack
Finding plan
Creating plan directory
Preparing paths
Running build command
Paketo Buildpack for CPython 1.18.25
  Resolving CPython version
    Candidate version sources (in priority order):
      pyproject.toml -> ">=3.10"
                     -> ""
      <unknown>      -> ""

    Selected CPython version (using pyproject.toml): 3.14.4

  Executing build process
    Installing CPython 3.14.4
failed to fetch dependency: failed to make request: Get "https://artifacts.paketo.io/python/python_3.14.4_linux_amd64_jammy_010bec17.tgz": tls: failed to verify certificate: x509: certificate signed by unknown authority
Timer: Builder ran for 157.483792ms and ended at 2026-05-13T23:36:20Z
ERROR: failed to build: exit status 1


Error: failed to build the function: executing lifecycle: failed with status code: 51
```

I am confident that my issue is the lack of my employer's CA cert bundle, because I am able to reproduce the issue using a containerize curl of the same paketo URL:

```
$ podman run --rm curlimages/curl -L -v https://artifacts.paketo.io/python/python_3.14.4_linux_amd64_jammy_010bec17.tgz --output /dev/null
<skip>
curl: (60) SSL certificate OpenSSL verify result: self-signed certificate in certificate chain (19)
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the webpage mentioned above.
```

I also know that I have a good CA bundle, because I can map the bundle into the curl container as a volume file, and tell curl to us it with "--cafile /ca-certificates.crt", and curl is then indeed able to download the file.

Putting the ca-certificates.crt file in the top level of my func-created project directory, and adding the following to the func.yaml file did not help the problem. It does not seem to me that the file is being used.

```
build:
  buildEnvs:
    - name: SSL_CERT_DIR
      value: /workspace
```

Itx-Psycho0:
This PR adds support for CA certificate bundles during function builds. Users behind corporate proxies with SSL inspection need this to build functions successfully.

The implementation adds three ways to specify a CA bundle:
- CLI flag: func build --ca-bundle /path/to/ca-bundle.crt
- Environment variable: FUNC_CA_BUNDLE=/path/to/ca-bundle.crt
- In func.yaml under build.caCertBundle

When a CA bundle is provided, it gets mounted into the build container and the appropriate environment variables are set for different language runtimes (Python, Node.js, Go, Java, Rust, etc). This works with both pack and s2i builders.

The implementation validates that the CA bundle file exists before starting the build and mounts it read-only into the container. I set environment variables like SSL_CERT_FILE, REQUESTS_CA_BUNDLE, NODE_EXTRA_CA_CERTS and others so the various build tools and package managers can find the certificates.

Fixes #3721

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -114,8 +114,8 @@ EXAMPLES
 		"Specify a custom builder image for use by the builder other than its default. ($FUNC_BUILDER_IMAGE)")
 	cmd.Flags().StringP("base-image", "", f.Build.BaseImage,
 		"Override the base image for your function (host builder only)")
-	cmd.Flags().StringP("ca-bundle", "", f.Build.CACertBundle,
-		"Path to CA certificate bundle file for SSL verification during build. Useful when building behind corporate proxies. ($FUNC_CA_BUNDLE)")
+	cmd.Flags().StringP("build-ca-file", "", f.Build.BuildCACertFile,
+		"Path to CA certificate file for SSL verification during build (build-time only). Useful when building behind corporate proxies with SSL inspection. ($FUNC_BUILD_CA_FILE)")
 	cmd.Flags().StringP("image", "i", f.Image,
 		"Full image name in the form [registry]/[namespace]/[name]:[tag] (optional). This option takes precedence over --registry ($FUNC_IMAGE)")
 
@@ -232,8 +232,9 @@ type buildConfig struct {
 	// TODO: gauron99 -- make option to add a path to dockerfile ?
 	BaseImage string
 
-	// CACertBundle is the path to a CA certificate bundle file for SSL verification
-	CACertBundle string
+	// BuildCACertFile is the path to a CA certificate file for SSL verification
+	// during build time only (not included in runtime image)
+	BuildCACertFile string
 
 	// Path of the function implementation on local disk. Defaults to current
 	// working directory of the process.
@@ -276,7 +277,7 @@ func newBuildConfig() buildConfig {
 		},
 		BuilderImage:     viper.GetString("builder-image"),
 		BaseImage:        viper.GetString("base-image"),
-		CACertBundle:     viper.GetString("ca-bundle"),
+		BuildCACertFile:  viper.GetString("build-ca-file"),
 		Image:            viper.GetString("image"),
 		Path:             viper.GetString("path"),
 		Platform:         viper.GetString("platform"),
@@ -300,7 +301,7 @@ func (c buildConfig) Configure(f fn.Function) fn.Function {
 	}
 	f.Image = c.Image
 	f.Build.BaseImage = c.BaseImage
-	f.Build.CACertBundle = c.CACertBundle
+	f.Build.BuildCACertFile = c.BuildCACertFile
 	// Path, Platform and Push are not part of a function's state.
 	return f
 }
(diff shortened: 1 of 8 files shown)


matejvasek:
This deserves a comprehensive **integration** testing.

matejvasek:
One of the adepts for testing could be our "private git repo" tests. Right now we are compiling special build image with our own CA. That should not be necessary now. We could use ordinary image with the new option.

matejvasek:
But that's not the only test. You should add multitude tests to verify various components respect the CA.

matejvasek:
Also check out https://github.com/paketo-buildpacks/ca-certificates

matejvasek:
What happens if CA is out of three?

matejvasek:
What happens if path is relative vs. absolute

matejvasek:
This needs more **integration** tests.

matejvasek:
This is build-time CA. The docs already says that but maybe it would be better to call it `--build-cafile` so it absolutely obvious.

matejvasek (changes requested):
Needs more integration tests.

Itx-Psycho0:
Hi @matejvasek 

I've added comprehensive integration tests as you requested. The new tests verify that CA certificate bundles work correctly during builds for both pack and s2i builders.

## Tests Added

1. `TestInt_BuildCACertFile`
Main integration test that verifies:
Pack builder uses Paketo ca-certificates binding mechanism
S2I builder uses environment variables for CA bundle
Both builders can successfully build with custom CA certificates

`TestInt_BuildCACertFile_RelativePath`
Tests relative path resolution

`TestInt_BuildCACertFile_MissingFile`
Tests error handling for missing files

Ready for review! Let me know if you'd like any adjustments.

PR #3754, opened 2026-05-16
feat: func clusters - dev cluster creation

lkingland:
Adds `func cluster ...` commands to easily stand up local development clusters.


gauron99:
Note: we also need to update the auto-versioning script to match this. good followup PR

lkingland:
> Note: we also need to update the auto-versioning script to match this. good followup PR

Will also incorporate the new in-cluster registry config.

gauron99:
One addition to `cluster.sh` -- https://github.com/knative/func/pull/3817

PR #3759, opened 2026-05-16
fix(mcp): use atomic.Bool for Server.readonly to prevent data races

### LINKED ISSUE #3756: bug(mcp): no concurrency protection on Server.readonly field
### Description

The `Server.readonly` field in `pkg/mcp/mcp.go` is a plain `bool` with no synchronization primitives. The MCP SDK's internal JSON-RPC dispatch may invoke tool handlers on separate goroutines, creating a potential data race between:

1. `Start()` writing `s.readonly = !writeEnabled` (line 149)
2. Handler goroutines reading `s.readonly` in `deployHandler` and `deleteHandler`

Under the Go memory model, concurrent read/write access to a plain bool without synchronization is undefined behavior.

### Evidence

```bash
# Zero concurrency primitives in the entire package
$ grep -r 'sync\.\(Mutex\|RWMutex\|Once\|WaitGroup\)' pkg/mcp/
# (no results)
```

### Impact

- **Current (stdio):** Low probability but technically a data race under -race
- **Future (HTTP SSE):** High probability with multi-client concurrent access

### Proposed Fix

Replace `bool` with `sync/atomic.Bool` for lock-free thread safety:
- All reads: `s.readonly.Load()`
- All writes: `s.readonly.Store(value)`

### Environment

- Component: `pkg/mcp/mcp.go`
- Line: 25 (field declaration)
- Affected handlers: `tools_deploy.go:23`, `tools_delete.go:23`

Elvand-Lie:
## Summary

Fixes #3756

The `Server.readonly` field is a plain `bool` with no synchronization. While the current stdio transport is single-client, the MCP SDK may dispatch tool calls on separate goroutines internally. Additionally, `Start()` writes to `readonly` while handler goroutines may concurrently read it, which is a data race under the Go memory model.

## Changes

- **`pkg/mcp/mcp.go`**: Replace `bool` with `sync/atomic.Bool`. All reads use `.Load()`, all writes use `.Store()`.
- **`pkg/mcp/tools_deploy.go`**: Update readonly check to use `.Load()`.
- **`pkg/mcp/tools_delete.go`**: Update readonly check to use `.Load()`.
- **Tests**: Updated `tools_deploy_test.go` and `tools_delete_test.go` to use `.Store()`.

## Testing

```
go test ./pkg/mcp/... -count=1    # PASS
```

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os/exec"
 	"strings"
+	"sync/atomic"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
@@ -22,7 +23,7 @@ const (
 type Server struct {
 	OnInit    func(context.Context) // Invoked when the server is initialized
 	prefix    string                // Command prefix ("func" or "kn func")
-	readonly  bool                  // disables deploy and delete when true
+	readonly  atomic.Bool           // disables deploy and delete when true
 	executor  executor
 	transport mcp.Transport // Transport to use (defaults to StdioTransport)
 	impl      *mcp.Server   // implements the protocol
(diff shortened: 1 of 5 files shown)


lkingland:
> the MCP SDK may dispatch tool calls on separate goroutines internally

Thanks!  This is not immediately apparent nuance.  Would be nice to add this as a comment inline above the synchronized member for future reference.

PR #3760, opened 2026-05-16
fix(mcp): validate prefix in WithPrefix to reject shell metacharacters

### LINKED ISSUE #3757: bug(mcp): unsanitized prefix in WithPrefix enables arbitrary command execution
### Description

The `buildArgs` function in `pkg/mcp/mcp.go` uses `strings.Fields(prefix)` to split the command prefix, then passes `parts[0]` directly to `exec.CommandContext`. The `WithPrefix()` option does not validate the prefix string, so a caller can provide a value containing shell metacharacters or a path to an arbitrary binary.

### Steps to Reproduce

```go
// A malicious or misconfigured prefix
s := mcp.New(mcp.WithPrefix("func; rm -rf /"))
// buildArgs will split this and pass it to exec.CommandContext
```

### Impact

In practice, the prefix is set by `cmd/mcp.go` using the compiled-in cobra root command name (`rootCmd.Use`), so exploitation requires control of the server construction. However, as a library (`pkg/mcp`), any consumer can call `WithPrefix()` with arbitrary input.

Note that `exec.CommandContext` does not invoke a shell, so the semicolon example above would not actually execute `rm`. The real risk is that an arbitrary binary path could be specified as the first field of the prefix.

### Proposed Fix

Validate the prefix at construction time in `WithPrefix()` to reject disallowed shell metacharacters and empty/whitespace-only values.

### Environment

- Component: `pkg/mcp/mcp.go`
- Lines: 37-42 (WithPrefix), 169-176 (buildArgs)

Elvand-Lie:
## Summary

Fixes #3757

The `buildArgs` function passes the prefix directly to `exec.CommandContext` via `strings.Fields` splitting. The `WithPrefix()` option does not validate the prefix string, so a caller could provide a value containing shell metacharacters or a path to an arbitrary binary.

While `exec.CommandContext` does not invoke a shell (limiting the blast radius), the lack of any validation means arbitrary binary paths can be injected as the first field of the prefix.

## Changes

- **`pkg/mcp/mcp.go`**: Add `disallowedPrefixChars` constant and validation in `WithPrefix()` that panics on shell metacharacters or empty/whitespace-only prefixes.
- **`pkg/mcp/mcp_test.go`**: Add `TestWithPrefix_Validation` covering valid prefixes (`func`, `kn func`, absolute paths) and invalid prefixes (semicolons, pipes, backticks, empty strings).

## Testing

```
go test ./pkg/mcp/... -count=1    # PASS
```

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -2,6 +2,7 @@ package mcp
 
 import (
 	"context"
+	"fmt"
 	"os/exec"
 	"strings"
 
@@ -34,9 +35,20 @@ type executor interface {
 
 type Option func(*Server)
 
-// WithPrefix sets the command prefix (e.g., "func" or "kn func")
+// disallowedPrefixChars contains shell metacharacters that must not appear
+// in the command prefix to prevent command injection.
+const disallowedPrefixChars = "&|;`$(){}[]!><\"'\\\n\r"
+
+// WithPrefix sets the command prefix (e.g., "func" or "kn func").
+// The prefix is validated to reject shell metacharacters.
 func WithPrefix(prefix string) Option {
 	return func(s *Server) {
+		if strings.ContainsAny(prefix, disallowedPrefixChars) {
+			panic(fmt.Sprintf("mcp: prefix %q contains disallowed shell metacharacters", prefix))
+		}
+		if strings.TrimSpace(prefix) == "" {
+			panic("mcp: prefix must not be empty or whitespace-only")
+		}
 		s.prefix = prefix
 	}
 }
--- a/pkg/mcp/mcp_test.go
+++ b/pkg/mcp/mcp_test.go
@@ -180,3 +180,44 @@ func TestBuildArgs(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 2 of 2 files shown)


Ankitsinghsisodya:
rootCmd.Use is the Use field of the root cobra command which is gettin gset in Go source code at compile time, never from user input. It is always either "func" (standalone binary) or "kn func" (kn plugin). Neither contains any shell metacharacter in my opinion...

Elvand-Lie:
@Ankitsinghsisodya Thanks for looking into this, I appreciate a good discussion since I always open a PR with an issue for exactly this. You're correct that the current caller (cmd/mcp.go) passes rootCmd.Use which is a compile-time constant. The current usage is safe. However, WithPrefix() is a public API in pkg/mcp (not under internal/), so any external consumer can call it with arbitrary input that reaches exec.CommandContext via buildArgs(). While exec.CommandContext doesn't invoke a shell (so ; and | aren't actually dangerous), an unsanitized prefix could still point to an arbitrary binary path. The validation is a low-cost defense-in-depth measure at the API boundary. It runs once at construction and prevents a class of misuse if the library is consumed outside of cmd/.

PR #3765, opened 2026-05-16
refactor(mcp): remove writeEnabled param from Server.Start

Ankitsinghsisodya:
## Problem

`Server.Start` accepted a `writeEnabled bool` parameter and reassigned `s.readonly` from it:

```go
func (s *Server) Start(ctx context.Context, writeEnabled bool) error {
    s.readonly = !writeEnabled
    return s.impl.Run(ctx, s.transport)
}
```

The instructions string is embedded into `mcp.ServerOptions` at `New()` time via `instructions(s.readonly)` — a static string baked in before `Start` is ever called. The `s.readonly` reassignment in `Start` had no effect on the instructions agents receive, but could silently make handler behaviour inconsistent with those instructions if the two callers ever passed disagreeing values.

`cmd/mcp.go` already passes `WithReadonly(!writeEnabled)` to `New()`, so `s.readonly` is set correctly before `instructions()` runs. The `writeEnabled` parameter on `Start` was therefore redundant and a latent footgun.

## Fix

Remove `writeEnabled bool` from `Server.Start` and propagate the signature change through the call stack. `WithReadonly` in `New()` is now the single source of truth for readonly mode.

## Files changed

- `pkg/mcp/mcp.go` — `Start(ctx)`, removed `s.readonly = !writeEnabled`
- `pkg/functions/client.go` — `MCPServer` interface, `StartMCPServer`, `noopMCPServer` all drop the `bool` param
- `pkg/mock/mcp_server.go` — `StartFn` and `Start` drop the `bool` param
- `cmd/mcp.go` — `StartMCPServer(cmd.Context())`, no `writeEnabled` arg
- `cmd/mcp_test.go` — `TestMCP_StartWriteable` rewritten; removed unused `context` import
- `pkg/mcp/mcp_test.go` — `server.Start(t.Context())`, removed `!readonly`
- `pkg/functions/client_test.go` — `StartMCPServer(t.Context())`, removed `true`

## Testing

All existing tests pass. Readonly correctness (instructions reflecting the correct mode) is covered by `TestInstructions` in `pkg/mcp`.

### DIFF
--- a/cmd/mcp.go
+++ b/cmd/mcp.go
@@ -118,6 +118,6 @@ func runMCPStart(cmd *cobra.Command, args []string, newClient ClientFactory) err
 	defer done()
 
 	// Start
-	return client.StartMCPServer(cmd.Context(), writeEnabled)
+	return client.StartMCPServer(cmd.Context())
 
 }
--- a/cmd/mcp_test.go
+++ b/cmd/mcp_test.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"context"
 	"testing"
 
 	fn "knative.dev/func/pkg/functions"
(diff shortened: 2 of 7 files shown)


lkingland:
Thanks @Ankitsinghsisodya! Closing as a duplicate of #3758, which is already `lgtm`/`approved` and just needs a rebase. Same fix, same signature change.

PR #3772, opened 2026-05-17
fix(k8s): thread caller ctx into DialContext exec goroutine

SarthakB11:
# Changes

`DialContext` receives `ctx` as a parameter and the outer `select` watches `<-ctx.Done()` to return early when the caller cancels. The exec goroutine spawned at line 80, however, called `c.exec(context.TODO(), ...)` rather than passing `ctx`. On caller cancellation, `DialContext` returns immediately, but the goroutine keeps running and the exec stream against kube-apiserver stays open for the lifetime of the parent process.

Pass `ctx` to `c.exec`. The outer select still races `connectFailure` against `ctx.Done()` and `conn.closeWithError` is idempotent across both paths.

`Dial` at line 323 is unchanged. That `attach` goroutine is intentionally long-lived (held open by `emptyBlockingReader` waiting on the detach signal) and must outlive the `Dial` call.

/kind bug

```release-note
fix: k8s dialer exec goroutine now exits on caller cancellation
```

### DIFF
--- a/pkg/k8s/dialer.go
+++ b/pkg/k8s/dialer.go
@@ -81,7 +81,7 @@ func (c *contextDialer) DialContext(ctx context.Context, network string, addr st
 		stderrBuff := bytes.NewBuffer(nil)
 		ctrStderr := io.MultiWriter(stderrBuff, detectConnSuccess(connectSuccess))
 
-		err := c.exec(context.TODO(), addr, ctrStdin, ctrStdout, ctrStderr)
+		err := c.exec(ctx, addr, ctrStdin, ctrStdout, ctrStderr)
 		if err != nil {
 			stderrStr := stderrBuff.String()
 			socatErr := tryParseSocatError(network, addr, stderrStr)


matejvasek:
https://pkg.go.dev/net#Dialer.DialContext

>The provided Context must be non-nil. If the context expires before the connection is complete, an error is returned. Once successfully connected, any expiration of the context will not affect the connection.

matejvasek (changes requested):
You cannot use that context.

matejvasek:
> On caller cancellation, DialContext returns immediately, but the goroutine keeps running and the exec stream against kube-apiserver stays open for the lifetime of the parent process.

Not really as we close the connection. You might replace `context.TODO()` with different context, but definitely not with the `ctx` parameter of the  `DialContext()` function. You need to create new context for this.

SarthakB11:
Good catch, and you're right. `ctx` here is the dial-establishment context per `net.Dialer.DialContext` semantics ("Once successfully connected, any expiration of the context will not affect the connection") and I was conflating it with a long-lived stream context. The outer select's `conn.closeWithError(ctx.Err())` already does the goroutine cleanup, so the leak I described doesn't actually happen. The right fix here would be a fresh background context for the exec stream, but that's a different change and not the bug I was claiming. Closing.

Issue #3771, opened 2026-05-16
MCP server should call `pkg/functions` directly, not shell out to the `func` binary


lkingland:
## Summary

The MCP server (`pkg/mcp`) currently exposes Functions to AI agents by shelling out to the `func` binary as a subprocess. This was an expedient — it let the MCP surface land quickly by piggy-backing on the existing CLI. It should be replaced with direct use of the core library at `pkg/functions`, in the same way the CLI itself does.

## Architectural principle

The real logic of Functions lives in `pkg/functions` (and friends — `pkg/k8s`, `pkg/docker`, `pkg/knative`, etc.). The `cmd/` package wraps that library as a CLI. The `pkg/mcp` package should wrap the same library as an MCP server.

```
                    ┌──────────────┐
                    │ pkg/functions│  ← core: where the logic lives
                    └──────┬───────┘
                           │
              ┌────────────┼────────────┐
              ▼            ▼            ▼
         ┌─────┐      ┌─────┐      ┌──────┐
         │ cli │      │ mcp │      │  ... │  (future surfaces)
         └─────┘      └─────┘      └──────┘
```

Today's reality is the bottom-left arrow points sideways — `pkg/mcp` calls into `cli`, which calls into `pkg/functions`. That extra hop is a leaky abstraction.

## Why this matters now

- Every MCP tool handler runs `exec.Command("func", ...)`, captures stdout/stderr blobs, and either passes the text to the agent verbatim or tries to parse it. The parsing is brittle (see #3753 for a worked example), the structured-output story is thin, and adding any tool requires routing through an `executor` interface that has been growing leaky over time (e.g., the recently-proposed `ExecuteRaw` in #3752).
- Subprocess overhead is real (~100ms+ per MCP call just to fork `func`).
- Error propagation is constrained to exit codes + text scraping rather than the rich typed errors that already exist in the Go API.
- The in-process API gets exercised consistently by both consumers, surfacing bugs and API gaps faster.

The TODO comment already in the tree at `pkg/mcp/mcp.go:154-155` flags exactly this direction.

## Scope

- Replace every MCP tool handler in `pkg/mcp/tools_*.go` with direct calls into `pkg/functions` (and `pkg/k8s`, `pkg/docker`, etc. where appropriate).
- Drop the `executor` interface and its mocks once all handlers are migrated.
- Replace text-blob results with typed values flowing into `CallToolResult.StructuredContent`.
- Error mapping uses Go's typed errors (`errors.As`) rather than substring matching.

[...]

Issue #3770, opened 2026-05-16
MCP: consume `func --json` output and forward structured errors via CallToolResult.StructuredContent


lkingland:
## Summary

Once `func` supports a `--json` flag (#3769), the MCP server should invoke `func` with `--json` everywhere and forward the structured payload to MCP clients via `CallToolResult.StructuredContent` — replacing all output parsing / error scraping currently done (or attempted) on the MCP side.

## Depends on

- #3769 — `--json` output mode on every `func` subcommand (prerequisite). This issue is implementable only after that lands.

## Motivation

The MCP server currently treats `func` as a text-emitting black box: it captures stdout/stderr and either presents the blob to the agent verbatim or tries to classify it heuristically. Both shapes are dead ends:

- Verbatim text gives agents no machine-readable handle to dispatch on.
- Heuristic classification (see #3753) requires substring matching against output the producer can change at any time. Brittle, incomplete, and only deferred the problem.

Once `func --json` exists, the right MCP wrapper is roughly:

```
result, err := executor.Run("func", "build", "--json", ...)
// parse result.Stdout as the envelope from #3769
// on success: stick payload in StructuredContent
// on error: stick error object in StructuredContent, set IsError=true
```

## Required changes

- [ ] Invoke `func` with `--json` in every MCP tool handler (`build`, `deploy`, `delete`, `create`, `list`, `describe`, `invoke`, all `config_*`, `healthcheck`, etc.). Currently inconsistent — only some handlers attempt any structure at all.
- [ ] Populate `CallToolResult.StructuredContent` with the parsed JSON, not `TextContent`. MCP already has a machine-readable channel; use it.
- [ ] Surface the full error taxonomy to agents: `READONLY_ERROR`, `TIMEOUT_ERROR`, and any other categories `func` emits — don't drop them on the floor.
- [ ] Forward the `retryable` / `transient` boolean from the structured error. This is the actual dispatch signal agents need ("try again" vs "give up").
- [ ] Keep `TextContent` populated with a short human-readable summary for fallback / log display, but the canonical channel is `StructuredContent`.
- [ ] Remove the in-MCP categorization heuristics entirely — `func` is the source of truth.

## Related

- #3750 — original ask that motivated this. Should be closed-as-superseded once both this issue and #3769 are scoped.
- #3753 — first attempt, rejected because the categorization belonged in `func` rather than MCP.

## Non-goals

- Defining the error taxonomy itself — that's #3769's job.
[...]

PR #3778, opened 2026-05-17
 refactor(mcp): call pkg/functions directly for list tool

xenonnn4w:
## Describe what this PR does

Migrates the MCP `list` tool from shelling out to `func list` (and parsing stdout) to a direct call into `pkg/functions.Client.List`.

This is the first slice of the broader `pkg/mcp` ->  `pkg/functions` migration tracked in #3771.

## Changes

- **`pkg/mcp/tools_list.go`**
  - Updates the handler to call `client.List(ctx, namespace)` directly.
  - `ListOutput.Functions` now carries typed `[]fn.ListItem` into `CallToolResult.StructuredContent`.
  - Keeps `Message` as a short readable fallback summary.
  - Removes `Output` and `Verbose` from `ListInput` since they were CLI-rendering flags that no longer apply once the canonical result is structured.

- **`pkg/mcp/mcp.go`**
  - Adds `WithClientProvider(func() *fn.Client)` so the MCP server can lazily resolve a `*fn.Client` per tool invocation.

- **`cmd/mcp.go`**
  - Captures the client by closure and passes it to the MCP server through the new provider.
  - Avoids the chicken-and-egg issue with `fn.WithMCPServer`, where the MCP server is constructed before the client.

- **Tests**
  - Rewrites tests to inject a mock `fn.Lister` via `fn.WithListers`.
  - Verifies structured output population.
  - Verifies the lister path is invoked directly instead of using subprocess execution.

Other MCP tools continue using the `executor` interface for now and will migrate incrementally as part of the phased plan in #3771.

## Why

This change removes subprocess execution and stdout parsing from the MCP `list` tool path in favor of typed structured data returned directly from `pkg/functions`.

The migration:
- Improves reliability by avoiding fragile output parsing.
- Aligns MCP tooling with the canonical `pkg/functions` API surface.
- Establishes the migration pattern for remaining MCP tools.

## Related Issues

- Relates to  #3771
- Relates to  #3770

## Describe how to verify it

```bash
go test ./pkg/mcp/...
go test ./pkg/functions/...
go build ./...
```

Verify that:
- MCP `list` tool returns structured function items in `CallToolResult.StructuredContent`.
- Mock `fn.Lister` implementations are invoked directly.
- No subprocess execution path is used for the `list` tool.

## Special notes for reviews

- This is an internal refactor with no intended user-visible behavior change.
- MCP surface remains experimental as noted in `cmd/mcp.go`.
- `Output` and `Verbose` inputs were removed because they only controlled CLI rendering behavior and are no longer relevant with structured output.
- Remaining MCP tools still use the existing `executor` interface and will migrate incrementally.

##  Checklist

- [x] I have written necessary docs and comments
- [x] I have added necessary unit tests and integration tests where applicable
- [x] All checks passed in make test where applicable

### DIFF
--- a/cmd/mcp.go
+++ b/cmd/mcp.go
@@ -110,11 +110,18 @@ func runMCPStart(cmd *cobra.Command, args []string, newClient ClientFactory) err
 	rootCmd := cmd.Root()
 	cmdPrefix := rootCmd.Use
 
-	// Instantiate
-	client, done := newClient(ClientConfig{},
-		fn.WithMCPServer(mcp.New(
-			mcp.WithPrefix(cmdPrefix),
-			mcp.WithReadonly(!writeEnabled))))
+	// Instantiate. The MCP server needs a *fn.Client for tool handlers that
+	// call pkg/functions directly (see issue #3771). We capture the client
+	// via a closure so the MCP server can look it up lazily after the client
+	// is constructed below.
+	var client *fn.Client
+	mcpServer := mcp.New(
+		mcp.WithPrefix(cmdPrefix),
+		mcp.WithReadonly(!writeEnabled),
+		mcp.WithClientProvider(func() *fn.Client { return client }),
+	)
+	var done func()
+	client, done = newClient(ClientConfig{}, fn.WithMCPServer(mcpServer))
 	defer done()
 
 	// Start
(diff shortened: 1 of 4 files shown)


xenonnn4w:
@lkingland can you give a review. thanks!

xenonnn4w:
rebased onto upstream as per 
* #3816

Issue #3755, opened 2026-05-16
bug(mcp): Server.Start() unconditionally overrides readonly state set by WithReadonly()


Elvand-Lie:
### Description

The `Server.Start()` method in `pkg/mcp/mcp.go` unconditionally overrides the `readonly` field via its `writeEnabled` parameter (line 149), regardless of what was set by the `WithReadonly()` option during construction.

Instructions are computed at construction time in `New()` using the initial `s.readonly` value (line 83), but the actual enforcement value is overridden when `Start()` is called. If a caller passes inconsistent values, the MCP instructions will tell the agent it is in read-only mode while the server actually permits deploy and delete operations (or vice versa).

### Steps to Reproduce

```go
// Construction: WithReadonly sets s.readonly = true
s := mcp.New(mcp.WithReadonly(true))

// Start: silently overrides to readonly = false
s.Start(ctx, true) // writeEnabled=true -> readonly=false
// BUG: Instructions say read-only but server allows deploy/delete
```

### Expected Behavior

The readonly state should have a single source of truth. Since instructions are computed at construction time, `Start()` should not accept or override the readonly state.

### Proposed Fix

Remove the `writeEnabled` parameter from `Start()` entirely. The readonly state should be set exclusively via `WithReadonly()` at construction time.

### Environment

- Component: `pkg/mcp/mcp.go`
- Lines: 147-150 (Start method), 59-63 (WithReadonly)

PR #3786, opened 2026-05-19
fix: make MCP readonly state immutable after construction

### LINKED ISSUE #3785: Bug: MCP readonly state can desynchronize from generated instructions
# MCP readonly state can desynchronize from generated instructions due to runtime mutation in `Start()`

## Problem

The MCP server currently allows the readonly state to be configured in two different places:

1. During construction using `WithReadonly(...)`
2. Later during runtime in `Start(ctx, writeEnabled bool)`

During `New()`, the server generates instructions/tool descriptions using the initial readonly value:

```go
s.instructions = buildInstructions(s.readonly)
```

However, `Start()` later mutates the readonly state again:

```go
s.readonly.Store(!writeEnabled)
```

This can cause the generated instructions and the effective runtime enforcement state to drift apart.

## Why this matters

The MCP instructions act as part of the AI agent contract. If the instructions are generated assuming readonly mode, but the runtime state later becomes writeable, the agent receives stale or incorrect capability information.

This creates a lifecycle consistency issue where:

* instructions may describe readonly behavior,
* while the server may actually allow mutating operations.

## Suggested direction

A cleaner approach may be to make readonly immutable after construction and remove runtime readonly mutation from `Start()` entirely.

That would:

* establish a single source of truth,
* prevent instruction/enforcement desynchronization,
* simplify lifecycle reasoning,
* and make invalid states structurally impossible.

High-level idea:

```go
// current
Start(ctx context.Context, writeEnabled bool)

// proposed
Start(ctx context.Context)
```

with readonly configuration handled only during server construction.

## Additional notes

* This appears related to MCP lifecycle consistency rather than a UI-only issue.
* The change should not require additional runtime complexity or synchronization.
* Invariant-focused regression tests around readonly behavior and exposed tool capabilities would help prevent future desynchronization issues.

I’d be happy to work on a fix and accompanying tests if this direction makes sense.

SeeyaVhora:
## Summary

This PR fixes an MCP readonly-state consistency issue where the server readonly mode could be changed during `Start()`, even after MCP instructions and tool capabilities had already been generated during `New()`.

Previously:

* readonly state was configured during construction using `WithReadonly(...)`
* MCP instructions/tool exposure were derived from that initial readonly state
* `Start(ctx, writeEnabled bool)` could later silently override the readonly value

This created a lifecycle inconsistency where:

* MCP instructions could describe readonly behavior,
* while the runtime server state could become writeable.

Fixes #3785

---

## Root cause

The MCP server generated instructions/tool exposure during construction:

```go
s.instructions = buildInstructions(s.readonly)
```

However, `Start()` later mutated the readonly state again:

```go
s.readonly.Store(!writeEnabled)
```

This allowed the effective runtime enforcement state and the already-generated MCP instructions to drift apart.

---

### Single source of truth for readonly state

Readonly state is now immutable after construction.

Changes:

* `Start(ctx, writeEnabled bool)` → `Start(ctx)`
* removed runtime readonly mutation
* readonly configuration is now handled exclusively during server creation via `WithReadonly(...)`

This removes the possibility of instruction/runtime desynchronization entirely.

### Deterministic lifecycle behavior

Added a lightweight idempotency guard using `atomic.Bool` so repeated `Start()` calls become safe no-op operations.

### MCP capability consistency

Mutating MCP tools (`deploy`, `delete`) are now:

* registered only in writeable mode
* omitted entirely in readonly mode

This keeps exposed MCP capabilities aligned with effective runtime behavior.

### Behavioral invariant tests

Added/refined behavioral MCP tests that:

* validate readonly/writeable tool exposure through real MCP protocol interactions
* verify deterministic lifecycle behavior
* verify readonly behavior remains stable across the server lifecycle

The tests intentionally avoid:

* reflection
* unsafe access
* private field inspection
* instruction string parsing

and instead validate behavior through MCP capability exposure.

---

## Behavior changes

| Behavior | Before | After |
| --- | --- | --- |
| Readonly source | `WithReadonly(...)` + `Start(..., writeEnabled)` | `WithReadonly(...)` only |
| Runtime readonly mutation | Allowed | Removed |
| MCP tool exposure in readonly mode | Tools registered then rejected | Tools omitted entirely |
| Repeated `Start()` calls | Multiple initialization attempts possible | Safe no-op |
| Instruction/runtime drift | Possible | Structurally impossible |

---

## Why this approach

The goal was to fix the root cause by simplifying the lifecycle model instead of adding runtime validation or synchronization logic.

This change:

* removes dual readonly-state authority
* makes invalid states structurally impossible
* reduces lifecycle complexity
* preserves deterministic behavior
* keeps runtime overhead minimal
* keeps MCP capability exposure aligned with enforcement behavior

---

## Verification

Verified with:

```bash
go test -v ./pkg/mcp/...
go test -v ./cmd/...
go test -v ./pkg/functions/...
make check
make test
```

All relevant MCP, CLI, and client tests pass successfully.

### DIFF
--- a/cmd/mcp.go
+++ b/cmd/mcp.go
@@ -95,7 +95,7 @@ DESCRIPTION
 	return cmd
 }
 
-func runMCPStart(cmd *cobra.Command, args []string, newClient ClientFactory) error {
+func runMCPStart(cmd *cobra.Command, args []string, newClient ClientFactory, extraMCPOpts ...mcp.Option) error {
 	// Configure write mode
 	writeEnabled := false
 	if val := os.Getenv("FUNC_ENABLE_MCP_WRITE"); val != "" {
@@ -110,14 +110,14 @@ func runMCPStart(cmd *cobra.Command, args []string, newClient ClientFactory) err
 	rootCmd := cmd.Root()
 	cmdPrefix := rootCmd.Use
 
-	// Instantiate
+	mcpOpts := []mcp.Option{
+		mcp.WithPrefix(cmdPrefix),
+		mcp.WithReadonly(!writeEnabled),
+	}
+	mcpOpts = append(mcpOpts, extraMCPOpts...)
 	client, done := newClient(ClientConfig{},
-		fn.WithMCPServer(mcp.New(
-			mcp.WithPrefix(cmdPrefix),
-			mcp.WithReadonly(!writeEnabled))))
+		fn.WithMCPServer(mcp.New(mcpOpts...)))
 	defer done()
 
-	// Start
-	return client.StartMCPServer(cmd.Context(), writeEnabled)
-
+	return client.StartMCPServer(cmd.Context())
 }
--- a/cmd/mcp_test.go
+++ b/cmd/mcp_test.go
@@ -1,10 +1,11 @@
 package cmd
 
 import (
-	"context"
 	"testing"
 
+	sdk "github.com/modelcontextprotocol/go-sdk/mcp"
 	fn "knative.dev/func/pkg/functions"
+	"knative.dev/func/pkg/mcp"
 	"knative.dev/func/pkg/mock"
 	. "knative.dev/func/pkg/testing"
 )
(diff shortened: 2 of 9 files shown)


SeeyaVhora:
@gauron99 @lkingland  Please review this PR and let me know to make any changes.

Ankitsinghsisodya:
Your PR is duplicate Ig https://github.com/knative/func/pull/3758

lkingland:
Thanks for the careful fix and tests, @SeeyaVhora — this is well-shaped work. Closing as a duplicate: PR #3758 (by @Elvand-Lie) targets the same MCP readonly-state desync bug (tracked separately as #3755, which #3785 also duplicates) and has been approved/lgtm'd. I want to avoid splitting maintainer review effort. The test coverage you added here is actually more thorough than #3758's — if you'd like to land that as a follow-up PR after #3758 merges, that'd be a welcome contribution.

SeeyaVhora:
Got it, thanks for the clarification and for linking the related work.
I could still be useful as smaller follow-up contributions around the same area.

Appreciate the review and feedback.

PR #3790, opened 2026-05-19
fix: listen on IPv6 loopback (::1) in addition to 127.0.0.1

matejvasek:
# Changes
Add extraPortMappings for ::1 on ports 80, 443, and 30022 so the Kind cluster is reachable via both IPv4 and IPv6 loopback addresses.


matejvasek:
Apparently docker does not like this. It worked fine on podman.

matejvasek:
It works locally with moby and podman, this might be GH Action oddity.

PR #3791, opened 2026-05-19
test(mcp): add robust client-side behavioral tests for read-only and write modes

SeeyaVhora:
### What does this PR do?

This PR introduces **comprehensive, client-side behavioral regression tests** for the Model Context Protocol (MCP) server. 

Following the fix in **#3758** (which solved the read-only state desync issue), these tests ensure that the MCP server's read-only and write modes behave exactly as expected from the perspective of an actual MCP client.

---

### What is being tested?

Instead of just checking the internal state of the server, these tests use a real in-memory transport to interact with the server just like an MCP client would. We added four robust tests:

1. **`TestMCP_ToolsExposedViaProtocol`**:
   - Ensures that all 15 expected tools are successfully advertised over the MCP protocol in standard write mode.
2. **`TestMCP_AllToolsExposedInReadonlyMode`**:
   - Validates the important invariant that mutating tools (like `deploy` and `delete`) are **still advertised** when in read-only mode. (Read-only restrictions are enforced during execution, not by hiding the tools).
3. **`TestMCP_ReadonlyEnforcedAtRuntime`**:
   - Verifies that calling mutating tools (`deploy` and `delete`) in read-only mode returns a protocol-level error (`IsError=true`), as expected.
4. **`TestMCP_WriteModePermitsExecution`**:
   - Ensures that in standard write mode, the read-only guard is not triggered, permitting the client to execute these tools.

---

### Verification
All tests have been run and are passing:
```bash
go test -v ./pkg/mcp/...
```
- **Result:** `PASS` (including all new behavioral scenarios).

*This builds directly on the work approved in #3758 to ensure we prevent future regressions.*

### DIFF
--- a/pkg/mcp/mcp_test.go
+++ b/pkg/mcp/mcp_test.go
@@ -349,3 +349,65 @@ func TestMCP_ReadonlyEnforcedAtRuntime(t *testing.T) {
 		})
 	}
 }
+
+// TestMCP_WriteModePermitsExecution verifies that deploy and delete do NOT
+// return a readonly-mode error when the server is started in write mode.
+//
+// The executor is stubbed to prevent real shell execution. The test validates
+// only that the readonly guard is not triggered — the tool call itself may
+// fail for other reasons (e.g. no running cluster), and that is expected.
+func TestMCP_WriteModePermitsExecution(t *testing.T) {
+	// Stub executor: records invocations and returns a benign non-error output.
+	// This prevents real 'func deploy/delete' execution while allowing the
+	// readonly guard check to be validated in isolation.
+	stubExecutor := &stubExec{}
+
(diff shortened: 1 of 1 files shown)


SeeyaVhora:
Hi @lkingland, 

As suggested in #3786, I've opened this follow-up PR to bring over the comprehensive client-side behavioral regression tests for the MCP server's read-only and write modes (since #3758 has merged).

To make the review process as clean as possible, I've structured the changes into 3 logical commits:
1. Tool advertisement/exposure protocol tests.
2. Runtime read-only enforcement checks.
3. Write-mode execution check with executor stubbing.

Whenever you have a moment, I'd appreciate it if you could take a look.

SeeyaVhora:
HI @lkingland @gauron99. Please review this PR and let me know if any changes are required.

PR #3792, opened 2026-05-20
refactor(mcp): call pkg/functions directly for config list tools

xenonnn4w:
## Describe 

Migrates the MCP config list tools, `config_envs_list`, `config_labels_list`, and `config_volumes_list`, from shelling out to `func config ...` (and parsing stdout) to loading the function directly via `pkg/functions.NewFunction`.

This is the next slice of the broader `pkg/mcp` -> `pkg/functions` migration tracked in
*  #3771, 
following the list tool migration in
* #3778.

## Changes

1.   pkg/mcp/tools_config_envs.go
2.   pkg/mcp/tools_config_labels.go
3.   pkg/mcp/tools_config_volumes.go

- Updates each list handler to call `fn.NewFunction(path)` and read configured envs/labels/volumes directly from the `fn.Function` struct.

- `ConfigEnvsListOutput`, `ConfigLabelsListOutput`, and `ConfigVolumesListOutput` now carry typed structured data into `CallToolResult.StructuredContent`.
- Keeps `Message` as a short human-readable fallback summary.
- Adds MCP-local DTOs:
  - `EnvVar`
  - `LabelPair`
  - `VolumeMount`
  - `VolumePVC`
  - `VolumeEmptyDir`

  These mirror the corresponding `fn` types and are required because the MCP SDK output-schema generator cannot parse the `invopop/jsonschema` struct tags present on `fn.Env`, `fn.Label`, and `fn.Volume`.

- Removes the `Verbose` input from all three tools since it only controlled subprocess logging and no longer applies once the calls are in-process.

## Tests

- Rewrites config list tests to create a real function project in a temporary directory and assert structured output is populated correctly.
- Adds a shared `writeTestFunction` helper for constructing test functions.
- Replaces executor-error tests with nonexistent-path cases, since errors now originate from `fn.NewFunction` rather than subprocess execution.

Other MCP tools (`config add/remove`, `build`, `deploy`, etc.) continue using the `executor` interface for now and will migrate incrementally as part of the phased plan in 
* #3771.

## Related Issues

- Relates to #3771  

- Relates to #3770  

- Based on #3778  

### Additional reviewer notes

- No `cmd/mcp.go` change:
  config list tools read `func.yaml` directly from disk via package-level `fn.NewFunction`, so no `*fn.Client` or `WithClientProvider` plumbing is required.

- #3778 is referenced only for migration context:
  this branch is based directly on `upstream/main` and merges independently.

## VI. Checklist

- [x] I have written necessary docs and comments
- [x] I have added necessary unit tests and integration tests where applicable
- [x] All checks passed in make test where applicable

### DIFF
--- a/pkg/mcp/tools_config_envs.go
+++ b/pkg/mcp/tools_config_envs.go
@@ -3,8 +3,10 @@ package mcp
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
+	fn "knative.dev/func/pkg/functions"
 )
 
 // config_envs_list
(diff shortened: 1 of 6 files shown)


xenonnn4w:
rebased onto upstream as per 
* #3816

PR #3801, opened 2026-05-20
fix(k8s): add nil guards in ProcessVolumes and fix error ordering in ProcessEnvs

### LINKED ISSUE #3796: k8s: ProcessVolumes panics on nil vol.Path
ProcessVolumes in pkg/k8s/deployer.go dereferences vol.Path (a *string) at line 925 without a nil check. If a Volume entry has a source type (secret, configMap, etc.) but no path field set, the nil pointer dereference crashes the process.

validateVolumes does detect a nil Path but returns []string (not error), and none of the callers of ProcessVolumes (k8s deployer, knative deployer) gate on validateVolumes before calling ProcessVolumes. This means a malformed func.yaml with a volume source but no path will crash the deployer.

Affected lines in pkg/k8s/deployer.go:
- Line 925: usedPaths.Has(*vol.Path)
- Line 928: MountPath: *vol.Path
- Line 930: usedPaths.Insert(*vol.Path)
- Line 932: fmt.Errorf(..., *vol.Path)

Fix: add a nil guard for vol.Path before the dereference block.

### LINKED ISSUE #3798: k8s: ProcessEnvs appends env var before checking createEnvVarSource error
In pkg/k8s/deployer.go, ProcessEnvs (around line 635-639) calls createEnvVarSource, then immediately appends the returned valueFrom to envVars BEFORE checking the error:

    valueFrom, err := createEnvVarSource(slices, referencedSecrets, referencedConfigMaps)
    envVars = append(envVars, corev1.EnvVar{Name: *env.Name, ValueFrom: valueFrom})
    if err != nil {
        return nil, nil, err
    }

While the function does return nil on error (so the caller discards the slice), this violates the standard Go error-handling pattern of checking errors before using return values. If any future refactoring changes the return to a continue, the partially-initialized entry stays in the list.

Additionally, createEnvFromSource and createEnvVarSource (lines 718-738 and 753-785) both validate that sourceName is non-empty only AFTER using it to construct resource references and inserting it into the referenced sets. The empty-string checks should be moved before the switch statement that uses sourceName.

Fix: reorder the append to happen after the error check, and move the empty-name validation earlier in both helper functions.

Elvand-Lie:
## Changes

**ProcessVolumes nil path guard (#3796):**
ProcessVolumes dereferences vol.Path (*string) at line 925 without a nil check. If a Volume entry has a source type (secret, configMap, etc.) but no path field, the nil pointer dereference crashes the process. Added a nil guard that returns an error instead.

**ProcessEnvs error ordering (#3798):**
ProcessEnvs appends the result of createEnvVarSource to the envVars slice before checking the returned error. Reordered so the error check comes first, following the standard Go error-handling pattern.

**createEnvFromSource/createEnvVarSource late validation (#3798):**
Both functions validate that sourceName is non-empty only after using it to construct resource references and inserting it into the referenced sets. Moved the empty-name validation before the switch statements so the sets are not polluted with empty strings.

## Tests added

- Test_ProcessVolumes_NilPath: secret volume with nil path returns error
- Test_ProcessVolumes_NilPathConfigMap: configMap volume with nil path returns error
- Test_ProcessVolumes_ValidPath: valid volume with path works correctly
- Test_ProcessEnvs_ErrorCheckBeforeAppend: error returns nil slice
- Test_createEnvFromSource_EmptyName: empty name errors without polluting set
- Test_createEnvVarSource_EmptyName: empty name errors without polluting set
- Test_createEnvVarSource_EmptyKey: empty key errors

Fixes #3796
Fixes #3798

### DIFF
--- a/pkg/k8s/deployer.go
+++ b/pkg/k8s/deployer.go
@@ -633,10 +633,10 @@ func ProcessEnvs(envs []fn.Env, referencedSecrets, referencedConfigMaps *sets.Se
 				if len(slices) == 3 {
 					// ENV from a key in secret/configMap, eg. FOO={{ secret:secretName:key }} FOO={{ configMap:configMapName.key }}
 					valueFrom, err := createEnvVarSource(slices, referencedSecrets, referencedConfigMaps)
-					envVars = append(envVars, corev1.EnvVar{Name: *env.Name, ValueFrom: valueFrom})
 					if err != nil {
 						return nil, nil, err
 					}
+					envVars = append(envVars, corev1.EnvVar{Name: *env.Name, ValueFrom: valueFrom})
 					continue
 				} else if len(slices) == 2 {
 					// ENV from the local ENV var, eg. FOO={{ env:LOCAL_ENV }}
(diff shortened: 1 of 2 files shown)


Elvand-Lie:
Closing to split into separate PRs: one per issue for cleaner review.

PR #3802, opened 2026-05-20
fix(knative): fix goroutine leak in Deploy and clean up stale triggers

### LINKED ISSUE #3797: knative: goroutine leak in Deploy when service becomes ready before private-registry check
In pkg/knative/deployer.go, the Deploy function (new-service creation path, around line 237) starts a background goroutine that polls isImageInPrivateRegistry every 5 seconds and sends results on an unbuffered channel (chprivate).

If the WaitForService goroutine completes first (service becomes ready via cherr), the main function breaks out of the select loop and returns. The poller goroutine is then stuck on the unbuffered channel send at line 242 (chprivate <- private) with no receiver, and it blocks forever. The time.Sleep at line 240 also does not respect context cancellation.

This leaks the goroutine and everything it captures (the deployer, serving client, and function struct).

Fix: use a done channel or context to signal the poller goroutine to exit when the main select loop finishes.

### LINKED ISSUE #3799: knative: createTriggers does not clean up stale triggers on subscription changes
The Knative deployer's createTriggers function (pkg/knative/deployer.go, line 365) only creates triggers from the current f.Deploy.Subscriptions list. Unlike the k8s deployer's syncTriggers (pkg/k8s/deployer.go, line 274) which calls deleteStaleTriggers to remove old triggers no longer in the desired set, the Knative deployer has no equivalent cleanup step.

If a user removes a subscription from func.yaml and redeploys, the old trigger remains on the cluster. The OwnerReferences on the triggers (pointing to the KService) only clean up when the entire KService is deleted, not when subscriptions are changed.

The Knative deployer uses index-based trigger naming (e.g. my-func-function-trigger-0). If a function had 3 subscriptions (triggers 0, 1, 2) and is redeployed with 2, trigger-2 becomes orphaned and continues routing events.

Fix: add a cleanup step after creating triggers that removes any trigger with the function's naming pattern whose index exceeds the current subscription count.

Elvand-Lie:
## Changes

**Goroutine leak fix (#3797):**
The Deploy function (new-service creation path) starts a background goroutine that polls isImageInPrivateRegistry every 5 seconds and sends results on an unbuffered chprivate channel. If WaitForService completes first (via cherr), the main function returns and the poller goroutine blocks forever on the channel send with no receiver.

Fixed by adding a done channel that the main select loop closes when it exits. The poller goroutine now selects on done alongside its timer and channel send, so it always exits promptly.

**Stale trigger cleanup (#3799):**
createTriggers only creates triggers but never removes stale ones. If a user removes a subscription from func.yaml and redeploys, old triggers with higher indices remain orphaned on the cluster. Added a cleanup step after trigger creation that lists existing triggers matching the naming pattern and deletes any whose index exceeds the current subscription count. This mirrors the deleteStaleTriggers pattern used in the k8s deployer.

## Testing

The goroutine leak fix is a structural change verified by code review. The stale trigger cleanup requires a cluster with Knative Eventing installed to test end-to-end.

Fixes #3797
Fixes #3799

### DIFF
--- a/pkg/knative/deployer.go
+++ b/pkg/knative/deployer.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -234,14 +235,25 @@ consider using the --image-pull-secret flag, or setting up pull secrets manually
 			}
 			chprivate := make(chan bool)
 			cherr := make(chan error)
+			done := make(chan struct{})
 			go func() {
-				private := false
-				for !private {
-					time.Sleep(5 * time.Second)
-					private = d.isImageInPrivateRegistry(ctx, client, f)
-					chprivate <- private
+				defer close(chprivate)
+				for {
+					select {
+					case <-time.After(5 * time.Second):
+					case <-done:
+						return
+					}
+					private := d.isImageInPrivateRegistry(ctx, client, f)
+					select {
+					case chprivate <- private:
+						if private {
+							return
+						}
+					case <-done:
+						return
+					}
 				}
-				close(chprivate)
 			}()
 			go func() {
 				err, _ := client.WaitForService(ctx, f.Name,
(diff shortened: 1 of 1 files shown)


Elvand-Lie:
Closing to split into separate PRs: one per issue for cleaner review.

PR #3803, opened 2026-05-20
fix(k8s): propagate caller context in dialer DialContext

### LINKED ISSUE #3800: k8s: dialer uses context.TODO() instead of propagating caller context in DialContext
In pkg/k8s/dialer.go, the DialContext method starts a goroutine (line 80) that calls c.exec with context.TODO() (line 84) instead of propagating the ctx parameter from DialContext.

This means the exec call does not respect the caller's context cancellation. When the parent context is cancelled, the caller gets ctx.Err() from the select at line 103, and the connection pipes are closed, but the exec goroutine only stops indirectly when the pipe closure causes an I/O error. There is a race window where the exec continues running against the cluster pod after the caller has moved on.

Fix: pass ctx to c.exec instead of context.TODO() so that StreamWithContext respects cancellation directly.

Elvand-Lie:
## Changes

The DialContext method in pkg/k8s/dialer.go starts a goroutine (line 80) that calls c.exec with context.TODO() instead of the ctx parameter from DialContext.

This means the exec call does not directly respect the caller's context cancellation. When the parent context is cancelled, the caller gets ctx.Err() from the select, and the pipes are closed, which eventually causes exec to error out. But passing ctx directly to exec lets StreamWithContext respond to cancellation immediately without relying on the indirect pipe-closure path.

This is a one-line change from context.TODO() to ctx.

Fixes #3800

### DIFF
--- a/pkg/k8s/dialer.go
+++ b/pkg/k8s/dialer.go
@@ -81,7 +81,7 @@ func (c *contextDialer) DialContext(ctx context.Context, network string, addr st
 		stderrBuff := bytes.NewBuffer(nil)
 		ctrStderr := io.MultiWriter(stderrBuff, detectConnSuccess(connectSuccess))
 
-		err := c.exec(context.TODO(), addr, ctrStdin, ctrStdout, ctrStderr)
+		err := c.exec(ctx, addr, ctrStdin, ctrStdout, ctrStderr)
 		if err != nil {
 			stderrStr := stderrBuff.String()
 			socatErr := tryParseSocatError(network, addr, stderrStr)


gauron99:
closing wrt https://github.com/knative/func/pull/3772/

matejvasek:
Look at https://github.com/knative/func/pull/3772

matejvasek:
@Elvand-Lie  We could reopen this if you address the concerns raised in the prior PR.

matejvasek:
Also you should adjust the dialer test to verify that connection works after "dial context" has expired.

Elvand-Lie:
@matejvasek I've updated this PR to address your feedback from #3772 I gave the exec stream its own background context and added an integration test (TestInt_DialContextExpiry) to verify the connection survives after the dial context expires

PR #3806, opened 2026-05-20
test: add E2E matrix test for deploy -e flag (fixes #3514)

### LINKED ISSUE #3514: Environment variables passed via -e flag not applied on deploy

Itx-Psycho0:
## What This PR Does

This PR adds an end-to-end matrix test that verifies environment variables passed via the `-e` flag during deployment actually work across all supported runtimes and builders.

## Background

Issue #3514 reported that environment variables passed with `func deploy -e KEY=VALUE` weren't showing up in the deployed service. The user could see them working with `func run -e`, but when deploying to the cluster, the variables were missing from the container.

After looking through the codebase, I noticed we have tests for `func config envs add`, but there's no E2E test that actually verifies the `-e` flag works during deployment. We're testing the config management commands but not the actual deployment behavior with environment variables.

@lkingland suggested adding an E2E matrix test that creates a function which reports back whether it received the environment variables, deploys it with `-e` flags, invokes it, and checks if the variables made it through. This would run across all our supported combinations to catch any builder or runtime-specific issues.

## Implementation

I created `e2e/e2e_matrix_deploy_envs_test.go` with a new test function `TestMatrix_Deploy_Envs()` that does the following for each runtime/builder/template combination:

1. Initializes a function with the specific runtime and template
2. Replaces the generated code with a simple implementation that reads `TEST_VAR` and `ANOTHER_VAR` from the environment and returns their values in the response
3. Deploys using `func deploy -e TEST_VAR=test_value -e ANOTHER_VAR=another_value --builder <builder>`
4. Waits for the function to become ready
5. Makes an HTTP request to the deployed function
6. Verifies the response contains both variables with correct values

The test includes language-specific implementations for reading environment variables since each runtime does it differently:

- **Go**: `os.Getenv()`
- **Node/TypeScript**: `process.env`
- **Python**: `os.getenv()`
- **Rust**: `env::var()`
- **Quarkus**: `@ConfigProperty` injection
- **SpringBoot**: `Environment.getProperty()`

Each implementation returns HTTP 500 if either variable is missing, and HTTP 200 with the values if both are present. This makes failures easy to detect.

## Why This Approach

The test follows the same pattern as existing matrix tests like `TestMatrix_Deploy`, using the `forEachPermutation` helper to run across all valid combinations. It respects the same matrix exceptions and timeout configurations.

By testing all combinations, we ensure the `-e` flag works consistently whether you're using host builder with Go, pack builder with Python, or s2i builder with Node. If there's a bug in how a specific builder handles environment variables, this test will catch it.

## Expected Results

When this runs in CI:

- If it passes: the `-e` flag works correctly and issue #3514 can be closed
- If it fails: we'll know exactly which builder/runtime combinations have the bug

The test runs with `FUNC_E2E_MATRIX=true go test -v ./e2e -run TestMatrix_Deploy_Envs` and will automatically run in CI when the matrix test environment variable is set.

## Testing

I've verified:
- Code compiles without errors
- Test structure follows existing matrix test patterns
- Each language implementation is syntactically correct
- Test properly cleans up resources

I haven't run the full matrix locally since it requires a Kubernetes cluster with all builders and runtime images, but the test follows the established patterns from other working matrix tests.

Fixes #3514

### DIFF
--- a/e2e/e2e_matrix_deploy_envs_test.go
+++ b/e2e/e2e_matrix_deploy_envs_test.go
@@ -1,0 +1,340 @@
+//go:build e2e
+
+package e2e
+
+import (
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+)
+
+// TestMatrix_Deploy_Envs ensures that environment variables passed via -e flag
+// on deploy are correctly applied to the deployed function across all builders
+// and runtimes.
+//
+// This test verifies that `func deploy -e KEY=VALUE` works correctly for:
+// - All supported runtimes (go, python, node, typescript, rust, quarkus, springboot)
+// - All supported builders (host, pack, s2i)
+// - Both http and cloudevents templates
+//
+// The test creates a function that echoes back environment variables, deploys
+// it with `func deploy -e TEST_VAR=test_value -e ANOTHER_VAR=another_value`,
+// then invokes the function and verifies the env vars are present.
+//
+// Related: Issue #3514
+func TestMatrix_Deploy_Envs(t *testing.T) {
+	forEachPermutation(t, "deploy-envs", func(t *testing.T, name, runtime, builder, template string) {
+		root := fromCleanEnv(t, name)
+
+		// Register cleanup functions
+		t.Cleanup(func() {
+			cleanImages(t, name)
+		})
+		t.Cleanup(func() {
+			clean(t, name, Namespace)
+		})
+
+		// Initialize function
+		initArgs := []string{"init", "-l", runtime, "-t", template}
+		initArgs, timeout := matrixExceptionsLocal(t, initArgs, runtime, builder, template)
+		if err := newCmd(t, initArgs...).Run(); err != nil {
+			t.Fatalf("Failed to create %s function with %s template: %v", runtime, template, err)
+		}
+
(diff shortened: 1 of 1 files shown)


Itx-Psycho0:
Hi @lkingland sir,

I've implemented the E2E matrix test you suggested for issue #3514. The test creates a function that echoes back environment variables, deploys it with `func deploy -e TEST_VAR=test_value -e ANOTHER_VAR=another_value`, and verifies the variables are present in the deployed container.

The test covers all 7 runtimes (go, python, node, typescript, rust, quarkus, springboot), all 3 builders (host, pack, s2i), and both templates (http, cloudevents). Each language has its own implementation for reading and returning the environment variables.

This should help us verify whether the `-e` flag works correctly across all combinations, or if there's a specific builder/runtime pair that has the issue reported in #3514.

Would appreciate your review when you get a chance. Thanks!

matejvasek:
Some check if failing.

PR #3807, opened 2026-05-20
fix(k8s): fix error ordering in ProcessEnvs and late validation

### LINKED ISSUE #3798: k8s: ProcessEnvs appends env var before checking createEnvVarSource error

Elvand-Lie:
## Problem

**ProcessEnvs error ordering:**
ProcessEnvs appends the result of createEnvVarSource to the envVars slice before checking the returned error. This violates standard Go error-handling pattern.

**createEnvFromSource/createEnvVarSource late validation:**
Both functions validate that sourceName is non-empty only after using it to construct resource references and inserting it into the referenced sets.

## Fix

Reordered ProcessEnvs so the error check comes first. Moved the empty-name validation before the switch statements in createEnvFromSource and createEnvVarSource so the sets are not polluted with empty strings.

## Tests added

- Test_ProcessEnvs_ErrorCheckBeforeAppend: error returns nil slice
- Test_createEnvFromSource_EmptyName: empty name errors without polluting set
- Test_createEnvVarSource_EmptyName: empty name errors without polluting set
- Test_createEnvVarSource_EmptyKey: empty key errors

Fixes #3798

### DIFF
--- a/pkg/k8s/deployer.go
+++ b/pkg/k8s/deployer.go
@@ -633,10 +633,10 @@ func ProcessEnvs(envs []fn.Env, referencedSecrets, referencedConfigMaps *sets.Se
 				if len(slices) == 3 {
 					// ENV from a key in secret/configMap, eg. FOO={{ secret:secretName:key }} FOO={{ configMap:configMapName.key }}
 					valueFrom, err := createEnvVarSource(slices, referencedSecrets, referencedConfigMaps)
-					envVars = append(envVars, corev1.EnvVar{Name: *env.Name, ValueFrom: valueFrom})
 					if err != nil {
 						return nil, nil, err
 					}
+					envVars = append(envVars, corev1.EnvVar{Name: *env.Name, ValueFrom: valueFrom})
 					continue
 				} else if len(slices) == 2 {
 					// ENV from the local ENV var, eg. FOO={{ env:LOCAL_ENV }}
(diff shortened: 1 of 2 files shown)


matejvasek:
there is conflict @Elvand-Lie

Elvand-Lie:
Rebased onto main. The conflict was from the ProcessVolumes nil-path tests that were merged in the meantime. Kept both test sets in. @matejvasek

PR #3808, opened 2026-05-20
fix(knative): fix goroutine leak in Deploy

### LINKED ISSUE #3797: knative: goroutine leak in Deploy when service becomes ready before private-registry check

Elvand-Lie:
## Problem

The Deploy function (new-service creation path) starts a background goroutine that polls isImageInPrivateRegistry every 5 seconds and sends results on an unbuffered chprivate channel.

If WaitForService completes first (via cherr), the main function returns and the poller goroutine blocks forever on the channel send with no receiver, leaking the goroutine and captured context.

## Fix

Fixed by adding a done channel that the main select loop closes when it exits. The poller goroutine now selects on done alongside its timer and channel send, so it always exits promptly.

Fixes #3797

### DIFF
--- a/pkg/knative/deployer.go
+++ b/pkg/knative/deployer.go
@@ -45,6 +45,8 @@ const (
 	KnativeDeployerName = "knative"
 )
 
+var errPrivateRegistry = stdErrors.New("your function image is unreachable. It is possible that your docker registry is private. If so, make sure you have set up pull secrets https://knative.dev/docs/developer/serving/deploying-from-private-registry")
+
 type DeployerOpt func(*Deployer)
 
 type Deployer struct {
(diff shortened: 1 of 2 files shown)


gauron99 (commented):
Thanks for the PR, I think there is the same issue in the waiter goroutine. Couldn't cherr also be blocked forever? We could make cherr buffered `make(chan error, 1)` or add the same done check in select as you did above.

Elvand-Lie:
didn't think about the waiter goroutine blocking on cherr if the main loop exits early. I've updated cherr to be buffered as you suggested and pushed the fix. Thank you @gauron99

PR #3814, opened 2026-05-20
fix: pin registry.localtest.me to IPv4 to unbreak CI Prepare Images step

gauron99:


matejvasek:
WTF how are not Happy Eyeballs working here? It worked for me when I added registry.localtest.me ?

Is the registry starting up inside? Maybe the registry deployment is not healthy (e.g. image pull err)?

matejvasek:
This is odd `::1` flake happened before https://github.com/knative/func/pull/3642#issuecomment-4319566382 even when we were running registry as container at localhost:50000.

matejvasek:
This is very weird. Happy eyeballs should be employed. both `127.0.0.1` and `::` should be dialed at the same time (or maybe `::1` is dialed with some small (100ms) advance). And whatever responds first should be used.

matejvasek:
This must have always be the case as locahost and localtest.me always had AAAA.

matejvasek:
Only thing that could break this out of nowhere (and it is out of nowhere) is that suddenly something started to listening on `[::1]:80`, something other than our ingress.

matejvasek:
Prepare images seem to be stuck here: https://github.com/knative/func/blob/main/hack/images.sh#L16-L21 nothing to do with `::1`.

matejvasek:
```
crane pull quay.io/buildah/stable:v1.31.0 /dev/stdout| pv > /dev/null
```
Seems to hang. There is our issue.

matejvasek:
The "Prepare Images" issue is caused by https://github.com/google/go-containerregistry/issues/2309.

PR #3818, opened 2026-05-21
mcp: add config_git_set and config_git_remove tools

Ankitsinghsisodya:
## Summary

Adds two MCP tools — `config_git_set` and `config_git_remove` — to configure Git source repository settings for pipeline-based Function builds. This completes the `config` tool group alongside the existing volumes, labels, and envs tools.

> **Note:** This is a re-submission of #3795, which was accidentally closed when the head repository was deleted.

## Changes

* **`pkg/mcp/tools_config_git.go`** — two tools following the existing `config_*` pattern:
  * `config_git_set`: configures Git URL, branch, subdirectory, and optional Pipelines-as-Code (PAC) resources on the Function. Requires `path`, `git_url`, and `git_branch`. Defaults `git_dir` to `"."` to avoid the CLI's interactive prompt when no subdirectory is needed. Automatically injects `--config-local` when no `config_*` flags are provided to prevent the webhook confirmation prompt in a non-TTY subprocess.
  * `config_git_remove`: removes Git settings and associated pipeline resources from a Function. Automatically injects `--delete-local` when neither `delete_local` nor `delete_cluster` is set, preventing the interactive "delete all?" prompt.
* **`pkg/mcp/tools_config_git_test.go`** — nine unit tests covering args, defaults, anti-hang behavior, and error propagation.
* **`pkg/mcp/mcp.go`** — registers both tools and three help resources (`func://help/config/git`, `func://help/config/git/set`, `func://help/config/git/remove`).
* **`pkg/mcp/instructions.md`** — adds `config_git_set` / `config_git_remove` agent guidance covering required fields, PAC resource flags, GitHub token requirements, and destructive-operation warning for `delete_cluster`.

## Notes

* No `config_git_list` tool is added because `func config git` (the root command) currently prints "Not implemented yet." and provides no useful output. Git settings can be inspected via the `func://function` resource which exposes the full `func.yaml` state.

## Test plan

* `make test` passes
* `make check` passes
* Manual: call `config_git_set` with `path`, `git_url`, `git_branch` and confirm `func.yaml` is updated with Git config
* Manual: call `config_git_remove` with `path` only and confirm local pipeline resources are cleaned up without prompting

### DIFF
--- a/pkg/mcp/instructions.md
+++ b/pkg/mcp/instructions.md
@@ -148,7 +148,7 @@ A first-time deploy can be detected by checking the func.yaml for a value in the
   - **REQUIRED:** `git_url` (repository URL) and `git_branch` (branch or tag, e.g. `main`)
   - **OPTIONAL:** `git_dir` (subdirectory in the repo; defaults to repository root when omitted)
   - **OPTIONAL:** `git_provider` (auto-detected from URL; override only if detection fails)
-  - **OPTIONAL PAC:** `config_local`, `config_cluster`, `config_remote` (boolean flags to control which pipeline resources are created; defaults to local-only when none are specified)
+  - **OPTIONAL:** `config_local`, `config_cluster`, `config_remote` (boolean flags to control which pipeline resources are created; defaults to local-only when none are specified)
   - **OPTIONAL:** `gh_access_token` — required only when `config_remote` is true to create a GitHub webhook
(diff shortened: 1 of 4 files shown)


lkingland (approved):
/lgtm

We shuold remember to put in agent instructions that a token should never be passed verbatim as a flag, but be read from a CLI secret manager or via env variable, lest the token end up in cli history and/or logfiles.

PR #3819, opened 2026-05-21
mcp: Add logs tool to retrieve deployed Function logs

Ankitsinghsisodya:
# Changes

- :gift: Adds a `logs` MCP tool returning a snapshot of a deployed Function's logs, identified by exactly one of `path` or `name`.
- :gift: Bounds the output: `tail` defaults to 1000 lines per pod (the CLI's default is unlimited), and the payload is capped at 256 KiB, keeping the most recent lines and setting `truncated`.
- :gift: Returns the CLI's stderr notices as `warnings`, so a Function scaled to zero is not mistaken for one which printed nothing.
- :broom: Registers the tool and the `func://help/logs` resource, and documents its contract in `instructions.md`.

Streaming (`--follow`) is not exposed, as it never terminates. `namespace` is forwarded as given; in `path` mode `func logs` reads the namespace from the Function's deploy identity and ignores the flag.

/kind enhancement

Relates to #3999

**Release Note**

```release-note
The MCP server now provides a 'logs' tool, which returns a snapshot of a deployed Function's logs.
```

**Docs**

```docs

```

### DIFF
--- a/pkg/mcp/instructions.md
+++ b/pkg/mcp/instructions.md
@@ -57,6 +57,7 @@ This is essential because:
 - Before 'deploy' → Read `func://help/deploy`
 - Before 'build' → Read `func://help/build`
 - Before 'list' → Read `func://help/list`
+- Before 'logs' → Read `func://help/logs`
 - Before 'delete' → Read `func://help/delete`
 
 The help text provides authoritative parameter information and usage context.
@@ -130,6 +131,17 @@ A first-time deploy can be detected by checking the func.yaml for a value in the
 - Does NOT use path parameter (operates on cluster, not local files)
 - Optional `namespace` parameter to list Functions in specific namespace
 - Returns list of deployed Functions in current/specified namespace
+
+### logs
+
+- **FIRST:** Read `func://help/logs` for authoritative usage information
+- Identify the Function by **path** (reads func.yaml) OR by **name** — never both at the same time
+- `path` must be an absolute path to the Function project directory
+- `name` is the deployed Function name on the cluster
+- Optional `namespace` parameter to target a specific Kubernetes namespace
+- Optional `since` parameter controls the time window of returned logs (e.g. `30s`, `5m`, `2h`; default is `1m`)
+- This tool is **read-only** — it never modifies any state
+- Use logs to diagnose a deployed Function after `deploy`, especially when combined with `invoke` to trigger the Function and observe its output
 
 ### delete
 
--- a/pkg/mcp/instructions.md
+++ b/pkg/mcp/instructions.md
@@ -63,6 +63,7 @@ This is essential because:
 - Before 'invoke' → Read `func://help/invoke`
 - Before 'list' → Read `func://help/list`
 - Before 'describe' → Read `func://help/describe`
+- Before 'logs' → Read `func://help/logs`
 - Before 'delete' → Read `func://help/delete`
 - Before 'run' → Read `func://help/run`
 
@@ -167,6 +168,17 @@ A first-time deploy can be detected by checking the func.yaml for a value in the
 - Exactly ONE of 'path' or 'name' must be provided, not both
 - 'namespace' is only valid together with 'name'; providing both 'path' and 'namespace' is rejected, since path mode already determines the namespace from the Function's own deploy identity
 - Read-only; does not modify local files or cluster resources
+
+### logs
+
+- **FIRST:** Read `func://help/logs` for authoritative usage information
+- Identify the Function by **path** (reads func.yaml) OR by **name** — never both at the same time
+- `path` must be an absolute path to the Function project directory
+- `name` is the deployed Function name on the cluster
+- Optional `namespace` parameter to target a specific Kubernetes namespace
+- Optional `since` parameter controls the time window of returned logs (e.g. `30s`, `5m`, `2h`; default is `1m`)
+- This tool is **read-only** — it never modifies any state
+- Use logs to diagnose a deployed Function after `deploy`, especially when combined with `invoke` to trigger the Function and observe its output
 
 ### delete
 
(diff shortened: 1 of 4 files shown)


pkg/mcp/tools_logs.go
```diff
@@ -0,0 +1,68 @@
...
+	},
+}
+
+func (s *Server) logsHandler(ctx context.Context, r *mcp.CallToolRequest, input LogsInput) (result *mcp.CallToolResult, output LogsOutput, err error) {
+	if input.Path != nil && input.Name != nil {
+		err = fmt.Errorf("'path' and 'name' are mutually exclusive: provide at most one")
+		return
+	}
+
+	out, err := s.executor.Execute(ctx, "logs", input.Args()...)
```
lkingland: Unfortunately I don't think we can do this in MCP, right?  It would hang forever?

We probably need a snapshot, and currently `func logs` is streaming only (like always running `kubectl logs -f`)

Opening a separate issue for this, because `func logs` should really operate like `kubectl logs` by default anyway, where you have to opt-in to an output stream with `-f`...

pkg/mcp/tools_logs.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		err = fmt.Errorf("%w\n%s", err, string(out))
+		return
+	}
+	output = LogsOutput{
+		Logs: string(out),
```
lkingland: Let's use your new `ExecuteSplit` here so that the logs don't get the
STDERR output from the `logs` command, which (right now at least) is something like 
```
Streaming logs for function '…' in namespace '…'...
Press Ctrl+C to stop.
```
But after the new "snapshot" ability will probably be something like:

```
Logs for function '…' in namespace '…' since '{timestamp'...
```

pkg/mcp/tools_logs.go
```diff
@@ -0,0 +1,68 @@
...
+	if i.Path != nil {
+		args = append(args, "--path", *i.Path)
+	} else if i.Name != nil {
+		args = append(args, "--name", *i.Name)
+	}
+
+	args = appendStringFlag(args, "--namespace", i.Namespace)
+	args = appendStringFlag(args, "--since", i.Since)
+	args = appendBoolFlag(args, "--verbose", i.Verbose)
+	return args
```
lkingland, line 178: Once `func logs` is by default a snapshot instead of a strem, you _shouldn't_ need to change this.  It'll just be the default.

pkg/mcp/tools_logs_test.go
```diff
@@ -0,0 +1,163 @@
...
+		t.Fatal(err)
+	}
+	if !result.IsError {
+		t.Fatal("expected an error result when both path and name are provided")
+	}
+}
+
+// TestTool_Logs_NoArgs ensures the logs tool works without any arguments,
+// falling back to the server's working directory (CLI default behaviour).
+func TestTool_Logs_NoArgs(t *testing.T) {
```
lkingland: I am thinking the MCP should generally not fall back to CWD like the CLI, with a better default being to enforce name or path.  So a test like `TestTool_Logs_RequiresPathOrName`, and then enforce that if neither is provided it errors.

pkg/mcp/instructions.md
```diff
@@ -131,6 +132,17 @@ A first-time deploy can be detected by checking the func.yaml for a value in the
...
+### logs
+
+- **FIRST:** Read `func://help/logs` for authoritative usage information
+- Identify the Function by **path** (reads func.yaml) OR by **name** — never both at the same time
+- `path` must be an absolute path to the Function project directory
+- `name` is the deployed Function name on the cluster
+- Optional `namespace` parameter to target a specific Kubernetes namespace
+- Optional `since` parameter controls the time window of returned logs (e.g. `30s`, `5m`, `2h`; default is `1m`)
+- This tool is **read-only** — it never modifies any state
+- Use logs to diagnose a deployed Function after `deploy`, especially when combined with `invoke` to trigger the Function and observe its output
```
lkingland, line 188: We can tighten up the wording a bit here:

- Exactly one of `path` or `name` is required (same shape as `delete` / `describe`)
- This tool returns a "finite snapshot" of recent logs (not a live stream); `since` bounds the window
- Reminder that the Function must already be deployed (path mode still talks to the cluster via the project's deploy identity)
- Optional: note knative-deployer-only limitation if the CLI still errors for other deployers (agents will hit that after deploy with other deploy modes)

The deploy -> invoke -> logs workflow line is good product intent.  We just need to get the "snapshot of logs" functionality in there to make this work via MCP 👍🏻

pkg/mcp/tools_logs.go
```diff
@@ -0,0 +1,68 @@
...
+
+var logsTool = &mcp.Tool{
+	Name:        "logs",
+	Title:       "Get Function Logs",
+	Description: "Retrieve recent logs from a deployed Function. Use the since argument to control the time window (e.g. '5m', '1h'). Identify the Function by path (reads func.yaml) or by name.",
+	Annotations: &mcp.ToolAnnotations{
+		Title:           "Get Function Logs",
+		ReadOnlyHint:    true,
+		DestructiveHint: ptr(false),
+		IdempotentHint:  true,
```
lkingland: nit:  Once we have the snapshot logs functionality added,  `IdempotentHint: true` is a stretch :)

lkingland (commented):
Looking very good!

Opened #3999 to give `func logs` the right behavior for MCP (snapshot by default, `-f` only for interactive streaming).

pkg/mcp/tools_logs.go
Ankitsinghsisodya: I have remove the `IdempotentHint: true`

pkg/mcp/instructions.md
Ankitsinghsisodya: I have added this in the instructions.md.

pkg/mcp/tools_logs_test.go
Ankitsinghsisodya: okey

pkg/mcp/tools_logs.go
Ankitsinghsisodya: okey

PR #3820, opened 2026-05-21
fix: forward CLI overrides to in-cluster deploy for Git-based remote deploys

### LINKED ISSUE #3768: Remote Git-based deploy doesn't forward CLI overrides like `--image-pull-secret` to the in-cluster `func deploy`

Ankitsinghsisodya:
Fixes #3768

> **Note:** Re-submission of #3777, accidentally closed when the head repository was deleted.

## Problem

When `func deploy` uses a Git-based pipeline (`f.Build.Git.URL` is set), CLI flags like `--image-pull-secret`, `--service-account`, and `--deployer` are silently ignored. The on-cluster `func-deploy` task step reads `func.yaml` directly from the cloned Git repo, which never contains in-memory CLI overrides — unlike the PVC-upload path fixed by #3663.

## Solution

Forward the three CLI overrides as discrete Tekton params on the pipeline run and apply them as environment variables on the `func-deploy` task step. The in-cluster `deploy` binary reads these env vars and applies them over the values loaded from `func.yaml`.

## How it works

```
func deploy --image-pull-secret my-secret --deployer raw
  → templateData.ImagePullSecret / Deployer populated
  → PipelineRun param: imagePullSecret=my-secret, deployer=raw
  → Pipeline threads params to Task: IMAGE_PULL_SECRET, DEPLOYER
  → func-deploy step env: FUNC_IMAGE_PULL_SECRET, FUNC_DEPLOYER
  → deploy binary: os.Getenv overrides f loaded from committed func.yaml
```

The env vars are only applied when non-empty, so existing behaviour is unchanged when no overrides are passed.

## Changes

* `cmd/func-util/main.go` — read `FUNC_IMAGE_PULL_SECRET`, `FUNC_SERVICE_ACCOUNT`, `FUNC_DEPLOYER` env vars after loading `func.yaml` and apply them over the struct before deploying
* `task-buildpack.yaml.tmpl` / `task-s2i.yaml.tmpl` — add `IMAGE_PULL_SECRET`, `SERVICE_ACCOUNT`, `DEPLOYER` params; set corresponding `FUNC_*` env vars on the `func-deploy` step
* `templates_pack.go` / `templates_s2i.go` — add the three params to the Pipeline spec and thread them to the task; add them to the PipelineRun params
* `templates.go` — add `ImagePullSecret`, `ServiceAccountName`, `Deployer` to `templateData`; populate from `f.Deploy.*` in `createAndApplyPipelineRunTemplate`

## Notes

* The `DeployerImage` is `scratch`-based with no shell, so env vars are used instead of a bash script with conditional flag appending.
* This is the Git-path equivalent of #3663, which fixed the same bug for the PVC-upload path.

### DIFF
--- a/cmd/func-util/main.go
+++ b/cmd/func-util/main.go
@@ -153,6 +153,22 @@ func deploy(ctx context.Context) error {
 	if f.Deploy.Image == "" {
 		f.Deploy.Image = f.Image
 	}
+
+	// For Git-based remote deploys the on-cluster func.yaml comes from the
+	// committed repo and never contains CLI overrides supplied by the user
+	// (--image-pull-secret, --service-account, --deployer).  The pipeline run
+	// forwards those overrides as environment variables so they can be applied
+	// here before deploying.
+	if v := os.Getenv("FUNC_IMAGE_PULL_SECRET"); v != "" {
+		f.Deploy.ImagePullSecret = v
+	}
+	if v := os.Getenv("FUNC_SERVICE_ACCOUNT"); v != "" {
+		f.Deploy.ServiceAccountName = v
+	}
+	if v := os.Getenv("FUNC_DEPLOYER"); v != "" {
+		f.Deploy.Deployer = v
+	}
+
 	if f.Deploy.Deployer == "" {
 		f.Deploy.Deployer = knative.KnativeDeployerName
 	}
--- a/cmd/func-util/main.go
+++ b/cmd/func-util/main.go
@@ -155,9 +155,25 @@ func deploy(ctx context.Context) error {
 	if f.Deploy.Image == "" {
 		f.Deploy.Image = f.Image
 	}
+	// For Git-based remote deploys the on-cluster func.yaml comes from the
+	// committed repo and never contains CLI overrides supplied by the user
+	// (--image-pull-secret, --service-account, --deployer).  The pipeline run
+	// forwards those overrides as environment variables so they can be applied
+	// here before deploying.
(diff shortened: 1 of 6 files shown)


Ankitsinghsisodya:
> /lgtm /approve

Can you do it again?

Ankitsinghsisodya:
@lkingland

PR #3776, opened 2026-05-17
feat: enable volume mounts for local func run

Ankitsinghsisodya:
Fixes #2940

## Summary

- Updated `pkg/docker/runner.go` -> `newHostConfig()` to configure volume mounts in addition to port bindings when running functions locally with `func run`
- Added `volumeMounts` and `toMount` helpers to convert `func.yaml` volume definitions into Docker mount specs
- Volumes that cannot be mapped locally emit a warning to stderr and are skipped; the run is not aborted

## Volume mapping

| func.yaml type | Docker mount | Local path |
|---|---|---|
| `secret: <name>` | bind | `.func/run/secrets/<name>` (created if absent) |
| `configMap: <name>` | bind | `.func/run/configmaps/<name>` (created if absent) |
| `emptyDir: {medium: Memory}` | tmpfs | — |
| `emptyDir: {}` | anonymous volume | — |
| `persistentVolumeClaim` | bind | `.func/run/pvcs/<claimName>` (created if absent) |

All bind-mount sources are rooted inside `.func/`, satisfying the security constraint in the issue (no paths outside `.func` are ever mounted).

## Notes

- Picks up where #3143 left off. The core logic is the same; this version uses the current moby API (`mount.Mount`) instead of the legacy `Binds []string` + `Tmpfs` fields, and resolves local paths from `f.Root` rather than the process working directory, which fixes a subtle bug when `func run` is invoked from outside the function root.
- The flaky Java SSL test that blocked #3143 is unrelated to this change.

PR #3143, opened 2025-10-24
Enable local volume mounts in func run for ConfigMaps, Secrets, EmptyDirs, and PVCs

intojhanurag:
fix: #2940 
### Changes
- Updated `pkg/docker/runner.go -> newHostConfig()` to mount the following volume types:
  - **ConfigMap** → `.func/run/configmaps/<name>` directory
  - **Secret** → `.func/run/secrets/<name>` directory
  - **EmptyDir** → Docker tmpfs
  - **PersistentVolumeClaim** → `.func/run/pvcs/<claimName>` directory
- Added warnings for volumes that attempt to map outside of `.func` to prevent accidental access to sensitive host directories.
- Preserves existing behavior for port bindings and other container settings.

PR #3821, opened 2026-05-21
feat: enable volume mounts for local func run

### LINKED ISSUE #2940: Volume Mounts for Local Runs
Currently, runtime volumes defined in `func.yaml` are ignored when running functions locally with `func run`. 

We should update `pkg/docker/runner.go` -> `newHostConfig()` to configure volumes in addition to the ports it currently defines.

1. **ConfigMap** → Local `.func/run/configmaps/<name>` directory
2. **Secret** → Local `.func/run/secrets/<name>` directory  
3. **EmptyDir** → Docker tmpfs or anonymous volume
4. **PersistentVolumeClaim** → Named Docker volume or local directory

Additionally, we could consider adding a new volume type specifically for local development:
```yaml
volumes:
  - hostPath: /local/path
    path: /container/path
```

Implementation Notes:

There should be warnings when defined resources are note available locally, but not fail.

Should disallow mapping of directories outside of .func for security (example risk:  user downloads a malicious function, runs it locally, accessing sensitive areas of the FS and exfiltrating via the funciton source implementation)

## Current Workarounds

1. Use the "host" builder (loses container isolation)
2. Run `func build` then manually `docker run -v local:container image`
3. Deploy to a test cluster for volume testing

Ankitsinghsisodya:
Fixes #2940

> **Note:** Re-submission of #3776, accidentally closed when the head repository was deleted.

## Summary

* Updated `pkg/docker/runner.go` -> `newHostConfig()` to configure volume mounts in addition to port bindings when running functions locally with `func run`
* Added `volumeMounts` and `toMount` helpers to convert `func.yaml` volume definitions into Docker mount specs
* Volumes that cannot be mapped locally emit a warning to stderr and are skipped; the run is not aborted

## Volume mapping

| func.yaml type             | Docker mount     | Local path                                      |
| -------------------------- | ---------------- | ----------------------------------------------- |
| secret: \<name\>             | bind             | .func/run/secrets/\<name\> (created if absent)    |
| configMap: \<name\>          | bind             | .func/run/configmaps/\<name\> (created if absent) |
| emptyDir: {medium: Memory} | tmpfs            | —                                               |
| emptyDir: {}               | anonymous volume | —                                               |
| persistentVolumeClaim      | bind             | .func/run/pvcs/\<claimName\> (created if absent)  |

All bind-mount sources are rooted inside `.func/`, satisfying the security constraint in the issue (no paths outside `.func` are ever mounted).

## Notes

* Picks up where #3143 left off. The core logic is the same; this version uses the current moby API (`mount.Mount`) instead of the legacy `Binds []string` + `Tmpfs` fields, and resolves local paths from `f.Root` rather than the process working directory, which fixes a subtle bug when `func run` is invoked from outside the function root.
* The flaky Java SSL test that blocked #3143 is unrelated to this change.

### DIFF
--- a/pkg/docker/runner.go
+++ b/pkg/docker/runner.go
@@ -330,12 +330,12 @@ func volumeMounts(root string, volumes []fn.Volume, out io.Writer) []mount.Mount
 	var mounts []mount.Mount
 	for _, vol := range volumes {
 		if vol.Path == nil {
-			fmt.Fprintf(out, "warning: skipping volume %v: missing path\n", vol)
+			fmt.Fprintf(out, "warning: skipping volume %s: missing path\n", vol.String())
 			continue
 		}
 		m, err := toMount(root, vol)
 		if err != nil {
-			fmt.Fprintf(out, "warning: skipping volume %v: %v\n", vol, err)
+			fmt.Fprintf(out, "warning: skipping volume %s: %v\n", vol.String(), err)
 			continue
 		}
 		mounts = append(mounts, m)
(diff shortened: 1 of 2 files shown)


pkg/docker/runner.go
```diff
@@ -317,7 +320,99 @@ func newHostConfig(host, port string) (c container.HostConfig, err error) {
...
+		src := filepath.Join(runDir, "configmaps", *vol.ConfigMap)
+		if err := os.MkdirAll(src, 0750); err != nil {
+			return mount.Mount{}, fmt.Errorf("cannot create local configmap dir %q: %w", src, err)
+		}
+		return mount.Mount{Type: mount.TypeBind, Source: src, Target: target}, nil
+	}
+
+	if vol.EmptyDir != nil {
+		// Both mediums map to tmpfs locally: EmptyDir is ephemeral (pod-lifetime)
+		// and anonymous Docker volumes persist across runs, leaking storage.
```
lkingland, line 397: Example of the body mismatch: this is tmpfs for EmptyDir (including default medium), not an anonymous volume as the PR table says.

pkg/docker/runner.go
```diff
@@ -303,7 +306,7 @@ func newContainerConfig(f fn.Function, _ string, verbose bool) (c container.Conf
 	return
 }
 
-func newHostConfig(host, port string) (c container.HostConfig, err error) {
+func newHostConfig(host, port string, f fn.Function, out io.Writer) (c container.HostConfig, err error) {
```
lkingland, line 309: naming it `out` is misleading, as it's realy error out, and there's no comments.  At a minimum it should be `errOut`

lkingland (commented):
The `.func/`-rooted secret/configMap binds and skip-and-warn on bad volumes look right. Pointer soup on `fn.Volume` is the existing YAML model; not something this PR invented.

The review table in the PR body does not match the code. Worth fixing so nobody reviews from the table:

* `emptyDir: {}` is described as an anonymous volume; the code mounts **tmpfs** (same as `medium: Memory`; the comment about avoiding leaks is the reason).
* PVC is described as a bind of `.func/run/pvcs/<claim>`; the code uses a **named Docker volume** keyed by `claimName`.

A few Copilot notes I agree with:

* Secret and ConfigMap mounts in Kubernetes are read-only. Setting `ReadOnly: true` on those binds would match cluster semantics (and not mask "I wrote the secret file" surprises locally).
* `safeLocalName` rejecting any `".."` substring is a bit eager (`foo..bar` is not traversal). Rejecting path separators, empty names, and real `..` segments is enough.
* `toMount` assuming a non-nil `Path` is fine as long as only `volumeMounts` calls it (it already skips nil). A one-line guard would make the helper reusable. Same idea if `out` can be nil; `io.Discard` is cheap.

The ConfigMap test not asserting `Target` is tiny; nice if you are already touching tests.

Once the body matches the code (and ReadOnly if you agree), this looks good to me.

PR #3774, opened 2026-05-17
feat: add --json output mode to all func subcommands

Ankitsinghsisodya:
## Summary

Implements structured JSON output for every `func` subcommand, as described in [#3769](https://github.com/knative/func/issues/3769).

All JSON is written to **stdout** using a versioned envelope:
```json
{"apiVersion":"v1","status":"ok","data":{...}}
{"apiVersion":"v1","status":"error","error":{"category":"...","code":"...","retryable":false,"message":"...","hint":"..."}}
```

The flag is opt-in — all existing human-readable output is unchanged.

## Changes

**New files**
- `cmd/json.go` — shared `JSONResponse`/`JSONError` structs, `WriteJSONSuccess`, `WriteJSONError`, `errorToJSONError` (maps 14+ typed errors to category/code/retryable), and `isJSONEnabled` helper
- `cmd/json_test.go` — 20 unit tests covering every error-type mapping, envelope shape, `apiVersion` presence, and round-trip validity

**Global plumbing**
- `cmd/root.go` — `--json` registered as a persistent flag on the root command (inheritable by all subcommands), bound to Viper
- `pkg/app/app.go` — top-level error sink writes `WriteJSONError` to stdout when `--json` is set, instead of plain text to stderr

**Per-command success payloads**

| Command | `data` fields |
|---|---|
| `build` | `name`, `image` |
| `deploy` | `name`, `namespace`, `url`, `image` |
| `create` | `name`, `path`, `runtime`, `template` |
| `delete` | `name`, `namespace` |
| `describe` | existing `fn.Instance` object (reuses `Formatter.JSON()`) |
| `list` | array of `listItems` (reuses `Formatter.JSON()`) |
| `run` | `address`, `host`, `port` (upgrades existing ad-hoc JSON to envelope) |
| `invoke` | `response` (HTTP body) |
| `version` | existing `Version` struct |
| `languages` | array of runtime strings |
| `templates` | map of runtime → template array |
| `environment` | existing `Environment` struct |
| `subscribe` | `name`, `source`, `filters` |
| `repository list` | array of `{name, url}` |
| `repository add/rename/remove` | operation result fields |
| `logs` | explicitly unsupported — returns error if `--json` is set |

**Updated tests**
- `cmd/languages_test.go` — updated `TestLanguages_JSON` to assert envelope shape
- `cmd/templates_test.go` — updated `TestTemplates_JSON` and `TestTemplates_ByLanguage` to assert envelope shape

## Test plan

[...]

Issue #3769, opened 2026-05-16
Add `--json` output mode to every `func` subcommand, including errors


lkingland:
## Summary

Every `func` subcommand should support a `--json` flag that emits structured output suitable for machine consumption. Errors must also be returned as JSON when `--json` is set, not as human-formatted strings.

## Motivation

Today the only structured signal callers (MCP server, scripts, IDEs, CI, future Console plugin) can extract from `func` is the exit code plus stdout/stderr text. This forces every consumer to either substring-match output (brittle — see #3753 for a worked example of why this fails) or invent its own taxonomy by guessing at `func`'s intent. The result is a contract that's secretly the consumer's job to maintain, even though the producer (`func`) is the only code that actually knows what happened.

Putting structured output behind a `--json` flag fixes this for every consumer at once. Same pattern as `kubectl -o json`, `gh --json`, `aws --output json` — users already expect it from modern CLIs.

## Scope

- Every existing `func` subcommand accepts `--json`.
- On success: command-specific structured payload (e.g., `func list --json` returns the function list; `func describe --json` returns the deployed-function spec; `func deploy --json` returns the resulting service URL/image/etc).
- **On failure: a structured error object** (this is the part most consumers actually need), with at minimum:
  - `category` (e.g. `REGISTRY_ERROR`, `CLUSTER_ERROR`, `BUILD_ERROR`, `VALIDATION_ERROR`, `AUTH_ERROR`, `READONLY_ERROR`, `TIMEOUT_ERROR`, `UNKNOWN_ERROR`)
  - `code` (stable string identifier; finer-grained than category)
  - `retryable` / `transient` boolean
  - `message` (human-readable summary)
  - `hint` (suggested remediation)
  - `context` (resource name, namespace, registry, etc. — whatever's relevant to the failure)
- Default output (no flag) stays exactly as it is. `--json` is opt-in.

## Open design questions (decide before implementing)

1. **Envelope shape.** Suggested: `{"status": "ok"|"error", "data": ..., "error": {...}}` so callers can dispatch with one key.
2. **Stream choice.** All JSON to stdout (recommended — `status` carries success/failure, no need to merge streams) vs error JSON to stderr.
3. **Streaming commands** (`func logs`, `func deploy --verbose`): pick NDJSON (one event per line) or explicitly mark `--json` unsupported for these. "Explicit and unsupported" beats "clever and surprising."
[...]

PR #3822, opened 2026-05-21
feat: add --json output mode to all func subcommands

Ankitsinghsisodya:
> **Note:** Re-submission of #3774, accidentally closed when the head repository was deleted.

## Summary

Implements structured JSON output for every `func` subcommand, as described in #3769.

All JSON is written to **stdout** using a single versioned envelope:

```json
{"apiVersion":"v1","status":"ok","data":{...}}
{"apiVersion":"v1","status":"error","error":{"category":"...","code":"...","retryable":false,"message":"...","hint":"..."}}
```

### One dialect, two spellings

Per review feedback: `func` previously had **two** raw JSON shapes — `--output json`
on `describe`/`list`/`version` (parsed directly by `pkg/mcp`) and `--json` on
`languages`/`templates`. Rather than add a third, both spellings now emit the
**same** envelope:

* `--json` is the shorthand spelling of `--output json`; the two produce byte-identical output (`TestJSON_DialectParity`).
* Wrapping happens in one place — the `JSON` case of `cmd/format.go`'s `write()` — so the spellings cannot drift apart again.
* `pkg/mcp` (`describe`, `version`) unwraps `.data`, and surfaces a `status:"error"` envelope as the CLI's own `category`/`code`/`hint` instead of a parse failure.

**Breaking:** `func describe|list|version --output json` and `func languages|templates --json`
now return the envelope rather than a bare payload. Consumers read `.data`.
Human, plain, yaml and url formats are unchanged.

## Changes

**New files**
* `cmd/json.go` — `JSONResponse`/`JSONError`, `WriteJSONSuccess`, `WriteJSONError`, `errorToJSONError` (maps 14+ typed errors to category/code/retryable), `isJSONEnabled`/`outputFormat`/`JSONOutputRequested`
* `cmd/json_test.go` — unit tests for every error-type mapping, envelope shape, `apiVersion` presence, round-trip validity
* `cmd/json_envelope_test.go` — envelope decode helper; dialect-parity and `$FUNC_JSON` tests
* `pkg/mcp/envelope.go` + test — envelope unwrapping for the MCP tools that parse CLI JSON

**Global plumbing**
* `cmd/root.go` — `--json` as a persistent flag on the root command, bound to viper with env lookup configured so `$FUNC_JSON` works for every subcommand
* `cmd/json.go` — JSON mode is the *effective* setting (flag, `$FUNC_JSON`, or `--output json`), not merely "was the flag typed"
* `pkg/app/app.go` — top-level error sink writes the error envelope to stdout in JSON mode, falling back to plain text on stderr if the envelope itself cannot be written

**Per-command success payloads**

| Command | data fields |
| ------- | ----------- |
| build | name, image |
| deploy | name, namespace, url, image |
| create | name, path, runtime, template |
| delete | name, namespace |
| describe | existing `fn.Instance` object |
| list | array of listItems (empty array, not "no functions found", when none) |
| run | address, host, port |
| invoke | response (HTTP body) |
| logs | name, namespace, lines — a **finite snapshot**, not a stream |
| version | existing Version struct |
| languages | array of runtime strings |
| templates | map of runtime → template array |
| environment | existing Environment struct |
| subscribe | name, source, filters |
| config envs / config labels | the configured entries |
| repository list | array of {name, url} |
| repository add/rename/remove | operation result fields |
| completion, tkn-tasks, mcp start | rejected — raw script / raw YAML / long-running stdio protocol |
| config git | rejected — not implemented yet, so it has no result to report |

### `func logs --json`

A structured response has to be finite, so `--json` reports a snapshot of the
logs available at the time of the call (`Follow: false`, no pod watch) rather
than refusing to run. New `k8s.GetPodLogsSnapshotBySelector` /
`knative.GetKServiceLogsSnapshot` back this. Streaming remains the default for
the human path until #3999 flips it, at which point `--json` simply rides the
same snapshot code.

## Test plan

* `make test` — green
* `make check` — 0 issues
* `TestJSON_DialectParity` — `--json` and `--output json` emit identical bytes
* `TestJSON_EnvVar` — `$FUNC_JSON` enables JSON mode on the success path
* `TestTemplates_JSON`, `TestLanguages_JSON`, `TestListEnvs` — assert the decoded **payload**, not just that `data` is non-nil
* `pkg/mcp` envelope tests — unwrapping, `status:"error"` classification, malformed and payload-less output

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -202,7 +202,26 @@ func runBuild(cmd *cobra.Command, _ []string, newClient ClientFactory) (err erro
 	}
 	// Stamp is a performance optimization: treat the function as being built
 	// (cached) unless the fs changes.
-	return f.Stamp()
+	if err = f.Stamp(); err != nil {
+		return
+	}
+	if isJSONEnabled(cmd) {
+		image := f.Build.Image
+		if image == "" {
+			image = f.Image
+		}
+		err = writeJSONSuccess(cmd.OutOrStdout(), buildJSONResult{
+			Name:  f.Name,
+			Image: image,
+		})
+	}
+	return
+}
+
+// buildJSONResult is the data payload emitted on success when --json is set.
+type buildJSONResult struct {
+	Name  string `json:"name"`
+	Image string `json:"image,omitempty"`
 }
 
 // warnRegistryInsecureChange checks if the registry has changed but
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -210,7 +210,7 @@ func runBuild(cmd *cobra.Command, _ []string, newClient ClientFactory) (err erro
 		if image == "" {
 			image = f.Image
 		}
-		err = writeJSONSuccess(cmd.OutOrStdout(), buildJSONResult{
+		err = WriteJSONSuccess(cmd.OutOrStdout(), buildJSONResult{
 			Name:  f.Name,
 			Image: image,
 		})
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"errors"
+	"fmt"
 	"os"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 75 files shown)


lkingland (commented):
This one probably needs to be revisited in light of our recent MCP updates.

describe/list/version already have --output json (raw Instance / ListItem / Version). This adds a second JSON dialect:

`{"apiVersion":"v1","status":"ok","data":{...}}`

MCP (describe, list, version) already parses the first dialect.  We need to pick one and go with it across the board.

Copilot is right on FUNC_JSON.

Tests only check the envelope, not the payload (_ = cmp.Diff), which is a regression.

func logs --json errors as unsupported (stream) needs to be updated to use our "snapshot" (default).

PR #3827, opened 2026-05-21
fix(mcp): prevent hang in config add/remove handlers when required fields are omitted

Ankitsinghsisodya:
> **Note:** Re-submission of #3761, accidentally closed when the head repository was deleted.

## Problem

The `config_envs_add`, `config_envs_remove`, `config_labels_add`, `config_labels_remove`, and `config_volumes_remove` MCP tool handlers could hang indefinitely when required fields were omitted by the caller.

The CLI commands these handlers invoke fall through to interactive survey prompts when flags are absent. These prompts block on stdin. The MCP subprocess has no TTY, so the call deadlocks and never returns.

## Root cause

The affected fields were declared as `*string` with `omitempty` in the JSON schema, making them appear optional to MCP clients. The CLI, however, requires them to operate non-interactively.

## Fix

Correct the schema to reflect the actual requirements of each command:

* `config_envs_add`: `value` → required `string`. `name` stays `*string` (optional) for bulk-import syntax.
* `config_envs_remove`: `name` → required `string`.
* `config_labels_add`: both `name` and `value` → required `string`.
* `config_labels_remove`: `name` → required `string`.
* `config_volumes_remove`: `mountPath` → required `string`.

## Tests

* Updated `_Error` tests to supply all required fields
* Added missing-field tests for every affected tool: `TestTool_ConfigEnvsAdd_MissingValue`, `TestTool_ConfigEnvsRemove_MissingName`, `TestTool_ConfigLabelsAdd_MissingName`, `TestTool_ConfigLabelsAdd_MissingValue`, `TestTool_ConfigLabelsRemove_MissingName`, `TestTool_ConfigVolumesRemove_MissingMountPath`
* Added `TestTool_ConfigEnvsAdd_BulkImport` confirming the optional-name bulk-import path still works

### DIFF
--- a/pkg/mcp/tools_config_envs.go
+++ b/pkg/mcp/tools_config_envs.go
@@ -188,6 +188,10 @@ type ConfigEnvsRemoveOutput struct {
 }
 
 func (s *Server) configEnvsRemoveHandler(ctx context.Context, r *mcp.CallToolRequest, input ConfigEnvsRemoveInput) (result *mcp.CallToolResult, output ConfigEnvsRemoveOutput, err error) {
+	if input.Name == "" {
+		err = fmt.Errorf("'name' must not be empty")
+		return
+	}
 	out, err := s.executor.Execute(ctx, "config", input.Args()...)
 	if err != nil {
 		err = fmt.Errorf("%w\n%s", err, string(out))
--- a/pkg/mcp/tools_config_envs.go
+++ b/pkg/mcp/tools_config_envs.go
@@ -86,6 +86,9 @@ type ConfigEnvsAddInput struct {
 // constructed and forwarded to the CLI. Character-set validation is deferred to
 // func's own parser, keeping this layer as a thin pass-through.
 func (i ConfigEnvsAddInput) validate() error {
+	if i.Value == nil && i.SecretName == nil && i.ConfigMapName == nil {
+		return fmt.Errorf("at least one of value, secretName, or configMapName must be provided")
+	}
 	if i.Value != nil && (i.SecretName != nil || i.ConfigMapName != nil) {
 		return fmt.Errorf("value is mutually exclusive with secretName/configMapName; provide one or the other")
 	}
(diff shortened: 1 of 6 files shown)


pkg/mcp/tools_config_envs.go
```diff
@@ -86,6 +86,9 @@ type ConfigEnvsAddInput struct {
 // constructed and forwarded to the CLI. Character-set validation is deferred to
 // func's own parser, keeping this layer as a thin pass-through.
 func (i ConfigEnvsAddInput) validate() error {
+	if i.Value == nil && i.SecretName == nil && i.ConfigMapName == nil {
+		return fmt.Errorf("at least one of value, secretName, or configMapName must be provided")
```
lkingland, line 90: Copilot's right on this one: `value: ""` (non-nil empty) still counts as provided and can reach the CLI. Let's reject empty strings for value/secretName/configMapName, same as the empty-name guards on labels/volumes.

pkg/mcp/tools_config_volumes_test.go
```diff
@@ -175,6 +175,50 @@ func TestTool_ConfigVolumesRemove(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	result, err := client.CallTool(t.Context(), &mcp.CallToolParams{
+		Name:      "config_volumes_remove",
+		Arguments: map[string]any{"path": "."},
+	})
+	// Schema validation may produce a protocol-level error or a tool-level error result.
+	if err == nil && !result.IsError {
```
lkingland, line 214: nit: if `err` and `result` are both nil this panics. `err == nil && (result == nil || !result.IsError)` is enough (same in the other Missing* tests).

pkg/mcp/tools_config_volumes.go
```diff
@@ -111,14 +111,13 @@ var configVolumesRemoveTool = &mcp.Tool{
 }
 
 type ConfigVolumesRemoveInput struct {
-	Path      string  `json:"path" jsonschema:"required,Path to the function project directory"`
-	MountPath *string `json:"mountPath,omitempty" jsonschema:"Mount path of the volume to remove"`
-	Verbose   *bool   `json:"verbose,omitempty" jsonschema:"Enable verbose logging output"`
+	Path      string `json:"path" jsonschema:"required,Path to the function project directory"`
```
lkingland, line 114: Could this be needed on `config volumes add` as well.

lkingland (commented):
In the right ballpark.

A few issues from me, and a few from Copilot as well, to address.

PR #3828, opened 2026-05-21
fix(mcp): block build --push in read-only MCP mode

Ankitsinghsisodya:
> **Note:** Re-submission of #3741, accidentally closed when the head repository was deleted.

## Summary

When the MCP server runs in read-only mode (default unless `FUNC_ENABLE_MCP_WRITE` enables writes), `deploy` and `delete` are already rejected. `build` with `push: true` could still push images to a registry. This change rejects that path with a clear error.

## Motivation

Read-only mode is meant to avoid mutating external state. Pushing a built image to a registry is such a mutation and should align with the same policy as deploy/delete.

## Changes

* `pkg/mcp/tools_build.go`: in `buildHandler`, return an error when `s.readonly` is true and the tool input requests push (`Push != nil && *input.Push`).

## Testing

* `go test ./pkg/mcp/...` or `make test`

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -14,6 +14,11 @@ const (
 	name    = "func"
 	title   = "func"
 	version = "0.1.0"
+
+	// EnvMCPWrite is the environment variable that enables write operations
+	// (build, deploy, delete) on the MCP server. Set to "true" to allow
+	// mutations; the server runs in read-only mode by default.
+	EnvMCPWrite = "FUNC_ENABLE_MCP_WRITE"
 )
 
 // NOTE: Invoking prompts in some interfaces (such as Claude Code) when all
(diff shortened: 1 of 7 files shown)


matejvasek:
There is a merge conflict.
Also copilot suggested some changes.

pkg/mcp/mcp.go
```diff
@@ -14,6 +14,11 @@ const (
 	name    = "func"
 	title   = "func"
 	version = "0.1.0"
+
+	// EnvMCPWrite is the environment variable that enables write operations
+	// (build, deploy, delete) on the MCP server. Set to "true" to allow
+	// mutations; the server runs in read-only mode by default.
```
lkingland: After this PR, local `build` (no push) is not a write. Please drop build from this list, and add a line to `instructions_warning.md` that push stays disabled in RO (alongside deploy/delete).

lkingland (commented):
Looks good.

Just needs a small change to the warnings that `build` is not RO, but `build --push` is (see below).

And a rebase.

lkingland (commented):
/approve
/lgtm

PR #3831, opened 2026-05-21
pkg/mcp: add tests for readonly enforcement, delete validation, and help resources

Ankitsinghsisodya:
* `tools_deploy_test.go` and `tools_delete_test.go` each had a single happy-path test that manually set `server.readonly = false`, leaving the readonly guard in both handlers completely untested.
* The mutual-exclusion validation in `deleteHandler` (exactly one of `path` or `name` must be provided) had no test coverage for either branch.
* All 14 `func://help/*` resources registered in `mcp.go` had zero tests; the `newHelpResource` handler — URI construction, executor dispatch, args assembly, and response shape — was entirely unexercised.

### DIFF
--- a/pkg/mcp/tools_delete_test.go
+++ b/pkg/mcp/tools_delete_test.go
@@ -8,6 +8,70 @@ import (
 	"knative.dev/func/pkg/mcp/mock"
 )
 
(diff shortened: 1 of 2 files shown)


Ankitsinghsisodya:
cc: @lkingland

PR #3832, opened 2026-05-21
mcp: use build-time version instead of hardcoded "0.1.0"

Ankitsinghsisodya:
> **Note:** Re-submission of #3662, accidentally closed when the head repository was deleted.

## Summary

* The MCP server advertised a static `version = "0.1.0"` constant to every MCP client regardless of the actual binary version, causing stale version metadata in clients that display server info.
* Replace the constant with `version.Vers` from `pkg/version`, which is already injected at build time via ldflags (`-X knative.dev/func/pkg/version.Vers=$(VERS)`).
* Fall back to `"0.0.0+source"` when the variable is empty (source builds that bypass the Makefile), consistent with how `cmd/root.go` handles the same case.

## Changes

* `pkg/mcp/mcp.go`: remove `const version = "0.1.0"`, import `knative.dev/func/pkg/version`, and use `version.Vers` (with fallback) when constructing the `mcp.Implementation` passed to `mcp.NewServer`.

## Test plan

* `go test ./pkg/mcp/...` passes
* `func version` and `func mcp start` report the same version string
* MCP client (e.g. Claude Code) shows the correct version after `func mcp start`


pkg/mcp/mcp.go
```diff
@@ -79,6 +80,7 @@ func WithReadonly(readonly bool) Option {
 func New(options ...Option) *Server {
 	s := &Server{
 		prefix:    "func",
+		version:   version.Get().String(),
```
lkingland: **Makefile does not inject a clean semver — this is the important bit.**

`make build` exports `GOFLAGS` with:

```make
VERS ?= $(shell git describe --tags --match 'v*' 2>/dev/null)
LDFLAGS := -X knative.dev/func/pkg/version.Vers=$(VERS) ...
```

and the `func` recipe is just `go build ./cmd/func` (versioning is only via those ldflags).

What `git describe --match 'v*'` actually produces:

| Situation | `version.Vers` |
|-----------|----------------|
| HEAD is a `v*` tag | `v0.49.0` |
| **Typical day** (N commits after a tag) | `v0.49.0-139-gf71e9cb5` |
| No matching tags / `go build` without `make` | empty → CLI already falls back to `v0.0.0+source` |

(The Makefile comment about `'tip'` / `TAG` / `VTAG` is stale — those vars are computed and never used. Not this PR's job to fix, but it means the *code* is `git describe`, not "semver or tip".)

Masterminds v1 *will* parse `v0.49.0-139-gf71e9cb5` (prerelease = `139-gf71e9cb5`). Then `String()` **strips the leading `v`**:

| Surface | String |
|---------|--------|
| `func version` / ldflags / `Original()` | `v0.49.0-139-gf71e9cb5` |
| MCP `Implementation.Version` + healthcheck (`Get().String()`) | `0.49.0-139-gf71e9cb5` |
| #3996 `version` tool (CLI JSON) | `v…` again |

The PR test plan says *"`func version` and `func mcp start` report the same version string."* That's false for almost every `make`-built binary (anything not exactly on a tag).

The git-describe suffix is the useful part of the stamp and is **not** a product version. Parsing as semver doesn't buy MCP anything except splitting the spelling from the CLI.

**Please use the ldflags string as-is**, same contract as `cmd/version.go`:

```go
vers := version.Vers
if vers == "" {
    vers = version.DefaultVers // "v0.0.0+source"
}
s.version = vers
```

Then Implementation, healthcheck, `func version`, and the #3996 tool all rhyme. No `Get()`, no parse, no `v`-strip.

(Caching on `Server` at `New()` is the right idea either way — keep that.)

pkg/version/version.go
```diff
@@ -1,3 +1,43 @@
...
+}
+
+// Get returns the parsed semver for this binary.  When no build-time version
+// was injected via ldflags, DefaultVers is used.  If the injected string is
+// unparseable, DefaultVers is used as a safe fallback.
+// String() returns a clean semver without the leading "v" (e.g. "0.0.0+source"),
+// suitable for machine-readable consumers such as the MCP server.
+// Original() round-trips the injected string verbatim, preserving the leading
+// "v" preferred by human-readable output.
+func Get() *semver.Version {
```
lkingland: This `Get()` layer is the source of the split: `String()` is defined as "clean semver without leading `v`", but Makefile `VERS` is `git describe`, not a clean tag.

Two more concerns if this stays:

1. **Invalid-ldflags fallback** — if `Vers` is somehow unparseable, we silently advertise `v0.0.0+source` while `func version` still prints the raw stamp. Hiding a broken release stamp is worse than showing it. The Makefile already has a fallback (empty → `DefaultVers` in the CLI). Don't add a second, quieter one.

2. **Scope** — MCP advertising `0.1.0` doesn't require a new version API, `pkg/app` changes, or `cmd.DefaultVersion` aliasing. A 4-line empty-check next to `New()` is the whole bugfix. Happy to have a shared helper *if* it just does "Vers or DefaultVers, verbatim."

The init-time parse of `DefaultVers` (panic if the constant is wrong) is fine *if* we keep a constant — just don't parse the injected `git describe` string.

pkg/app/app.go
```diff
@@ -32,7 +32,7 @@ func Main() {
 	cfg := cmd.RootCommandConfig{
 		Name: "func",
 		Version: cmd.Version{
-			Vers: version.Vers,
+			Vers: version.Get().Original(),
```
lkingland: This is a no-op when `Vers` is a valid describe/tag string (`Original()` == what ldflags injected), and a **behavior change** when it isn't: CLI would have printed the raw `Vers`, now it prints DefaultVers.

I'd drop this change unless we're committing to `Get()` as the one version API (see other comments — I'd rather we didn't, for MCP). `pkg/app` can keep passing `version.Vers`; `cmd/version.go` already fills empty with `DefaultVersion`.

pkg/mcp/tools_healthcheck.go
```diff
@@ -21,7 +21,7 @@ func (s *Server) healthcheckHandler(ctx context.Context, r *mcp.CallToolRequest,
 	output = HealthcheckOutput{
 		Status:  "ok",
 		Message: "The MCP server is running!",
-		Version: version,
+		Version: s.version,
```
lkingland: Good that healthcheck and `Implementation.Version` share `s.version` — that's the right pairing.

Once `s.version` is the raw ldflags string (or `DefaultVers`), please also make sure the schema/description still says this is the **MCP server / this binary**, not the remote cluster, and that it matches `func version`. #3996 is the tool for "client I'm driving"; this field is "server that's answering." After both land, agents should see the *same* spelling on both (today this PR would teach them `1.2.3` vs `v1.2.3`).

lkingland:
Let's keep the `version` package a simple "string-stamp" library.  Move the default from the CLI in there like you did.  Set `Vers` to that default.  Then it's up to the user if they want it as a Semver or not.

Finally, we'd need a patch the Makefile to ensure it doesn't overwrite the default with an empty value (`-X pkg.Var=` with an empty value *clears* the Go initializer).

Here's the whole `version` package:

```go
package version

const Default = "v0.0.0+source"

var (
	Vers = Default // overwritten by ldflags when make actually has a describe
	Kver string
	Hash string
)
```

No `Get()`, no Masterminds. Semver stays an exercise for the caller (`func version` already parses if it needs to).

And here's a proper conditional for the Makefile:

```make
LDFLAGS :=
ifneq ($(VERS),)
LDFLAGS += -X knative.dev/func/pkg/version.Vers=$(VERS)
endif
ifneq ($(KVER),)
LDFLAGS += -X knative.dev/func/pkg/version.Kver=$(KVER)
endif
ifneq ($(HASH),)
LDFLAGS += -X knative.dev/func/pkg/version.Hash=$(HASH)
endif
```

MCP server can load it in on start:
```Go
s := &Server{
	prefix:  "func",
	version: version.Vers,
	// ...
}

```

Healthcheck can just use the member directly:
```go
output = HealthcheckOutput{
	Status:  "ok",
	Message: "The MCP server is running!",
	Version: s.version,
}
```

CLI can use the strings directly, no `.Get().Original()` confusion:
```go
// pkg/app/app.go
Version: cmd.Version{
	Vers: version.Vers,
	Kver: version.Kver,
	Hash: version.Hash,
}
```

```go
// cmd/root.go
const DefaultVersion = version.Default
```

Furthermore, the check in `cmd.version.go` can be removed because of the Makefile hardening:

`if v.Vers == "" { v.Vers = DefaultVersion }` 

Lastly, we can keep the leading `v` for now because a semver is a simple `strings.TrimPrefix` away, and this way we're coherent with Git/Go/Knative/Kubernetes.

lkingland (changes requested):
Let's keep the `version` package a simple "string-stamp" library.  Move the default from the CLI in there like you did.  Set `Vers` to that default.  Then it's up to the user if they want it as a Semver or not.

Finally, we'd need a patch the Makefile to ensure it doesn't overwrite the default with an empty value (`-X pkg.Var=` with an empty value *clears* the Go initializer).

Here's the whole `version` package:

```go
package version

const Default = "v0.0.0+source"

var (
	Vers = Default // overwritten by ldflags when make actually has a describe
	Kver string
	Hash string
)
```

No `Get()`, no Masterminds. Semver stays an exercise for the caller (`func version` already parses if it needs to).

And here's a proper conditional for the Makefile:

```make
LDFLAGS :=
ifneq ($(VERS),)
LDFLAGS += -X knative.dev/func/pkg/version.Vers=$(VERS)
endif
ifneq ($(KVER),)
LDFLAGS += -X knative.dev/func/pkg/version.Kver=$(KVER)
endif
ifneq ($(HASH),)
LDFLAGS += -X knative.dev/func/pkg/version.Hash=$(HASH)
endif
```
(be sure to keep the existing `LDFLAGS += -X` ... SocatImage / TarImage / FuncUtilImage lines after it)

MCP server can load it in on new:
```Go
s := &Server{
	prefix:  "func",
	version: version.Vers,
	// ...
}

i := mcp.NewServer(&mcp.Implementation{
    Name:    name,
    Title:   title,
    Version: s.version, // same stamp
}, ...)
```

Healthcheck can just use the member directly:
```go
output = HealthcheckOutput{
	Status:  "ok",
	Message: "The MCP server is running!",
	Version: s.version,
}


```

CLI can use the strings directly, no `.Get().Original()` confusion:
```go
// pkg/app/app.go
Version: cmd.Version{
	Vers: version.Vers,
	Kver: version.Kver,
	Hash: version.Hash,
}
```

```go
// cmd/root.go
const DefaultVersion = version.Default
```

Furthermore, the check in `cmd/version.go` can be removed because of the Makefile hardening:

`if v.Vers == "" { v.Vers = DefaultVersion }` 

Lastly, we can keep the leading `v` for now because a semver is a simple `strings.TrimPrefix` away in most cases, and this way we're coherent with Git/Go/Knative/Kubernetes.

Ankitsinghsisodya:
@lkingland Can you re run the tests?

Ankitsinghsisodya:
> @lkingland Can you re run the tests?

@lkingland

PR #3834, opened 2026-05-21
fix(k8s): propagate caller context in dialer DialContext

Elvand-Lie:
Replaces #3803 with the requested fixes from #3772.

1. Gives the exec stream its own background-derived context instead of reusing the dial connection context.
2. Cancels the exec stream when the connection pipes close.
3. Adds an integration test (\TestInt_DialContextExpiry\) to verify the connection survives after the dial context expires.

### DIFF
--- a/pkg/k8s/dialer.go
+++ b/pkg/k8s/dialer.go
@@ -106,8 +106,10 @@ func (c *contextDialer) DialContext(ctx context.Context, network string, addr st
 	case <-connectSuccess:
 		return conn, nil
 	case err := <-connectFailure:
+		execCancel()
 		return nil, err
 	case <-ctx.Done():
+		execCancel()
 		_ = conn.closeWithError(ctx.Err())
 		return nil, ctx.Err()
 	}
--- a/pkg/k8s/dialer.go
+++ b/pkg/k8s/dialer.go
@@ -77,11 +77,18 @@ func (c *contextDialer) DialContext(ctx context.Context, network string, addr st
 	ctrStdin, ctrStdout, conn := newConn()
 	connectSuccess := make(chan struct{})
 	connectFailure := make(chan error, 1)
+	// The exec stream must outlive the dial context: per net.Dialer.DialContext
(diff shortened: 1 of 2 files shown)


pkg/k8s/dialer.go
```diff
@@ -77,11 +77,18 @@ func (c *contextDialer) DialContext(ctx context.Context, network string, addr st
...
 	connectSuccess := make(chan struct{})
 	connectFailure := make(chan error, 1)
+	// The exec stream must outlive the dial context: per net.Dialer.DialContext
+	// semantics, ctx governs connection *establishment* only. Once connected,
+	// expiry of ctx must not tear down the tunnel. We therefore give exec its
+	// own background-derived context and cancel it when the goroutine exits
+	// (which happens when the connection's pipes are closed).
+	execCtx, execCancel := context.WithCancel(context.Background())
 	go func() {
+		defer execCancel()
```
matejvasek, line 87: At this point in execution the `c.exec()` call already exited so canceling the context does not do much (but still should be done just to be sure and so linters do not complain.

However you might want to cancel that context in the error return branches.

pkg/k8s/dialer_int_test.go
```diff
@@ -247,3 +248,131 @@ func TestInt_DialUnreachable(t *testing.T) {
...
+
+	transport := &http.Transport{
+		DialContext: dialer.DialContext,
+	}
+	client := http.Client{Transport: transport}
+
+	svcURL := fmt.Sprintf("http://%s.%s.svc", svc.Name, svc.Namespace)
+
+	// First request — establishes the connection while dialCtx is alive.
+	req, err := http.NewRequestWithContext(dialCtx, "GET", svcURL, nil)
```
matejvasek: I am afraid this context is not only used as context for the dial operation but also for whole request lifetime.
You might need to do HTTP request "by hand" using lower level API.

Elvand-Lie:
I see that now. Updated to use a raw et.Conn instead and added the cancel calls. Let me know if this solution works @matejvasek.

PR #3847, opened 2026-05-25
fix(cmd): move file read from newInvokeConfig to runInvoke

### LINKED ISSUE #3846: cmd: invoke reads --file data from disk twice (redundant I/O)
## Description

When `--file` is passed to `func invoke`, the file contents are read from disk twice:

1. In `newInvokeConfig()` (line 280-286) — reads file into `cfg.Data`
2. In `runInvoke()` (line 201-207) — reads the same file again into `m.Data`

The second read is redundant because `m.Data` is already initialized from `cfg.Data` at line 195, which already contains the file contents from the first read. The second read at line 206 just overwrites `m.Data` with identical content.

## Impact

- Wasted I/O (file read twice for every invocation with `--file`)
- TOCTOU consistency risk: if the file changes between the two reads, the behavior would be unpredictable

## Proposed Fix

Remove the second `os.ReadFile` block in `runInvoke()` (lines 200-207). The file is already read into `cfg.Data` during config creation.

/kind cleanup

Elvand-Lie:
Fixes https://github.com/knative/func/issues/3846

## Problem

When `--file` is passed to `func invoke`, the file contents were read from disk twice:

1. In `newInvokeConfig()` — reads file into `cfg.Data`
2. In `runInvoke()` — reads the same file again into `m.Data`

The first read in `newInvokeConfig()` is problematic because it occurs before the `--confirm` interactive prompts. If the user changes the file path during these prompts, the original file data has already been loaded, causing stale data to be sent. 

## Changes

This PR removes the early file read and ensures the file is read only once, at the correct time:

- **Removed early read**: Eliminated the `os.ReadFile` block from `newInvokeConfig()`.
- **Relocated read logic**: Moved the file read into `runInvoke()`, executing it after the configuration is finalized and immediately before client initialization for fail-fast behavior.
- **Added testing**: Introduced `TestInvoke_FileFlag` and `TestInvoke_FileFlagNonExistent` with `sync.Mutex` synchronization to explicitly test file-based invocation and error handling.

### Testing

- `TestInvoke_FileFlag` — Verifies that providing a valid file correctly reads and sends its contents to the mock server.
- `TestInvoke_FileFlagNonExistent` — Verifies that providing a non-existent file path immediately returns the appropriate error.

### DIFF
--- a/cmd/invoke.go
+++ b/cmd/invoke.go
@@ -176,6 +176,15 @@ func runInvoke(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 			return fmt.Errorf("--extension (-e) is only valid with cloudevents")
 		}
 	}
+	// If --file was specified, read the file contents into cfg.Data now
+	// that all config (including interactive --confirm prompts) is finalized.
+	if cfg.File != "" {
+		content, err := os.ReadFile(cfg.File)
+		if err != nil {
+			return err
+		}
+		cfg.Data = content
+	}
 
 	// Client instance from env vars, flags, args and user prompts (if --confirm)
 	client, done := newClient(ClientConfig{Verbose: cfg.Verbose, InsecureSkipVerify: cfg.Insecure})
@@ -196,15 +205,6 @@ func runInvoke(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 		Data:        cfg.Data,
 		Format:      cfg.Format,
 		Extensions:  exts,
-	}
-
-	// If --file was specified, use its content for message data
-	if cfg.File != "" {
-		content, err := os.ReadFile(cfg.File)
-		if err != nil {
-			return err
-		}
-		m.Data = content
 	}
 
 	// Invoke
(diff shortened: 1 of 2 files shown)


gauron99:
The removed `os.ReadFile` in `runInvoke()` is the only code that reads the file after `--confirm` interactive prompts may have changed `cfg.File`. Two scenarios break:

1. `func invoke --file=a.txt --confirm` → user changes file to `b.txt` in the prompt → `a.txt` contents get sent
2. `func invoke --confirm` → user types `mydata.txt` in the File prompt → file is never read, default `{"message":"Hello World"}` gets sent

I suspect the first `if cfg.File != ""` read is actually redundant. We want to read the file after the whole config is constructed and its values finalized. The first read in `newInvokeConfig()` most likely served a purpose of displaying data from a file if it was already provided previously, but if File is provided the Data prompt is skipped entirely. There might be some simplification possible here. Make sure to test this locally and if no integration tests exist, create them.

Elvand-Lie:
@gauron99 Right. Since this was meant to be a cleanup, I saw the two os.ReadFile blocks and assumed the second one in runInvoke() was redundant because cfg.Data was already populated. I didn't account for the interactive --confirm prompts mutating cfg.File between the two reads, meaning the first read pulls stale data. My bad.

I've pushed a fix:
1. Removed the early read from newInvokeConfig().
2. Moved the os.ReadFile back into runInvoke() so it reads the file only after the config and prompts are finalized.
3. Added TestInvoke_FileFlag and TestInvoke_FileFlagNonExistent to explicitly test this behavior against the mock runner.
4. Fixed formatting with goimports to pass the CI style check.

PR #3851, opened 2026-05-25
fix(cmd): add deployer type guard to logs command

### LINKED ISSUE #3850: cmd: logs command silently fails for keda/raw deployed functions (hardcoded Knative selector)
## Description

The `func logs` command unconditionally calls `knative.GetKServiceLogs()`, which uses a `serving.knative.dev/service` label selector internally:

```go
// pkg/knative/logs.go:20
selector := fmt.Sprintf("serving.knative.dev/service=%s", kServiceName)
```

This selector only matches pods created by the Knative deployer. However, functions can be deployed with three different deployers (`knative`, `raw`, or `keda`), as defined in `DeploySpec.Deployer`.

For `raw` or `keda` deployed functions, no pods match the selector, causing the command to **silently produce no output** or return a confusing Kubernetes API error. The user gets no indication that the command simply doesn't support their deployer type.

The `logs` command already retrieves the function's deployed details via `client.Describe()`, which returns an `Instance` struct containing the `Deployer` field — but this field is never checked.

## Expected Behavior

When running `func logs` against a function deployed with a non-Knative deployer, the user should see a clear error message explaining the limitation and suggesting alternatives (e.g., `kubectl logs`).

## Proposed Fix

After retrieving the function instance via `client.Describe()`, check `instance.Deployer`:
- If `"knative"` or `""` (default): proceed with existing `knative.GetKServiceLogs()`
- Otherwise: return a clear error message

/kind bug

Elvand-Lie:
Fixes https://github.com/knative/func/issues/3850

## Changes

The `func logs` command unconditionally called `knative.GetKServiceLogs()`, which uses a `serving.knative.dev/service` label selector that only matches pods created by the Knative deployer.

For functions deployed with the `keda` or `raw` deployer, no pods match the selector, causing the command to silently produce no output. The user gets no indication that the command doesn't support their deployer type.

### What this PR does:

- After retrieving the function instance via `client.Describe()`, store the `Deployer` field
- Before calling `knative.GetKServiceLogs()`, check the deployer type
- If the deployer is not `"knative"` (or empty, which defaults to knative), return a clear error message directing users to use `kubectl logs` instead

### Example error output for unsupported deployer:

```
Error: 'func logs' is not yet supported for the "keda" deployer.
Currently only functions deployed with the default Knative deployer are supported.
You can use 'kubectl logs' directly to view logs for keda-deployed functions
```

### Testing

The guard is a simple string check on the `Deployer` field already returned by `client.Describe()`. No new dependencies or API calls are introduced.

/kind bug

### DIFF
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -71,8 +71,9 @@ func runLogs(cmd *cobra.Command, newClient ClientFactory) error {
 	client, done := newClient(ClientConfig{Verbose: cfg.Verbose})
 	defer done()
 
-	// Get function details
+	// Get function details and deployer type
 	var f fn.Function
+	var deployer string
 	if cfg.Name != "" {
 		// Get function by name
 		instance, err := client.Describe(cmd.Context(), cfg.Name, cfg.Namespace, fn.Function{})
@@ -82,6 +83,7 @@ func runLogs(cmd *cobra.Command, newClient ClientFactory) error {
 		f.Name = instance.Name
 		f.Namespace = instance.Namespace
 		f.Image = instance.Image
+		deployer = instance.Deployer
 	} else {
 		// Load function from path
 		f, err = fn.NewFunction(cfg.Path)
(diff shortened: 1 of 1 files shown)


gauron99 (approved):
/lgtm

PR #3852, opened 2026-05-25
fix: return error for malformed --extension flag values

Ankitsinghsisodya:
## Problem

`extensionsMap()` in `cmd/invoke.go` silently dropped any `--extension` value that did not contain `=`. The invocation proceeded with fewer extensions than the user specified, with no feedback — particularly confusing when debugging CloudEvent routing.

## Changes

**`cmd/invoke.go`**
- Changed `extensionsMap()` signature from `map[string]string` to `(map[string]string, error)`.
- Returns a descriptive error for any entry missing `=` (e.g. `invalid --extension "badformat": must be in key=value format`).
- Renamed the shadowing local variable `extensionsMap` → `result` for clarity.
- Updated the call site to propagate the error before constructing `InvokeMessage`.

**`cmd/invoke_test.go`**
- `TestInvokeExtensionsMapValid`: verifies correct parsing including values that themselves contain `=` (e.g. base64 strings like `foo=bar=baz`).
- `TestInvokeExtensionsMapMalformed`: verifies that a missing `=` produces an error rather than a silent drop.

## Testing

```
go test ./cmd/... -run TestInvokeExtensionsMap
```

### DIFF
--- a/cmd/invoke.go
+++ b/cmd/invoke.go
@@ -186,7 +186,7 @@ func runInvoke(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 
 	// Message to send the running function built from parameters gathered
 	// from the user (or defaults)
-	exts, err := cfg.extensionsMap()
+	exts, err := cfg.parseExtensions()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


cmd/invoke.go
```diff
@@ -315,15 +319,16 @@ func newInvokeConfig() (cfg invokeConfig, err error) {
...
 
-func (c invokeConfig) extensionsMap() map[string]string {
-	extensionsMap := make(map[string]string)
+func (c invokeConfig) parseExtensions() (map[string]string, error) {
+	result := make(map[string]string)
 	for _, ext := range c.Extensions {
 		parts := strings.SplitN(ext, "=", 2)
-		if len(parts) == 2 {
-			extensionsMap[parts[0]] = parts[1]
+		if len(parts) != 2 || strings.TrimSpace(parts[0]) == "" {
```
matejvasek, line 326: What if parts is empty array here?

matejvasek: OK, that probably should not happen, my bad.

PR #3665, opened 2026-05-06
Fix remote Quarkus builds failing on s390x and ppc64le clusters

Itx-Psycho0:
This fix addresses the issue where remote Quarkus builds fail on s390x and ppc64le clusters.

The problem was that FSGroup permissions were not being honored on the build PersistentVolume, which prevented the build process from accessing its workspace correctly.

The solution adds a podTemplate with proper security context settings to both the Pack and S2I PipelineRun templates. This ensures that all Tekton task pods inherit the correct security context for volume access. On s390x and ppc64le architectures, some storage backends require explicit FSGroup configuration to properly handle volume permissions.

What changed:
- Added podTemplate with securityContext to the Pack builder PipelineRun templates
- Added podTemplate with securityContext to the S2I builder PipelineRun templates
- Added a test to verify the podTemplate with securityContext is present in generated PipelineRuns

The security context is configured with:
- runAsUser: 1001 (matches the Tekton buildpack task requirements)
- runAsGroup: 0 (matches the Tekton buildpack task requirements)
- fsGroup: 1002 (ensures proper volume ownership for non-root users)

All tests pass and linting checks are clean. This fix is compatible with all architectures and will not affect existing functionality on amd64 systems.

Fixes #3515

PR #3855, opened 2026-05-25
Add strict Tekton schema validation for Pipeline and PipelineRun templates

matejvasek:
### Summary

- Add `TestPipelineTemplatesValidate` and `TestPipelineRunTemplatesValidate` which render all Tekton YAML templates and decode them into typed `tektonv1.Pipeline` / `tektonv1.PipelineRun` structs using a **strict deserializer** that rejects unknown fields
- Replace the string-based `Test_PipelineRunHasPodTemplateSecurityContext` (from #3665) with these schema-level validations

### Motivation

PR #3665 added `spec.podTemplate` to PipelineRun templates, but this field is only valid in the Tekton `v1beta1` API. In `v1`, it must be nested under `spec.taskRunTemplate.podTemplate`. The existing tests used `strings.Contains` on rendered YAML, which cannot catch misplaced fields — the string was present in the file, but Kubernetes silently ignored it (`Warning: unknown field "spec.podTemplate"`).

These new tests would have caught that bug at CI time by failing with a strict decode error on the unknown field.

### What changed

- **`TestPipelineTemplatesValidate`**: renders `packPipelineTemplate` and `s2iPipelineTemplate` with real inline task specs, strict-decodes into `tektonv1.Pipeline`
- **`TestPipelineRunTemplatesValidate`**: renders all 4 PipelineRun templates (pack/s2i × standard/PAC), strict-decodes into `tektonv1.PipelineRun`
- Shared helpers `strictTektonDecoder` and `renderTemplate` extracted to reduce duplication
- Removed `Test_PipelineRunHasPodTemplateSecurityContext` and its unused imports (`os`, `strings`)

### Test plan

- [x] `go test ./pkg/pipelines/tekton/ -run TestPipelineTemplatesValidate` passes
- [x] `go test ./pkg/pipelines/tekton/ -run TestPipelineRunTemplatesValidate` passes
- [x] All existing tests in the package continue to pass

### DIFF
--- a/pkg/pipelines/tekton/templates_test.go
+++ b/pkg/pipelines/tekton/templates_test.go
@@ -328,6 +328,80 @@ func Test_createAndApplyPipelineRunTemplate(t *testing.T) {
 	}
 }
 
+// strictTektonDecoder returns a strict deserializer that rejects unknown fields
+// in Tekton v1 resources (Pipeline, PipelineRun, etc.).
+func strictTektonDecoder(t *testing.T) runtime.Decoder {
+	t.Helper()
+	myScheme := runtime.NewScheme()
+	if err := tektonv1.AddToScheme(myScheme); err != nil {
+		t.Fatal(err)
+	}
+	codecs := serializer.NewCodecFactory(myScheme, serializer.EnableStrict)
+	return codecs.UniversalDeserializer()
+}
+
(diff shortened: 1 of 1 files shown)


gauron99 (approved):
/lgtm

PR #3856, opened 2026-05-26
cluster: migrate registry from host container to in-cluster deployment

lkingland:
## Summary

Migrates `func cluster create` from running a standalone `func-registry` container on the host to deploying the registry as in-cluster Kubernetes resources, matching the approach already taken in `hack/cluster.sh` (PR #3718).

**What changed:**
- Registry is now a Deployment + ClusterIP Service + Contour Ingress inside the Kind cluster (was a host-side `docker run` container with ExternalName Service)
- `registryAddr` is `registry.localtest.me` (was `localhost:50000`)
- Containerd mirrors point at `http://localhost:5000` via hostPort (was `http://func-registry:5000` via Docker network DNS)
- Each cluster owns its own registry, destroyed automatically by `kind delete cluster`
- Delete flow simplified: no shared container teardown needed
- Removed: `ensureRegistry`, `registryStatus`, `teardownRegistry`, `setupPodmanMacOSForwarding`, and all host-container lifecycle constants

**What's preserved:**
- Host trust config (Docker `daemon.json` / Podman `registries.conf`) — now uses `registry.localtest.me`
- `--skip-registry-config` flag behavior unchanged
- Last-cluster check for host trust revert
- Kubeconfig stat guard in Delete() to suppress noisy warnings on orphan-cleanup path


CLAUDE.local.md
```diff
@@ -0,0 +1,136 @@
+# Functions — Hugr Agent Instructions
+
+Repository: https://github.com/knative/func
+
+This file is kept in the hugr repo and symlinked into func src workspaces for
+agents.
```
matejvasek: What does this mean? What is `hugr`?

pkg/cluster/registry.go
```diff
@@ -15,34 +15,84 @@ import (
...
+// installRegistry deploys the container registry as in-cluster Kubernetes
+// resources (Deployment + ClusterIP Service + Contour Ingress), configures
+// host-side trust, and applies the local-registry-hosting ConfigMap.
 func installRegistry(ctx context.Context, cfg ClusterConfig, out io.Writer) error {
 	start := time.Now()
 	status(out, "Creating Registry")
 
-	if err := ensureRegistry(ctx, cfg, out); err != nil {
-		return err
+	registryManifest := `apiVersion: apps/v1
```
matejvasek: possible improvement: make this Go struct instead of string literal.

pkg/cluster/registry.go
```diff
@@ -51,114 +101,25 @@ func installRegistry(ctx context.Context, cfg ClusterConfig, out io.Writer) erro
...
-spec:
-  type: ExternalName
-  externalName: %s
-`, registryContainerName)
-
-	if err := applyManifest(ctx, out, cfg, registrySvc); err != nil {
-		return fmt.Errorf("applying registry service: %w", err)
-	}
-
 	success(out, "Registry", time.Since(start))
```
matejvasek, line 59: potential improvement: we could wait until `http://registry.localtest.me/v2/` returns `200`.

pkg/cluster/registry.go
```diff
@@ -15,34 +15,84 @@ import (
...
-var registryAddr = fmt.Sprintf("localhost:%d", registryHostPort)
+const registryAddr = "registry.localtest.me"
 
-// installRegistry starts the shared local container registry, configures
-// host-side trust for it, and applies the in-cluster ConfigMap + Service
-// the kind cluster uses to reach it.
+// installRegistry deploys the container registry as in-cluster Kubernetes
+// resources (Deployment + ClusterIP Service + Contour Ingress), configures
+// host-side trust, and applies the local-registry-hosting ConfigMap.
 func installRegistry(ctx context.Context, cfg ClusterConfig, out io.Writer) error {
```
matejvasek, line 31: Not sure when/how `installRegistry()` is called but it preferably should be called after `Contour` installation (or in parallel to it). But it's not necessary.

matejvasek:
unrelated to this PR: why we use `exec()` instead of using `Go` code, kind/kubectl is `Go` code. Also why call docker via CLI `exec()`? Why not API? We already extensively use k8s and docker API already.

pkg/cluster/registry.go
matejvasek: >But it's not necessary.

Because we use generic `k8s` ingress not some specific `Contour` resource.

lkingland:
> unrelated to this PR: why we use `exec()` instead of using `Go` code, kind/kubectl is `Go` code. Also why call docker via CLI `exec()`? Why not API? We already extensively use k8s and docker API already.

Exactly right!  This is simply a stepping-stone.  Translating script->Go code with this structure is much less error-prone.  Next step is to replace the exec's with code wherever possible!  Same way we did with the builders: wrap the cli, get it working, replace with the stdlib 👍🏻

CLAUDE.local.md
lkingland: That's my clanker.
...  He's still learning.  🤣

pkg/cluster/delete.go
```diff
@@ -8,37 +8,29 @@ import (
...
 	defer setKubeconfig(cfg.Kubeconfig())()
 
-	status(out, "Deleting Cluster")
-
-	if err := run(ctx, out, "",
-		cfg.kind(), "delete", "cluster",
-		"--name="+cfg.Name,
-		"--kubeconfig="+cfg.Kubeconfig()); err != nil {
-		warnf(out, "failed to delete cluster %q: %v", cfg.Name, err)
+	if _, err := os.Stat(cfg.Kubeconfig()); err == nil {
```
gauron99: do we want to not gate the "delete" here on existing config? if there was some transient error during kubeconfig startup that created resources but not the file? 
wouldnt the delete be idempotent?
-- im not sure, just thinking out loud

pkg/cluster/registry.go
```diff
@@ -304,34 +264,9 @@ func configurePodmanHTTP(out io.Writer) error {
...
 
-	setupPodmanMacOSForwarding(out)
 	return nil
 }
 
-// setupPodmanMacOSForwarding sets up SSH port forwarding on macOS so the
-// Podman VM can access the host's local registry. Idempotent: detects an
-// existing backgrounded ssh forwarder and skips rather than spawning
-// another (which would leak or fail to bind).
-func setupPodmanMacOSForwarding(out io.Writer) {
```
gauron99, line 315: do we not need this anymore?

gauron99 (commented):
note from AI: stale comment in`cmd/cluster.go` -> `Deletes a local development cluster and its associated registry container` 😁

lkingland:
🤖 @gauron99 re stale delete help ("registry container"): fixed in f5a260a1 — docs now match in-cluster registry + last-cluster host-trust revert.

pkg/cluster/delete.go
lkingland: 🤖 Keeping the kubeconfig gate: skips noisy kind-delete when nothing was created; host-trust revert still runs below. Full always-delete would be more idempotent but mostly silence vs rare orphan — happy to open a follow-up if you want the louder path.

pkg/cluster/registry.go
lkingland: 🤖 Correct — no longer needed. That SSH forward existed so the Podman VM could reach a host-side registry container. Registry is in-cluster now (hostPort + registry.localtest.me); no host container to tunnel to.

lkingland:
🤖 @matejvasek [re your note on exec vs Go APIs](https://github.com/knative/func/pull/3856#issuecomment-4544007828): agreed this PR stays on the script→Go stepping-stone. Follow-up for kind/kubectl/docker library calls: https://github.com/knative/func/issues/3964

pkg/cluster/registry.go
lkingland: 🤖 Done in 47c768b4 — Deployment/Service/Ingress/ConfigMap are typed k8s objects marshaled to YAML; same apply path.

pkg/cluster/registry.go
lkingland: 🤖 Done in ed43102a — poll until /v2/ returns 200. Uses http://127.0.0.1:5000 (hostPort) rather than registry.localtest.me because Contour installs in parallel with the registry goroutine; hostPort is ready with the Deployment and matches the containerd mirror endpoint.

pkg/cluster/registry.go
lkingland: 🤖 Agreed — leaving order as-is (generic Ingress; Contour still parallel). No change.

gauron99:
Tested end-to-end: built from branch, created a default cluster, verified registry deployment, host access via `registry.localtest.me`, and image push — all working.

Two minor follow-ups filed:
- #3979 — dead `--container-engine` flag on delete
- #3980 — registry unreachable from host with `--serving=false` (no Contour)

/lgtm
/approve

PR #3862, opened 2026-05-28
fix: increase default pvc size

matejvasek:
The previous 256Mi is too low. It worked mostly by accident: KinD uses hostpath storage so there is no size enforcement. On AWS OCP we probably use storage with minimum 1Gi size.

When I tried LVM based volumes I got no space left on device.

/kind bug

```release-note
fix: increased size of pvc used for on cluster build
```


gauron99 (approved):
/lgtm

PR #3864, opened 2026-05-28
feat: implement AWS ECR credentials loader in Kubernetes keychains

### LINKED ISSUE #3863: Feature : Implement AWS ECR credentials loader in Kubernetes keychains
## Description

The Kubernetes keychain implementation currently lacks programmatic AWS Elastic Container Registry (ECR) credential loading support.

While GCP (GCR) and Azure (ACR) registries already support automatic credential resolution through dedicated credential loaders, the ECR loader implementation (`GetECRCredentialLoader`) currently returns an empty slice, preventing Kubernetes keychains workflows from resolving ECR credentials programmatically.

As a result:

* OCI image pull/push workflows relying on Kubernetes keychains cannot automatically authenticate against AWS ECR.
* Users must rely on external Docker authentication setup or manual credential management.
* Behavior is inconsistent compared to existing GCR and ACR integrations.

## Proposed solution

Implement AWS ECR credential resolution support in Kubernetes keychains by:

* Adding registry detection logic for AWS ECR endpoints, including:

  * `public.ecr.aws`
  * Commercial ECR registries
  * FIPS endpoints
  * China partitions
  * GovCloud/Secret partitions

* Integrating the official `amazon-ecr-credential-helper/ecr-login` helper programmatically using `authn.NewKeychainFromHelper()`.

* Supporting graceful fallback behavior by returning `creds.ErrCredentialsNotFound` when no ambient AWS credentials are available, allowing subsequent keychain loaders to continue authentication attempts.

## Expected behavior

Kubernetes keychains should automatically resolve AWS ECR credentials similarly to existing GCR and ACR integrations, enabling seamless OCI registry authentication workflows without requiring manual Docker credential configuration.

## Additional context

Suggested test coverage:

* Validate ECR registry hostname detection across supported AWS partitions.
* Verify non-ECR registries are ignored.
* Ensure fallback behavior works correctly when ambient AWS credentials are unavailable.
* Confirm keychain authentication attempts execute correctly for ECR registries.

Nachiket-Roy:
### Summary 
This change adds programmatic AWS Elastic Container Registry (ECR) authentication support when resolving OCI registry keychains, aligning it with the existing GCP (Google Container Registry) and Azure (ACR) patterns. Previously, the ECR credentials loader was left unimplemented (GetECRCredentialLoader returned an empty slice), which prevented automatic, programmatic ECR credential resolution when pushing/pulling images in Kubernetes keychains workflows.

### What changes were made?
1. Registry Matching (isECRRegistry): Added matching logic to detect if a registry hostname corresponds to an AWS ECR registry. This includes `public.ecr.aws` as well as private registry formats across various AWS partitions.
2. ECR Helper Integration (`GetECRCredentialLoader`): Leveraged the official `amazon-ecr-credential-helper/ecr-login` library programmatically, wrapped via `authn.NewKeychainFromHelper()`.
3. Graceful Fallback: Silenced standard library logging with `io.Discard` and ensured that if no ambient credentials exist, `creds.ErrCredentialsNotFound` is returned, letting subsequent credential loaders in the chain attempt authentication.

### Testing

- Unit test was added (`pkg/k8s/keychains_test.go`)
- Test has been verified locally

Closes : #3863


pkg/k8s/keychains.go
```diff
@@ -44,8 +48,77 @@ func GetGoogleCredentialLoader() []creds.CredentialsCallback {
...
+	isKnownECRDomain := strings.HasSuffix(registry, ".amazonaws.com") ||
+		strings.HasSuffix(registry, ".amazonaws.com.cn") ||
+		strings.HasSuffix(registry, ".sc2s.sgov.gov") ||
+		strings.HasSuffix(registry, ".c2s.ic.gov")
+	if !isKnownECRDomain {
+		return false
+	}
+
+	return strings.Contains(registry, ".dkr.ecr.") || strings.Contains(registry, ".dkr.ecr-fips.")
+}
```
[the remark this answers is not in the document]
Nachiket-Roy: This would be unnecessary, current code matches the GCP and ACR pattern correclty

pkg/k8s/keychains.go
```diff
@@ -44,8 +48,77 @@ func GetGoogleCredentialLoader() []creds.CredentialsCallback {
...
+	return []creds.CredentialsCallback{
+		func(registry string) (oci.Credentials, error) {
+			if !isECRRegistry(registry) {
+				return oci.Credentials{}, creds.ErrCredentialsNotFound
+			}
+
+			res, err := name.NewRegistry(registry)
+			if err != nil {
+				return oci.Credentials{}, fmt.Errorf("parse registry: %w", err)
+			}
```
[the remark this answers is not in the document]
Nachiket-Roy: Not needed as name.NewRegistry() is from the go-containerregistry library

pkg/k8s/keychains.go
```diff
@@ -44,8 +49,104 @@ func GetGoogleCredentialLoader() []creds.CredentialsCallback {
...
 func GetECRCredentialLoader() []creds.CredentialsCallback {
-	return []creds.CredentialsCallback{} // TODO: Implement ECR credentials loader
+	ecrHelper := ecr.NewECRHelper(ecr.WithLogger(io.Discard))
+	keychain := authn.NewKeychainFromHelper(ecrHelper)
+
+	return []creds.CredentialsCallback{
+		func(registry string) (oci.Credentials, error) {
+			if !isECRRegistry(registry) {
+				return oci.Credentials{}, creds.ErrCredentialsNotFound
+			}
```
matejvasek: Do we really need this pre-check? Won't `keychain.Resolve(res)` do the job on it's own?

Nachiket-Roy: To fail fast and same pattern is followed in GCP and ACR

pkg/k8s/keychains.go
```diff
@@ -44,8 +52,165 @@ func GetGoogleCredentialLoader() []creds.CredentialsCallback {
...
+		strings.Contains(errStr, "NoCredentialProviders") ||
+		strings.Contains(errStr, "no AWS credentials")
+}
+
+var (
+	ecrHelper   *ecr.ECRHelper
+	ecrKeychain authn.Keychain
+	ecrInitOnce sync.Once
+	ecrCache    sync.Map // registry (string) -> ecrCacheEntry
+)
```
matejvasek: Why is this global state?

Nachiket-Roy: In non-AWS environments, credential discovery can take ~5 seconds before failing. Caching it globally ensures this only happens once, and all subsequent lookups fail instantly instead of repeating the delay.

matejvasek:
## via Claude

The goroutine+timeout pattern here has a structural problem: when the 5s timeout fires, the function returns but the goroutine running `Resolve()` keeps going — it leaks permanently because `Resolve` doesn't accept a context and there's no way to cancel the underlying HTTP call. Additionally, caching timeout errors means a single transient network blip permanently breaks ECR auth for the process lifetime.

The root cause is that neither `authn.Keychain.Resolve()` nor the ECR helper's `Get()` accept a `context.Context`, so there's no way to do cooperative cancellation through those APIs. Implementing a custom `ClientFactory` to inject HTTP timeouts is also disproportionately complex (5-method interface).

I'd suggest bypassing the ECR helper entirely and calling the AWS ECR SDK directly. Both `LoadDefaultConfig` and `GetAuthorizationToken` accept a context natively — no goroutine, no leak, no fake timeout. Something along these lines:

```go
func extractRegion(registry string) string {
	// hostname format: <account>.dkr.ecr[-fips].<region>.<partition-domain>
	parts := strings.Split(registry, ".")
	for i, p := range parts {
		if (p == "ecr" || p == "ecr-fips") && i+1 < len(parts) {
			return parts[i+1]
		}
	}
	return ""
}

func getECRCredentials(ctx context.Context, registry string) (oci.Credentials, error) {
	if registry == "public.ecr.aws" {
		return getPublicECRCredentials(ctx)
	}
	return getPrivateECRCredentials(ctx, registry)
}

func getPrivateECRCredentials(ctx context.Context, registry string) (oci.Credentials, error) {
	region := extractRegion(registry)
	if region == "" {
		return oci.Credentials{}, fmt.Errorf("cannot extract region from %q", registry)
	}

	opts := []func(*config.LoadOptions) error{config.WithRegion(region)}
	if strings.Contains(registry, ".dkr.ecr-fips.") {
		opts = append(opts, config.WithUseFIPSEndpoint(aws.FIPSEndpointStateEnabled))
	}

	cfg, err := config.LoadDefaultConfig(ctx, opts...)
	if err != nil {
		return oci.Credentials{}, creds.ErrCredentialsNotFound
	}

	client := ecr.NewFromConfig(cfg)
	out, err := client.GetAuthorizationToken(ctx, &ecr.GetAuthorizationTokenInput{})
	if err != nil {
		return oci.Credentials{}, creds.ErrCredentialsNotFound
	}

	if len(out.AuthorizationData) == 0 || out.AuthorizationData[0].AuthorizationToken == nil {
		return oci.Credentials{}, creds.ErrCredentialsNotFound
	}

	return decodeAuthToken(*out.AuthorizationData[0].AuthorizationToken)
}

func getPublicECRCredentials(ctx context.Context) (oci.Credentials, error) {
	cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion("us-east-1"))
	if err != nil {
		return oci.Credentials{}, creds.ErrCredentialsNotFound
	}

	client := ecrpublic.NewFromConfig(cfg)
	out, err := client.GetAuthorizationToken(ctx, &ecrpublic.GetAuthorizationTokenInput{})
	if err != nil {
		return oci.Credentials{}, creds.ErrCredentialsNotFound
	}

	if out.AuthorizationData == nil || out.AuthorizationData.AuthorizationToken == nil {
		return oci.Credentials{}, creds.ErrCredentialsNotFound
	}

	return decodeAuthToken(*out.AuthorizationData.AuthorizationToken)
}

func decodeAuthToken(token string) (oci.Credentials, error) {
	decoded, err := base64.StdEncoding.DecodeString(token)
	if err != nil {
		return oci.Credentials{}, fmt.Errorf("decode auth token: %w", err)
	}

	username, password, ok := strings.Cut(string(decoded), ":")
	if !ok || username == "" || password == "" {
		return oci.Credentials{}, creds.ErrCredentialsNotFound
	}

	return oci.Credentials{Username: username, Password: password}, nil
}

func GetECRCredentialLoader() []creds.CredentialsCallback {
	return []creds.CredentialsCallback{
		func(registry string) (oci.Credentials, error) {
			if !isECRRegistry(registry) {
				return oci.Credentials{}, creds.ErrCredentialsNotFound
			}

			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
			defer cancel()

			return getECRCredentials(ctx, registry)
		},
	}
}
```

This eliminates the goroutine, the channel, the `sync.Once`, the `sync.Map` cache, the `ecrCacheEntry` type, and the `isAWSCredentialsNotFound` helper — all replaced by ~70 lines of straightforward sequential code with proper context propagation. The `aws-sdk-go-v2` packages are already transitive dependencies, and the `amazon-ecr-credential-helper` could be dropped from `go.mod`.

The error handling could be refined to distinguish "no credentials configured" from transient failures if you want to selectively cache or retry, but even as-is this is safer than the current approach of caching all errors permanently.

pkg/k8s/keychains.go
matejvasek: OK, but could not we keep this state in `GetECRCredentialLoader()` (closure), instead of having it global?

matejvasek:
Ideally we should submit PR to the helper repository. The PR would add the missing context param. Then we could use the helper as it is nicely.

matejvasek:
> I'd suggest bypassing the ECR helper entirely and calling the AWS ECR SDK directly.

This however may miss some features of the helper, e.g. caching.

matejvasek:
btw, I belive that if you install the ecr helper as binary and set it up for docker in docker config we should pick it up.

matejvasek:
https://github.com/awslabs/amazon-ecr-credential-helper/pull/1092

matejvasek:
The semaphore introduces a cross-registry blocking problem. Consider:

1. Goroutine A starts for registry X, acquires semaphore (1/2 used)
2. `Resolve()` hangs on IMDS, 5s timeout fires, function returns error to caller
3. Goroutine A is still running, holding its semaphore slot (1/2 used)
4. Goroutine B starts for registry Y (different registry), acquires semaphore (2/2 used)
5. Same thing — timeout, goroutine still running (2/2 used)
6. Any subsequent ECR lookup for any registry blocks on `Acquire`, then fails with "queue full"

After just 2 timeouts — potentially for completely unrelated registries — all ECR credential lookups are dead until one of the leaked goroutines' `Resolve()` calls eventually return (which depends on the AWS SDK's internal HTTP timeout, typically 30s+).

The semaphore doesn't prevent the goroutine leak (the leaked goroutine holds the slot until `Resolve` completes), it just caps the damage at 2 leaked goroutines while introducing a new failure mode where unrelated registries block each other.

I'd suggest dropping it. The goroutine leak is inherent to `Resolve()` / `Get()` not accepting a `context.Context` — it can't be fixed at this layer. The TTL cache and not caching timeouts (both good changes in this revision) are sufficient mitigation: a leaked goroutine runs to completion eventually, and the next lookup retries cleanly.

matejvasek:
The caching here is more complex than it needs to be. The ECR helper already has its own file-based credential cache (`cache.BuildCredentialsCache`) that stores successful auth tokens with expiration and even falls back to expired tokens when the API fails. So caching successes in the loader is redundant — and the current code already avoids that, which is good.

The only value of caching at this layer is avoiding repeated 5-second timeout hangs when credentials aren't configured. For that, a TTL'd `ecrCacheEntry` struct is overkill. A simple set of failed registries would do:

```go
var failedRegistries sync.Map

// early return
if _, ok := failedRegistries.Load(registry); ok {
    return oci.Credentials{}, creds.ErrCredentialsNotFound
}

// after the lookup, only cache "no credentials" errors (not timeouts)
if errors.Is(resErr, creds.ErrCredentialsNotFound) {
    failedRegistries.Store(registry, struct{}{})
}
```

No TTL is needed — if AWS credentials aren't configured, they won't appear mid-process. The closure scoping already ensures the state is fresh per `GetECRCredentialLoader()` call, so a process restart clears it.

Alternatively, drop the cache entirely. The 5-second timeout is bounded, the helper caches its own successes, and `isECRRegistry` filters out non-ECR registries. The worst case is a few 5-second waits during a single push when credentials aren't configured — which is arguably correct behavior.

matejvasek:
`isECRRegistry` is duplicating what the helper already does internally. `ECRHelper.Get()` calls `api.ExtractRegistry()` as its very first operation — a regex match, no network calls, no AWS SDK initialization — and returns `credentials.NewErrCredentialsNotFound()` for non-ECR registries.

Removing `isECRRegistry` and letting the helper handle registry detection would:

- Fix the missing partition coverage (`.amazonaws.eu`, `.on.aws`, `.cloud.adc-e.uk`, `.csp.hci.ic.gov`, etc.) that the helper's regex already supports
- Fix the missing `ecr-public.aws.com` dual-stack hostname
- Keep detection automatically in sync with the helper across version upgrades
- Remove ~15 lines of code and the associated tests

The only argument for keeping it is avoiding the goroutine spawn for non-ECR registries, but that's a micro-optimization around the goroutine pattern itself.

matejvasek:
My PR in aws ecr was merged.

Nachiket-Roy:
@matejvasek will update my pr in few hours.

matejvasek:
Please rebase on the main branch to fix the CI issues.

pkg/k8s/keychains.go
```diff
@@ -44,8 +63,129 @@ func GetGoogleCredentialLoader() []creds.CredentialsCallback {
 	}
 }
 
+func isECRRegistry(registry string) bool {
```
matejvasek: Please remove this. As mentioned earlier the helper itself does cheap quick regex check.

matejvasek: previous comment https://github.com/knative/func/pull/3864#issuecomment-4611789242

matejvasek: Otherwise PR looks good.

PR #3873, opened 2026-06-03
enable jj for make check

gauron99:


lkingland (approved):
Nice!

gauron99:
@lkingland can you recheck pls i had to rebase for gitlab test fix

matejvasek:
Nice cleanup — deduplicating those four identical pipelines is a clear win, and jj support is a good addition.

A couple of things I noticed:

**Bug: unbound variable in the fallback path**

The script uses `set -euo pipefail`, but if neither git nor jj is detected (and `FUNC_VCS` is unset), `VCS` is never assigned. The first `[ "$VCS" = git ]` in `cmd_ls_sources` will crash with `bash: VCS: unbound variable` before the `find` fallback can be reached. Adding `VCS=""` before the detection block would fix it.

**Hardcoded generated-file list**

The git path dynamically reads `linguist-generated` from `.gitattributes`, but the jj/find paths hardcode the two filenames (`zz_filesystem_generated.go`, `zz_close_guarding_client_generated.go`). If a new generated file is added to `.gitattributes`, `vcs.sh` would also need updating. Worth at least a `# keep in sync with .gitattributes` comment, or parsing `.gitattributes` directly.

matejvasek:
Doesn't `jj` have something akin to `git check-attr --stdin linguist-generated`? If not cannot we still call `git check-attr --stdin linguist-generated` even for `jj` instead of hardcoding ?

gauron99:
adding `git` command to `jj` repo does nothing if `.git` is not present. Point for this is `jj` can actually run without `.git` repo completely (like jj workspace which is equivalent for git worktrees). Unfortunately the support for `.gitignore` is not yet in `jj` but I added a comment there for tracking

PR #3874, opened 2026-06-03
fix: GitLab integration test insecure registry detection

gauron99:
AI analysis (with my tweaks)
## Summary
- Fix `TestInt_Gitlab` build failure caused by upstream change to `.cluster.local` HTTPS handling
- Improve `awaitBuildCompletion` to detect build failures immediately instead of waiting the full 10-minute timeout

## Root cause

**Apr 29:** [`google/go-containerregistry#2281`](https://github.com/google/go-containerregistry/pull/2281) changed `.local` domain handling to only treat `.localhost` as HTTP. Previously, any `.local` domain (including `.cluster.local`) automatically used HTTP.

**May 18:** Released in [`go-containerregistry` v0.21.6](https://github.com/google/go-containerregistry/releases/tag/v0.21.6).

**Jun 1 13:45 UTC:** Paketo lifecycle [bumped to that version](https://github.com/buildpacks/lifecycle/pull/1654), released as [`lifecycle` v0.21.10](https://github.com/buildpacks/lifecycle/releases/tag/v0.21.10).

**Jun 1 15:07 UTC:** [Last passing run](https://github.com/knative/func/actions/runs/26763452535) (before builder rebuild).

**Jun 1 17:57 UTC:** Our scheduled [`Update builder-jammy-full image` workflow](https://github.com/knative/func/actions/runs/26772351443) rebuilt `builder-jammy-tiny:v2`, downloading lifecycle v0.21.10.

**Jun 1 20:40 UTC:** [First failing run](https://github.com/knative/func/actions/runs/26780714415) (after builder rebuild).

The build step (`step-create`) fails with:
```
ERROR: failed to initialize analyzer: validating registry read access:
Get "https://registry.default.svc.cluster.local:5000/v2/":
http: server gave HTTP response to HTTPS client
```

The test set `Image` but not `Registry` on the function. Our template code checks `f.RegistryInsecure || isInsecureRegistry(f.Registry)` to decide whether to set `CNB_INSECURE_REGISTRIES` — but with both unset, the check never fired. Previously this was masked by the library silently falling back to HTTP for all `.local` domains.

Other projects hit the same issue: [`tektoncd/chains`](https://github.com/ab-ghosh/chains/commit/6d97fced7ec4c808266966056720d2e1f1899bed).

## Changes

1. **Set `RegistryInsecure: true` on the test function** — this is the correct flag for an in-cluster plaintext registry. A follow-up will fix the template code to also derive the registry from `Image` when `Registry` is not set, so this works automatically.

2. **Detect build failures in `awaitBuildCompletion`** — returns `error` instead of `struct{}`, checks `ConditionSucceeded=False` in addition to `True`. Before this, a failed build silently waited the full 10-minute timeout. Now it fails in ~80s with the actual TaskRun reason and message.

## Test plan
- [x] Reproduced failure locally with `gitlab-ce:latest` (19.0.1) and lifecycle 0.21.10
- [x] Verified fix passes: `TestInt_Gitlab` completes in ~155s
- [x] Verified improved error reporting: build failures surface in ~80s with TaskRun reason/message

### DIFF
--- a/pkg/pipelines/tekton/gitlab_int_test.go
+++ b/pkg/pipelines/tekton/gitlab_int_test.go
@@ -76,13 +76,14 @@ func TestInt_Gitlab(t *testing.T) {
 	funcImg := fmt.Sprintf("registry.default.svc.cluster.local:5000/fn-%s", uuid.NewUUID())
 
 	f := fn.Function{
-		Root:     projDir,
-		Name:     glabEnv.ProjectName,
-		Runtime:  "test-runtime",
-		Template: "test-template",
-		Image:    funcImg,
-		Created:  time.Now(),
-		Invoke:   "none",
+		Root:             projDir,
+		Name:             glabEnv.ProjectName,
+		Runtime:          "test-runtime",
+		Template:         "test-template",
+		RegistryInsecure: true,
+		Image:            funcImg,
+		Created:          time.Now(),
+		Invoke:           "none",
 		Build: fn.BuildSpec{
 			Git: fn.Git{
 				URL:      strings.TrimSuffix(glabEnv.HTTPProjectURL, ".git"),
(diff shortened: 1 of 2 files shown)


pkg/pipelines/tekton/gitlab_int_test.go
```diff
@@ -80,6 +80,7 @@ func TestInt_Gitlab(t *testing.T) {
 		Name:     glabEnv.ProjectName,
 		Runtime:  "test-runtime",
 		Template: "test-template",
+		Registry: "registry.default.svc.cluster.local:5000",
```
matejvasek: Could we use `RegistryInsecure: true,` here?

PR #3876, opened 2026-06-03
test: reduce GitLab test instance memory footprint (~400-500MB)

matejvasek:
Disable unused services (monitoring, registry, pages, KAS), reduce
sidekiq concurrency, tune PostgreSQL, disable unneeded Rails features,
and update gitaly config to remove dead ruby_max_rss setting and fix
deprecated RPC names.


gauron99 (approved):
/lgtm

PR #3878, opened 2026-06-03
Duplicate env and volume validation

### LINKED ISSUE #3877: Fail Fast on Duplicate Environment Variables and Volume Mount Paths
Currently, duplicate environment variable names and volume mount paths are not validated locally.

- Duplicate environment variables are silently overwritten during interpolation, which can lead to unexpected runtime behavior.
- Duplicate volume mount paths are only rejected later by Kubernetes during deployment, after image build and push steps have already completed.

Adding validation during `f.Validate()` would provide earlier feedback, prevent ambiguous configurations, and avoid wasted build/deployment time.

Nachiket-Roy:
### Changes
Validation Logic
- Modified `ValidateEnvs` and `ValidateBuildEnvs` to track encountered variable names using a set/map.
- Returns a detailed error if duplicate names are detected, showing the index of the duplicate and the index of the first definition.

Volume Mount Paths

- Modified `validateVolumes` to track encountered volume paths.
- Returns a detailed error if multiple volumes are mounted to the exact same path.

Unit Tests

- Fixed legacy test cases that reused name pointer references across slice items to use unique variable names.
- Added new test cases verifying that duplicate environment variable name and volume mount paths are successfully rejected.

Closes : #3877

### DIFF
--- a/pkg/functions/function_volumes.go
+++ b/pkg/functions/function_volumes.go
@@ -81,8 +81,17 @@ func (v Volume) String() string {
 //   - emptyDir: {}                                         # mount EmptyDir as Volume
 //     path: /etc/configMap-volume
 func validateVolumes(volumes []Volume) (errors []string) {
-
+	seenPaths := make(map[string]int)
 	for i, vol := range volumes {
+		if vol.Path != nil {
+			path := *vol.Path
+			if firstIdx, seen := seenPaths[path]; seen {
+				errors = append(errors, fmt.Sprintf("volume entry #%d has duplicate path %q (first defined at entry #%d)", i, path, firstIdx))
+			} else {
+				seenPaths[path] = i
+			}
+		}
+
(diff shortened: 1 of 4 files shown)


pkg/functions/function_volumes.go
```diff
@@ -81,8 +81,17 @@ func (v Volume) String() string {
 //   - emptyDir: {}                                         # mount EmptyDir as Volume
 //     path: /etc/configMap-volume
 func validateVolumes(volumes []Volume) (errors []string) {
-
+	seenPaths := make(map[string]int)
```
matejvasek, line 87: What about paths like /etc/foo/ vs. /etc/foo (trailing slash)?

Nachiket-Roy: Resolved. Thanks!

PR #3885, opened 2026-06-08
fix: host builder can push to OCP internal registry

matejvasek:
## Summary
- Fix OCI pusher failing to push images to OpenShift's internal registry when the repository doesn't exist yet
- OpenShift's registry returns HTTP 500 for manifest HEAD requests by digest on non-existent repos, which breaks `remote.WriteIndex` since it checks child manifests by digest
- Before writing the index, do a HEAD by tag to check repo existence; on 404, push an empty image to create the repo first, then proceed with the full index push

### DIFF
--- a/pkg/oci/pusher.go
+++ b/pkg/oci/pusher.go
@@ -2,6 +2,7 @@ package oci
 
 import (
 	"context"
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


gauron99 (approved):
/lgtm

PR #3888, opened 2026-06-09
gitignore the whole .claude + .ai dirs 

gauron99:
ignore whole `.claude` and new `.ai` directories

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,9 +34,8 @@ __pycache__
 
 # AI
 AGENTS.override.md
-.claude/*.local.*
-.claude/pr/*
-.claude/plans/*
+.ai/*
+.claude/*
 .roo/*
 .mcp*
 


gauron99:
actually we have skills in the `.claude` repo i just noticed/remember :facepalm:

matejvasek:
We also could check presents of these in PR in the pre-checks.

gauron99:
yes i was thinking the same, atleast as a warning

gauron99:
but then again, the real problem would also be someone fetching the PR via claude or other ai and it reading those files. So for example just commenting "hey careful theres changes to these files" isnt really sufficient.

I thought you can `git add --force <file>` so this is no way a "defense" against someone with malicious intent. Its just a good practice for us to not commit our personal files really - which makes me think that the .gitignore is fine and we should probably try a bit more "drastic measures".

This could be a `vouch` system where only certain people can edit the `.claude` etc. -> meaning we run a workflow which will check 1. files being changed 2. some text file -> anyone who is on this special list (`vouched.txt`) will have permission to edit these "ai" related files. -> check PR author against list, if not there, auto-close PR.

This way we dont go the full `vouch` (think `ghostty` repo/Hashicorp) system for the whole repo but we select handful of files we deem are critical and only allow some people to touch those. whereever some kind of injection like this would most likely occur.

lkingland:
For the time being just cease using AI to injest any data (PRs, issues, etc) from GitHub.  We have an open Slack chat about how we can harden our sytems.

lkingland (approved):

lkingland:
We could provide "suggested skills" for clients to use, in a dedicated directory like ./ai/skills, and include in our instructions they're available.

I don'tw ant to push agent configuration on others.  It's tantamount to altering their IDE when they pull your source code.  Rather, suggestions of skills to use with the system they can symlink or copy seems like a nice middle-ground.

This not as solved problem as far as I am aware.  For example, if I have a settings.local.json or CLAUDE.local.md... and that's of course gitignored, it doesn't come along when creating workspaces.

THis is something I've solved personally using explicit autogeneration of these files when creating workspaces, but I don't think we can expect developers to create their own tooling just to work around this contention.

PR #3895, opened 2026-06-16
fix: gracefully shutdown host-built functions

gauron99:
- fixes`Stop` hook on functions built via `host` builder (go+python today)


twoGiants (approved):
/approve
/lgtm

PR #3896, opened 2026-06-16
cmd: apply function context to invoke/logs flag defaults

gauron99:
Bring `NewInvokeCmd` and `NewLogsCmd` in line with the `cfg.Apply(f)` idiom already used by deploy and build -> fold the current function's state into the global config before registering flags, so `func.yaml` values become flag defaults.

### DIFF
--- a/cmd/invoke.go
+++ b/cmd/invoke.go
@@ -118,6 +118,12 @@ EXAMPLES
 		fmt.Fprintf(cmd.OutOrStdout(), "error loading config at '%v'. %v\n", config.File(), err)
 	}
 
+	// Function Context
+	f, _ := fn.NewFunction(effectivePath())
+	if f.Initialized() {
+		cfg = cfg.Apply(f)
+	}
+
 	// Flags
 	cmd.Flags().StringP("format", "f", "", "Format of message to send, 'http' or 'cloudevent(s)'.  Default is to choose automatically. ($FUNC_FORMAT)")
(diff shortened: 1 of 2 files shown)


cmd/invoke.go
```diff
@@ -118,6 +118,12 @@ EXAMPLES
 		fmt.Fprintf(cmd.OutOrStdout(), "error loading config at '%v'. %v\n", config.File(), err)
 	}
 
+	// Function Context
+	f, _ := fn.NewFunction(effectivePath())
+	if f.Initialized() {
```
twoGiants, line 123: Judging from your description this could benefit from a test. Same for `logs.go`.

twoGiants (approved):
Looks good! But I'd add tests to cover the new behavior.

/approve
/lgtm
/hold for others and for comment reply

cmd/invoke.go
gauron99: Hey thanks for the test pointer.

hmm this PR was made as extraction from larger refactor PR meant to reduce the review mental load.

It doesn't actually change functionality because we don't set the functions values (from `func.yaml` which this `fn.NewFunction` loads) in the flags' defaults. (few lines below) -> the process is slightly different for these commands

It becomes meaningful with that larger-refactor commits which add new flags that would go through this defaulting -> something to test :grin: 

I will try to extract differently so its a "more meanigful" PR or fold back to the original, thanks.

PR #3897, opened 2026-06-16
pipelines: never upload .func to the cluster in direct remote deploy

gauron99:
- Direct remote (on-cluster) builds tar the function source and upload it to a PVC. Honor the same ignoring system that local builds already implement -> Using the default ignored + `.funcignore` + `.gitignore`. 

- Scaffolding under `.func/build` is regenerated on-cluster by the func-scaffold step, so nothing needed is lost

### DIFF
--- a/pkg/pipelines/tekton/pipelines_provider.go
+++ b/pkg/pipelines/tekton/pipelines_provider.go
@@ -281,14 +281,17 @@ func (pp *PipelinesProvider) Run(ctx context.Context, f fn.Function) (string, fn
 
 // Creates tar stream with the function sources as they were in "./source" directory.
 func sourcesAsTarStream(f fn.Function) *io.PipeReader {
-	ignored := func(p string) bool { return strings.HasPrefix(p, ".git") }
-	if gi, err := gitignore.CompileIgnoreFile(filepath.Join(f.Root, ".gitignore")); err == nil {
-		ignored = func(p string) bool {
-			if strings.HasPrefix(p, ".git") {
-				return true
-			}
-			return gi.MatchesPath(p)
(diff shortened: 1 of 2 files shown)


pkg/pipelines/tekton/pipelines_provider.go
```diff
@@ -281,14 +281,17 @@ func (pp *PipelinesProvider) Run(ctx context.Context, f fn.Function) (string, fn
...
-			}
-			return gi.MatchesPath(p)
+	// Apply the same ignore policy as local builds: fn.IsIgnored always excludes
+	// the DefaultIgnored plus the user's .funcignore patterns. This guarantees local
+	// runtime data under .func — scaffolding (regenerated on-cluster by the
+	// func-scaffold step)
+	userPatterns := fn.ParseFuncIgnore(f.Root)
+	gi, giErr := gitignore.CompileIgnoreFile(filepath.Join(f.Root, ".gitignore"))
+	ignored := func(p string) bool {
+		if fn.IsIgnored(p, userPatterns) {
```
twoGiants, line 291: nit

You can inline `userPatterns` here, it's not called anywhere else:


```go
		if fn.IsIgnored(p, fn.ParseFuncIgnore(f.Root)) {
```

twoGiants (approved):
Looks good. I have only a nit.

/approve
/lgtm
/hold for the nit

pkg/pipelines/tekton/pipelines_provider.go
gauron99: that would actually re-read `.funcignore` file every `ignored` function call inside the `Walk` below. Its extracted so that read is only fired once

lkingland (approved):
🎉

PR #3899, opened 2026-06-17
e2e: add lifecycle structure + tests for existing python + go

gauron99:
e2e lifecycle tests
- start,ready,alive hooks in one function
- stop separate to deliberately test stop hook works on `func delete`


lkingland (approved):
🎉

PR #3903, opened 2026-06-17
fix: autoscaling annotation format and empty image guard

Elvand-Lie:
## Description
Two bug fixes:

### 1. Autoscaling annotations formatted with `%f` produce invalid values
`setServiceOptions` in `pkg/knative/deployer.go` uses `fmt.Sprintf("%f", ...)` to format scale `Target` and `Utilization` annotations, producing strings like `"100.000000"` instead of `"100"`. Changed `%f` to `%g` which strips trailing zeros.

**Files changed:** `pkg/knative/deployer.go` (lines 592, 598)

### 2. `ImageNameWithDigest` produces invalid OCI reference when `Build.Image` is empty
No guard against empty `Build.Image` in `ImageNameWithDigest`. When called with non-empty digest but empty image, produces invalid `@sha256:...` reference. Added early return for empty image.

**Files changed:** `pkg/functions/function.go` (line 864)

### Regression test
Added `TestSetServiceOptions_ScaleAnnotationFormat` that verifies scale annotations contain clean values (`"100"` not `"100.000000"`).

**Files changed:** `pkg/knative/deployer_test.go`

### Verification
- `go vet ./pkg/knative ./pkg/functions` - clean
- `go test ./pkg/knative/ -count=1` - all 20 tests pass
- `go test ./pkg/functions/ -run TestFunction_ImageWithDigest -count=1` - all 5 tests pass

### DIFF
--- a/pkg/functions/function.go
+++ b/pkg/functions/function.go
@@ -862,6 +862,9 @@ func (f Function) ImageNameWithDigest(newDigest string) string {
 		return f.Build.Image
 	}
 	image := f.Build.Image
+	if image == "" {
+		return ""
+	}
 
 	// overwrite current digest
 	shaIndex := strings.Index(image, "@sha256:")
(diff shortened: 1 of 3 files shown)


Elvand-Lie:
Closing in favor of separate PRs per bug fix.

PR #3912, opened 2026-06-20
fix(invoke): validate extension format and remove dead branch

### LINKED ISSUE #3346: Allow us to set extensions on `func invoke`

shreyaabaranwal:
Addresses review follow-ups from #3347:
- Remove unreachable second branch in cloudevent validation
- Return an error on malformed --extension input (missing '=') instead of silently dropping it
- Simplify extension map iteration with k, v range
- Add tests for the cloudevent guard, malformed input, and multi-extension cases

Fixes #3346

# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/cmd/invoke.go
+++ b/cmd/invoke.go
@@ -175,14 +175,16 @@ func runInvoke(cmd *cobra.Command, _ []string, newClient ClientFactory) (err err
 		if effectiveFormat != "cloudevent" {
 			return fmt.Errorf("--extension (-e) is only valid with cloudevents")
 		}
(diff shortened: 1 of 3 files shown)


shreyaabaranwal:
Looks like #3839 and #3852 already addressed the dead branch and malformed-extension validation. Closing this as the work is now covered. Thanks!

PR #3913, opened 2026-06-20
refactor(invoke): simplify extension map iteration

shreyaabaranwal:
Use `for k, v := range` instead of indexed map lookup when setting CloudEvent extensions.

# Changes

- Simplify CloudEvent extension iteration using `for k, v := range` instead of looking up each value by key

/kind cleanup

Relates to #3346

**Release Note**

```release-note

```

**Docs**

```docs

```


shreyaabaranwal:
The E2E failure looks unrelated to this PR — `TestCore_PythonUserDepsRemote` fails with an image-not-found error caused by `git describe` failing in CI (`FUNC_VERSION` ends up unset). This PR only changes invoke extension map iteration. Could a maintainer re-run the job? 
Thanks! 🙏

PR #3914, opened 2026-06-22
e2e cleanup: RemoteUpdate, TriggerSync simplified, fix test prefix to run in general e2e suite

gauron99:
- test `Remote_Update` mimics the local deploy -> update function -> re-deploy to make sure we cover that scenario for direct deploy as well
- test `TriggerSync` was simplified using 1 function and updating the subscriptions (fewer resources, faster, simpler test)
- Few days ago I added explicit instanced tests but forgot we use the explicit test Prefixes that are run as part of the general e2e suite, just a fix to make that part of `TestCore` -- what it does: two sequential requests, assert the second count is greater than the first to prove state persistance

Had to also add an optional keda removal for `C/M/R job` - e2e test (Core|Metadata|Remote) because we were hitting the CPU usage max 4000m that github CI provisions. removing keda (since its not used in any of the e2e C/M/D tests reduces that by 1000m :nice:)

### DIFF
--- a/.github/workflows/functions.yaml
+++ b/.github/workflows/functions.yaml
@@ -208,6 +208,7 @@ jobs:
       FUNC_CLUSTER_RETRIES: 5
       FUNC_E2E_CLEAN: false # cluster only used once
       FUNC_E2E_VERBOSE: true
+      FUNC_CLUSTER_KEDA: "false" # these tests don't use KEDA; skip it to free cluster CPU (PR #3914 'Insufficient cpu')
     steps:
       - uses: actions/checkout@v4
       - uses: knative/actions/setup-go@main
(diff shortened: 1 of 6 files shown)


lkingland (approved):

gauron99:
had to update the instance test (which did not run previously in CI because of the wrong name format, had to adjust for what we have for tests. It discovered a bug where the test was checking for a absolute sequential number but unfrotunately there are other services that make requests to this function which then failed -- we poll for the first "function is up we can start testing" + probably some knative health checks are hitting the function too. so now instead we send 2 requests and check "is the second one higher than first" to say "state is persisted"

gauron99:
we are hitting 
```
Error: deploy error. knative deployer failed to wait for the Knative Service to become ready: RevisionFailed: Revision "func-e2e-test-metadata-volumes-00001" failed with message: 0/1 nodes are available: 1 Insufficient cpu. no new claims to deallocate, preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod..
```

lkingland (approved):

PR #3922, opened 2026-06-25
fix(ci): make git describe robust against shallow checkouts

shreyaabaranwal:
# Changes

- :bug: Make `git describe` resilient to shallow checkouts so the version string is never empty

While working on #3913 I noticed the `E2E - Core, Metadata, and Remote` check kept failing on a change that was completely unrelated to it (just an extension map iteration cleanup). Digging into the logs of the failing run (job 83049870138), the actual cause turned out to be in CI itself, not the PR.

`actions/checkout` does a shallow clone by default and doesn't fetch tags. Because of that, `git describe --tags --match 'v*'` exits with status 128 and the version ends up empty. That empty version flows into the image tags, so the e2e tests later fail looking up an image that was never built under the expected name — which is why `TestCore_PythonUserDepsRemote` reports "image not found".

This PR fixes the root cause rather than the symptom:

- Set `fetch-depth: 0` on all `actions/checkout@v4` steps so tags are actually available when `git describe` runs.
- Add a safe fallback in `Makefile` and `hack/release.sh` so the version degrades to a commit SHA (or a placeholder) instead of being empty if no matching tag is found.

This should unblock e2e on PRs like #3913 that don't touch CI but were failing because of it.

/kind bug

Relates to #3913

**Release Note**
```release-note
NONE
```

**Docs**
```docs
NONE
```

### DIFF
--- a/.github/workflows/functions.yaml
+++ b/.github/workflows/functions.yaml
@@ -38,6 +38,8 @@ jobs:
     timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - uses: knative/actions/setup-go@main
 
       - name: Check and Lint
@@ -72,6 +74,8 @@ jobs:
         run: git config --global core.autocrlf false
 
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - uses: knative/actions/setup-go@main
 
       - name: Run Unit Tests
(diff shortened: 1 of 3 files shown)


shreyaabaranwal:
this one fixes a CI issue that's currently failing e2e on unrelated PRs (like #3913). Would a maintainer mind dropping /ok-to-test so the checks can run? Thanks!

PR #3927, opened 2026-07-02
Python: build and deploy legacy parliament functions (deprecated)

gauron99:
Old parliament-style Python functions (bare func.py with def main(context), Procfile "web: python -m parliament .", requirements.txt) keep building and deploying on modern func instead of breaking in the ASGI scaffolding:

- pack + s2i detect the legacy layout and skip modern scaffolding, building the pre-v1.18 way from the user's own Procfile + requirements.txt.
- cloudevents is pinned below 2.0 during the build (parliament-functions leaves it unpinned; cloudevents 2.x removed cloudevents.http and crashes the stack): pack via constraints.txt + PIP_CONSTRAINT (local and remote/Tekton), s2i via the legacy assemble script. The s2i builder image is pinned back to ubi8/python-39 unless the user overrides it.
- host builder and host runner reject with a clear error pointing at pack/s2i.
- the other pre-v1.18 Procfile-based python layouts (the old flask/wsgi templates: root Procfile, no root pyproject.toml) are detected and rejected with a clear migration error instead of failing confusingly inside modern scaffolding.
- a deprecation warning is emitted on every build, pointing at the current Python function layout docs.

All legacy logic lives in per-package legacy_parliament.go files; every inline touch is a marked one-liner (grep "LEGACY PYTHON") so sunset removal is: delete the files, grep, delete the guards, compile.

### DIFF
--- a/e2e/e2e_core_test.go
+++ b/e2e/e2e_core_test.go
@@ -417,6 +417,77 @@ func TestCore_StaticSignature_Python(t *testing.T) {
 	}
 }
 
+// LEGACY PYTHON: old parliament functions (func.py def main(context) + parliament
+// Procfile) must build and serve via pack and s2i, built locally and in-cluster.
+func TestCore_LegacyParliamentSignature_Python(t *testing.T) {
+	// The fixtures in testdata/legacy-parliament were generated with func
+	// v1.17.0 (the last parliament-era release):
+	//
+	//	func create -l python -t http parliament-http
+	//	func create -l python -t cloudevents parliament-cloudevents
+	//
+	// The cells pair builders, flavors and localities so that each builder
+	// builds both locally and remotely (in-cluster, where scaffolding runs
+	// inside the Tekton pipeline) and each flavor passes through both builders.
+	for _, tc := range []struct {
+		builder  string
+		template string
+		remote   bool
+	}{
+		{"pack", "http", false},
+		{"s2i", "cloudevents", false},
+		{"pack", "cloudevents", true},
+		{"s2i", "http", true},
+	} {
+		cell := tc.builder
+		if tc.remote {
+			cell += "-remote"
+		}
+		t.Run(cell, func(t *testing.T) {
+			name := "func-e2e-parliament-" + cell
+			root := fromCleanEnv(t, name)
+
+			// Copy in the v1.17.0-generated fixture, renaming it for this run;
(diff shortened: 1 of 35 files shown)


e2e/e2e_core_test.go
```diff
@@ -417,6 +417,56 @@ func TestCore_StaticSignature_Python(t *testing.T) {
...
+
+
+def main(context: Context):
+    return "OK", 200
+`
+			writeAll := map[string]string{
+				"func.py":          parliamentPy,
+				"Procfile":         "web: python -m parliament .\n",
+				"requirements.txt": "parliament-functions==0.1.0\n",
+			}
```
lkingland: We're 100% sure this accurately represents what pre-scaffolding (WSGI) Parliament-based functions looked like?

I am tempted to say you should actually use an earlier version of `func` to generate an HTTP and CloudEvent python Function from that era, and stick them in `./e2e/testdata/` somewhere, and use that.

pkg/buildpacks/builder.go
```diff
@@ -221,7 +221,10 @@ func (b *Builder) Build(ctx context.Context, f fn.Function, platforms []fn.Platf
 	// Python scaffolding via inline pre-buildpack.
 	// Injects a script that rearranges user code into fn/ and copies
 	// scaffolding from .func/build/ to the workspace root.
-	if f.Runtime == "python" {
+	//
+	// LEGACY PYTHON: skip the pre-buildpack for old parliament functions; inject
+	// the cloudevents<2 pip constraint instead (else-if below).
+	if f.Runtime == "python" && !f.IsLegacyParliament() {
```
lkingland: nit:  I would switch this conditional.  The _exception_ is `f.isLegacyParliament`, so the longer, uglier if statement shuld be the exeception.

```
if f.Runtime == "python" && f.IsLegacyParliament() {
  .. do legacy stuff
} else if f.Runtime == "python" {
  .. contineu one with the default, modern flow for python runtimes
}
```

lkingland (commented):
🎉   This is really looking great.  More was needed than I expected, that's for sure

e2e/e2e_core_test.go
gauron99: sure! i can pu that into testdata

lkingland (approved):
🎉

PR #3951, opened 2026-07-17
fix: skip knative backends on forbidden so list/describe/delete work for non-knative specific use-case

gauron99:
dont error on `forbidden` error. If user wants to use bare `k8s` deployment and has 0 knative installed - standard user will most likely not have grants for non-existing groups/resources.

### DIFF
--- a/pkg/knative/describer.go
+++ b/pkg/knative/describer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"os"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


lkingland (approved):

gauron99:
Integration Tests is failing here on TestInt_Gitlab, unrelated to this PR's changes:

https://github.com/knative/func/actions/runs/29567897179/job/87876478422

`cannot create new personal access token, unexpected status: 422`

Cause: `gitlab/gitlab-ce:latest` moved to 19.2.0 on 2026-07-16 and changed the sign-in/PAT web flow that TestInt_Gitlab's getAPIToken scrapes. Fix pinning the image is up in #3952.

PR #3953, opened 2026-07-17
feat: persistent deployer in func.yaml, block incompatible deployer switch, func delete removes f.Deploy values to match reality

gauron99:
### global deployer
- make deployer as a globally configurable -- mirror builder -> its now always persisted in func.yaml
- this makes it so its easier to work with `deployers.X` value instead of hardcoded values
- for backwards compat if we hit `deployer == ""` (as it was before) we default to Knative as that was the default all along
- note: I wouldnt be opposed to removing the "backwards compat" the only time you hit this is you deployed function with older binary, then you fetch new binary and try to "redeploy" again... its adding extra machinery which is ugly 😄 but I suppose it is the "standard upgrade path"

### deployer switch
- second part is blocking switch re-deploy early -- we now have 3 distinct deployers of which most dont share resources.
- we dont want to make deployers aware of other deployers resources therefore stop this early.
- the only documented compatible switch right now is `raw` -> `keda` because keda deployer embeds the raw deployer -- eg same resources.

### func delete change
- on successful undeploy -- `func delete` now actually removes the `f.Deploy.Namespace` and `f.Deploy.Deployer` values so that `func.yaml` correctly represents the state of the function on cluster (only when deleting by project! not name since theres no local files for that flow)

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -21,7 +21,8 @@ This command undeploys a function from the cluster. By default the function from
 the project in the current directory is undeployed. Alternatively either the name
 of the function can be given as argument or the project path provided with --path.
 
-No local files are deleted.
+No local files are deleted. When undeploying by project (not by name), on success
+the local func.yaml is updated to reflect that the function is no longer deployed.
 `,
 		Example: `
 # Undeploy the function defined in the local directory
@@ -82,13 +83,19 @@ func runDelete(cmd *cobra.Command, args []string, newClient ClientFactory) (err 
 	defer done()
 
 	if cfg.Name != "" { // Delete by name if provided
-		return client.Remove(cmd.Context(), cfg.Name, cfg.Namespace, fn.Function{}, cfg.All)
+		_, err = client.Remove(cmd.Context(), cfg.Name, cfg.Namespace, fn.Function{}, cfg.All)
+		return err
 	} else { // Otherwise; delete the function at path (cwd by default)
 		f, err := fn.NewFunction(cfg.Path)
 		if err != nil {
 			return err
 		}
-		return client.Remove(cmd.Context(), "", "", f, cfg.All)
+		// updates f.Deploy.<Deployer|Namespace> (clears them on success)
+		f, err = client.Remove(cmd.Context(), "", "", f, cfg.All)
+		if err != nil {
+			return err
+		}
+		return f.Write()
 	}
 }
 
--- a/cmd/delete_test.go
+++ b/cmd/delete_test.go
@@ -7,7 +7,9 @@ import (
 	"strings"
 	"testing"
 
+	"knative.dev/func/pkg/deployers"
 	fn "knative.dev/func/pkg/functions"
+	"knative.dev/func/pkg/keda"
 	"knative.dev/func/pkg/mock"
 	. "knative.dev/func/pkg/testing"
 )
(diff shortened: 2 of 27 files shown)


cmd/delete.go
```diff
@@ -88,7 +89,17 @@ func runDelete(cmd *cobra.Command, args []string, newClient ClientFactory) (err
...
+			return err
+		}
+		// Successful undeploy: clear live-deployment state so func.yaml
+		// reflects reality. Deploy-section config (image, options, etc.) is kept.
+		f.Deploy.Namespace = ""
+		f.Deploy.Deployer = ""
+		if err = f.Write(); err != nil {
+			return fmt.Errorf("function was undeployed from the cluster, but updating the local configuration failed: %w", err)
+		}
+		return nil
```
lkingland: This might actually be part of the implementation of `client.Remove`

cmd/delete_test.go
```diff
@@ -340,3 +342,135 @@ func TestDelete_NoFunctionAtPath(t *testing.T) {
 		t.Fatalf("unexpected error message: %v", err)
 	}
 }
+
+// TestDelete_ByProjectClearsDeployedMarker ensures that a successful
+// path-based delete clears both Deploy.Namespace and Deploy.Deployer -
+// live-deployment state that's stale once undeployed. The next flagless
+// deploy correctly falls back to the global config's deployer default.
+func TestDelete_ByProjectClearsDeployedMarker(t *testing.T) {
```
lkingland, line 349: Likewise, this could be a test in core, because it would be doing this clearing for every deployer (even the mock on the test)

cmd/delete_test.go
```diff
@@ -340,3 +342,135 @@ func TestDelete_NoFunctionAtPath(t *testing.T) {
...
+		t.Fatalf("expected Deploy.Deployer cleared after a successful undeploy, got %q", loaded.Deploy.Deployer)
+	}
+}
+
+// TestDelete_ByProjectThenRedeployWithDifferentDeployerNotBlocked:
+// a keda function, undeployed by project, must not have its subsequent
+// redeploy with a different deployer blocked by the deployer-switch
+// guard - the guard only fires on an already-deployed function, and
+// undeploy should've cleared up that function & func.yaml.
+func TestDelete_ByProjectThenRedeployWithDifferentDeployerNotBlocked(t *testing.T) {
```
lkingland: This might also be a feature/test for core:  two mock deployers, different names, ensure the guard fires correctly.

cmd/delete_test.go
```diff
@@ -340,3 +342,135 @@ func TestDelete_NoFunctionAtPath(t *testing.T) {
...
+	if !deployer.DeployInvoked {
+		t.Fatal("expected the deployer to be invoked - the guard must not fire on an undeployed function")
+	}
+}
+
+// TestDelete_ByNameDoesNotTouchLocalFuncYaml ensures delete-by-name never
+// reads or writes the func.yaml in the current directory, even when one
+// exists there - the deployed-marker clearing added for path-based delete
+// is scoped to that branch only.
+func TestDelete_ByNameDoesNotTouchLocalFuncYaml(t *testing.T) {
```
lkingland: The `cmd` package should have no knowledge of the fact func.yaml even exists.

The test is implemented correctly, but the comment is misleading.

cmd/deploy.go
```diff
@@ -284,6 +285,25 @@ func runDeploy(cmd *cobra.Command, newClient ClientFactory) (err error) {
...
+		from, to := f.Deploy.Deployer, cfg.Deployer
+		if from == "" {
+			// backwards compatible function without deployer persisted.
+			// Pin to Knative deployer, not Default deliberately.
+			from = deployers.Knative
+		}
+		safeSwitch := from == k8s.KubernetesDeployerName && to == keda.KedaDeployerName
+		if from != to && !safeSwitch {
+			return fmt.Errorf("function %q was deployed with the %q deployer; redeploying with %q would orphan the old deployer's resources on the cluster - run func delete first to remove them, then redeploy", f.Name, from, to)
+		}
```
lkingland: This is an interesting case.

My first thought is that the `if from != to` check would be part of core, but the part which makes a "safeSwitch" could be a backwards-compatibility `if` statement here in the CLI which just sets `f.Deploy.Deployer` to `knative` if it is deployed but no value was recorded.

So, essentially splitting this whole block into two pieces 1) the assertion in core and 2) the back-compat default in the cli (where implementation names are expected).

cmd/deploy_test.go
```diff
@@ -2547,3 +2550,301 @@ func TestDeploy_ValidDomain(t *testing.T) {
...
+			t.Fatalf("expected deployer to remain %q after a flag-less redeploy, got %q", other, loaded.Deploy.Deployer)
+		}
+	})
+}
+
+// TestDeploy_DeployerSeededFromGlobalConfig ensures the deployer flag's
+// default is sourced from the global config chain, same as builder. A global
+// config file with a deployer set seeds a flagless deploy of a fresh function
+// with that value.
+func TestDeploy_DeployerSeededFromGlobalConfig(t *testing.T) {
```
lkingland: This one I do think is in the right place, since it's the CLI that loads and applies global config

pkg/config/config.go
```diff
@@ -27,12 +28,16 @@ const (
 
 	// DefaultBuilder is statically defined by the builders package.
 	DefaultBuilder = builders.Default
+
+	// DefaultDeployer is statically defined by the deployers package.
+	DefaultDeployer = deployers.Default
```
lkingland, line 33: Nice to see the (very sparse) global config system holding water 👍🏻

lkingland (commented):
Nice!  My only suggestion is to move most of that logic into core from the CLI pkg

cmd/delete.go
gauron99: ah thanks this is good feedback! for this to be in client some further code work was required first... Im addressing after rebase

cmd/deploy.go
gauron99: The "safe switch check" needs to live in the CLI because the `to` is a `cfg.Deployer` value. The only way we can check for already deployed deployer value is from cluster running service - we'd need to determine what the service is, get its annotations and check. 

OR we could add f.Deployer type of "what to deploy with" and f.Deploy.Deployer "what is deployed with"...

gauron99:
to resolve the tension between unclear boundaries of func.yaml purpose in terms of "half status/ half declared intention" we introduced different means to interpret some keys. usually having one yaml key represent both is fine but sometimes its not. Sometime the current state is a dependency for some action.

Knowing if a function was deployed is essential to deliver "was the deployer changed". We can determine if function was deployed by checking the `f.Depoy.Namespace` because we intentionally created 2 keys for this (the other f.Namespace`) Now we need to introduce the same for deployer because the difference between intent and current state requires it.

cmd/deploy.go
```diff
@@ -159,6 +160,8 @@ EXAMPLES
 	// contextually relevant function; but sets are flattened via cfg.Apply(f)
 	cmd.Flags().StringP("builder", "b", cfg.Builder,
 		fmt.Sprintf("Builder to use when creating the function's container. Currently supported builders are %s.", KnownBuilders()))
+	cmd.Flags().String("deployer", cfg.Deployer,
+		fmt.Sprintf("Type of deployment to use: '%s' for Knative Service, '%s' for Kubernetes Deployment, or '%s' for Deployment with a KEDA HTTP scaler ($FUNC_DEPLOYER)", deployers.Knative, deployers.Kubernetes, deployers.Keda))
```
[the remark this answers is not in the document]
gauron99: resolved

pkg/functions/client.go
```diff
@@ -1150,8 +1165,12 @@ func (c *Client) List(ctx context.Context, namespace string) ([]ListItem, error)
 // function defined at root is used if it exists. If calling this directly
 // namespace must be provided in .Deploy.Namespace field except when using mocks
 // in which case empty namespace is accepted because its existence is checked
-// in the sub functions remover.Remove and pipelines.Remove
-func (c *Client) Remove(ctx context.Context, name, namespace string, f Function, all bool) error {
+// in the sub functions remover.Remove and pipelines.Remove.
+//
+// Returns structure 'f' with f.Deploy.Namespace cleared if the removal was
+// successful and error returned is nil. If error was encountered, returns
+// 'f' unmodified.
```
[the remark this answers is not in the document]
gauron99: resolved

lkingland (approved):
🎉

PR #3967, opened 2026-07-25
test: skip gpg signing in test gitInit

lkingland:
- :bug: Local-disable `commit.gpgsign` and `tag.gpgsign` in the config-ci e2e helper's temp git repo so a host global GPG signing config cannot fail `git commit` with `No secret key` for the synthetic test identity.

/kind bug

Functions#61


gauron99 (approved):
/lgtm

PR #3968, opened 2026-07-25
feat(cluster): install act for local GitHub Actions workflow runs

lkingland:
# Changes

- :gift: Install pinned **act** (nektos/act v0.2.88) via `func cluster create` into `$XDG_CONFIG_HOME/func/bin`, alongside kubectl and kind — so `func config ci` workflows can be exercised locally.
- :gift: Generic `.tar.gz` install path in `pkg/cluster/binaries.go` (`ArchiveEntry`): verify the archive SHA-256, then extract a regular-file entry atomically (size/type guards). Windows: close archive handles before renaming over the download path.
- :broom: `ActOverride` + exported `ClusterConfig.Act()` (managed BinDir, then PATH); CLI samples `FUNC_TEST_ACT` like kubectl/kind.
- :broom: e2e config-ci resolves act via `cluster.ClusterConfig{…}.Act()` (not a hand-built `config.Dir()` path).
- :broom: Unit tests for `extractFromTarGz`; cache hit is existence-only (archive pin ≠ extracted binary).

Checksums match the official v0.2.88 `checksums.txt` for linux/darwin amd64+arm64.

/kind enhancement

# Release Note

```release-note
func cluster create now installs act for local GitHub Actions workflow runs
```

# Docs

- [ ] Docs PR required (link below)
- [x] Docs not required

Functions#62

### DIFF
--- a/e2e/e2e_config_ci_test.go
+++ b/e2e/e2e_config_ci_test.go
@@ -12,7 +12,7 @@ import (
 
 	"gopkg.in/yaml.v3"
 
-	"knative.dev/func/pkg/config"
+	"knative.dev/func/pkg/cluster"
 )
 
 func TestConfigCI_DeployFuncViaGeneratedGitHubWorkflow(t *testing.T) {
@@ -95,11 +95,8 @@ func runGitHubWorkflow(t *testing.T, dir string) {
 	if strings.Contains(Registry, "registry.localtest.me") {
 		args = append(args, "--env", "FUNC_REGISTRY_INSECURE=true")
 	}
-	// Prefer act installed by `func cluster create` ($XDG…/func/bin) over bare PATH.
-	act := "act"
-	if managed := filepath.Join(config.Dir(), "bin", "act"); fileExists(managed) {
-		act = managed
-	}
+	// Resolve act via cluster.Act() (managed BinDir, then PATH; FUNC_TEST_ACT override).
+	act := cluster.ClusterConfig{ActOverride: os.Getenv("FUNC_TEST_ACT")}.Act()
 	cmd := exec.Command(act, args...)
 	cmd.Dir = dir
 	cmd.Stdout = os.Stdout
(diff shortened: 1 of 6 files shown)


e2e/e2e_config_ci_test.go
```diff
@@ -89,7 +91,12 @@ func runGitHubWorkflow(t *testing.T, dir string) {
 	if strings.Contains(Registry, "registry.localtest.me") {
 		args = append(args, "--env", "FUNC_REGISTRY_INSECURE=true")
 	}
-	cmd := exec.Command("act", args...)
+	// Prefer act installed by `func cluster create` ($XDG…/func/bin) over bare PATH.
+	act := "act"
+	if managed := filepath.Join(config.Dir(), "bin", "act"); fileExists(managed) {
```
gauron99: The PR description says there's an `ActOverride` / `act()` accessor, but the diff doesn't have one. The e2e test builds the path by hand through `config.Dir()` instead. This seems to be missing?

lkingland:
@gauron99 Good catch — they were dropped when golangci flagged an unused unexported `act()` after e2e still hand-built the path.

Restored and wired properly in the latest commit:
- `ActOverride` + exported `ClusterConfig.Act()` (BinDir → PATH)
- CLI samples `FUNC_TEST_ACT` (same family as `FUNC_TEST_KUBECTL` / `FUNC_TEST_KIND`)
- e2e config-ci now uses `cluster.ClusterConfig{ActOverride: os.Getenv("FUNC_TEST_ACT")}.Act()`

Thanks!

pkg/cluster/binaries.go
```diff
@@ -46,6 +52,22 @@ var bins = []struct {
 			return fmt.Sprintf("https://github.com/kubernetes-sigs/kind/releases/download/v%s/kind-%s-%s", kindVersion, goos, goarch)
 		},
 	},
+	{
+		Name:         "act",
+		Version:      actVersion,
+		Checksums:    actChecksums,
+		ArchiveEntry: "act",
+		URL: func(goos, goarch string) string {
+			// GitHub release asset names are case-insensitive, so goos
```
gauron99, line 61: note: I verified the case-insensivity too but did not find any documentation on it; if GH ever changes that CDN behaviour, we might get some failures :D

gauron99 (approved):
/lgtm
/approve

PR #3969, opened 2026-07-25
fix(cluster): run Delete when no clusters are tracked

lkingland:
# Changes

- :bug: `func cluster delete` no longer hard-errors with `no clusters exist` when nothing is tracked. It hands off to `cluster.Delete`, which already performs empty-list shared-resource cleanup and returns nil (**idempotent / `rm -f` semantics** — exit 0 where it previously exited non-zero).
- :broom: Quiet `kind delete` warnings when no kubeconfig is present, so empty-system delete does not print a scary "failed to delete cluster" for a cluster that never existed.

Wrong name with existing clusters still errors as before.

/kind bug

# Release Note

```release-note
func cluster delete is idempotent when no clusters are tracked (exit 0, shared cleanup)
```

# Docs

- [ ] Docs PR required (link below)
- [x] Docs not required

Functions#63

### DIFF
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -249,17 +249,24 @@ func runClusterDelete(cmd *cobra.Command, args []string) error {
 		cfg.Name = args[0]
 	}
 
+	// If clusters are tracked and a non-matching name was given, refuse with
+	// a clear message — otherwise hand off to Delete, which handles both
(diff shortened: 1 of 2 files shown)


gauron99 (approved):
/lgtm & the bot

PR #3985, opened 2026-08-03
cluster: Contour when --serving=false; drop dead delete --container-engine

### LINKED ISSUE #3979: cluster delete: remove dead --container-engine flag
## Context

PR #3856 migrated the dev cluster registry from a host-side container to an in-cluster Kubernetes deployment. This removed `teardownRegistry`, which was the only consumer of `ContainerEngineOverride` in the delete path.

## Problem

The `--container-engine` flag on `func cluster delete` (`cmd/cluster.go:233-234`) is now dead code. It parses and stores a value in `ContainerEngineOverride`, but nothing in `Delete()` reads it.

**Evidence:**
- `Delete()` (`pkg/cluster/delete.go:21-55`) uses `cfg.Kubeconfig()`, `cfg.kind()`, `cfg.Name`, and `cfg.SkipRegistryConfig`. It never calls `cfg.ContainerEngine()`.
- The only callsite of `ContainerEngine()` in the cluster package is `getKindNodeIPs` (`kubernetes.go:143`), which runs during **create**, not delete.
- `revertHostRegistry()` auto-detects engines via `hasCommand()`, completely ignoring the override.

**Tested empirically:** `func cluster delete --container-engine=podman` behaves identically to `func cluster delete` — the flag is silently ignored.

## Suggested fix

Remove the `--container-engine` flag and `ContainerEngineOverride` from `newClusterDeleteConfig()` in `cmd/cluster.go`. Also remove it from the `PreRunE: bindEnv(...)` call and the usage template.

/kind cleanup

### LINKED ISSUE #3980: cluster create: registry unreachable from host with --serving=false
## Context

PR #3856 migrated the dev cluster registry from a host-side container (`localhost:50000`) to an in-cluster Kubernetes deployment accessed via a Contour Ingress at `registry.localtest.me`.

## Problem

The registry Ingress uses `ingressClassName: contour-external`, but Contour is only installed when `--serving=true` (the default). With `--serving=false`, no Ingress controller exists, so the host cannot reach the registry.

**Before #3856:** The registry was a host-side container on `localhost:50000` — always reachable regardless of `--serving`.

**After #3856:** Host registry access requires Contour → requires `--serving=true`.

The help text (`cmd/cluster.go:109-110`) documents `--serving=false --eventing=false` as a valid "minimal cluster (just Kubernetes + registry)" use case, implying the registry should be functional.

**Empirical test results** (from `func cluster create --serving=false --eventing=false`):
- Registry deployment: `1/1 Available` inside the cluster ✅
- Ingress object: exists but has **no ADDRESS** (no controller to reconcile it) ❌
- `contour-external` namespace: does not exist ❌
- `curl http://registry.localtest.me/v2/`: **Connection reset by peer** ❌
- In-cluster containerd mirrors via `hostPort:5000`: work fine ✅

Note: `hack/cluster.sh` doesn't have this problem because it always installs serving + Contour unconditionally.

## Suggested fixes (pick one)

- **(a) Document it** — note that `--serving=false` means no host push, only in-cluster access
- **(b) Always install Contour** — decouple Contour installation from `--serving` so the registry Ingress always has a controller
- **(c) Fallback port mapping** — add a Kind `extraPortMappings` entry that maps a host port to the registry's `hostPort:5000`, bypassing Ingress entirely

/kind bug

lkingland:
## Summary

Follow-up to #3856 (squash-merged at `c5310074` before these review fixes landed on the PR tip).

**S1:** Remove dead `--container-engine` from `func cluster delete` — nothing in the delete path read `ContainerEngineOverride` after host registry container teardown was removed.

**S2:** When `--serving=false`, Contour was not installed, but the registry still applied an Ingress with `ingressClassName: contour-external`. Host `docker push registry.localtest.me/...` then failed. Extract `installContour` from `installNetworking`; on the registry path, install Contour before the registry when Serving is off. Default `--serving=true` unchanged (matches `hack/cluster.sh`, which always installs Contour via `networking`). Help documents Contour on minimal create.

## Test plan

- [ ] `go build ./cmd/ ./pkg/cluster/`
- [ ] Default `func cluster create` path unchanged (Serving + Contour via svr)
- [ ] `func cluster create --serving=false --eventing=false` installs Contour then registry; host registry hostname resolves via Ingress

Functions#43 follow-up.

Fixes #3979
Fixes #3980

### DIFF
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -54,7 +54,7 @@ EXAMPLES
 	o Create a default development cluster
 	  $ {{rootCmdUse}} cluster create
 
-	o Create a minimal cluster (just Kubernetes + registry)
+	o Create a minimal cluster (Kubernetes + Contour + registry; no Serving/Eventing)
 	  $ {{rootCmdUse}} cluster create --serving=false --eventing=false
 
 	o List existing clusters
@@ -99,14 +99,17 @@ DESCRIPTION
 	  export KUBECONFIG=~/.config/func/clusters/func.local/kubeconfig.yaml
 
 	Installed components are controlled by the --serving, --eventing,
-	--tekton, and --keda flags. Required binaries are downloaded into the
-	func config directory on first use unless --skip-binaries is set.
+	--tekton, and --keda flags. --serving and --eventing default to true
+	(hack/cluster.sh always installs Serving). Contour is always present for
+	the registry host Ingress: with Serving (default path) or standalone when
+	--serving=false. Required binaries are downloaded into the func config
+	directory on first use unless --skip-binaries is set.
 
 EXAMPLES
-	o Create a default development cluster
+	o Create a default development cluster (Serving + Eventing + registry)
 	  $ {{rootCmdUse}} cluster create
 
-	o Create a minimal cluster (just Kubernetes + registry)
+	o Create a minimal cluster (Kubernetes + Contour + registry; no Serving/Eventing)
 	  $ {{rootCmdUse}} cluster create --serving=false --eventing=false
 
 	o Create a cluster with a custom name and domain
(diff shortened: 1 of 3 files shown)


gauron99 (approved):
/lgtm
/approve
some tests might be useful, can be followup

PR #3986, opened 2026-08-03
feat(mcp): add 'invoke' tool and update documentation

Ankitsinghsisodya:
# Changes

- :gift: Add an `invoke` MCP tool that lets agents send a test request to a local or remote running Function, mirroring `func invoke` (path, target, format, id, source, type, data, content-type, request-type, file, insecure, verbose)
- :gift: Add a `func://help/invoke` MCP resource exposing `func invoke --help`
- :gift: Document the `invoke` tool in the MCP agent instructions, including the deploy-then-verify workflow (invoke a Function after deploying it to confirm it responds successfully before reporting success)

/kind enhancement

**Release Note**
```release-note
Added an `invoke` tool to the func MCP server, allowing AI agents to send a test request to a local or remote running Function and verify it responds successfully (e.g. after a deploy).
```

**Docs**

```docs

```

### DIFF
--- a/pkg/mcp/instructions_warning.md
+++ b/pkg/mcp/instructions_warning.md
@@ -15,13 +15,12 @@ The Functions MCP server is currently running in **read-only mode**.
 **Disabled operations:**
 - Deploy to cluster
 - Delete from cluster
-- Invoke a Function
 
-These operations are disabled to prevent unintended cluster modifications and side effects from invoking a Function's handler.
+These write operations are disabled to prevent unintended cluster modifications.
 
 ## Enabling Write Mode
 
-If the user needs to deploy, delete, or invoke Functions, you MUST inform them to enable write mode:
+If the user needs to deploy or delete Functions, you MUST inform them to enable write mode:
 
 1. Close/exit this application completely
 2. Set the environment variable: `FUNC_ENABLE_MCP_WRITE=true`
--- a/pkg/mcp/instructions_warning.md
+++ b/pkg/mcp/instructions_warning.md
@@ -15,12 +15,13 @@ The Functions MCP server is currently running in **read-only mode**.
 **Disabled operations:**
 - Deploy to cluster
 - Delete from cluster
+- Invoke a Function
 
-These write operations are disabled to prevent unintended cluster modifications.
+These operations are disabled to prevent unintended cluster modifications and side effects from invoking a Function's handler.
 
 ## Enabling Write Mode
 
-If the user needs to deploy or delete Functions, you MUST inform them to enable write mode:
+If the user needs to deploy, delete, or invoke Functions, you MUST inform them to enable write mode:
 
 1. Close/exit this application completely
 2. Set the environment variable: `FUNC_ENABLE_MCP_WRITE=true`
--- a/pkg/mcp/tools_invoke.go
+++ b/pkg/mcp/tools_invoke.go
@@ -1,0 +1,72 @@
+package mcp
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+var invokeTool = &mcp.Tool{
+	Name:        "invoke",
+	Title:       "Invoke Function",
+	Description: "Invoke a local or remote Function with a test request.",
+	Annotations: &mcp.ToolAnnotations{
+		Title:          "Invoke Function",
+		ReadOnlyHint:   false,
+		IdempotentHint: false, // Invoking a Function may trigger arbitrary side effects in the Function's handler.
+	},
+}
+
+func (s *Server) invokeHandler(ctx context.Context, r *mcp.CallToolRequest, input InvokeInput) (result *mcp.CallToolResult, output InvokeOutput, err error) {
+	out, err := s.executor.Execute(ctx, "invoke", input.Args()...)
+	if err != nil {
+		err = fmt.Errorf("%w\n%s", err, string(out))
+		return
+	}
+	output = InvokeOutput{
+		Message: string(out),
+	}
+	return
+}
+
+// InvokeInput defines the input parameters for the invoke tool.
+type InvokeInput struct {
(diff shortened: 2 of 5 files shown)


Ankitsinghsisodya:
cc: @lkingland

pkg/mcp/instructions_warning.md
```diff
@@ -15,12 +15,13 @@ The Functions MCP server is currently running in **read-only mode**.
 **Disabled operations:**
 - Deploy to cluster
 - Delete from cluster
+- Invoke a Function
 
-These write operations are disabled to prevent unintended cluster modifications.
+These operations are disabled to prevent unintended cluster modifications and side effects from invoking a Function's handler.
```
lkingland: No need to change this.  We're concerned with literally `write operations`, and `invoke` needs not be disabled (it's a non-mutative call).  Deploy, Delete, etc., however, are.

pkg/mcp/instructions_warning.md
```diff
@@ -15,12 +15,13 @@ The Functions MCP server is currently running in **read-only mode**.
...
 
 ## Enabling Write Mode
 
-If the user needs to deploy or delete Functions, you MUST inform them to enable write mode:
+If the user needs to deploy, delete, or invoke Functions, you MUST inform them to enable write mode:
```
lkingland: As per above, invoke need not be disabled when in RO mode.

Think of it this way:  does invoking a function mutate the cluster state?  That's the real question when deciding if it's disabled 👍🏻

pkg/mcp/tools_invoke.go
```diff
@@ -0,0 +1,90 @@
...
+
+// InvokeInput defines the input parameters for the invoke tool.
+type InvokeInput struct {
+	Path        string            `json:"path" jsonschema:"required,Path to the function project directory"`
+	Target      *string           `json:"target,omitempty" jsonschema:"Function instance to invoke: local, remote, or a URL (default: auto-discovery; prefers local when both local and remote are running)"`
+	Format      *string           `json:"format,omitempty" jsonschema:"Format of message to send: http or cloudevent (default: auto-detected)"`
+	ID          *string           `json:"id,omitempty" jsonschema:"CloudEvent id for the request data"`
+	Source      *string           `json:"source,omitempty" jsonschema:"CloudEvent source for the request data"`
+	Type        *string           `json:"type,omitempty" jsonschema:"CloudEvent type for the request data"`
+	Data        *string           `json:"data,omitempty" jsonschema:"Data (content) to send in the request"`
```
lkingland: When `data` (and `file`) are omitted, `func invoke` still sends a real default payload — currently `{"message":"Hello World"}` with content-type `application/json` (and default source/type of `/boson/fn` / `boson.fn`, method POST). See `DefaultInvoke*` in `pkg/functions/invoke.go`.

The schema here only says "Data (content) to send in the request", so an agent can reasonably assume a bare invoke is an empty probe and then misread handler behavior.

Let's include the actual CLI defaults in the jsonschema description here (and ideally also in the `### invoke` section of `instructions.md`) so agents know what they will actually get when they omit these fields and avoid confusion.

pkg/mcp/tools_invoke.go
```diff
@@ -0,0 +1,90 @@
...
+	ContentType *string           `json:"contentType,omitempty" jsonschema:"MIME type of the data"`
+	RequestType *string           `json:"requestType,omitempty" jsonschema:"HTTP method override (e.g., GET, POST)"`
+	File        *string           `json:"file,omitempty" jsonschema:"Path to a file whose content is used as the request data (overrides data)"`
```
lkingland: `--file` is resolved relative to the **MCP server process CWD** (the `func` child inherits that), not the agent's working directory and not relative to `path`.

That is the same class of footgun that made `path` required + absolute-path-only in the agent instructions: a human can `cd myfunc && func invoke --file=payload.json`, but an agent that passes `path=/abs/myfunc` with `file=payload.json` will often fail or read the wrong file.

A simple documentation addition stating that `file` must be an absolute path (in this jsonschema string and/or the `### invoke` instructions), matching the absolute-path guidance already given for `path`, should do the trick!

pkg/mcp/tools_invoke.go
```diff
@@ -0,0 +1,90 @@
...
+type InvokeInput struct {
+	Path        string            `json:"path" jsonschema:"required,Path to the function project directory"`
+	Target      *string           `json:"target,omitempty" jsonschema:"Function instance to invoke: local, remote, or a URL (default: auto-discovery; prefers local when both local and remote are running)"`
+	Format      *string           `json:"format,omitempty" jsonschema:"Format of message to send: http or cloudevent (default: auto-detected)"`
+	ID          *string           `json:"id,omitempty" jsonschema:"CloudEvent id for the request data"`
+	Source      *string           `json:"source,omitempty" jsonschema:"CloudEvent source for the request data"`
+	Type        *string           `json:"type,omitempty" jsonschema:"CloudEvent type for the request data"`
+	Data        *string           `json:"data,omitempty" jsonschema:"Data (content) to send in the request"`
+	ContentType *string           `json:"contentType,omitempty" jsonschema:"MIME type of the data"`
+	RequestType *string           `json:"requestType,omitempty" jsonschema:"HTTP method override (e.g., GET, POST)"`
```
lkingland: Minor schema faithfulness nit: the CLI accepts only `GET` or `POST` (`http request type '…' not supported` otherwise), with default POST — see `cmd/invoke.go` / `pkg/functions/invoke.go`.

"HTTP method override (e.g., GET, POST)" both invents "override" and invites other methods via the open-ended "e.g.".

Prefer something like: `HTTP method: GET or POST (default: POST)`.

lkingland (commented):
Very good!

I only found a few minor things (see below)

💯  Nice work so far

pkg/mcp/instructions_warning.md
Ankitsinghsisodya: yes, I got it that invoke doesn't mutate the cluster state.

Ankitsinghsisodya:
@lkingland I have done all the improvements as suggested

PR #3988, opened 2026-08-04
added portmapping for registory to bypass ingress

### LINKED ISSUE #3980: cluster create: registry unreachable from host with --serving=false

nXtCyberNet:
# Changes

-  added port mapping for 5000 for registry for ingress bypass
- Changed Magic DNS configuration to wait for coredns deployment rollout instead of all  pods to avoid race conditions 

/kind bug
/kind enhancement

Fixes #3980

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/pkg/cluster/dns.go
+++ b/pkg/cluster/dns.go
@@ -34,10 +34,9 @@ func configureMagicDNS(ctx context.Context, cfg ClusterConfig, out io.Writer) er
 		return err
 	}
 	if err := run(ctx, out, "",
-		cfg.kubectl(), "wait", "pod",
-		"--for=condition=Ready", "-l", "!job-name",
+		cfg.kubectl(), "rollout", "status", "deploy/coredns",
 		"-n", "kube-system", "--timeout=60s"); err != nil {
-		return fmt.Errorf("waiting for coredns: %w", err)
+		return fmt.Errorf("waiting for coredns rollout: %w", err)
 	}
 
 	success(out, "Magic DNS", time.Since(start))
--- a/pkg/cluster/kubernetes.go
+++ b/pkg/cluster/kubernetes.go
@@ -27,6 +27,10 @@ nodes:
     - containerPort: 30022
       hostPort: 30022
       listenAddress: "127.0.0.1"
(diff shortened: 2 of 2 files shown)


gauron99:
Sorry about the overlap.

I assigned you these issues as follow-ups to #3856. The original author also landed the same fixes in a larger follow-up (#3985), which will close #3979 / #3980.

Please close this as superseded. Happy to find you another issue if you want one.

PR #3991, opened 2026-08-04
ocp-route: new WithExposer() on deployer impl

gauron99:
Functions deployed with the **raw** and **keda** deployers can now get a public URL through an OpenShift Route.
- Removed the "slightly convenient" deployer switch raw -> keda allowance -> all deployer switching now prints an error that resources would be stranded and to undeploy first
- raw deployer adds Route to function's namespace, adds ownerReference, all related resources live as long as function service exists and/or is redeployed otherwise.

- keda deployer deploys the Route in keda-derived namespace (`keda` or `openshift-keda`) because it needs to target the intereceptor in order for keda's scaling to work. (Route cannot target services across namespaces) - ownerRef also cannot target across namespaces -> this requires explicit deletion.

- we add selectors (what does not change for a deployed function), labels and annotations to identify the function and if it was deployed/how. - this is convenient so that commands like `list` and `describe` can only fetch the function service as usual and learn all about it via its metadata, no need to fetch other resources around the cluster.

# AI info

## Usage

```sh
func deploy --deployer=raw --expose=route                       # public https URL, hostname minted by the router
func deploy --deployer=keda --expose=route --domain=fn.corp.com # custom domain
func deploy --deployer=raw --expose=none                        # back to cluster-local; the Route is removed
```

- **Cluster-local stays the default.** Nothing is exposed unless you ask.
- The chosen mode persists in func.yaml and survives flagless redeploys; `--expose=""` clears it.
- `func describe` leads with the public URL of an exposed function.
- **Custom domains**: DNS and certificates stay user's responsibility. Point DNS at the router; cert-manager's openshift-routes flow works, and injected certificates survive redeploys.
- On a cluster without the Route API, `--expose=route` is refused before anything is built, with a distinct message when the cluster could not be reached. The knative deployer ignores the flag (it has its own exposure) with a warning.
- `func delete` removes the exposure with the function, including keda's Route, which nothing garbage collects.

## How it works

<details>
<summary>Mechanism, for reviewers</summary>

- `pkg/deployer.Exposer` is a small pluggable interface; `pkg/ocproute` implements it through the dynamic client (no `github.com/openshift/api` dependency).
- Raw: the Route sits beside the function, owned by its Service, so garbage collection removes it with the function. The owner reference is hand-built without `BlockOwnerDeletion`, which OpenShift's `OwnerReferencesPermissionEnforcement` would reject for ordinary users.
- Keda: the Route lives in the interceptor's namespace and targets the interceptor, preserving scale-from-zero. Cross-namespace owner references are forbidden, so removal is explicit: on `--expose=none` and on `func delete`, with a best-effort candidate-namespace sweep for Routes a crash left unrecorded. A namespace the account cannot read is warned about and skipped, so deleting a never-exposed function needs no interceptor-namespace permissions.
- The interceptor's namespace is probed (`openshift-keda` for CMA, `keda` for the upstream chart). Exposure refuses when the probe cannot confirm one; cluster-local deploys proceed on the platform's guess.
- func.yaml records intent (`expose`) and applied state (`deploy.expose`) separately. Remote deploys record what the pipeline's describer observed, never the intent: a published func-util that predates this feature deploys cluster-local, and the CLI warns instead of recording a Route that does not exist. A Route whose record cannot be written is rolled back within the same deploy.
- The `func.domain` label is excluded from the immutable Deployment and Service selectors; updates preserve a legacy Deployment's live selector, and changing a domain pinned by one refuses with recreation instructions. A domain change replaces the Route (in-place host edits need `routes/custom-host` update permission project admins lack; the old certificate names the old host). Updates reconcile func's owned fields onto the live Route, so third-party state survives.
- The same platform gate runs in func-util, covering git/PAC deploys no CLI ever sees. The first commit removes the special-cased deployer switch handling in preparation.

</details>

## Testing

- Unit: exposure reconcile and records, recording rollback, interceptor resolution, remover sweep, and guards for failures only OpenShift surfaces (selector immutability, `BlockOwnerDeletion` admission).
- E2E, split by platform: refusal tests run on KinD in CI; Route round trips, keda toggle and delete, custom domain with TLS, and remote deploys run on OpenShift (manual leg, `make test-e2e-expose`).


pkg/ocproute/route.go
```diff
@@ -0,0 +1,423 @@
...
+// and reports the real limit.
+func (x *Exposer) generate(e deployer.Exposure) (*unstructured.Unstructured, error) {
+	labels, err := deployer.GenerateCommonLabels(e.Function, e.Decorator)
+	if err != nil {
+		return nil, err
+	}
+
+	labels[fnlabels.FunctionNamespaceKey] = e.FunctionNamespace
+
+	annotations := deployer.GenerateCommonAnnotations(e.Function, e.Decorator, false /* dapr n/a for routing */, x.deployerName)
```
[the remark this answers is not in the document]
gauron99: Fixed: the stamp is now asserted after user annotations in `GenerateCommonAnnotations` (same order `GenerateCommonLabels` uses), with a unit test.

pkg/k8s/deployer.go
```diff
@@ -469,7 +662,7 @@ func (d *Deployer) generateDeployment(f fn.Function, namespace string, daprInsta
 		Spec: appsv1.DeploymentSpec{
 			Replicas: &replicas,
 			Selector: &metav1.LabelSelector{
-				MatchLabels: labels,
+				MatchLabels: deployer.SelectorLabels(labels),
```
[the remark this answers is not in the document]
gauron99: Fixed: updates now preserve the live Deployment's selector, so legacy Deployments pinning `func.domain` redeploy fine while the domain is unchanged. Changing or removing a pinned domain is refused with recreation instructions (`func delete` + redeploy), since no in-place update can alter an immutable selector. Covered by unit tests for all four cases (preserved, changed, removed, current-era).

pkg/ocproute/route.go
```diff
@@ -0,0 +1,423 @@
...
+		}
+		maps.Copy(labels, route.GetLabels())
+		merged.SetLabels(labels)
+
+		annotations := merged.GetAnnotations()
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		maps.Copy(annotations, route.GetAnnotations())
+		merged.SetAnnotations(annotations)
```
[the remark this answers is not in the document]
gauron99: Fixed: the update overlay now reasserts `metadata.ownerReferences` from the generated Route, so the owner tracks the current Service UID. Keda's ownerless Exposure generates no owner reference and its Route stays untouched. Unit-tested both ways.

pkg/keda/remover.go
```diff
@@ -32,36 +52,83 @@ func (remover *Remover) Remove(ctx context.Context, name, ns string) error {
...
+// is never read as empty.
+//
+// dynClient is a parameter so a test can reach this.
+func (remover *Remover) unexpose(ctx context.Context, dynClient dynamic.Interface, recordedNS, name, ns string) error {
+	if recordedNS == "" {
+		for _, candidate := range interceptorNamespaceCandidates() {
+			if err := remover.exposer.Unexpose(ctx, dynClient, interceptorExposureRef(name, ns, candidate)); err != nil {
+				return fmt.Errorf("no record said where this function's Route was, and namespace %q "+
+					"could not be checked for one: %w", candidate, err)
+			}
```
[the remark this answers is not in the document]
gauron99: Fixed: the no-record sweep is now best-effort - a candidate that cannot be checked is skipped with a warning instead of failing the delete, so cluster-local functions delete without interceptor-namespace permissions. Denial on a recorded namespace stays fatal: the record means removal is owed. Unit-tested.

pkg/k8s/deployer.go
```diff
@@ -243,16 +287,190 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+		if err := d.exposer.Unexpose(ctx, dynClient, ref); err != nil {
+			return "", "", fmt.Errorf("failed to remove external exposure: %w", err)
+		}
+		if err := SetRouteHostname(ctx, clientset, namespace, f.Name, "", ""); err != nil {
+			return "", "", err
+		}
+		return defaultURL, "", nil
+
+	default: // nothing wanted, nothing recorded: Route API untouched
+		return defaultURL, "", nil
```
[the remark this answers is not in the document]
gauron99: Fixed: a recording failure now rolls the just-created Route back within the same deploy, and a failed rollback reports both errors. A kill between create and record still orphans; delete's garbage collection remains the backstop. Unit-tested.

pkg/keda/deployer.go
```diff
@@ -116,24 +158,81 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+		// record outlives the Route, matching the raw deployer and the remover,
+		// at the price of a dead hostname in describe until the retry lands.
+		// recordedNS is read from appService, fetched before the exposure work
+		// above.
+		if appliedExpose == "" {
+			if recordedNS := appService.Annotations[k8s.RouteNamespaceAnnotation]; recordedNS != "" {
+				if err := d.exposer.Unexpose(ctx, dynClient, interceptorExposureRef(f.Name, namespace, recordedNS)); err != nil {
+					return fn.DeploymentResult{}, fmt.Errorf("failed to remove external exposure: %w", err)
+				}
+			}
```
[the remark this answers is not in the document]
gauron99: Fixed with a rollback rather than the suggested opt-out sweep: recording failure takes the just-created Route down via the interceptor namespace already in hand. Sweeping on opt-out was deliberately avoided - every cluster-local redeploy has no record, so it would demand interceptor-namespace permissions on ordinary deploys. Delete's sweep remains the backstop for a hard kill. Unit-tested.

pkg/ocproute/route.go
```diff
@@ -0,0 +1,399 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to generate Route: %w", err)
+	}
+
+	name, err := x.ensure(ctx, client, e, route)
+	if err != nil {
+		return "", err
+	}
+
+	return waitForAdmitted(ctx, client, e.Namespace, name, admissionTimeout)
```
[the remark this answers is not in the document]
gauron99: Fixed in e590351e. ensure now reports whether the call brought the Route into existence (the plain create and the domain-change replacement) and Expose deletes that Route again when admission fails, folding a failed deletion into the returned error. A pre-existing Route that was only updated stays: it was admitted once and may be serving. Pinned by Test_Expose_RollsBackUnadmittedRoute and Test_Expose_KeepsPreexistingRouteOnAdmissionFailure.

pkg/ocproute/route.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	// A changed domain replaces the Route rather than editing it: an
+	// injected certificate names the old host, so it must die with the old
+	// Route instead of surviving as a mismatch on the new one. Editing
+	// spec.host in place is also closed off, needing "update" on
+	// routes/custom-host, which project admins lack by default; delete and
+	// create need only "create", which they have.
+	if existing.GetLabels()[deployer.DomainLabel] != e.Function.Domain {
+		if err := routes.Delete(ctx, existing.GetName(), metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
gauron99: Not changing this. The scenario needs a managed Route under a name func would not generate, and no released func has ever generated a different name; Routes are new in this PR and both deployers have one naming rule. find is label-based so removal stays robust, not to migrate names. If a name is taken at create, func refuses and creates nothing, which is the existing AlreadyExists path.

pkg/keda/exposure.go
```diff
@@ -0,0 +1,191 @@
...
+		interceptorServiceName, strings.Join(candidates, " or "))
+}
+
+// interceptorExposureName builds the name of the object func creates to
+// expose a keda function; the interceptor itself is keda's, func only routes
+// through it. Every keda function's exposure lands in the one interceptor
+// namespace, so the name carries the function's namespace too; without it two
+// functions of the same name in different namespaces would collide.
+func interceptorExposureName(name, namespace string) string {
+	return fmt.Sprintf("%s-%s", name, namespace)
```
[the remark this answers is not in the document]
gauron99: Leaving this as is. A collision here lands in the same rule every name conflict in this feature follows: func refuses and creates nothing. Labels keep the two functions' Routes distinct, so nothing is damaged or wrongly adopted; the second function reports the conflict and renaming either function unblocks it. A collision-proof scheme (hash suffix) is planned together with the handling for over-long Route names, which needs the same mechanism.

cmd/client.go
```diff
@@ -172,22 +175,30 @@ func newKnativeDeployer(verbose bool) fn.Deployer {
...
-
-	return k8s.NewDeployer(options...)
+		k8s.WithExposer(ocproute.New(deployers.Kubernetes)),
+	)
 }
 
+// newKedaDeployer builds the keda deployer. The Exposer is keda's own, never
+// the embedded raw deployer's, so Routes point at the interceptor rather than
+// bypassing it. Attached unconditionally for the reason in newK8sDeployer,
+// which bites harder here: keda's Route has no owner reference, so a Route
```
lkingland, line 195: Honest take — definitely load-bearing — a sharp call, even.

cmd/client.go
```diff
@@ -70,7 +72,8 @@ func NewClient(cfg ClientConfig, options ...fn.Option) (*fn.Client, func()) {
 			fn.WithRepositoriesPath(config.RepositoriesPath()),
 			fn.WithScaffolder(buildpacks.NewScaffolder(cfg.Verbose)),
 			fn.WithBuilder(buildpacks.NewBuilder(buildpacks.WithVerbose(cfg.Verbose))),
-			fn.WithRemovers(knative.NewRemover(cfg.Verbose), k8s.NewRemover(cfg.Verbose), keda.NewRemover(cfg.Verbose)),
+			fn.WithRemovers(knative.NewRemover(cfg.Verbose), k8s.NewRemover(cfg.Verbose),
+				keda.NewRemover(cfg.Verbose, keda.WithRemoverExposer(ocproute.New(deployers.Keda)))),
```
lkingland: "RemoverExposer" leads to more questions than answers to me.

cmd/client.go
```diff
@@ -172,22 +175,30 @@ func newKnativeDeployer(verbose bool) fn.Deployer {
...
+// wiring time cannot know.
 func newK8sDeployer(verbose bool) fn.Deployer {
-	options := []k8s.DeployerOpt{
+	return k8s.NewDeployer(
 		k8s.WithDeployerVerbose(verbose),
 		k8s.WithDeployerDecorator(deployDecorator{}),
-	}
-
-	return k8s.NewDeployer(options...)
+		k8s.WithExposer(ocproute.New(deployers.Kubernetes)),
```
lkingland, line 188: It seems to me like in both cases (keda and k8s deployers), the exposer could be defaulted to this, as there would presumably be no import cycle with ocproute, and it's compatible with both OpenShift and vanilla Kubernetes.

That would greatly simplify the CLI (and anyone wanting to use Functions via the library), since a simple
`deployer := k8s.NewDeployer()` would be sufficient.

Is there a reason we need the deploy decorator and ocproute exposer instantiated here in the CLI?

cmd/deploy.go
```diff
@@ -570,6 +615,11 @@ type deployConfig struct {
 
 	// ManagementDisabled disables automatic Function CR sync after deploy.
 	ManagementDisabled bool
+
+	// Expose is the intended external exposure mode from --expose / FUNC_EXPOSE
+	// (maps to Function.Expose). Defaults to cluster-local; "none" is explicit;
+	// "route" is OpenShift-only.
+	Expose string
```
lkingland, line 620: I'm still wondering if this could be a boolean.

A Function is either configured such that we want the system to "do whatever we can" to expose it, which has varying levels of difficulty depending on the cluster distribution and configuration.

Maybe I'm missing a case here.

cmd/errors.go
```diff
@@ -217,6 +220,32 @@ func (e *ErrInvalidDomain) Unwrap() error {
...
+func NewErrInvalidExpose(err error) error {
+	return &ErrInvalidExpose{Err: err}
+}
+
+func (e *ErrInvalidExpose) Error() string {
+	return fmt.Sprintf(`%v
+
+Try this:
+  func deploy --expose=route  Create an OpenShift Route (OpenShift clusters only)
+  func deploy --expose=none   Cluster-local, no external exposure
```
lkingland, line 236: For example, if we also set the default to `true`, then we're even lingustically compatible with our other supported cluster configurations:
```
func deploy --expose=false Do not create an external route (currently OpenShift only)
```

cmd/errors.go
```diff
@@ -217,6 +220,32 @@ func (e *ErrInvalidDomain) Unwrap() error {
...
+
+deploy.expose takes effect with the raw and keda deployers only (--deployer=raw or --deployer=keda).
+Functions are cluster-local by default; external exposure is opt-in.
+For more options, run 'func deploy --help'`, e.Err)
```
lkingland, line 241: Functions are actually "expose by default" right now (if you're configured correctly).  What's _not_ done by default is the DomainMapping (you get a default, less pretty, URL).

pkg/deployer/expose.go
```diff
@@ -0,0 +1,96 @@
...
+	// Namespace holds the exposing object, not always the function's own:
+	// keda targets the interceptor Service in the operator's namespace, and an
+	// exposing object can only target a Service beside it.
+	Namespace string
+}
+
+// Exposure describes the external address wanted for one function: which
+// Service to send traffic to, and what to name the object that does it.
+type Exposure struct {
+	Function fn.Function
```
lkingland: If this structure is "data", then we probably don't want to embed the entire Function struct.

If it's a "doer" object, then the canonical place for the FunctionNamespace is on the function struct, so either we don't want to reference the entire `Function`, or we should use a convenience accessor like
```
ns := myExposure.FunctionNamespace() ->

func (e Exposure) FunctionNamespace() string {
  return e.Function.Deploy.Namespace
}
```

Looks to me like this structure is more like "data", so maybe removing the Function member, and passing that around as an argument would make this a bit more encapsulated.

pkg/deployer/expose.go
```diff
@@ -0,0 +1,96 @@
...
+	Owner *metav1.OwnerReference
+
+	Decorator DeployDecorator
+}
+
+// Ref is what identifies this Exposure's object on the cluster, as opposed to
+// what describes how to build it.
+func (e Exposure) Ref() ExposureRef {
+	return ExposureRef{
+		FunctionName:      e.Function.Name,
```
lkingland: I'd probably make this `e.FunctionName` just like `FunctionNamespace`.

pkg/k8s/openshift.go
```diff
@@ -130,40 +131,60 @@ func GetOpenShiftDockerCredentialLoaders() []creds.CredentialsCallback {
...
+// IsOpenShift is a convenient wrapper for getting simple yes/no for openshift
+// cluster. The inner function should run in the cmd layer once to resolve the
+// detectOnce.Do(), any call after is cached so we dont have to call API all the
+// time.
+//
+// note: gauron99: this might change after restructuring to kubeconfig resolution
+// at the start of program instead of adhoc API calls of kube client throughout
+// the codebase
+func IsOpenShift() bool {
+	ok, _ := DetectOpenShift()
```
lkingland, line 177: Could capture the error and write it to STDERR as a warning, just in case.

pkg/keda/deployer.go
```diff
@@ -116,24 +158,77 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+			url = fmt.Sprintf("https://%s", exposedHost)
+			appliedExpose = f.Expose
+		}
+	}
 
 	if err := d.ensureHTTPScaledObject(ctx, f, namespace, deployment, appService, hosts); err != nil {
 		return fn.DeploymentResult{}, fmt.Errorf("failed to ensure http scaled object exists: %w", err)
 	}
 
+	if d.exposer != nil {
```
lkingland: This Deploy method is getting a bit long.  Notice there's two blocks with the same conditional (`if d.exposer != nil`).

I'd take a little while to try extracting this method down by ~50% and when coupled with turning `expose` into a boolean, I bet a nontrivial amount of complexity drops out; and with it, places for bugs to hide

pkg/keda/deployer.go
```diff
@@ -205,11 +300,38 @@ func (d *Deployer) httpScaledObject(f fn.Function, namespace string, deployment
...
+	err := k8s.SetRouteHostname(ctx, clientset, namespace, name, exposedHost, routeNS)
+	if err == nil {
+		return nil
+	}
+	if exposedHost == "" {
+		// A clear, not a record: nothing was created this deploy, so there
+		// is nothing to roll back.
+		return err
+	}
+	if rbErr := d.exposer.Unexpose(ctx, dynClient, interceptorExposureRef(name, namespace, interceptorNS)); rbErr != nil {
```
lkingland: A method tasked with "recording" an exposure should probably not be responsible for _removing_ an exposure if it fails.  That's an unintended side-effect.

Perhaps capture the failure to record exposure one level up and then, as part of the error handling, clean up.

pkg/keda/remover.go
```diff
@@ -32,36 +52,80 @@ func (remover *Remover) Remove(ctx context.Context, name, ns string) error {
...
+
+// unexpose removes the function's Route, found by its name and namespace
+// labels. A record (recordedNS non-empty) names the one namespace to act
+// on, and failing there fails the delete: the record says removal is owed.
+// With no record, the candidate namespaces are swept best-effort for a
+// Route a crash left unrecorded: a candidate that cannot be checked is
+// warned about and skipped. dynClient is a parameter so a test can reach
+// this.
+func (remover *Remover) unexpose(ctx context.Context, dynClient dynamic.Interface, recordedNS, name, ns string) error {
+	if recordedNS == "" {
```
lkingland: (minor observation)

In opposition to some purists who believe every function should have exactly one exit line, I believe quick exit is often better.

It might be worth adding to your agent's code style instructions to exit rather than nest.
```
if recordedNS != "" {
  return nil
}
// assume recordedNS == ""
```

Really helps keep things easier to read.

lkingland (commented):
+5,218 -217, that's quite a PR!

I didn't maually read every line, but overall it's looking clean and good where it matters.

By perhaps revisiting if we can use a boolean for `expose`, and shifting around where the openshift package is used, I suspect a good deal of complexity can get pushed down out of the CLI and Core.

I'll try to do a deeper review when it's out of Draft form manually again, and throw some tokens at the review then as well.  looking great so far 🚀 👍🏻

cmd/deploy.go
gauron99: Yes and No -- I have a note in my "followups" to make this a string AND boolean (thats for the flag). it would smartly default to an exposer mechanism we define (later gateway for vanilla k8s/route for openshift) when you simply write flag `--expose`. BUT the problem is you can still use gateway on Openshift so its not one stop shop even per platform.

Therefore this obviously cant be ONLY boolean if we introduce a different exposure mechanism. therefore playing ahead.
We will have `route` (OpenShift) and `gateway`

cmd/errors.go
gauron99: yes, see note above

pkg/k8s/openshift.go
gauron99: We currently have a Sync.Once mechanism where we ask the API via the `DetectOpenShift` directly upfront (where err is returned) and then with the cached return we run this boolean only (`IsOpenshift`).
 
Im hoping to refactor this with the kube client injection through the whole system instead of calling the current kubeconfig everywhere.

pkg/deployer/expose.go
gauron99: Sure! will take a look

cmd/client.go
gauron99: We currently default to no exposure because of simplicity. as the first commit I removed my "one deployer switch" which was ok, the `deployed raw` -> `redeploy with keda` because the complexity of keeping that alive was more trouble at the time with adding the exposer and felt low ROI

If we make route the default now, when we later introduce the gateway it will require to make some form of this again because the expose is platform specific.
`OpenShift -> route`
`Vanilla -> gateway`

Then there is also the question of... "I need to undeploy old exposer! OK I need the... exposer" - this whole thing is to be refactored with the gateway (second exposer in mind) but I deferred for later in interest of time to make this PR

cmd/client.go
gauron99: I argued with AI on this quite a bit. Its because its a "special case" for keda, theres no ownerRef so you have to manually go in and take a look for your `Route` that is associated with your Function. `func delete` is the last line of defense for this. I intentionally made the deployer - unexpose mechanism try to target the `Route` as much as possible via Function Service labels etc -> so its targeted removal. the `func delete` is actually going to list the Routes and see which one could be associated with it. It does some discovery on top - because, precisely, this is a unfortunate situation of resource without `ownerRef`

cmd/client.go
gauron99: gotta love the AI wording!

cmd/errors.go
gauron99: Yeah youre right, what I meant to say is that for `raw` and `keda` deployers, default is cluster-local. meaning the --expose="" (empty) value is `none` meaning unexposed. I did this, again for simplicity, though the original design was expose by default because I think you would want to have a Function exposed without having to do any extra legwork. `func create - func deploy` and you have an externally exposed, running function accessible via URL that you can curl right from your terminal.

pkg/keda/deployer.go
gauron99: I think I have a way to optimize this better. Thanks for pointing this out, will update in next push

cmd/client.go
lkingland: OK, so maybe chalk this up to being a stepping-stone before a refactor after we have all of the pieces in place to make "exposing" the default (or more likely, keyed off the domain chosen for the Function name).

lkingland:
ALl my nitpicks have been addressed, and this looks good enough to me to merge.
Also, we're going to be mutating this all a little bit when the "Func Domains" epic is implemented, so that will give us another chance to give it all a once-over.
/hold just in case you're waiting for anyone else, but as far as I am concerned:
/lgtm
/approve
:shipit: 🎉

PR #3994, opened 2026-08-06
feat: pin func-utils image to a per-release-branch tag

gauron99:
proposal: pin func-utils img per release so that atleast per branch we have a more stable func-utils so we are not constantly breaking older functions because of the dynamic tag we currently have everywhere

### DIFF
--- a/.github/workflows/functions.yaml
+++ b/.github/workflows/functions.yaml
@@ -1,11 +1,6 @@
 name: Functions
 
 # To run this workflow's tests locally, see ./hack/test-full.sh
-#
-# Fork CI note (Hamr#395 external-dev / Functions#57): coverage upload goes through
-# ./.github/actions/codecov — no-ops when CODECOV_TOKEN is empty so forks without
-# the secret stay green after tests. Upstream knative/func still uploads when set.
-
 
 permissions:
   id-token: write # Required for signing
@@ -460,7 +455,9 @@ jobs:
       - test-e2e-podman
       - test-e2e-runtimes
       - test-e2e-config-ci
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    if: >-
+      (github.event_name == 'push' || github.event_name == 'workflow_dispatch') &&
+      (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release-'))
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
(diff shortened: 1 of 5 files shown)


gauron99:
@matejvasek @lkingland WDYT?

lkingland (approved):

matejvasek:
## Automated Code Review

> 🤖 _This review was generated by Claude Code (AI). Treat findings as suggestions to verify, not verdicts._

This PR pins the `func-utils` image to a per-release-branch tag. It's well-constructed — the findings below are mostly low-severity robustness/consistency issues.

### Findings

**1. Broad release-branch gate vs. strict tag regex** — `.github/workflows/functions.yaml`
The job-level `if` accepts any `refs/heads/release-*` ref, but the "Derive Image Tag" step only accepts `^release-([0-9]+\.[0-9]+)$` and `exit 1`s otherwise. A push to `release-1.24.0`, `release-v1.24`, or `release-next` passes the gate, runs the full build + e2e, then hard-fails at derive — turning the workflow red instead of cleanly skipping.

**2. Detached-HEAD fallback yields floating `v2`** — `Makefile`
Release-tag pinning falls back to `git branch --show-current`, which is empty in detached-HEAD checkouts (git worktree, SHA checkout, Prow's non-TAG_RELEASE path). This silently embeds `ghcr.io/knative/func-utils:v2` into a release-branch binary, defeating the pin. Official releases are safe only because `release.sh` passes an exact KVER.

**3. `workflow_dispatch` republishes utils but skips CLI image** — `.github/workflows/functions.yaml`
`publish-image` (ko func CLI image) gates on `push && main` only, while `build` and `publish-utils-image` also allow `workflow_dispatch`. A manual dispatch on main republishes the utils image but silently skips the CLI image.

**4. `FUNC_UTILS_IMG` repo-path override breaks mirror mapping** — `hack/images.sh`
`images.sh` reuses only the tag from `make func-utils-image` and hardcodes `registry.localtest.me/knative/func-utils`. A `FUNC_UTILS_IMG` override that changes the repo/registry path (the documented "always wins" override) makes the binary embed one path while images.sh pushes to another — in-cluster e2e pulls fail to resolve.

**5. Scope creep: deleted "Fork CI note"** — `.github/workflows/functions.yaml`
The PR removes the unrelated comment documenting why codecov no-ops on forks (Hamr#395 / Functions#57). Rationale is lost and unrelated to the func-utils pin.

**6. Redundant `if` duplicated from `needs`** — `.github/workflows/functions.yaml`
`publish-utils-image`'s new `if` duplicates the identical condition on `build`, which it already declares in `needs`. When build skips, the job skips via `needs` anyway. Two copies to keep in sync.

**7. Inaccurate KVER comment** — `Makefile`
The comment says Prow passes KVER as `v1.24.0`, but `hack/release.sh` passes `KVER="${TAG}"` (the `knative-vX.Y.Z` form). No functional impact — the regex tolerates both — but the comment could mislead.

---

Highest-value items: **#1** (workflow fails red instead of skipping) and **#2** (pin silently defeated in detached-HEAD builds).

PR #3995, opened 2026-08-07
feat(mcp): add 'describe' tool and update documentation

Ankitsinghsisodya:
# Changes

- 🎁 Add a `describe` MCP tool that wraps `func describe --output json` and returns structured JSON (name, namespace, URL, routes, image, readiness, deployer, labels, subscriptions, revision) instead of requiring agents to parse text output
- 🎁 Add a `describe` help resource and document the tool's path/name mutual-exclusivity in `pkg/mcp/instructions.md`
- 🎁 Add tests covering argument passing, JSON parsing, path/name mutual exclusivity, and malformed CLI output handling

/kind enhancement

Fixes #

**Release Note**

​```release-note
Added a `describe` MCP tool that returns structured JSON (URL, image, namespace, labels, subscriptions, readiness) for a deployed Function, so AI agents no longer need to parse `func describe`'s text output.
​```

**Docs**

​```docs

​```

### DIFF
--- a/pkg/mcp/instructions.md
+++ b/pkg/mcp/instructions.md
@@ -39,7 +39,7 @@ This is essential because:
 
 **Exceptions:**
 - The `list` tool operates on the cluster, not local files, so it does NOT use a path parameter (it uses namespace instead)
-- The `delete` tool requires exactly one of `path` or `name`; it does NOT support a no-argument CWD mode (the MCP server process has its own working directory unrelated to the Function being managed)
+- The `delete` and `describe` tools each require exactly one of `path` or `name`; they do NOT support a no-argument CWD mode (the MCP server process has its own working directory unrelated to the Function being managed)
 
 ## Deployment Behavior
 
@@ -57,6 +57,7 @@ This is essential because:
 - Before 'deploy' → Read `func://help/deploy`
 - Before 'build' → Read `func://help/build`
 - Before 'list' → Read `func://help/list`
+- Before 'describe' → Read `func://help/describe`
 - Before 'delete' → Read `func://help/delete`
 
 The help text provides authoritative parameter information and usage context.
@@ -130,6 +131,15 @@ A first-time deploy can be detected by checking the func.yaml for a value in the
 - Does NOT use path parameter (operates on cluster, not local files)
 - Optional `namespace` parameter to list Functions in specific namespace
 - Returns list of deployed Functions in current/specified namespace
+
+### describe
+
+- **FIRST:** Read `func://help/describe` for authoritative usage information
+- Supports TWO modes (mutually exclusive):
+  1. **Describe by PATH:** Provide 'path' parameter (reads function name from func.yaml at that path)
+  2. **Describe by NAME:** Provide 'name' parameter (describes named function from cluster)
+- Exactly ONE of 'path' or 'name' must be provided, not both
+- Read-only; does not modify local files or cluster resources
 
 ### delete
 
--- a/pkg/mcp/instructions.md
+++ b/pkg/mcp/instructions.md
@@ -135,10 +135,12 @@ A first-time deploy can be detected by checking the func.yaml for a value in the
 ### describe
 
 - **FIRST:** Read `func://help/describe` for authoritative usage information
+- Describes a **deployed** Function instance on the cluster; it never just reads local `func.yaml`. This tool will fail if the Function has not yet been deployed (e.g. calling it right after 'create' but before 'deploy' is a usage error, not a tool bug)
 - Supports TWO modes (mutually exclusive):
-  1. **Describe by PATH:** Provide 'path' parameter (reads function name from func.yaml at that path)
-  2. **Describe by NAME:** Provide 'name' parameter (describes named function from cluster)
+  1. **Describe by PATH:** Provide 'path' parameter (reads the Function's name/namespace from func.yaml at that path, then describes whatever is deployed for that Function)
+  2. **Describe by NAME:** Provide 'name' parameter (describes the named Function on the cluster)
 - Exactly ONE of 'path' or 'name' must be provided, not both
+- 'namespace' is only valid together with 'name'; providing both 'path' and 'namespace' is rejected, since path mode already determines the namespace from the Function's own deploy identity
 - Read-only; does not modify local files or cluster resources
 
 ### delete
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -1,6 +1,7 @@
 package mcp
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os/exec"
@@ -32,6 +33,14 @@ type Server struct {
 
 type executor interface {
 	Execute(ctx context.Context, subcommand string, args ...string) ([]byte, error)
+	// ExecuteSplit runs the command and returns stdout and stderr captured
+	// into separate buffers. Unlike Execute (which uses CombinedOutput and
+	// therefore offers no guarantee about the relative ordering of stdout
+	// and stderr bytes - they're copied by two independently-scheduled
+	// goroutines), ExecuteSplit gives each stream its own buffer, so callers
+	// that need to parse structured output (e.g. JSON) from stdout can do so
+	// without risk of stderr content (warnings, etc.) corrupting the parse.
+	ExecuteSplit(ctx context.Context, subcommand string, args ...string) (stdout, stderr []byte, err error)
 }
 
 type Option func(*Server)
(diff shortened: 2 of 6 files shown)


pkg/mcp/tools_describe.go
```diff
@@ -0,0 +1,102 @@
...
+		IdempotentHint: true, // Describing the same function multiple times returns consistent results at any point in time.
+	},
+}
+
+func (s *Server) describeHandler(ctx context.Context, r *mcp.CallToolRequest, input DescribeInput) (result *mcp.CallToolResult, output DescribeOutput, err error) {
+	// Validate: exactly one of Path or Name must be provided
+	if (input.Path != nil && input.Name != nil) || (input.Path == nil && input.Name == nil) {
+		err = fmt.Errorf("exactly one of 'path' or 'name' must be provided")
+		return
+	}
```
lkingland, line 29: Here's a bit of a tricky one...

`func describe` has two different ways to find a Function, and they use namespace differently:

1. By path
"look at this Function project on disk (`func.yaml`), and describe whatever is deployed for *that* Function." The name and namespace come from the project itself. The CLI does **not** want a separate `--namespace` in this mode.

2. By name
"describe the Function called X on the cluster." Here `--namespace` is meaningful (which cluster namespace to look in).

So if someone passes both `path` and `namespace`, the CLI rejects it: namespace without a name is treated as inconsistent (`must also specify a name when specifying namespace`). Path already implies "use the project's own deploy identity."

This handler currently only checks "exactly one of path or name." It should also reject `path` + `namespace` as being nonsensical.  Without this, agent might get confused because it can send a combination that _looks_ sensible in the schema but always fails at the CLI.

Suggested tweak:

Extend the validation so that when `path` is set, `namespace` must not be set (mirror the CLI). Optionally also document on the `namespace` field that it only applies with `name`. A small unit test enforcing that when `path`+`namespace`  are both provided, an error is generated, would help future proof it as well.

pkg/mcp/tools_describe.go
```diff
@@ -0,0 +1,102 @@
...
+	// the CLI can write warnings to stderr on an otherwise-successful call
+	// (e.g. permission warnings from the knative describer), and stdout and
+	// stderr copied via CombinedOutput have no guaranteed relative ordering.
+	// Parsing JSON only ever out of a clean, unmixed stdout avoids that
+	// entirely rather than relying on any heuristic about stream ordering.
+	stdout, stderr, err := s.executor.ExecuteSplit(ctx, "describe", input.Args()...)
+	if err != nil {
+		err = fmt.Errorf("%w\nstdout: %s\nstderr: %s", err, string(stdout), string(stderr))
+		return
+	}
```
lkingland, line 50: `ExecuteSplit` is the right call.  A clean stdout for JSON is exactly what we need!

One small nit:  on the success path, `stderr` is currently just discarded. The knative describer can warn on stderr while still returning partial JSON (e.g. RBAC: can't list triggers → empty subscriptions). The agent then sees "no subscriptions" with no hint that listing failed.

You can surface non-empty stderr using an optional `warnings` field on `DescribeOutput`, or a short note in the structured result itself, whichever feels better.

pkg/mcp/tools_describe.go
```diff
@@ -0,0 +1,102 @@
...
+	Name          string            `json:"name" jsonschema:"Function name"`
+	Namespace     string            `json:"namespace,omitempty" jsonschema:"Kubernetes namespace"`
+	URL           string            `json:"url,omitempty" jsonschema:"Primary route URL"`
+	Routes        []string          `json:"routes,omitempty" jsonschema:"All route URLs"`
+	Image         string            `json:"image,omitempty" jsonschema:"Deployed container image"`
+	Ready         string            `json:"ready,omitempty" jsonschema:"Overall readiness (true/false/unknown)"`
+	Deployer      string            `json:"deployer,omitempty" jsonschema:"Deployer backend (knative, k8s, keda)"`
+	Labels        map[string]string `json:"labels,omitempty" jsonschema:"Function labels"`
+	Subscriptions []fn.Subscription `json:"subscriptions,omitempty" jsonschema:"Active event subscriptions"`
+	Revision      string            `json:"revision,omitempty" jsonschema:"Source commit SHA, read from the OCI revision label baked into the built image"`
```
lkingland, line 124: `fn.Instance` also carries `Middleware`.  Let's add that one as well in the mapping.

pkg/mcp/tools_describe.go
```diff
@@ -0,0 +1,102 @@
...
+		args = append(args, "--path", *i.Path)
+	} else if i.Name != nil {
+		args = append(args, *i.Name)
+	}
+
+	args = appendStringFlag(args, "--namespace", i.Namespace)
+	args = appendBoolFlag(args, "--verbose", i.Verbose)
+
+	// The tool's contract is structured JSON, regardless of caller input.
+	args = append(args, "--output", "json")
```
lkingland, line 108: Agree with this decision to just go straight to --output json wit no commentary.  That's the right shape.

pkg/mcp/mcp.go
```diff
@@ -108,6 +117,7 @@ func New(options ...Option) *Server {
 	mcp.AddTool(i, buildTool, s.buildHandler)
 	mcp.AddTool(i, deployTool, s.deployHandler)
 	mcp.AddTool(i, listTool, s.listHandler)
+	mcp.AddTool(i, describeTool, s.describeHandler)
```
lkingland, line 120: By the way, I am glad to see this dichotomy of endpoint types beginning to emerge.

This PR is moving toward a better shape for MCP tools in general:  when the CLI has a machine-readable mode:

- "Passthrough", where `Args()` just `Execute`, then `Message: string` (build, deploy, delete, config_*, etc.)
- "Semantic", where we use a machine format, parse, remap into structured fields (`describe` is the first!)

In general, this is a good fist step to our final goal (later) of replacing this middle layer with a direct-to-api calls.

This does beg the question:  should we update the others soon?

`list` would be the highest priority; it already has `--output json` but still returns an opaque `Message` string. It should follow the same structured pattern as `describe`.

Any other command that already supports `--output json` / structured stdout should get the same treatment as we go along.

Looking good overall.

pkg/mcp/instructions.md
```diff
@@ -131,6 +132,15 @@ A first-time deploy can be detected by checking the func.yaml for a value in the
...
 - Returns list of deployed Functions in current/specified namespace
 
+### describe
+
+- **FIRST:** Read `func://help/describe` for authoritative usage information
+- Supports TWO modes (mutually exclusive):
+  1. **Describe by PATH:** Provide 'path' parameter (reads function name from func.yaml at that path)
+  2. **Describe by NAME:** Provide 'name' parameter (describes named function from cluster)
+- Exactly ONE of 'path' or 'name' must be provided, not both
+- Read-only; does not modify local files or cluster resources
```
lkingland, line 144: Possible agent confusion here...

Can we add a short agent-facing note here that `describe` is about a _deployed_ instance, not local source alone.

Path mode still talks to the cluster (it loads name/namespace from `func.yaml` and describes the remote Function). It will fail if the Function has never been deployed.  So it is not "print local func.yaml."
Mentioning this nuance explicitly will stops what I suspect would be a common agent mistake of calling `describe` right after `create` (before `deploy`) and treating the error as a tool bug.

pkg/mcp/mock/executor.go
```diff
@@ -27,3 +30,15 @@ func (m *Executor) Execute(ctx context.Context, subcommand string, args ...strin
 
 	return []byte(""), nil
 }
+
+// ExecuteSplit implements the executor interface, recording invocation
+// details and delegating to ExecuteSplitFn if provided.
+func (m *Executor) ExecuteSplit(ctx context.Context, subcommand string, args ...string) ([]byte, []byte, error) {
```
lkingland, line 36: Glad we have this.  Should have been the default from the beginning, but we can do that later.

pkg/mcp/tools_describe.go
```diff
@@ -0,0 +1,102 @@
...
+		return
+	}
+
+	var instance fn.Instance
+	if err = json.Unmarshal(stdout, &instance); err != nil {
+		err = fmt.Errorf("failed to parse describe output: %w\n%s", err, string(stdout))
+		return
+	}
+
+	output = DescribeOutput{
```
lkingland, line 63: Just a warning here that reuising the types from the core package might not always work, so in the future, you can create similar "copies" with MCP-specific fields and shape if needed.  That decoupling is worth the "copied" lines sometimes

lkingland (commented):
Looking good!  Comments below...

pkg/mcp/tools_describe.go
```diff
@@ -0,0 +1,126 @@
...
+	Name          string            `json:"name" jsonschema:"Function name"`
+	Namespace     string            `json:"namespace,omitempty" jsonschema:"Kubernetes namespace"`
+	URL           string            `json:"url,omitempty" jsonschema:"Primary route URL"`
+	Routes        []string          `json:"routes,omitempty" jsonschema:"All route URLs"`
+	Image         string            `json:"image,omitempty" jsonschema:"Deployed container image"`
+	Ready         string            `json:"ready,omitempty" jsonschema:"Overall readiness (true/false/unknown)"`
+	Deployer      string            `json:"deployer,omitempty" jsonschema:"Deployer backend (knative, k8s, keda)"`
+	Labels        map[string]string `json:"labels,omitempty" jsonschema:"Function labels"`
+	Subscriptions []fn.Subscription `json:"subscriptions,omitempty" jsonschema:"Active event subscriptions"`
+	Middleware    *fn.Middleware    `json:"middleware,omitempty" jsonschema:"Middleware backend (e.g. keda) applied at deploy time, if any"`
```
lkingland, line 123: Really shouldn't be a pointer.  Empty structs are serialized as equivalent to nil when omitempty.

lkingland (approved):
/lgtm

PR #3998, opened 2026-08-07
feat(mcp): add `run` and `run_stop` tools to the MCP server

Ankitsinghsisodya:
## Summary

Adds two new MCP tools that let clients start and stop a Function locally through the MCP server, mirroring `func run`:

- **`run`** — builds the Function if needed, starts it in the background, and returns once it's ready with its `pid` and `url`. Accepts optional `path`, `registry`, `build`, and `port` parameters. Only one run is allowed per Function path at a time.
- **`run_stop`** — gracefully stops a Function previously started with `run` (SIGTERM, then SIGKILL after a grace period), matched by the same absolute `path`.

Both tools are disabled when the server is running in read-only mode, consistent with `deploy`/`delete`.

## Implementation details

- `pkg/mcp/process.go`: new `processStarter` abstraction that spawns `func run --json` as a subprocess, parses its stdout for a JSON readiness line (`host`/`port`), and returns a `stop` func to terminate it. Also introduces a `runRegistry` to track active runs keyed by function path.
- `pkg/mcp/tools_run.go` / `pkg/mcp/tools_run_stop.go`: tool definitions, input/output schemas, and handlers for `run` and `run_stop`.
- `pkg/mcp/mock/process_starter.go`: mock `processStarter` for tests, injected via new `WithProcessStarter` server option.
- `pkg/mcp/mcp.go`: wires up the new tools and default process starter/run registry on the server.
- `pkg/mcp/instructions.md` / `pkg/mcp/instructions_warning.md`: document the new tools' usage/path requirements and note them as disabled in read-only mode.

## Test plan

- [x] `make test` — new unit tests in `pkg/mcp/process_test.go`, `pkg/mcp/tools_run_test.go`, and `pkg/mcp/tools_run_stop_test.go` cover readonly rejection, duplicate-run rejection, start errors, and stop success/failure paths using the mock process starter.
- [ ] `make check`


pkg/mcp/tools_run.go
```diff
@@ -0,0 +1,91 @@
...
+	output = RunOutput{
+		Pid: pid,
+		URL: fmt.Sprintf("http://%s:%s", host, port),
+	}
+	return
+}
+
+// RunInput defines the input parameters for the run tool.
+type RunInput struct {
+	Path     *string `json:"path,omitempty" jsonschema:"Absolute path to the function project directory (default: server's current working directory)"`
```
lkingland: Same note as on the other MCP tools: let's make absolute `path` required, dropping the "defaults to the MCP server's CWD" behavior since that's really an interactive-cli optimization.

Agents live in a different working directory than the MCP server process; omitting path almost always results in an unexpected path

pkg/mcp/process.go
```diff
@@ -0,0 +1,269 @@
...
+	//
+	// On success it returns the process ID, host, and port, along with a
+	// stop function which gracefully terminates the process (SIGTERM, then
+	// SIGKILL after a grace period) and blocks until it has fully exited.
+	// stop is idempotent and safe to call even if the process has already
+	// exited on its own.
+	//
+	// On error, any process that was started is terminated before
+	// returning; no process is left running.
+	Start(ctx context.Context, subcommand string, args ...string) (pid int, host, port string, stop func() error, err error)
```
lkingland, line 44: We need a teardown hook when the MCP server itself is going away.

If the server process exits hard (client disconnect, panic, SIGTERM on the server, etc.), any `func run` children still registered here will be left behind — ports held, maybe containers, etc.

Please add a hard-stop path that walks the registry and attempts to stop every active subprocess (same signal you'd send on a normal `run_stop`). On a true hard-kill of the *server* (e.g. SIGKILL), we can't run user code — that's fine; rely on the OS reaping / zombie handling in that case. What we *can* handle is the normal server shutdown / signal path: capture it and fan the same polite stop out to everything still registered.

pkg/mcp/process.go
```diff
@@ -0,0 +1,269 @@
...
+	// Send SIGTERM (not the default SIGKILL) on cancellation so the child's
+	// own signal handler can run its cleanup (job.Stop()). WaitDelay forces
+	// a SIGKILL if it has not exited within the grace period.
+	cmd.Cancel = func() error {
+		if cmd.Process == nil {
+			return nil
+		}
+		return cmd.Process.Signal(syscall.SIGTERM)
+	}
+	cmd.WaitDelay = runStopGrace
```
lkingland: I don't think we should ever escalate to SIGKILL here.

`WaitDelay` + force-kill short-circuits `func run`'s own cleanup (`job.Stop()`, container teardown, etc.). Prefer SIGTERM only and wait — if something is truly wedged, that's a separate bug to fix in the runner, not something MCP should paper over with kill -9.

So: drop the SIGKILL / `WaitDelay` escalation; stop = signal SIGTERM and wait for exit.

pkg/mcp/tools_run_stop.go
```diff
@@ -0,0 +1,60 @@
...
+
+var runStopTool = &mcp.Tool{
+	Name:        "run_stop",
+	Title:       "Stop Local Function Run",
+	Description: "Stop a Function previously started with the run tool.",
+	Annotations: &mcp.ToolAnnotations{
+		Title:           "Stop Local Function Run",
+		ReadOnlyHint:    false,
+		DestructiveHint: ptr(true),
+		IdempotentHint:  true, // stopping an already-stopped run at the same path fails clearly either way
```
lkingland: Nit: `IdempotentHint: true` here I think is right, but we'd want the call to exit 0 (perhaps with a nice message to STDERR stating "already stopped"), not an error.

pkg/mcp/tools_run.go
```diff
@@ -0,0 +1,91 @@
...
+		DestructiveHint: ptr(false),
+		IdempotentHint:  false, // calling run again for a path already running is rejected, not idempotent
+	},
+}
+
+func (s *Server) runHandler(ctx context.Context, r *mcp.CallToolRequest, input RunInput) (result *mcp.CallToolResult, output RunOutput, err error) {
+	if s.readonly.Load() {
+		err = fmt.Errorf("the server is currently in read-only mode; to enable write operations, set FUNC_ENABLE_MCP_WRITE in the server environment and restart the server")
+		return
+	}
```
lkingland: No need to gate this on FUNC_ENABLE_MCP_WRITE since it's local only.  No need for a check.

pkg/mcp/instructions_warning.md
```diff
@@ -15,6 +15,7 @@ The Functions MCP server is currently running in **read-only mode**.
 **Disabled operations:**
 - Deploy to cluster
 - Delete from cluster
+- Run/stop Functions locally
```
lkingland: Since these are non-cluster (local only), there's no need to have them as disabled.

pkg/mcp/tools_run.go
```diff
@@ -0,0 +1,91 @@
...
+
+	pid, host, port, stop, err := s.starter.Start(readyCtx, "run", input.Args(path)...)
+	if err != nil {
+		err = fmt.Errorf("unable to run function: %w", err)
+		return
+	}
+
+	if err = s.runs.add(path, pid, stop); err != nil {
+		_ = stop()
+		return
```
lkingland: Just double-checking I understand the race story correctly, and logging it so it's explicit in the PR:

Two concurrent `run` calls for the same path can both pass the fail-fast `get` (lines above), both successfully `Start` a process, then only one wins `runs.add`. The loser correctly `stop()`s its orphan?

That means for a brief window there can be two `func run` processes for the same Function (two ports, two builds, etc) until the loser is torn down.

Therefore, is `add` the arbiter of serizlization? Would you rather hold a per-path lock from before `Start` through `add` so the second caller waits or fails without spawning?

I'm fine with the current "start then register, stop loser" approach if it's intentional; I just want a line of comment (or a tiny lock) so the next reader doesn't have to reverse-engineer it!

pkg/mcp/tools_run_stop.go
```diff
@@ -0,0 +1,60 @@
...
+
+	output = RunStopOutput{
+		Message: fmt.Sprintf("stopped function at %q (pid %d)", path, entry.pid),
+	}
+	return
+}
+
+// RunStopInput defines the input parameters for the run_stop tool.
+type RunStopInput struct {
+	Path *string `json:"path,omitempty" jsonschema:"Absolute path to the function project directory (default: server's current working directory); must match the path used with the run tool"`
```
lkingland: Same as `run`: let's require absolute `path` here too.

lkingland (commented):
Love that this one is starting to show why an MCP server helps do things a simple skill cannot do as elegantly.

Small suggestions around error and exit code handling below:

pkg/mcp/tools_run.go
Ankitsinghsisodya: yeah your read of the race was right.

I went with the tiny lock instead of just a comment. Added `reserve`/`activate`/`release` on the registry: `reserve(path)` claims the path atomically before `Start` is even called, so a second concurrent `run` for the same path now fails right there and never spawns a process at all — no more "start two, stop the loser." `activate` fills in the pid/stop once the process is actually up, and `release` frees the reservation if `Start` fails.

lkingland:
Confirmed the follow-up against the earlier review — this landed cleanly:

- required absolute `path` on `run` and `run_stop` (no CWD default)
- `stopAll` on normal server shutdown (`Start` after `Run` returns)
- SIGTERM only (no `WaitDelay` / SIGKILL)
- not gated on `FUNC_ENABLE_MCP_WRITE`; RO instructions list local run as available
- `run_stop` is actually idempotent now
- race: `reserve`/`activate`/`release` so the second caller never spawns — nicer than stop-the-loser

Non-blocking: instructions tell agents to read `func://help/run`, but that help resource isn't registered yet (unlike create/build/deploy). Easy add if you want it in this PR or a follow-up.

/lgtm
/approve

lkingland:
Just needs a rebase, then can re-approve

Ankitsinghsisodya:
> Just needs a rebase, then can re-approve

I have rebased it

Ankitsinghsisodya:
> Just needs a rebase, then can re-approve

Can you approve it

PR #4003, opened 2026-08-11
refactor(listers): inject k8s.Client via DI

### LINKED ISSUE #4000: Better DI for listers
As of now Lister interface implementation create k8s client internally. The should not create it on their own. The client shall be injected via constructor function.

pmeida:
# Changes

- Introduce `k8s.Client` wrapper with lazy config resolution, supporting both CLI path (`clientcmd.ClientConfig`) and SDK path (direct `*rest.Config`)
- Inject `*k8s.Client` into k8s, knative, and keda listers via constructors instead of creating clientsets internally
- Make `NewClient` infallible - errors are deferred to when the config is actually needed
- Add unit tests for `k8s.Client`
- The `knative` and `keda` listers inline the serving/HTTP client creation from the injected `*k8s.Client` config instead of calling `NewServingClient`/`NewHTTPScaledObjectClientset`, since modifying those functions to accept `*k8s.Client` would cascade to deployer, describer, and remover callers which are outside the scope of this PR.

/kind enhancement

Fixes #4000

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -60,6 +60,7 @@ func NewTestClient(options ...fn.Option) ClientFactory {
 // 'Verbose' indicates the system should write out a higher amount of logging.
 func NewClient(cfg ClientConfig, options ...fn.Option) (*fn.Client, func()) {
 	var (
+		kc = k8s.NewClient(k8s.GetClientConfig())
 		t  = newTransport(cfg.InsecureSkipVerify)                                // may provide a custom impl which proxies
 		c  = newCredentialsProvider(config.Dir(), t, "", cfg.InsecureSkipVerify) // for accessing registries
 		d  = newKnativeDeployer(cfg.Verbose)                                     // default deployer (can be overridden via options)
@@ -76,7 +77,7 @@ func NewClient(cfg ClientConfig, options ...fn.Option) (*fn.Client, func()) {
 				k8s.NewDescriber(cfg.Verbose, k8s.WithDescriberTransport(t)),
 				keda.NewDescriber(cfg.Verbose, keda.WithDescriberTransport(t)),
 			),
-			fn.WithListers(knative.NewLister(cfg.Verbose), k8s.NewLister(cfg.Verbose), keda.NewLister(cfg.Verbose)),
+			fn.WithListers(knative.NewLister(kc, cfg.Verbose), k8s.NewLister(kc, cfg.Verbose), keda.NewLister(kc, cfg.Verbose)),
 			fn.WithDeployer(d),
 			fn.WithPipelinesProvider(pp),
 			fn.WithPusher(docker.NewPusher(
--- a/cmd/completion_util.go
+++ b/cmd/completion_util.go
@@ -16,9 +16,10 @@ import (
 )
 
 func CompleteFunctionList(cmd *cobra.Command, args []string, toComplete string) (strings []string, directive cobra.ShellCompDirective) {
+	kc := k8s.NewClient(k8s.GetClientConfig())
 	listers := []fn.Lister{
-		knative.NewLister(false),
-		k8s.NewLister(false),
+		knative.NewLister(kc, false),
+		k8s.NewLister(kc, false),
 	}
 
 	items := []fn.ListItem{}
--- a/e2e/e2e_core_test.go
+++ b/e2e/e2e_core_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 
 	fn "knative.dev/func/pkg/functions"
+	"knative.dev/func/pkg/k8s"
 	"knative.dev/func/pkg/knative"
 )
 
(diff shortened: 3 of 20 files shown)


matejvasek:
> *🤖 This review was generated by Claude Code.*

## Review: refactor(listers): inject k8s.Client via DI

**+191 / -27** across 20 files | **Fixes:** #4000

### Summary

Introduces a `k8s.Client` wrapper that centralizes Kubernetes config resolution and injects it into the k8s, knative, and keda listers via constructors instead of having each lister create its own clientset internally. The wrapper supports two construction paths: `NewClient(cc)` for CLI (lazy `clientcmd.ClientConfig` resolution) and `NewClientFromConfig(cfg)` for SDK/testing (direct `*rest.Config`).

### What's good

- **Clean DI pattern.** The `k8s.Client` wrapper is minimal and fits the stated goal. Making construction infallible and deferring errors to first use is the right call — it avoids forcing callers to handle errors at wiring time.
- **Config caching.** `ClientConfig()` resolves the `rest.Config` once and caches it, avoiding redundant kubeconfig parsing across listers sharing the same `*Client`.
- **Well-scoped.** The PR deliberately avoids cascading changes to deployer/describer/remover by inlining the serving/HTTP client creation in the knative and keda listers. This is a reasonable tactical choice and is clearly documented in the PR description.
- **Good test coverage.** Unit tests cover all four construction/error paths for the new `Client` type.
- **Nice drive-by fix.** Corrected the error message in `NewEventingClient` from "serving" to "eventing" (`pkg/knative/client.go:44`).

### Issues

**1. `Clientset()` creates a new `*kubernetes.Clientset` on every call** (`pkg/k8s/client.go:47-53`)

Unlike `ClientConfig()` which caches the result, `Clientset()` allocates a new clientset each time. This matches the pre-existing behavior (the old `NewKubernetesClientset()` did the same), so it's not a regression. But since the whole point of the wrapper is to share state, it would be natural to cache the clientset too. Worth considering as a follow-up.

**2. No thread-safety on `ClientConfig()` lazy initialization** (`pkg/k8s/client.go:32-45`)

There's a check-then-write race on `c.cfg`: two concurrent goroutines could both see `c.cfg == nil`, both call `c.cc.ClientConfig()`, and both write to `c.cfg`. The result is probably benign (both would get a valid config), but it's technically a data race. If this will only ever be called from a single goroutine, it's fine. If concurrent use is expected, a `sync.Once` would be a cleaner pattern:

```go
type Client struct {
    cc      clientcmd.ClientConfig
    cfg     *rest.Config
    cfgOnce sync.Once
    cfgErr  error
}
```

**3. `NewClientFromConfig(nil)` produces a confusing error** (`pkg/k8s/client.go:28-30`)

Passing `nil` to `NewClientFromConfig` would succeed at construction time, then fail at `ClientConfig()` with `"no kubernetes client configuration available"` — which is the error path for a missing `clientcmd.ClientConfig`, not a nil `rest.Config`. This is unlikely to happen in practice but could confuse someone debugging. A nil guard or a comment documenting the contract would help.

**4. `validateKubeconfigFile()` inconsistency across listers**

The knative lister calls `validateKubeconfigFile()` before using the config (`pkg/knative/lister.go:29`), but neither the k8s nor keda listers do. This is pre-existing behavior, not introduced by this PR, but since you're touching all three listers it might be worth either pulling this validation into `Client.ClientConfig()` or dropping it from the knative lister (it's arguably redundant given that `ClientConfig()` will fail if the kubeconfig is invalid).

### Minor nits

- `pkg/knative/lister.go:35` and `:40` both say `"failed to create serving client"` — the first is really a config resolution failure. A more specific message like `"failed to get kubernetes client config"` would make debugging easier (the keda lister gets this right at `pkg/keda/lister.go:35`).

### Verdict

Clean, well-scoped refactor. The core DI pattern is correct. The issues above are all minor and could be addressed as follow-ups. The PR achieves its goal of making listers testable via dependency injection without cascading changes to unrelated components.

pmeida:
Fixed:

- Nil `kc` guard in all three listers - prevents nil-pointer panic for SDK callers
- Error message in knative lister - "failed to create serving client" -> "failed to get kubernetes client config" (keda lister already had this right)
- `validateKubeconfigFile()` removed from knative lister - avoids false rejection when SDK path provides a valid injected config but `KUBECONFIG` env is bad.

Skipped:

- `Clientset() `caching - Pre-existing behavior, separate optimization
- `sync.Once` thread safety - CLI is single-threaded, not needed
- `NewClientFromConfig(nil)` confusing error - Unlikely misuse, error message is accurate enough
- `validateKubeconfigFile()` inconsistency across listers - Fixed by removing it from knative (item 3 above), and k8s/keda listers never had it

matejvasek:
> * `sync.Once` thread safety - CLI is single-threaded, not needed

I would not bet on this.

Also even if it were 100% true the `pkg` should be also usable as a library (actually whole point of this PR is that we use this as a library, right). We should be robust, we already use `sync.Once` as several places for safe lazy initialization.

pmeida:
Fair, it is low effort.

gauron99 (approved):
Did a quick review in GH only - this seems good enough for now if it works for your #4000 case, we can adjust as necessary when the whole injection fix (for all cmds) will be made in near future.

/lgtm
/approve

PR #4029, opened 2026-09-01
fix: source func CLI version from build ldflags

### LINKED ISSUE #4028: config ci: source func CLI version from build ldflags
## Problem

The `func config ci` command generates a GitHub Actions workflow that installs the func CLI using [functions-dev/action](https://github.com/functions-dev/action). The version passed to that action was [hardcoded](https://github.com/knative/func/blob/main/pkg/ci/github/workflow.go#L17) to `knative-v1.22.0`, meaning:

```yaml
- name: Install func cli
  uses: functions-dev/action@main
  with:
    name: func
    version: knative-v1.22.0
```

**Stale version**: The version in generated workflows would never update as new releases shipped - ⚠️ currently wrong on [latest release](https://github.com/knative/func/blob/release-1.23/pkg/ci/github/workflow.go#L17).

## Expected behavior

The generated workflow should install the func CLI version that matches the build that generated it. For dev builds between releases, it should fall back to the nearest real release tag.

/kind bug

pmeida:
# Changes

- Source `FuncCliVersion` from `version.Kver` (set via ldflags at build time) instead of a hardcoded string
- Strip the `git describe` suffix (e.g. `-22-g2871d3bf`) so dev builds between releases emit the nearest real release tag
- Expose `FuncCliVersion` as a field on `WorkflowConfig` so library consumers can override it directly
  - this is important because library consumers with `no ldflags` have empty `version.Kver`, so it falls back to `DefaultFuncCliVersion` which may be out of date as it has to be manually updated 

/kind bug

Fixes #4028

**Release Note**

```release-note
`func config ci` now sources the func CLI version used in generated GitHub Actions workflows from the binary's build-time version (set via ldflags), instead of a hardcoded string. Dev builds between releases emit the nearest real release tag by stripping the git describe suffix.
```

**Docs**

```docs

```

### DIFF
--- a/cmd/config_ci.go
+++ b/cmd/config_ci.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"io"
+	"regexp"
 	"strings"
 
 	"github.com/ory/viper"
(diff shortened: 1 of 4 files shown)


pmeida:
@matejvasek It should be backported to release 4.23 right?

matejvasek:
Yeah, we will charrypick it to release 1.23.

PR #4033, opened 2026-09-01
feat: configure CI workflow builder through WorkflowConfig and func.yaml

### LINKED ISSUE #4031: config ci: builder configured in func.yaml is ignored in generated workflow
## Summary

`func config ci` generates a GitHub Actions workflow with `FUNC_BUILDER` always derived from the runtime, silently ignoring any `build.builder` value configured in `func.yaml`.

## Steps to reproduce

1. Create a function: `func create -l node .`
2. Set `build.builder: s2i` in `func.yaml`
3. Run `FUNC_ENABLE_CI_CONFIG=true func config ci`
4. Inspect `.github/workflows/func-deploy.yaml`

## Expected behavior

`FUNC_BUILDER: s2i` - matching the value in `func.yaml`.

## Actual behavior

`FUNC_BUILDER: pack` - the runtime default, ignoring `func.yaml`.

## Root cause

`determineBuilder` only considered the runtime and remote flag. `f.Build.Builder` was never read, so library consumers had no way to influence the generated `FUNC_BUILDER` value without patching the YAML after the fact.

pmeida:
# Changes

- Add `Builder` field to `WorkflowConfig` so the builder flows through the established config pattern alongside all other workflow settings
- `Generate()` resolves the builder with priority: `WorkflowConfig.Builder` > `f.Build.Builder` (func.yaml) > runtime default; `g.cfg` is never mutated
- Validate explicit builder values against the known set; unknown values produce a clear error instead of a broken workflow
- Validate that builder `host` is incompatible with remote builds

/kind feature
/kind bug

Fixes #4031

**Release Note**

```release-note
`func config ci` now respects the builder configured in `func.yaml`. Previously, `FUNC_BUILDER` in the generated GitHub Actions workflow was always derived from the runtime, silently overriding any `build.builder` value set by the user. The builder can also be set programmatically via `WorkflowConfig.Builder`, which takes priority over `func.yaml`.
```

**Docs**

```docs

```

### DIFF
--- a/pkg/ci/github/generator_test.go
+++ b/pkg/ci/github/generator_test.go
@@ -9,6 +9,7 @@ import (
 
 	"gotest.tools/v3/assert"
 	"gotest.tools/v3/assert/cmp"
+	"knative.dev/func/pkg/builders"
 	"knative.dev/func/pkg/ci/github"
 	fn "knative.dev/func/pkg/functions"
 )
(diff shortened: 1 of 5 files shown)


twoGiants (changes requested):
@pmeida thanks for taking a look into this.

IMHO this is a feature and a bug, not just a bug. 

The builder should be set via the `WorkflowConfig`. Lets connect.

pkg/ci/github/generator_test.go
twoGiants, line 1: Can you update `assertDefaultWorkflow` to also assert for the builder env var?

twoGiants (approved):
/lgtm
/approve
/hold for the one small change (you can also do it in your next PR)

pkg/ci/github/generator_test.go
pmeida: I'll add it in the next round.

PR #4037, opened 2026-09-02
fix: normalize file modes in python image layer

matejvasek:
## Problem

Functions built with the OCI/host builder can fail to start on platforms that
run containers with an **arbitrary UID** (e.g. OpenShift's `restricted` SCC, and
Knative services running under it). For Python the failure is:

```
python: can't open file '/func/.func/build/service/main.py': [Errno 13] Permission denied
```

The same image runs fine under local `podman`, which runs it as the image's
configured `UID 1000`.

## Root cause

The OCI builder tars files into image layers using the **on-disk file modes
verbatim** (via `tar.FileInfoHeader`), overriding only `Uid`/`Gid`, never
`Mode`. Three layer-writers did this:

- `newPythonLibTarball` (`python_builder.go`) — the `.func/build` lib layer
- `newDataTarball` (`builder.go`) — the shared layer with the function source,
  static files and the `/func` directory itself (used by **both** runtimes)
- `newCertsTarball` (`builder.go`) — the CA certs layer

The build directory is created with `0774` which, under the default `umask 022`,
becomes `0754` — stripping the traverse (`x`) bit for group/other:

```
drwxr-xr-- 1000:1000  /func/.func/build      # 0754, no x for "other"
```

Baked into the image, this is only traversable/readable by `UID 1000`. Under
`podman` the process *is* `UID 1000` (the owner) → works. Under an arbitrary UID
(group `0`) it falls into the "other" class → `EACCES`.

For **Go** the same bug is latent: the exec layer already forced `0755`
(`go_builder.go`), and the binary is self-contained, so it works *as long as the
project directory happens to be `0755`*. With a stricter project-dir mode
(`0750`/`0700`) `/func` is baked in non-traversable and the container can't reach
`/func/f` under an arbitrary UID — failing before `main` runs.

## Fix

Two commits:

1. **python image layer** — normalize modes in `newPythonLibTarball`.
2. **shared data & certs layers** — normalize modes in `newDataTarball` and
   `newCertsTarball`, fixing the Go builder's latent case and hardening the
   Python source files shipped in the shared layer.

In all cases: directories and executables → `0755`, regular files → `0644`
(certs → `0644`), leaving symlink modes untouched. This mirrors the existing
`0755` normalization the Go exec layer already had (`go_builder.go`).

## Tests

- `TestNewPythonLibTarball_NormalizesModes` — builds the lib tarball from a tree
  with restrictive modes (`0754` dirs, `0600` file) and asserts normalization.
- `TestNewDataTarball_NormalizesModes` — builds the data tarball from a
  non-traversable (`0750`) project root and asserts `/func` and subdirs are
  `0755`, files `0644`, executables `0755`.
- Existing `TestBuilder_Files` continues to pass (symlink handling unchanged).

```release-note
fix: python host builder permission issues on OpenShift
```

### DIFF
--- a/pkg/oci/builder.go
+++ b/pkg/oci/builder.go
@@ -381,6 +381,21 @@ func newDataTarball(root, target string, ignored []string, verbose bool) error {
 		header.Name = slashpath.Join("/func", filepath.ToSlash(relPath))
 		header.Uid = DefaultUid
 		header.Gid = DefaultGid
+		// Normalize permissions so the image works on platforms that run
+		// containers with an arbitrary UID (e.g. OpenShift's restricted SCC).
+		// The on-disk mode is not portable: e.g. a project directory created
+		// under a stricter umask can be non-traversable by group/other, which
+		// would make /func (and thus /func/f) inaccessible to any UID other
+		// than the image's configured one. Directories and executables get
+		// 0755, regular files 0644. Symlink modes are not meaningful and are
+		// left untouched.
+		if info.Mode()&fs.ModeSymlink == 0 {
+			if info.IsDir() || info.Mode()&0o111 != 0 {
+				header.Mode = (header.Mode & ^int64(fs.ModePerm)) | 0o755
+			} else {
+				header.Mode = (header.Mode & ^int64(fs.ModePerm)) | 0o644
+			}
+		}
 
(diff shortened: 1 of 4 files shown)


gauron99 (commented):
/lgtm

Issue #4044, opened 2026-09-03
Expose KEDA ScaledObject and HTTP scaling configuration options


aliok:
## Problem

Several KEDA scaling configuration values are currently hardcoded and not configurable via func.yaml.

### ScaledObject (Kafka trigger)

Hardcoded in `pkg/keda/kafka_scaling.go`:
- `cooldownPeriod`: 300 (seconds) — how long before scaling down after triggers go idle
- `pollingInterval`: not set, defaults to 30 seconds — how often KEDA checks metrics

Not exposed from KEDA's Kafka scaler:
- `offsetResetPolicy`
- `allowIdleConsumers`
- `excludePersistentLag`
- `scaleToZeroOnInvalidOffset`
- `limitToPartitionsWithLag`

### HTTPScaledObject (HTTP trigger)

Hardcoded in `pkg/keda/deployer.go`:
- `CooldownPeriod`: 300
- `ScalingMetric.Rate.TargetValue`: 100 (requests/sec)
- `ScalingMetric.Rate.Window`: 1 minute
- `ScalingMetric.Rate.Granularity`: 1 second

## Proposed solution

Add optional fields to `scale.keda` in func.yaml for the common options:

```yaml
deploy:
  options:
    scale:
      keda:
        pollingInterval: 15
        cooldownPeriod: 60
        triggers:
          - type: kafka
            lagThreshold: 5
          - type: http
            targetValue: 50
```

Start with `pollingInterval`, `cooldownPeriod`, and HTTP `targetValue` as the most impactful. The Kafka-specific options (`allowIdleConsumers`, etc.) can be added later as needed.

Issue #4043, opened 2026-09-03
Migrate KEDA HTTP scaling from HTTPScaledObject to ScaledObject with external trigger


aliok:
## Problem

Currently, the KEDA deployer uses `HTTPScaledObject` for HTTP-based autoscaling. This creates an internal `ScaledObject` under the hood, which means we cannot create a second `ScaledObject` for the same Deployment (e.g. for Kafka consumer-lag scaling). KEDA's admission webhook rejects it:

```
admission webhook "vscaledobject.kb.io" denied the request: the workload 'my-func'
of type 'apps/v1.Deployment' is already managed by the ScaledObject 'my-func'
```

This makes it impossible to combine HTTP + Kafka triggers on the same function.

## Proposed solution

Replace `HTTPScaledObject` with a single `ScaledObject` that uses the KEDA HTTP add-on's external scaler as one of its triggers, alongside any other triggers (Kafka, cron, etc.).

The HTTP add-on exposes an external scaler gRPC endpoint that a regular `ScaledObject` can reference via an `external` trigger. The interceptor routing can be handled via `InterceptorRoute` (or the HTTP add-on's route CRD) instead of relying on `HTTPScaledObject` to set it up.

This would:
1. Allow combining HTTP + Kafka (or any other) triggers in a single `ScaledObject`
2. Give us direct control over all trigger configuration in one place
3. Remove the hidden internal `ScaledObject` that `HTTPScaledObject` creates

## Current workaround

Users must choose either HTTP-only or Kafka-only triggers — they cannot be combined.

## References
- KEDA HTTP add-on external scaler: the add-on runs a gRPC external scaler service
- KEDA limitation: one `ScaledObject` per workload

gauron99:
Notes from reviewing and testing #4051:
- `HTTPScaledObject` is deprecated since http add-on v0.14.0 in favor of `InterceptorRoute` (v1beta1) plus a user-owned ScaledObject with an `external-push` trigger (`scalerAddress: keda-add-ons-http-external-scaler.keda:9090`, `interceptorRoute: <name>`). See the [migration guide](https://keda.sh/http-add-on/0.16/operations/migrate-httpscaledobject-to-interceptorroute/). We're on v0.12.0 (go.mod and the test cluster); the latest is v0.16.0. So this needs a minimum add-on bump, since older installs lack the CRD.
[...]

PR #4045, opened 2026-09-04
[WIP] Scaling with KEDA Kafka scaler

### LINKED ISSUE #3934: Kafka functions: KEDA scaling with consumer-lag trigger
`func deploy` already supports `--deployer keda` for KEDA-based scaling. For Kafka functions, add a Kafka consumer-lag trigger so KEDA can scale based on consumer group lag — enabling scale-to-zero when there are no messages and scaling up under load. Currently Kafka functions require `min-scale: 1` to stay running.

Traffic-based autoscaling (e.g. Knative's KPA or KEDA HTTP scaler) does not work for Kafka functions because there is no inbound HTTP traffic — the function pulls messages from Kafka internally. Scaling must be driven by Kafka consumer group lag instead.

aliok:
## Summary

Adds KEDA Kafka consumer-lag scaling for the `keda` deployer, closes #3934.

- New `scale.keda.triggers` sub-key on `func.yaml`, supporting `http`, `kafka`, and `cron` trigger types. `deployer: keda` now requires at least one trigger to be declared explicitly (existing KEDA functions are migrated to an explicit `http` trigger to preserve current behavior).
- New `scale.kpa` sub-key for Knative-specific autoscaling config (`metric`/`target`/`utilization`), kept alongside the existing flat fields for backwards compatibility with older CLI versions.
- The `keda` deployer creates a `ScaledObject` + `TriggerAuthentication` (via the dynamic/unstructured client — no new dependency on `kedacore/keda`) when a `kafka` trigger is present, resolving SASL/TLS credentials from `run.kafka` and the function's secret-backed volumes.
- Fixed the raw deployer's readiness wait, which previously errored (`could not find current pod-template-hash`) on a deployment with 0 desired replicas — needed for `scale.min: 0` to work with Kafka-triggered scale-to-zero.

## func.yaml: old vs new structure

Before (implicit HTTP trigger, flat KPA fields only):

```yaml
deployer: keda
options:
  scale:
    min: 1
    max: 10
    metric: concurrency
    target: 200
    utilization: 80
```

After — `keda` deployer now requires explicit triggers. HTTP-only (equivalent to the old implicit behavior):

```yaml
deployer: keda
options:
  scale:
    min: 1
    max: 10
    keda:
      triggers:
        - type: http
```

After — Kafka consumer-lag trigger:

```yaml
deployer: keda
options:
  scale:
    min: 0
    max: 10
    keda:
      triggers:
        - type: kafka
          lagThreshold: 5
          activationLagThreshold: 0
run:
  kafka:
    brokers: "my-cluster-kafka-bootstrap.kafka.svc.cluster.local:9093"
    topic: "my-topic"
    consumerGroup: "my-function-group"
    securityProtocol: "SASL_SSL"
    tls:
      caCert: "/etc/kafka/ca/ca.crt"
    sasl:
      mechanism: "SCRAM-SHA-512"
      user: "my-kafka-user"
      password: "{{ secret:my-kafka-user:password }}"
  volumes:
    - secret: my-cluster-cluster-ca-cert
      path: /etc/kafka/ca
```

After — `knative` deployer, explicit KPA sub-key (old flat `metric`/`target`/`utilization` fields still work and are migrated into `kpa` automatically on deploy):

```yaml
deployer: knative
options:
  scale:
    min: 1
    max: 10
    kpa:
      metric: concurrency
      target: 200
      utilization: 80
```

A `specVersion` migration handles both transitions automatically: an old `keda` function with no triggers gets `keda.triggers: [{type: http}]` added, and an old `knative` function's flat KPA fields get copied into `scale.kpa` (the flat fields are left in place too, so older CLI versions can keep reading them).

## Known limitations (tracked separately)

- A Deployment can only be owned by one `ScaledObject`, so `http` and `kafka` triggers cannot currently be combined on the same function — the HTTP trigger's `HTTPScaledObject` creates its own internal `ScaledObject`. Tracked in #4043 (migrate the HTTP path to a `ScaledObject` with an external trigger, so triggers of any kind can be combined).
- Several KEDA/HPA tuning knobs (`pollingInterval`, `cooldownPeriod`, HTTP `targetValue`, etc.) are currently hardcoded rather than configurable via `func.yaml`. Tracked in #4044.

## Testing

- New unit tests: trigger resolution, secret-ref parsing, volume-path lookup, `TriggerAuthentication`/`ScaledObject` construction, SASL mechanism mapping, migration behavior, and cross-field validation.
- New integration test (`TestInt_KafkaScaling`, `-tags integration`): deploys a function with a Kafka-only trigger against a real cluster and verifies the `ScaledObject`/`TriggerAuthentication` are created with the expected spec and cleaned up on remove.
- Manually verified all `{raw, knative, keda} x {no kafka, with kafka}` deployer combinations end-to-end on Kind, including watching KEDA actually scale a Kafka consumer from 0 to N replicas under load, and the specVersion migration path from an old flat-scale `func.yaml`. Walkthroughs and captured resources for each scenario are under `docs/testing-deployments/`.

## Docs

- `docs/reference/func_yaml.md`: documented `deployer`, `run.kafka`, `scale.keda`, `scale.kpa`.
- `docs/reference/func_deploy.md` / `--deployer` flag help: updated to mention KEDA's Kafka/cron triggers, not just HTTP.

### DIFF
--- a/cmd/delete_test.go
+++ b/cmd/delete_test.go
@@ -353,7 +353,15 @@ func TestDelete_ByProjectClearsDeployedMarker(t *testing.T) {
 		Runtime:  "go",
 		Registry: TestRegistry,
 		Deployer: keda.KedaDeployerName, // intent - how to deploy
-		Deploy:   fn.DeploySpec{Namespace: "myns", Deployer: keda.KedaDeployerName},
+		Deploy: fn.DeploySpec{
+			Namespace: "myns",
+			Deployer:  keda.KedaDeployerName,
+			Options: fn.Options{
+				Scale: &fn.ScaleOptions{
+					KEDA: &fn.KEDAScaleOptions{Triggers: []fn.KEDATrigger{{Type: "http"}}},
+				},
+			},
+		},
 	}
 	f, err := fn.New().Init(f)
 	if err != nil {
@@ -441,7 +449,16 @@ func TestDelete_ByNameLeavesLocalFunctionUntouched(t *testing.T) {
 // after removal keeps the INTENT deployer intact and functional.
 func TestDelete_ByProjectPreservesDeployerForRedeploy(t *testing.T) {
 	root := FromTempDirectory(t)
-	if _, err := fn.New().Init(fn.Function{Runtime: "go", Root: root, Registry: TestRegistry}); err != nil {
+	f, err := fn.New().Init(fn.Function{Runtime: "go", Root: root, Registry: TestRegistry})
+	if err != nil {
+		t.Fatal(err)
+	}
+	// keda requires at least one trigger to be declared explicitly.
+	f.Deployer = keda.KedaDeployerName
+	f.Deploy.Options.Scale = &fn.ScaleOptions{
+		KEDA: &fn.KEDAScaleOptions{Triggers: []fn.KEDATrigger{{Type: "http"}}},
+	}
+	if err := f.Write(); err != nil {
 		t.Fatal(err)
 	}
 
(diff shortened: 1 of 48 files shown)


aliok:
/close

Closing in favor of https://github.com/knative/func/pull/4051

Issue #3203, opened 2025-11-18
Remote builds require checkout


lkingland:
Currently it is required that the Function be available locally in order to trigger a remote build because values from the func.yaml are required to create the pipeline.  This limitation should be removed such that the remote build triggered retrieves everything it needs from the remote git repository on-demand.  See the implementation of the `TestRemote_Ref` and `TestRemote_Dir` tests in the E2E tests for an example.

PR #4048, opened 2026-09-08
Name the remote source with --source, --source-dir and --revision

gauron99:
# Changes

- :broom: A remote build from a repository is named by its repository, a directory within it and a revision, and nothing about those values is specific to git: the revision is whatever git's fetch takes, a branch, a tag or a commit. `func deploy`'s `--git-url`, `--git-dir` and `--git-branch` become `--source`, `--source-dir` and `--revision` (`FUNC_SOURCE`, `FUNC_SOURCE_DIR`, `FUNC_REVISION`). The MCP `deploy` tool's `gitUrl`, `gitDir` and `gitBranch` become `source`, `sourceDir` and `revision`.
- :broom: In `func.yaml`, `build.git` (`url`, `revision`, `contextDir`) becomes `build.source` (`url`, `revision`, `dir`). A migration carries existing files over on load; the next write stores the new keys.
- `func config git set` keeps `--git-url`, `--git-branch` and `--git-dir`: the command is git-only, webhooks come from git hosts, and its branch is the one whose pushes Pipelines-as-Code builds, a trigger filter rather than a revision. A TODO at the flag records that this value should get a key of its own and that the PAC deploy step should stamp the revision it built from the push event. It writes the same `build.source` keys.

Behaviour is otherwise unchanged.

/kind cleanup

Relates to #3203

**Release Note**

```release-note
action required: `func deploy`'s `--git-url`, `--git-dir` and `--git-branch` are renamed `--source`, `--source-dir` and `--revision` (`FUNC_SOURCE`, `FUNC_SOURCE_DIR`, `FUNC_REVISION`); `--revision` takes a branch, a tag or a commit. The MCP `deploy` tool's `gitUrl`, `gitDir` and `gitBranch` become `source`, `sourceDir` and `revision`. In `func.yaml`, `build.git` becomes `build.source` with `url`, `revision` and `dir`; existing files are migrated on load. `func config git set` keeps its `--git-*` flags.
```

**Docs**

```docs

```


lkingland (commented):
Looking good!

I might consider going even more fundamental:  `--revision` (and later `--source`)

After all, there's no gurantee git will be the defacto choice forever, and we could in theory back --source with something other than a VCS.

lkingland (approved):
Looking good to me.  I see the bots identified a few typos we might want to /hold for

lkingland:
Well, that was fast.  I suppose we get the tweaks in a follow-up 😆

PR #4049, opened 2026-09-08
chore: remove the tkn-tasks subcommand

gauron99:
func tkn-tasks printed the embedded Tekton tasks as v1beta1 ClusterTasks for hand installation, added in #2396 for the OpenShift web console's static function plugin.


lkingland (approved):

gauron99:
as per discussion on slack, keeping this functionality on main  in case its needed in the near future. As I understand this should be possible to remove later (possibly early next year - 2027)  when we know if this is still needed or not.

PR #4051, opened 2026-09-08
Add KEDA Kafka consumer-lag scaling support

aliok:
# Changes

This PR adds Kafka consumer-lag autoscaling support via KEDA. When `deployer: keda` is set and a Kafka trigger is configured, the deployer creates a KEDA ScaledObject with a Kafka trigger that scales function pods based on consumer group lag, along with a TriggerAuthentication resource for SASL/TLS credentials.

The func.yaml scale configuration has been restructured:

- `scale` is now a top-level field (previously nested under `deploy.options.scale`), with `min`/`max` plus one of two mutually-exclusive sub-keys:
  - `scale.kpa` for the `knative` deployer: `metric`, `target`, `utilization`.
  - `scale.keda` for the `keda` deployer: `pollingInterval`, `cooldownPeriod`, and a `triggers` list. Each trigger (`http`, `kafka`, or `cron`) carries its own knobs, e.g. `lagThreshold` and `activationLagThreshold` for Kafka triggers.
- Kafka runtime configuration lives under `run.kafka` (brokers, topic, consumerGroup, security protocol, SASL/TLS settings).

A migration handles existing func.yaml files automatically: it moves `deploy.options.scale` to the top-level `scale`, lifts the old flat `metric`/`target`/`utilization` fields into `scale.kpa`, and renames `deploy.deployer` to `deploy.activeDeployer`.

The PR also includes testing deployment scenarios (A-G) with collected Kubernetes resource snapshots for each deployer/Kafka combination.

/kind enhancement

Supersedes #4045

Closes #2419

**Release Note**

```release-note
Functions can now consume from Apache Kafka using KEDA-based consumer-lag autoscaling. Configure Kafka connection details (brokers, topic, consumer group, SASL/TLS) under `run.kafka` in func.yaml and set `deployer: keda` to deploy with a ScaledObject that scales based on Kafka consumer lag. The `scale` config is now a top-level field with deployer-specific sub-keys: `scale.kpa` (knative) and `scale.keda` (keda), the latter supporting tuning knobs like pollingInterval, cooldownPeriod, and per-trigger lagThreshold/activationLagThreshold.
```

**Docs**

```docs

```


pkg/keda/kafka_scaling.go
```diff
@@ -0,0 +1,387 @@
...
+
+	if kafka.SASL != nil && kafka.SASL.Password != "" {
+		secretName, secretKey := parseSecretRef(kafka.SASL.Password)
+		if secretName != "" {
+			secretRefs = append(secretRefs, map[string]interface{}{
+				"parameter": "password",
+				"name":      secretName,
+				"key":       secretKey,
+			})
+		}
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/functions/function_scale.go
```diff
@@ -0,0 +1,109 @@
...
+		errors = append(errors, "scale.keda.pollingInterval must be >= 1")
+	}
+	if keda.CooldownPeriod != nil && *keda.CooldownPeriod < 1 {
+		errors = append(errors, "scale.keda.cooldownPeriod must be >= 1")
+	}
+
+	for i, t := range keda.Triggers {
+		switch t.Type {
+		case "http":
+			if t.TargetValue != nil && *t.TargetValue < 1 {
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/keda/kafka_scaling.go
```diff
@@ -0,0 +1,387 @@
...
+			continue
+		}
+		mountPath := *v.Path
+		if strings.HasPrefix(certPath, mountPath) {
+			rel, err := filepath.Rel(mountPath, certPath)
+			if err != nil {
+				continue
+			}
+			return *v.Secret, rel
+		}
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/keda/remover.go
```diff
@@ -50,24 +50,32 @@ func (remover *Remover) Remove(ctx context.Context, name, ns string) error {
...
 				name, recordedNS, err)
 		}
 	}
 
+	// Clean up Kafka scaling resources before deleting the Deployment.
+	// These have ownerReferences so they'd be garbage-collected, but
+	// explicit deletion avoids races with a slow GC.
+	// Ignore not-found: these resources may not exist (HTTP-only deploy).
+	_ = deleteScaledObject(ctx, dynClient, ns, scaledObjectName(name))
+	_ = deleteTriggerAuth(ctx, dynClient, ns, triggerAuthName(name))
```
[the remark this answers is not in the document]
aliok: Fixed

docs/reference/func_yaml.md
```diff
@@ -149,21 +188,83 @@ Options allows you to set specific configuration for the deployed function, allo
...
+        cpu: 100m
+        memory: 128Mi
+      limits:
+        cpu: 1000m
+        memory: 256Mi
+        concurrency: 100
+```
+
+Example using `deployer: keda` with an HTTP trigger and a Kafka trigger:
+
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/keda/deployer.go
```diff
@@ -139,39 +146,81 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+	// Kafka trigger path: TriggerAuthentication + ScaledObject
+	if wantKafka && f.Run.Kafka != nil {
+		if needsTriggerAuth(f.Run.Kafka) {
+			ta := buildTriggerAuth(f, deployment, namespace)
+			if ta != nil {
+				if err := ensureTriggerAuth(ctx, dynClient, ta); err != nil {
+					return fn.DeploymentResult{}, fmt.Errorf("failed to ensure TriggerAuthentication: %w", err)
+				}
+			}
+		}
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/keda/kafka_scaling.go
```diff
@@ -0,0 +1,422 @@
...
+		mountPath := filepath.Clean(*v.Path)
+		cp := filepath.Clean(certPath)
+		if cp == mountPath || strings.HasPrefix(cp, mountPath+string(filepath.Separator)) {
+			rel, err := filepath.Rel(mountPath, cp)
+			if err != nil {
+				continue
+			}
+			return *v.Secret, rel
+		}
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/keda/deployer.go
```diff
@@ -139,39 +146,90 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
 	} else {
-		if url, err = d.deployClusterLocal(ctx, target); err != nil {
-			return fn.DeploymentResult{}, err
+		// No HTTP trigger — URL is the app service
+		url = fmt.Sprintf("http://%s.%s.svc:8080", f.Name, namespace)
+	}
+
+	// Kafka trigger path: TriggerAuthentication + ScaledObject
+	if wantKafka && f.Run.Kafka != nil {
+		if needsTriggerAuth(f.Run.Kafka) {
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/functions/function_migrations.go
```diff
@@ -356,6 +357,86 @@ func migratePersistentVolumeTypoFixup(fn Function, m migration) (Function, error
...
+		if hasFlat && newScale.KPA == nil {
+			newScale.KPA = &KPAScaleOptions{
+				Metric:      old.Metric,
+				Target:      old.Target,
+				Utilization: old.Utilization,
+			}
+		}
+
+		f.Scale = newScale
+	}
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/functions/function_migrations.go
```diff
@@ -356,6 +357,97 @@ func migratePersistentVolumeTypoFixup(fn Function, m migration) (Function, error
...
+		}
+
+		hasFlat := old.Metric != nil || old.Target != nil || old.Utilization != nil
+		if hasFlat && newScale.KPA == nil {
+			newScale.KPA = &KPAScaleOptions{
+				Metric:      old.Metric,
+				Target:      old.Target,
+				Utilization: old.Utilization,
+			}
+		}
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/functions/function_options.go
```diff
@@ -7,15 +7,41 @@ import (
...
 	Resources *ResourcesOptions `yaml:"resources,omitempty"`
 }
 
 type ScaleOptions struct {
-	Min         *int64   `yaml:"min,omitempty" jsonschema_extras:"minimum=0"`
-	Max         *int64   `yaml:"max,omitempty" jsonschema_extras:"minimum=0"`
+	Min  *int64            `yaml:"min,omitempty" jsonschema_extras:"minimum=0"`
+	Max  *int64            `yaml:"max,omitempty" jsonschema_extras:"minimum=0"`
+	KEDA *KEDAScaleOptions `yaml:"keda,omitempty"`
+	KPA  *KPAScaleOptions  `yaml:"kpa,omitempty"`
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/keda/deployer.go
```diff
@@ -139,39 +156,122 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
-		appliedExpose = f.Expose
 	} else {
-		if url, err = d.deployClusterLocal(ctx, target); err != nil {
-			return fn.DeploymentResult{}, err
+		// No HTTP trigger — URL is the app service. The Service listens on
+		// port 80 (routing to the container's DefaultHTTPPort via
+		// targetPort), so the URL, like elsewhere in the codebase (e.g.
+		// pkg/k8s/describer.go), has no explicit port.
+		url = fmt.Sprintf("http://%s.%s.svc", f.Name, namespace)
+	}
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/keda/deployer.go
```diff
@@ -99,8 +100,49 @@ func (k *kedaDeployerDecorator) UpdateLabels(function fn.Function, labels map[st
...
+		if err := validateBridgeName(f.Name); err != nil {
+			return fn.DeploymentResult{}, err
+		}
+	}
+
+	// The following are all pure functions of f -- no cluster state needed --
+	// so they run before d.Deployer.Deploy creates anything. ValidateScale
+	// already rejects each of these, but Deploy is reachable without going
+	// through Function.Validate first (library callers, tests): failing
+	// before the raw Deployment/Service exist, rather than after, avoids
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/keda/deployer.go
```diff
@@ -139,42 +181,116 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+		// port 80 (routing to the container's DefaultHTTPPort via
+		// targetPort), so the URL, like elsewhere in the codebase (e.g.
+		// pkg/k8s/describer.go), has no explicit port.
+		url = fmt.Sprintf("http://%s.%s.svc", f.Name, namespace)
+
+		// A prior deploy may have exposed this function over HTTP. Nothing
+		// reconciles that exposure once the HTTP trigger is gone, so clear
+		// it the same way deployClusterLocal does -- otherwise the old
+		// Route and the Service's exposure annotations stay active,
+		// pointing at a function that no longer has anything serving HTTP.
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/keda/deployer.go
```diff
@@ -139,42 +181,116 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+		}
+		if err := d.clearExposure(ctx, target, appService.Annotations[k8s.RouteNamespaceAnnotation]); err != nil {
 			return fn.DeploymentResult{}, err
 		}
 	}
 
+	// Kafka trigger path: TriggerAuthentication + ScaledObject
+	if wantKafka && f.Run.Kafka != nil {
+		if needsTriggerAuth(f.Run.Kafka) {
+			ta := buildTriggerAuth(f, deployment, namespace)
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/keda/deployer.go
```diff
@@ -139,42 +181,116 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+		if so != nil {
+			if err := ensureScaledObject(ctx, dynClient, so); err != nil {
+				return fn.DeploymentResult{}, fmt.Errorf("failed to ensure ScaledObject: %w", err)
+			}
+		}
+	} else {
+		// The Kafka trigger was dropped (or never configured): remove any
+		// scaler resources a prior deploy left behind, so switching back to
+		// http-only doesn't leave a ScaledObject/TriggerAuthentication still
+		// acting on stale Kafka lag config. Not fatal to Deploy, same as
```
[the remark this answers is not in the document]
aliok: Fixed

pkg/functions/function_scale.go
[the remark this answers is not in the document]
aliok: Thanks — splitting this into the two concerns it raises:

- **`cron` and combined `http`+`kafka`**: agreed, these are unsupported and are now rejected by validation (a `cron` trigger and any `http`+`kafka` combination on the same workload both fail preflight). So the "silently does nothing / two scalers on one Deployment" cases are closed.

- **`http`-only trigger with `run.kafka` set**: this one is *intentionally allowed*, not a misconfiguration. `run.kafka` and the scaling trigger are orthogonal:
  - `run.kafka` describes **what the function does** — it consumes from Kafka (`FUNC_TRANSPORT=kafka`).
  - the trigger describes **how it scales** — HTTP traffic, Kafka lag, or just fixed min/max bounds.

  Consuming Kafka does not require lag-based autoscaling. "Run a Kafka consumer whose replica count is driven by HTTP traffic (or fixed min/max)" is a legitimate, supported choice, so validation deliberately does not reject it.

  One caveat we should document: with `http`-only + `run.kafka` + `min: 0`, KEDA can scale the workload to zero when there is no HTTP traffic, which stops the consumer and lets lag grow with nothing to wake it. `min: >= 1` keeps a consumer alive and avoids this. That is inherent to HTTP scale-to-zero and is the user's explicit choice, so it's a docs note rather than a validation error.

pkg/keda/deployer.go
```diff
@@ -99,8 +101,111 @@ func (k *kedaDeployerDecorator) UpdateLabels(function fn.Function, labels map[st
...
+		}
+	}
+	if wantKafka && f.Run.Kafka != nil && f.Run.Kafka.SASL != nil && f.Run.Kafka.SASL.Mechanism != "" && kedaSASLType(f.Run.Kafka.SASL.Mechanism) == "" {
+		// kedaSASLType returns "" for anything it doesn't recognize, and
+		// buildScaledObject would set the trigger's "sasl" metadata to that
+		// empty string, producing an invalid KEDA trigger.
+		return fn.DeploymentResult{}, fmt.Errorf(
+			"function %q: run.kafka.sasl.mechanism %q is not supported, must be one of PLAIN, SCRAM-SHA-256, SCRAM-SHA-512",
+			f.Name, f.Run.Kafka.SASL.Mechanism)
 	}
```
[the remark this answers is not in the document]
aliok: Addressed. Extracted the securityProtocol/TLS/SASL consistency checks out of `validateKafka` into an exported `fn.ValidateKafkaSecurity(kafka)` (single source of truth), and the KEDA deployer preflight now calls it instead of the previous narrow "non-empty unrecognized mechanism" check.

So a direct `Deployer.Deploy` caller that bypasses `Function.Validate` is now rejected before any resource is created for:
- `SASL_SSL`/`SASL_PLAINTEXT` with an empty/nil mechanism (previously slipped through and left the ScaledObject without `sasl` metadata),
- a `sasl` block with a non-SASL security protocol,
- an unrecognized mechanism,
- missing user/password, and an invalid security protocol.

Scoped deliberately to the security/SASL/TLS consistency subset — it excludes the runtime/invoke and broker/topic/consumerGroup checks (the former is a function-authoring concern, not a KEDA-resource concern; the latter are already validated in the preflight above) so it does not reject otherwise-valid direct-deploy inputs.

Covered by `TestDeploy_KafkaSASLPreflight` (empty mechanism, sasl-with-SSL, unrecognized mechanism).

pkg/keda/kafka_scaling.go
```diff
@@ -0,0 +1,475 @@
...
+		Version:  "v1alpha1",
+		Resource: "triggerauthentications",
+	}
+)
+
+func scaledObjectName(funcName string) string {
+	return funcName + "-kafka"
+}
+
+func triggerAuthName(funcName string) string {
```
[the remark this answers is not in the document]
aliok: Fixed. Added `validateKafkaResourceNames`, the Kafka-path counterpart to `validateBridgeName`: it validates the generated ScaledObject name (`<name>-kafka`) and, when credentials are configured, the TriggerAuthentication name (`<name>-kafka-auth`) against the DNS-1035 63-character limit before anything is created. `needsTriggerAuth` picks the longest suffix that will actually be created, per your "use the longest required suffix" note.

It runs under the `wantKafka` branch of the Deploy preflight, symmetric with the existing `validateBridgeName` call under `wantHTTP`, so it also covers direct `Deploy` callers that bypass `Function.Validate`.

Covered by `TestDeploy_KafkaResourceNamePreflight`.

pkg/functions/function_migrations.go
```diff
@@ -356,6 +357,97 @@ func migratePersistentVolumeTypoFixup(fn Function, m migration) (Function, error
...
+	type oldOptions struct {
+		Scale *oldScale `yaml:"scale,omitempty"`
+	}
+	type oldDeploy struct {
+		Options oldOptions `yaml:"options,omitempty"`
+	}
+	var disk struct {
+		Deploy oldDeploy     `yaml:"deploy,omitempty"`
+		Scale  *ScaleOptions `yaml:"scale,omitempty"`
+	}
```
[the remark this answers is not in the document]
aliok: Non-issue — there is no legacy `deploy.kafka` to migrate. Kafka config has lived at `run.kafka` (`RunSpec.Kafka`) since it was introduced (commit `c570624d`); it was never stored under `deploy.kafka`. `git log -S 'deploy.kafka'` across the history returns nothing.

Old func.yaml files deserialize their Kafka block straight into `f.Run.Kafka` through the normal unmarshal path, so nothing is dropped and no migration step is required. The literal observation (the migration input carries no legacy kafka field) is accurate but harmless. Resolving.

pkg/keda/deployer.go
```diff
@@ -139,42 +204,139 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
 
-	ref := deployer.NewExposureRef(f.Name, namespace, interceptorNS)
-	if err := ensureInterceptorBridgeService(ctx, k8sClientset, ref, deployment); err != nil {
-		return fn.DeploymentResult{}, fmt.Errorf("failed to ensure proxy service exists: %w", err)
+	// Delete stale scaler resources for whichever trigger type is NOT
+	// currently configured, before provisioning the type that is: creating
+	// a new scaler while an old one of the other kind still targets the
+	// same Deployment can trip KEDA's one-scaler-per-workload rule and fail
+	// the new scaler's readiness wait. Not fatal to Deploy, same as
```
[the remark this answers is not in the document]
aliok: Known and deliberately accepted — this is documented in the code at the delete site and at the transition site.

`deleteScaledObject`/`deleteTriggerAuth` return once the delete is *accepted*, not once the object is gone: KEDA attaches its own finalizer to the ScaledObject/TriggerAuthentication, so `Delete()` only sets `deletionTimestamp` and hands off to KEDA's controller, which finishes cleanup asynchronously. Polling for actual removal on every deploy would add latency to the overwhelmingly common single-type deploy (not just transitions) to close a race this narrow.

The transition is therefore treated as a rare, self-correcting race: if the new scaler's readiness trips over the old one still terminating, the next deploy succeeds once KEDA finishes. The reasoning and trade-off are written out in the comment above the delete block in `Deploy` (deployer.go, the "This narrows that race but doesn't close it" note) and on `deleteScaledObject`/`deleteTriggerAuth`. Not fatal to Deploy either way — both scalers are owned by the Deployment and garbage-collected regardless.

pkg/functions/function.go
```diff
@@ -336,10 +339,10 @@ type DeploySpec struct {
...
 	ImagePullSecret string `yaml:"imagePullSecret,omitempty"`
 
-	// Deployer records the deployer the Function is CURRENTLY DEPLOYED:
-	// observed state, written after successful deployment, and cleared on
-	// undeploy alongside Namespace.
-	Deployer string `yaml:"deployer,omitempty" jsonschema:"enum=knative,enum=raw,enum=keda"`
+	// ActiveDeployer records the deployer the Function is CURRENTLY DEPLOYED
+	// with: observed state, written after successful deployment, and cleared
+	// on undeploy alongside Namespace. User intent lives on Function.Deployer.
+	ActiveDeployer string `yaml:"activeDeployer,omitempty" jsonschema:"enum=knative,enum=raw,enum=keda"`
```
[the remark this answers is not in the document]
aliok: Intentional, and we're accepting the Go source break rather than adding deprecated aliases.

The rename fixes a real ambiguity: `DeploySpec.Deployer`/`Expose` were being used to mean *both* user intent and observed/deployed state. They now mean observed state only — `ActiveDeployer`/`ActiveExpose`, written after a successful deploy and cleared on undeploy — while user intent lives on the top-level `Function.Deployer`/`Function.Expose`. Keeping the old names as aliases would re-introduce exactly the intent-vs-observed collision the rename removes, so we're not doing that.

func.yaml (on-disk) compatibility is preserved: the YAML keys did change (`deploy.deployer`/`deploy.expose` → `deploy.activeDeployer`/`deploy.activeExpose`), but a dedicated migration recovers the legacy keys — see `function_migrations.go` (the migration that maps the old observed-state keys and back-fills `f.Deployer` from a pre-#3953 `deploy.deployer`). So existing func.yaml files load correctly.

The break is limited to Go consumers referencing these struct fields directly, which is acceptable for this library type. Documenting the break here rather than carrying aliases.

pkg/keda/deployer.go
```diff
@@ -99,8 +102,165 @@ func (k *kedaDeployerDecorator) UpdateLabels(function fn.Function, labels map[st
...
+		// kafka trigger's ScaledObject honors it. Warn so a caller who set it
+		// expecting it to apply isn't left wondering why nothing changed.
+		fmt.Fprintf(os.Stderr, "Warning: scale.keda.pollingInterval is ignored for an http trigger (it applies only to kafka triggers); function %q\n", f.Name)
+	}
+
+	if wantHTTP {
+		if err := validateBridgeName(f.Name); err != nil {
+			return fn.DeploymentResult{}, err
+		}
+	}
```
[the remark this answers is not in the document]
aliok: Intended and allowed — this is the same point raised (and resolved) on the earlier thread about `run.kafka` with an HTTP trigger. It conflates two orthogonal concerns:

- **`run.kafka`** = *what the function does* — it consumes from Kafka (the runtime sets `FUNC_TRANSPORT=kafka`).
- **the KEDA trigger** = *how the Deployment is scaled* — by HTTP traffic, by Kafka lag, or by fixed min/max bounds.

There is no rule that consuming Kafka forces lag-based autoscaling. "Run a Kafka consumer, but manage its replica count by HTTP traffic (or a fixed minimum)" is a coherent, supported choice — e.g. a consumer that also serves an HTTP API, or one you want held at a fixed replica count regardless of lag. Rejecting the combination would forbid those valid setups, so `ValidateScale` and the preflight deliberately allow it.

The `scale.min: 0` case is a genuine footgun, but it's the user's explicit choice, not an invalid configuration: with an HTTP-only trigger and `min: 0`, KEDA scales to zero when there's no HTTP traffic, which stops the consumer, and lag then grows with no HTTP request to wake it. The remedy is `min >= 1` (keeps at least one consumer alive), which is exactly what a user choosing HTTP-driven scale-to-zero for a Kafka consumer needs to know — so this belongs in the docs, not in a hard rejection. If lag-based scaling is what's wanted, that's precisely what a `kafka` trigger is for.

So: no validation change; will make sure the `min: 0` caveat is called out in the docs.

cmd/func-util/main.go
```diff
@@ -159,7 +159,7 @@ func deploy(ctx context.Context) error {
 	// honors --deployer, which travels in func.yaml as intent.
 	deployer := f.Deployer
 	if deployer == "" {
-		deployer = f.Deploy.Deployer
+		deployer = f.Deploy.ActiveDeployer
```
gauron99: forgotten rename back

cmd/delete_test.go
```diff
@@ -353,7 +353,13 @@ func TestDelete_ByProjectClearsDeployedMarker(t *testing.T) {
 		Runtime:  "go",
 		Registry: TestRegistry,
 		Deployer: keda.KedaDeployerName, // intent - how to deploy
-		Deploy:   fn.DeploySpec{Namespace: "myns", Deployer: keda.KedaDeployerName},
+		Deploy: fn.DeploySpec{
+			Namespace:      "myns",
+			ActiveDeployer: keda.KedaDeployerName,
```
gauron99: forgotten rename back

cmd/delete_test.go
```diff
@@ -382,8 +388,8 @@ func TestDelete_ByProjectClearsDeployedMarker(t *testing.T) {
 	if loaded.Deploy.Namespace != "" {
 		t.Fatalf("expected Deploy.Namespace cleared after a successful undeploy, got %q", loaded.Deploy.Namespace)
 	}
-	if loaded.Deploy.Deployer != "" {
-		t.Fatalf("expected Deploy.Deployer cleared after a successful undeploy, got %q", loaded.Deploy.Deployer)
+	if loaded.Deploy.ActiveDeployer != "" {
```
gauron99: same here

cmd/delete_test.go
```diff
@@ -401,7 +407,7 @@ func TestDelete_ByNameLeavesLocalFunctionUntouched(t *testing.T) {
 		Runtime:  "go",
 		Registry: TestRegistry,
 		Name:     "localfn",
-		Deploy:   fn.DeploySpec{Namespace: "myns", Deployer: keda.KedaDeployerName},
+		Deploy:   fn.DeploySpec{Namespace: "myns", ActiveDeployer: keda.KedaDeployerName},
```
gauron99: and here

cmd/delete_test.go
```diff
@@ -441,7 +447,16 @@ func TestDelete_ByNameLeavesLocalFunctionUntouched(t *testing.T) {
 // after removal keeps the INTENT deployer intact and functional.
 func TestDelete_ByProjectPreservesDeployerForRedeploy(t *testing.T) {
 	root := FromTempDirectory(t)
-	if _, err := fn.New().Init(fn.Function{Runtime: "go", Root: root, Registry: TestRegistry}); err != nil {
+	f, err := fn.New().Init(fn.Function{Runtime: "go", Root: root, Registry: TestRegistry})
+	if err != nil {
+		t.Fatal(err)
+	}
+	// keda requires at least one trigger to be declared explicitly.
```
gauron99:
```
❯ FUNC_REGISTRY=ttl.sh/myfunc-pr4051:1h $FUNC deploy --deployer keda
Error: 'func.yaml' contains errors:
	deployer keda requires at least one trigger in scale.keda.triggers
```

This should be possible to run. We should default to the old http trigger as is on main without any extra specs given from user. (Which I think now requires even manual edit of func.yaml)

cmd/delete_test.go
```diff
@@ -476,8 +491,8 @@ func TestDelete_ByProjectPreservesDeployerForRedeploy(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if loaded.Deploy.Deployer != keda.KedaDeployerName {
-		t.Fatalf("expected the flag-less redeploy to reuse the persisted %q deployer, got %q", keda.KedaDeployerName, loaded.Deploy.Deployer)
+	if loaded.Deploy.ActiveDeployer != keda.KedaDeployerName {
```
gauron99: also should not be here

cmd/delete_test.go
```diff
@@ -490,7 +505,7 @@ func TestDelete_ByProjectThenRedeployWithDifferentDeployerNotBlocked(t *testing.
 		Root:     root,
 		Runtime:  "go",
 		Registry: TestRegistry,
-		Deploy:   fn.DeploySpec{Namespace: "myns", Deployer: keda.KedaDeployerName},
+		Deploy:   fn.DeploySpec{Namespace: "myns", ActiveDeployer: keda.KedaDeployerName},
```
gauron99: should not be here

cmd/deploy.go
```diff
@@ -306,8 +306,8 @@ func runDeploy(cmd *cobra.Command, newClient ClientFactory) (err error) {
 
 	// Back-compat: a function deployed before the deployer was recorded has a
 	// namespace but no deployer, which historically could only mean knative.
-	if f.Deploy.Namespace != "" && f.Deploy.Deployer == "" {
-		f.Deploy.Deployer = deployers.Knative
+	if f.Deploy.Namespace != "" && f.Deploy.ActiveDeployer == "" {
+		f.Deploy.ActiveDeployer = deployers.Knative
```
gauron99: should not be here

cmd/deploy.go
```diff
@@ -377,7 +377,7 @@ func runDeploy(cmd *cobra.Command, newClient ClientFactory) (err error) {
 		// After a remote pipeline: intent was active, describer saw no
 		// applied expose. Typical cause is a func-util image that predates
 		// this field. Knative is excluded; it never applies expose.
-		if fn.ExposureRecordMissing(f.Expose, f.Deploy.Expose, f.Deploy.Deployer) {
+		if fn.ExposureRecordMissing(f.Expose, f.Deploy.ActiveExpose, f.Deploy.ActiveDeployer) {
```
gauron99: should not be here

cmd/deploy_test.go
```diff
@@ -1187,7 +1188,7 @@ func TestDeploy_NamespaceRedeployWarning(t *testing.T) {
 	f := fn.Function{
 		Runtime: "go",
 		Root:    root,
-		Deploy:  fn.DeploySpec{Namespace: "funcns", Deployer: deployers.Default},
+		Deploy:  fn.DeploySpec{Namespace: "funcns", ActiveDeployer: deployers.Default},
```
gauron99: same in this file too

pkg/config/config.go
```diff
@@ -133,8 +133,8 @@ func (c Global) Apply(f fn.Function) Global {
 		c.Builder = f.Build.Builder
 	}
 	// opt 1: last deployed deployer
-	if f.Deploy.Deployer != "" {
-		c.Deployer = f.Deploy.Deployer
+	if f.Deploy.ActiveDeployer != "" {
```
gauron99: should not be here

pkg/functions/function_migrations.go
```diff
@@ -406,6 +407,147 @@ func nestedString(v interface{}, keys ...string) string {
...
+	// field was introduced (commit 65029640). Every deploy since then sets
+	// f.Deployer explicitly (see cmd/deploy.go), so the only way to see it
+	// empty here while the legacy observed field is non-empty is a
+	// function deployed before that split existed. Treat the old value as
+	// the carried-forward intent too, not just observed state -- computed
+	// before anything below reads f.Deployer, so both the scale.kpa
+	// decision and the keda-defaults-to-http-trigger block see it. Read from
+	// disk rather than f.Deploy.ActiveDeployer so the migration is robust to
+	// callers that pass a Function not populated via the primary unmarshal.
+	if f.Deployer == "" && disk.Deploy.Deployer != "" {
```
gauron99: I think if we fix the validation (to allow empty scale because default should be the http scaler as is today on main) we can completely remove this deploy.deployer -> deployer  intent transition.

This is technically fixing one scenario if im not missing any and that is if a user has a function deployed and they run func delete with a new binary the deploy.deployer (which was the only key for them but now is status only) would get removed and the intent would be lost so re-running a func deploy would run with defaultDeployer instead of their declared one. I think that blast radius is extremely small and it does not need to be coupled with this PR

pkg/functions/function_scale.go
```diff
@@ -0,0 +1,144 @@
...
+import (
+	"fmt"
+	"math"
+)
+
+// ValidateScale validates the top-level scale configuration against the chosen
+// deployer and Kafka config. It replaces the previous validateScaleDeployer,
+// validateKEDAScale, and validateKPAScale functions with a single entry point.
+func ValidateScale(scale *ScaleOptions, deployer string, kafka *KafkaConfig) (errors []string) {
+	if deployer == "keda" && (scale == nil || scale.KEDA == nil || len(scale.KEDA.Triggers) == 0) {
```
gauron99: I think if we go with the rest of fixes we can drop this condition block and say that empty scale is valid for every deployer -> for keda deployer it simply means "use http scaler"

pkg/keda/deployer.go
```diff
@@ -99,8 +102,165 @@ func (k *kedaDeployerDecorator) UpdateLabels(function fn.Function, labels map[st
...
+	triggers := triggers(f)
+	if len(triggers) == 0 {
+		// triggers(f) only returns empty when scale.keda is present with an
+		// explicitly empty triggers list (the nil-Scale/nil-KEDA case falls
+		// back to a plain http trigger). ValidateScale already rejects this,
+		// but Deploy is reachable without Function.Validate first (library
+		// callers, tests): without this check, Deploy would silently skip
+		// both the HTTPScaledObject and Kafka ScaledObject paths and deploy
+		// with no scaler at all.
+		return fn.DeploymentResult{}, fmt.Errorf("function %q: deployer keda requires at least one trigger in scale.keda.triggers", f.Name)
```
gauron99, line 114: as mentioned above in some test i reviewed first I think we should build-in a default to always fallback to http scaler (as is currently on main now) when the scale.keda is nil (not explicitly [] empty) & `deployer` is keda. 
This will give us backwards compatibility and super ease of use for users because you dont need to worry about anything after `func deploy --deployer keda`.

pkg/keda/deployer.go
```diff
@@ -99,8 +102,165 @@ func (k *kedaDeployerDecorator) UpdateLabels(function fn.Function, labels map[st
 }
 
 func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResult, error) {
```
gauron99, line 104: the deployer seems to be doing a lot of validation that is already done beforehand duplicating. Also the deployer should be minimal "executor". I think we could trim this down to nil guards and name validators (for k8s resource char limits) and secret path?

gauron99 (commented):
First pass on the PR. Thanks for doing the migration and the scale yaml movement too!
Im not sure if my page is messed up but im seeing a lot of ActiveDeployer left over after that 4th commit -- Ive stopped naming them in the reviews after a couple.

Ive also though we could simplify the PR by removing the deployer migration (which seems unrelated and its there as a transient "fix") -> we fix the empty scale validator -> we can remove the deployer migration

pkg/functions/function_migrations.go
aliok: Agreed — done. Removed the `deploy.deployer` → `deployer` intent-recovery block entirely. The remaining edge you noted (a `func delete` with a newer binary dropping the intent so a re-deploy falls back to the default deployer) is pre-existing on `main` and decoupled from this PR as you suggested — tracked in https://github.com/knative/func/issues/4054.

pkg/functions/function_scale.go
aliok: Done — dropped the block. `scale == nil` now returns valid for every deployer; for `deployer: keda` it means "use the http scaler", supplied by the deployer's `triggers()` helper. Only an explicitly-written `scale.keda` with an empty `triggers: []` is still rejected.

pkg/keda/deployer.go
aliok: Done — `deployer: keda` with a nil `scale.keda` now defaults to a single `http` trigger (backwards-compatible with `main`). `func deploy --deployer keda` works with no scale config at all; only an explicitly-written empty `triggers: []` is rejected. Docs (`func_yaml.md`) updated to state this.

pkg/functions/function.go
```diff
@@ -337,9 +367,9 @@ type DeploySpec struct {
 	// More info: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
 	ImagePullSecret string `yaml:"imagePullSecret,omitempty"`
 
-	// Deployer records the deployer the Function is CURRENTLY DEPLOYED:
-	// observed state, written after successful deployment, and cleared on
-	// undeploy alongside Namespace.
+	// Deployer records the deployer the Function is CURRENTLY DEPLOYED
+	// with: observed state, written after successful deployment, and cleared
+	// on undeploy alongside Namespace. User intent lives on Function.Deployer.
```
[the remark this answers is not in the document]
aliok: This is intentional and reflects a later decision by the human reviewer. The `activeDeployer` YAML key was reverted: the persisted key stays `deploy.deployer` (observed state), and the observed-vs-intent distinction is documented in the field's doc comment rather than encoded in the key name. The Go field was renamed back from `ActiveDeployer` to `Deployer` across all call sites and tests accordingly. So the field tags, migration, and tests are consistent with the intended format — `deployer`, not `activeDeployer`.

pkg/keda/deployer.go
aliok: Agreed the deployer is doing too much and should be a minimal executor — but Id like to keep it for this PR and do the trim separately, tracked in https://github.com/knative/func/issues/4055.

The reason it can't just be deleted now: these checks aren't re-validation within one process — each guards a different deploy *process*, and `Function.Validate()` (which runs `ValidateScale`) is skipped in most of the flows that reach the deployer. `func-util deploy` (the on-cluster Tekton deploy step) calls `client.Deploy(ctx, f)` with no `Validate()`, and `Client.Deploy` itself only validates expose/switch, never scale. So for every on-cluster deploy (and any library caller) the deployer's preflight is currently the *only* validation that runs.

The genuinely deployer-only bits you called out — k8s resource-name length and secret/TLS paths — stay regardless. The rest duplicates `ValidateScale`/`ValidateKafkaSecurity`, and the clean fix is to validate once at the shared choke point (`Client.Deploy`, which both the CLI and `func-util` go through) and then shrink the deployer. That's a cross-deployer change with a scoping decision (deploy-relevant validation vs. the full authoring `Validate()`), so it's better as its own PR — #4055.

pkg/functions/function_migrations.go
```diff
@@ -406,6 +407,143 @@ func nestedString(v interface{}, keys ...string) string {
...
+	}
+	type oldDeploy struct {
+		Options  oldOptions `yaml:"options,omitempty"`
+		Deployer string     `yaml:"deployer,omitempty"`
+		Expose   string     `yaml:"expose,omitempty"`
+	}
+	var disk struct {
+		Deploy oldDeploy     `yaml:"deploy,omitempty"`
+		Scale  *ScaleOptions `yaml:"scale,omitempty"`
+	}
```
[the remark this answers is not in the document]
aliok: Won't fix in this PR — accepted as negligible risk. The path requires a *pre-0.34* func.yaml (the format from before the 0.34 specs restructure that moved scale under `deploy.options.scale`) that *also* carried flat KPA tuning fields (metric/target/utilization) under the old top-level `options.scale`. The migration's disk re-read deliberately targets the 0.34+ location (`deploy.options.scale`); teaching it to also parse the pre-0.34 top-level shape adds migration surface for a file layout we have no evidence exists in the wild. Min/max (the fields that layout realistically held) still carry through via the in-memory fallback. If a real affected file ever surfaces we can revisit, but it's not worth coupling to this PR.

pkg/functions/function_options.go
```diff
@@ -7,15 +7,60 @@ import (
...
+	Timezone               string `yaml:"timezone,omitempty"`
+	Start                  string `yaml:"start,omitempty"`
+	End                    string `yaml:"end,omitempty"`
+	DesiredReplicas        *int64 `yaml:"desiredReplicas,omitempty" jsonschema_extras:"minimum=1"`
+}
+
+type KPAScaleOptions struct {
 	Metric      *string  `yaml:"metric,omitempty" jsonschema:"enum=concurrency,enum=rps"`
-	Target      *float64 `yaml:"target,omitempty" jsonschema_extras:"minimum=0.01"`
+	Target      *float64 `yaml:"target,omitempty" jsonschema:"exclusiveMinimum=true" jsonschema_extras:"minimum=0.01"` // exclusiveMinimum=true: jsonschema_extras' "minimum" truncates "0.01" to 0 via strconv.Atoi, so this at least excludes the concrete invalid value (0) ValidateScale rejects
```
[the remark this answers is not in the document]
aliok: Won't fix — known generator limitation, already documented in the code comment on this line. The alecthomas/jsonschema generator parses `jsonschema_extras` "minimum" with `strconv.Atoi`, which truncates `0.01` to `0`, so the exact runtime bound can't be expressed in the schema; `exclusiveMinimum` at least rejects the one concrete invalid value (0). Values `0.001`–`0.009` pass the generated schema but are rejected by `ValidateScale` at runtime with a clear message, and the runtime validator is the source of truth. Working around the generator for this narrow gap isn't worth it.

pkg/keda/deployer.go
```diff
@@ -139,42 +299,163 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+	// deploy (the overwhelmingly common case, not just a transition)
+	// wasn't judged worth it for a race this narrow.
+	if !wantHTTP {
+		// No HTTP trigger: a prior deploy's HTTPScaledObject and
+		// interceptor bridge Service, if any, are now orphaned.
+		if err := deleteHTTPScaledObject(ctx, namespace, f.Name); err != nil {
+			fmt.Fprintf(os.Stderr, "warning: %v\n", err)
+		} else if d.verbose {
+			fmt.Fprintf(os.Stderr, "deleted HTTPScaledObject %s/%s (if it existed); KEDA's finalizer means removal may still be in progress\n", namespace, f.Name)
+		}
```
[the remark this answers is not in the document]
aliok: Fixed in 2daac7f8. The stale HTTP-scaler cleanup during an http→kafka transition now returns the error instead of logging a warning and continuing, so we no longer risk two scalers on one Deployment with Deploy reporting success. As you noted, `deleteHTTPScaledObject` already suppresses NotFound (returns nil), so steady-state deploys stay a no-op and only a real API failure mid-transition aborts. Added unit coverage that the delete helpers propagate non-NotFound errors.

pkg/keda/deployer.go
```diff
@@ -139,42 +299,163 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+	if !wantKafka {
+		// The Kafka trigger was dropped (or never configured): remove any
+		// scaler resources a prior deploy left behind, so switching back to
+		// http-only doesn't leave a ScaledObject/TriggerAuthentication
+		// still acting on stale Kafka lag config.
+		if err := deleteScaledObject(ctx, dynClient, namespace, scaledObjectName(f.Name)); err != nil {
+			fmt.Fprintf(os.Stderr, "warning: %v\n", err)
+		} else if d.verbose {
+			fmt.Fprintf(os.Stderr, "deleted ScaledObject %s/%s (if it existed); KEDA's finalizer means removal may still be in progress\n", namespace, scaledObjectName(f.Name))
+		}
```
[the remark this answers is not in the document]
aliok: Fixed in 2daac7f8, same change as the http→kafka direction. The stale Kafka `ScaledObject`/`TriggerAuthentication` cleanup during a kafka→http transition now returns the error instead of only logging it, so a stale Kafka scaler can no longer compete with the new HTTPScaledObject while Deploy reports success. `deleteScaledObject`/`deleteTriggerAuth` already treat NotFound as success, so steady-state deploys are unaffected. Unit coverage added for the non-NotFound error path.

gauron99:
this seems to be an issue
```
 deleted ScaledObject default/testremote-default-kafka (if it existed); KEDA's finalizer means removal may still be in progress
        ERROR: cannot deploy the function: deploy error. failed to remove stale Kafka TriggerAuthentication before switching triggers: failed to delete TriggerAuthentication default/testremote-default-kafka-auth: triggerauthentications.keda.sh "testremote-default-kafka-auth" is forbidden: User "system:serviceaccount:default:default" cannot delete resource "triggerauthentications" in API group "keda.sh" in the namespace "default"
```

pkg/keda/deployer.go
```diff
@@ -139,42 +299,172 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
-
-	labels, err := deployer.GenerateCommonLabels(f, d.decorator)
-	if err != nil {
-		return fn.DeploymentResult{}, fmt.Errorf("failed to generate common labels: %w", err)
+	if !wantKafka {
+		// The Kafka trigger was dropped (or never configured): remove any
+		// scaler resources a prior deploy left behind, so switching back to
+		// http-only doesn't leave a ScaledObject/TriggerAuthentication
+		// still acting on stale Kafka lag config.
+		if err := deleteScaledObject(ctx, dynClient, namespace, scaledObjectName(f.Name)); err != nil {
```
gauron99: I think this delete shouldn't run when nothing tells us the ScaledObject exists. Kubernetes checks RBAC before existence so with the roles from `on_cluster_build.md` even a plain http deploy fails:
```
failed to remove stale Kafka scaler before switching triggers: failed to delete ScaledObject c1test/pr4051-func-kafka: scaledobjects.keda.sh "pr4051-func-kafka" is forbidden: User "system:serviceaccount:c1test:default" cannot delete resource "scaledobjects" in API group "keda.sh" in the namespace "c1test"
```
(reproduced with a func-utils image built from this branch; the Deployment and Service stay up with no scaler). Your comment at 317-323 makes the same argument for the companion deletes.

What do you think about gating this on a record, like `--expose` does with `RouteNamespaceAnnotation` in `clearExposure`?
1. Record the active trigger type (and whether a TriggerAuthentication was created) on the function's Service before creating the scaler.
2. Treat a missing record as HTTPSO since thats what it was prior this PR.
3. Use the annotation record on func service to check if we are trying to swtich a scaler and if so, error early

pkg/keda/deployer.go
```diff
@@ -139,42 +299,172 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+		if !needsAuth {
+			// SASL/TLS credentials were removed from run.kafka while the kafka
+			// trigger stayed: a prior deploy may have left a
+			// TriggerAuthentication behind that nothing references anymore.
+			// Delete it only AFTER the ScaledObject above has been reconciled to
+			// drop its authenticationRef -- deleting first would, if that update
+			// then failed, leave the live ScaledObject pointing at a
+			// TriggerAuthentication that no longer exists. Not fatal, same
+			// treatment as the no-kafka-at-all cleanup above.
+			if err := deleteTriggerAuth(ctx, dynClient, namespace, triggerAuthName(f.Name)); err != nil {
```
gauron99: Same pattern here and at 353-354: a kafka deploy without credentials, and every plain http deploy, deletes `<name>-kafka-auth` just in case. Even the dev cluster's `default` SA has no rights on triggerauthentications, so each such deploy prints a warning about an object that never existed:
```
warning: failed to delete TriggerAuthentication default/kremote-kafka-auth: triggerauthentications.keda.sh "kremote-kafka-auth" is forbidden: User "system:serviceaccount:default:default" cannot delete resource "triggerauthentications" ...
```
The record from the comment above would cover this too.

pkg/keda/deployer.go
```diff
@@ -139,42 +299,172 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
...
+	// existence, so such a delete returns Forbidden (not NotFound) even when
+	// there is nothing to remove -- making it fatal would break every
+	// single-type deploy on a cluster with tighter RBAC.
+	//
+	// None of this waits for actual removal: KEDA attaches its own finalizer
+	// to ScaledObject/TriggerAuthentication, so a successful Delete() only
+	// sets deletionTimestamp and returns. A transition immediately followed
+	// by another deploy could still see the old scaler mid-termination; that
+	// narrow, self-correcting race resolves on retry and isn't worth a
+	// wait-for-actual-deletion poll on every deploy.
```
gauron99: This isn't really an edge-case,  a simple redeploy with upgrades
`func deploy --deployer=keda` with http (standard, old functions default etc)
then re-deploy with 
`func deploy --deployer=keda` with changed/added kafka trigger `scale.keda.triggers: ... kafka` hits 
```
Error: deploy error. failed to ensure ScaledObject: failed to create ScaledObject s3test/s3fn-kafka: admission webhook "vscaledobject.kb.io" denied the request: the workload 's3fn' of type 'apps/v1.Deployment' is already managed by the ScaledObject 's3fn'
```
and that leaves a function without a scaler at all until you run deploy again.

I think since we plan to move to breaking-change newer keda add-on version with that IntereceptorRoute I say :

- we simply say that its currently required to undeploy first before changing to a different-scaler-scaled function when re-deploying

 -> so we introduce an early check before anything is changed on cluster that reads something like what we already have for the deployer switch check (since the deployers use different resources we say that itsrequired to undeploy first before re-deploying with different deployer)

pkg/keda/deployer.go
```diff
@@ -139,42 +299,172 @@ func (d *Deployer) Deploy(ctx context.Context, f fn.Function) (fn.DeploymentResu
 		return fn.DeploymentResult{}, fmt.Errorf("failed to get service %s/%s: %v", namespace, f.Name, err)
 	}
 
-	ref := deployer.NewExposureRef(f.Name, namespace, interceptorNS)
-	if err := ensureInterceptorBridgeService(ctx, k8sClientset, ref, deployment); err != nil {
-		return fn.DeploymentResult{}, fmt.Errorf("failed to ensure proxy service exists: %w", err)
+	// Delete stale resources for whichever trigger type is NOT currently
```
gauron99: Minor: this comment is long compared to the rest of the codebase. I'd keep the rule and the reason and trim the rest. (or can be changed slighlty based on the other suggestions)

pkg/keda/remover.go
```diff
@@ -50,24 +50,39 @@ func (remover *Remover) Remove(ctx context.Context, name, ns string) error {
...
 	}
 
+	// Clean up Kafka scaling resources before deleting the Deployment.
+	// These have ownerReferences so they'd be garbage-collected, but
+	// explicit deletion avoids races with a slow GC. Errors here (both
+	// functions already ignore not-found) are not fatal to Remove: the
+	// owner reference still cleans these up eventually, but the user is
+	// warned so a persistent failure (e.g. missing RBAC) doesn't go
+	// unnoticed.
+	if err := deleteScaledObject(ctx, dynClient, ns, scaledObjectName(name)); err != nil {
```
gauron99: Keeping these explicit deletes makes sense to me for faster cleanup (the slow-GC reason above). One thing: they run for every keda function, including http-only ones that never had a ScaledObject or TriggerAuthentication. `func delete` runs this with the user's own credentials, so a user without delete rights on `keda.sh` resources gets a Forbidden warning on every delete, and needs extra grants for objects that never existed. With the record suggested on the deployer, I think we could delete here only what was recorded, and leave the rest to the owner references both objects already carry.

pkg/keda/kafka_scaling.go
```diff
@@ -0,0 +1,552 @@
...
+		return fmt.Errorf(
+			"function name %q is too long for the keda deployer: its Kafka scaler resource would be named %q, which is not a valid name (%s)",
+			name, longest, strings.Join(errs, "; "))
+	}
+	return nil
+}
+
+// triggers returns the function's explicitly configured KEDA triggers, or a
+// plain http trigger when none are configured at all. This only matters for
+// callers of Deploy that bypass fn.Function.Validate (which requires
```
gauron99, line 66: Minor: this still says `Function.Validate` requires keda triggers to be declared explicitly, which stopped being true with the default-http change.

pkg/k8s/wait.go
```diff
@@ -69,6 +69,10 @@ func checkIfDeploymentIsAvailable(ctx context.Context, clientset *kubernetes.Cli
 
 	desiredReplicas := *deployment.Spec.Replicas
 
+	if desiredReplicas == 0 {
```
gauron99, line 72: Nice! this also fixes a bug on main: redeploying a keda function that is scaled to zero fails there with `deployment did not become ready: could not find current pod-template-hash for deployment ...` (checked with a main build; this branch passes). 

Could we get a unit test for `checkIfDeploymentIsAvailable` with `Spec.Replicas = 0`, and a short comment that at zero the new pod template isn't verified? A broken image then only shows on the next wake-up, unlike knative, which starts each new revision once.

pkg/functions/function_migrations.go
```diff
@@ -406,6 +407,143 @@ func nestedString(v interface{}, keys ...string) string {
...
+
+	// Clear the old location so it doesn't get serialized.
+	f.Deploy.Options.Scale = nil
+
+	// keda deployer without triggers: default to http. Since ValidateScale no
+	// longer rejects a nil scale.keda for keda (the deployer defaults to the
+	// http scaler at runtime), this injection is now cosmetic -- it just
+	// materializes that default explicitly into migrated old files rather than
+	// leaving scale.keda absent. Kept so migrated files stay self-describing.
+	if f.Deployer == "keda" {
```
gauron99, line 515: This seems unneeded now that a nil `scale.keda` means http (`ValidateScale` accepts it and `triggers()` falls back to http), and it has a side effect: a keda function migrated from 0.37.0 can't move to another deployer without a manual edit. After `func delete`, `func deploy --deployer knative` fails because the injected block stays in func.yaml:
```
Error: 'func.yaml' contains errors:
	scale.keda requires deployer: keda
```
I think we should drop this block (510-528); migrated keda functions then look like new ones.

pkg/functions/function_migrations.go
```diff
@@ -406,6 +407,143 @@ func nestedString(v interface{}, keys ...string) string {
...
+		}
+	}
+
+	// deploy.deployer and deploy.expose keep their YAML tags. Populate the
+	// observed-state fields from disk so the migration is self-contained
+	// regardless of how f was constructed. f.Root == "" (library callers)
+	// needs no handling here: the in-memory value already reflects whatever
+	// was set via the Go field name, so there's nothing on disk to migrate
+	// from and nothing to fall back to.
+	if disk.Deploy.Deployer != "" {
```
gauron99, line 536: Seems like a leftover from the reverted `activeDeployer` rename: `deploy.deployer` and `deploy.expose` keep their tags, so the first unmarshal already fills these. I think this block and the two fields in `oldDeploy` (434-435) can go.

pkg/functions/function_migrations.go
```diff
@@ -406,6 +407,143 @@ func nestedString(v interface{}, keys ...string) string {
...
+	}
+	var disk struct {
+		Deploy oldDeploy     `yaml:"deploy,omitempty"`
+		Scale  *ScaleOptions `yaml:"scale,omitempty"`
+	}
+
+	if f.Root != "" {
+		bb, err := os.ReadFile(filepath.Join(f.Root, FunctionFile))
+		if err == nil {
+			_ = yaml.Unmarshal(bb, &disk)
```
gauron99, line 445: Minor: this drops the unmarshal error. Low risk, since `NewFunction` fails earlier on a broken file, but I'd rather not hide it.

pkg/functions/function_options.go
```diff
@@ -7,15 +7,60 @@ import (
 )
 
 type Options struct {
-	Scale     *ScaleOptions     `yaml:"scale,omitempty"`
+	// Scale is kept for YAML deserialization of old func.yaml files (pre-0.38.0
+	// stored scale under deploy.options.scale). The 0.34.0 migration writes to
+	// it and the 0.38.0 migration moves it to Function.Scale. Hidden from the
+	// JSON schema so new files never use this path.
+	Scale     *ScaleOptions     `yaml:"scale,omitempty" jsonschema:"-"`
```
gauron99, line 14: This seems to be why a new file can still use the old key with no feedback: a 0.38.0 func.yaml with `deploy.options.scale.min: 3` deploys with no warning and no min-scale (checked on a cluster), because the unknown-key warning in `Migrate()` sees a known field.

The other disk-reading migrations (`migrateGitToSource`, `migrateFromInvokeStructure`, ...) read func.yaml into their own throwaway old type, so removed keys vanish from the current struct. What do you think about doing the same here?
- remove `Scale` from `Options`
- give `migrateToSpecs_previousFunction` its own old options type (it reuses the current `Options`, function_migrations.go:590) and drop its scale copy (283-284)
- let `migrateScaleToTopLevel` read both old locations from disk: top-level `options.scale` (before 0.34) and `deploy.options.scale` (0.34-0.37). Its old type only needs min/max/metric/target/utilization, since no old file has `keda`/`kpa`
- drop the in-memory fallback (458-468): `Migrate()` only runs from `NewFunction` and `hasInitializedFunction`, and `Client.Init` gives new functions the latest spec version. (`hasInitializedFunction` takes the fallback today, but only uses `Initialized()`, so that's harmless.)

Then the existing unknown-key warning covers it, and it also fixes the pre-0.34 KPA loss from the earlier Copilot thread on this migration.

cmd/deploy_test.go
```diff
@@ -2970,9 +2981,20 @@ func TestDeploy_ExposeIgnoredByDeployerNote(t *testing.T) {
 			// route cases need OpenShift gate open; none/empty do not care.
 			cleanup := k8s.SetOpenShiftForTest(true, nil)
 			defer cleanup()
-			if _, err := fn.New().Init(fn.Function{Runtime: "go", Root: root}); err != nil {
+			f, err := fn.New().Init(fn.Function{Runtime: "go", Root: root})
+			if err != nil {
 				t.Fatal(err)
 			}
+			if slices.Contains(tt.args, "keda") {
+				// keda requires at least one trigger to be declared explicitly.
```
gauron99, line 2989: I think this setup, the one at 2751-2757, and `cmd/delete_test.go:360-362` are no longer needed since keda defaults to http. The tests pass without them, and this comment isn't true anymore.

e2e/e2e_expose_test.go
```diff
@@ -60,6 +60,30 @@ import (
 // and s2i. Remote deploys keep pack: host cannot build in-cluster.
 // ---------------------------------------------------------------------------
 
+// setKedaHTTPTrigger declares an explicit http KEDA trigger in the func.yaml
+// at root. keda defaults to the http scaler when no scale.keda is given, so
+// this is not required to reach a deployable state; it makes the http trigger
+// explicit in func.yaml for tests whose assertions read it back.
+func setKedaHTTPTrigger(t *testing.T, root string) {
```
gauron99, line 67: Same here: the comment says it's not required, and no test reads the trigger back. I'd drop the helper and its 6 calls.

gauron99 (commented):
thanks for dealing with the migrations of func.yaml, changing the scale keys around.

I have some ideas to simplify further keeping in mind the future add-on breaking change introduction which will simplify this whole thing even further when we switch to a single ScaledObject and we can simply CRUD that.

In the meantime I think we can simply have a guard for the scaler switch.
And as for the service annotations -> this will match the current exposer/deployer stamp which is on every service too which helps with other commands like describe/list etc which can simply fetch the functions service and should get the info about the function there

aliok:
/close

Closing this to create 2 separate PRs. Part 1 is here: https://github.com/knative/func/pull/4063

